PR #412, opened 2012-03-15
Added new method Str::unique().

swt83:
Generate a unique string based on time instead of random chance.

### DIFF
--- a/laravel/str.php
+++ b/laravel/str.php
@@ -292,6 +292,14 @@ class Str {
 	{
 		return substr(str_shuffle(str_repeat(static::pool($type), 5)), 0, $length);
 	}
+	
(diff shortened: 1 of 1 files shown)


najela:
@swt83 But the md5 algorithm is not guaranteed to generate a unique string each time right?

GrahamCampbell:
Neither is any hashing algorithm.

najela:
@GrahamCampbell  Then we could use this https://gist.github.com/dahnielson/508447#file-uuid-php-L74

GrahamCampbell:
@najela This PR isn't even up for discussion anymore as it's for Laravel 3, which has been EOL for a long time now.

PR #623, opened 2012-05-04
Fix View's use of Events so you can actually implement your own things

sparksp:
I've just tried to make use of the View events and ran into a few problems; here's how/why I fixed it:
1. Using `Event::first` returns the response of the first event, often `null`.  So instead use `Event::until` to get the first non-null response.
2. If none of the events return a value then `Event::until` returns `$responses` (an empty array) - `null` makes more sense.
3. Get Blade to return `null` instead of `false` if it can't handle the file, since Event watches for `null`, and `null` is what you get with no return too.  And updated View to check for `null` too.

**Note:** Tested on master and rebased for develop; **untested** on develop.

### DIFF
--- a/laravel/blade.php
+++ b/laravel/blade.php
@@ -42,7 +42,7 @@ class Blade {
 			// return false so the View can be rendered as normal.
 			if ( ! str_contains($view->path, BLADE_EXT))
 			{
-				return false;
+				return;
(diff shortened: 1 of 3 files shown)


ALLMA99:
### ZOHRAN

PR #678, opened 2012-05-16
Allow to add database specific fieldtype within Schema using Closure.

crynobone:
This allow developer that need database specific configuration to do so, without the need to do Schema, ALTER TABLE using Raw Query and so on.

For example:

```
Schema::create('users', function($table)
{
    $table->increments('id');
    $table->string('username', 50);
    $table->generic('gender', function ($column) {
        return "ENUM('m','f')";
    });
});
```

Initial discussion at #668


mikey0000:
geospatial is also part of mysql as well as of 5.6

PR #681, opened 2012-05-17
Added Output Buffering

bencorlett:
Hi,

So currently, if I have something like this:

```
Route::get('/', function()
{
    echo 'First';

    return Redirect::to('second');
});

Route::get('second', function()
{
    return 'Second';
});
```

The redirect never happens and I just get an empty response. That's because Laravel isn't utilising output buffering. I have added output buffering, which means you can echo anything, set session values or whatever then return an object that's an instance of Laravel\Redirect, and the redirect will happen.

Signed-off-by: Ben Corlett bencorlett@me.com

### DIFF
--- a/laravel/core.php
+++ b/laravel/core.php
@@ -29,7 +29,7 @@ define('MB_STRING', (int) function_exists('mb_get_info'));
 |
 */
 
-ob_start();
+ob_start('mb_output_handler');
 
 /*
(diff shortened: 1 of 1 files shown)


hammad74:
Hi Ben. Look I'm now on gthub. I need Pinter's on how to utilize it. And creait a page or code key. Or a template

PR #866, opened 2012-06-27
Added @assign Blade function for assigning variables inside Blade view f...

nikkobautista:
...iles

Signed-off-by: Nikko Bautista nikko@nikkobautista.com


AhmedRamadanAhmed:
I understand the overall code but how this use to assign value to variable @nikkobautista

PR #897, opened 2012-07-03
Support for MySQL character sets and collation

bluematt:
This commit adds support for setting character set and/or collations for MySQL.  This is useful where the database system's default character set and/or collation is not set to the character set/collation desired for the table/column.

The default character set is the value set in the charset option in the MySQL database configuration in /application/config/database.php

The default collation is set by the MySQL server.

Tables and columns can both have their character set/collation set individually.

Example migration:

``` php
/**
 * File: /application/migrations/2012_07_02_193607_create_mixed_charset_table.php
 */

class Create_Mixed_Charset_Table {

    /**
     * Make changes to the database.
     *
     * @return void
     */
    public function up()
    {
        Schema::create('my_mixed_table', function($table) {
            $table->charset('utf8'); // same as default
            // $table->collate('utf8_unicode_ci'); // default is 'utf_general_ci'

            $table->increments('id');

            // uses table default values
            $table->string('text_default', 100);

            // force Latin 1, collation is default for this charset
            $table->string('text_latin1', 100)->charset('latin1');

            // force a specific collation
            $table->string('text_utf8_unicode_ci', 100)->collate('utf8_unicode_ci');

            // force Latin 1 with a specific collation
            $table->string('text_latin1_german_ci', 100)->charset('latin1')->collate('latin1_german1_ci');
        });
    }

    /**
     * Revert the changes to the database.
     *
     * @return void
     */
    public function down()
    {
        Schema::drop('my_mixed_table');
    }

}
```

This results in the following SQL:

``` sql
CREATE TABLE `my_mixed_table` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `text_default` varchar(100) NOT NULL,
  `text_latin1` varchar(100) CHARACTER SET latin1 NOT NULL,
  `text_utf8` varchar(100) CHARACTER SET utf8 COLLATE utf8_unicode_ci NOT NULL,
  `text_latin1_german` varchar(100) CHARACTER SET latin1 COLLATE latin1_german1_ci NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8
```

Note that specifying an invalid character set or collation will result in a SQL error.

### DIFF
--- a/laravel/database/schema/grammars/mysql.php
+++ b/laravel/database/schema/grammars/mysql.php
@@ -33,6 +33,16 @@ class MySQL extends Grammar {
 			$sql .= ' ENGINE = '.$table->engine;
 		}
 
+		if ( ! is_null($table->charset))
+		{
+			$sql .= ' DEFAULT CHARACTER SET = '.$table->charset;
+		}
+
+		if ( ! is_null($table->collation))
+		{
+			$sql .= ' COLLATE = '.$table->collation;
+		}
+
 		return $sql;
 	}
 
@@ -77,7 +87,7 @@ class MySQL extends Grammar {
 			// types to the correct types.
 			$sql = $this->wrap($column).' '.$this->type($column);
 
-			$elements = array('unsigned', 'nullable', 'defaults', 'incrementer');
+			$elements = array('unsigned', 'charset', 'collate', 'nullable', 'defaults', 'incrementer');
 
 			foreach ($elements as $element)
 			{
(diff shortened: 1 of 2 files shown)


abepark01:
I realize this PR is closed, but I have a similar issue where I need to set the character set and collation at the table and column level. I checked the master branch repository and did notice this PR didn't make it in.
Can I create another PR based on work @bluematt did for Laravel 5 on the development branch?

thstarshine:
+1 for adding this to Laravel 5.

devWebMntq:
+1

shehi:
@abepark01 I believe its worth the effort. +1

Cyrille37:
:+1: for `$table->charset('charset id')`

Cyrille37:
ok, it presents as a property `$table->charset = 'utf8'`

PR #1192, opened 2012-09-07
Date format validation rule

franzliedke:
This allows for validating a PHP [date format](http://php.net/manual/en/datetime.createfromformat.php#refsect1-datetime.createfromformat-parameters).

Rules would look like this:

```
'start_time' => 'date_format:"H:i"',
```

Note the quotes wrapping the value so that the colon does not count as parameter separator.

---

Everything documented and properly unit-tested.

### DIFF
--- a/laravel/tests/cases/validator.test.php
+++ b/laravel/tests/cases/validator.test.php
@@ -494,11 +494,8 @@ class ValidatorTest extends PHPUnit_Framework_TestCase {
 		$rules = array('date' => 'date_format:j-M-Y');
 		$this->assertTrue(Validator::make($input, $rules)->valid());
 
-		$input['date'] = '2009-02-15 15:16:17';
-		$rules['date'] = 'date_format:Y-m-d H\\:i\\:s';
-		$this->assertTrue(Validator::make($input, $rules)->valid());
-
-		$rules['date'] = 'date_format:"Y-m-d H:i:s"';
+		$input['date'] = '2009-02-15,15:16:17';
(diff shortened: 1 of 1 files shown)


mjbrunme:
PHP's date_parse_from_format is either broken or is intentionally far too lenient to use to validate that a given date is formatted correctly.  You can not, for example, validate a 2 year vs 4 year date, or a month with or without leading zeroes.

For example, if you set up the rule:
```php
'year' => 'date_format:Y'
```

and pass "16" to it, it will not report any validation issues.

mansouralex:
Hello @franzliedke ,@mjbrunme I'm facing a related issue with the `Y` format:

The rule as: `date_format:d-m-Y` and it's accepting such date: "21-12-0020"

Any thoughts how we can restrict this and end up with valid year ?

ncpope:
0020 is a valid year

satodu:
I used CARBON. http://carbon.nesbot.com/docs/

It's made easier, you can install using the caomposer and call the object in your controller.

mansouralex:
@satodu I'm already aware and using CARBON. But how do you suggest we employ it in this case?

troccoli:
The date is a valid date. If it's not for your use case I suggest you
create your own validation rule.

satodu:
$dt = Carbon::createFromDate(0020, 10, 6);
if ($dt->year < 2012) {
    echo 'Man! it's too old';
}

satodu:
@mansouralex 

$dt = Carbon::createFromDate(0020, 10, 6);
if ($dt->year < 2012) {
     echo 'Man! it's too old';
}

PR #2977, opened 2014-08-22
[Proposal] More detailed logging

barryvdh:
When looking through errors it is not very clear where an error occured. For example 404 errors: from ` exception 'Symfony\Component\HttpKernel\Exception\NotFoundHttpException' in /var/www/html/project/vendor/laravel/framework/src/Illuminate/Routing/RouteCollection.php:140` you don't know on what path that happend (and beginners don't even get that this is just a 404).
`Error 404 on GET http://mydomain.com/my-route: exception 'Symfony\...` would make much more sense imho..

(Follow-up on https://github.com/laravel/laravel/pull/2764 for 4.3 branch)

### DIFF
--- a/app/Providers/ErrorServiceProvider.php
+++ b/app/Providers/ErrorServiceProvider.php
@@ -1,6 +1,6 @@
 <?php namespace App\Providers;
 
-use App, Log, Exception;
+use App, Log, Exception, Request, URL;
 use Illuminate\Support\ServiceProvider;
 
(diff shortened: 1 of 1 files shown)


joshhornby:
+1

barryvdh:
Okay now it changed to the Exception handler

	public function report(Exception $e)
	{
		return parent::report($e);
	}

Would be nice if we would have gotten the request there also, but I guess that's too late now.

rweas:
:+1:

PR #3019, opened 2014-09-15
Remove the composer.lock file (.gitignore)

tyloo:
This composer.lock would be useful to keep (for faster composer install).

I don't know why you set phpunit as a dev dependency. I have 5 minutes composer update because of this bad guy now :(

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 /vendor
 composer.phar
-composer.lock
 .env.*
 .DS_Store
 Thumbs.db


DmitryRomanenko:
I disagree.  
This file should be in any PHP project repository because it determines exact versions of dependencies.

If any dependency is broken in future, the project installation with composer will fail.  
That's why it's necessary.

GrahamCampbell:
This is already done on the develop branch. I'll will get merged into the master very early next week.

mattstauffer:
@GrahamCampbell Yah, I noticed my latest l5 install didn't have that anymore. That makes https://github.com/laravel/installer/issues/5 closable as well, I believe, right?

GrahamCampbell:
@mattstauffer Yeh.

PR #3026, opened 2014-09-19
Add support for configuring AWS S3 region.

paulofreitas:
See https://github.com/laravel/framework/pull/5773 for details.

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -52,6 +52,7 @@ return [
 			'driver' => 's3',
 			'key'    => 'your-key',
 			'secret' => 'your-secret',
+			'region' => 'your-region',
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
NB, https://github.com/laravel/framework/pull/6305 was merged to laravel/framework that adds this.

taylorotwell:
Can we set it to a default valid region like us-east-1 or whatever it is?

paulofreitas:
@GrahamCampbell Thanks for updating here! Glad to know it got implemented somehow. :)

@taylorotwell I think it's a reasonable default, many libraries (including AWS SDK for Java) uses `us-east-1` by default, and if I'm not wrong it's the default region used to create regions when no custom region was opted. ;)

PR #3145, opened 2014-11-20
Make trailing slash rewrite use relative path

jakezatecky:
The "Redirect Trailing Slash" `RewriteRule` does not use a relative path. Instead, it uses an absolute path to the web root. This causes an issue if you add `RewriteBase` to `.htaccess`.

To see this problem in action, consider the following `.htaccess:

```
RewriteBase /newbase

# Redirect Trailing Slashes...
RewriteRule ^(.*)/$ /$1 [L,R=301]
```

If we make a request to  `http://some-site.com/newbase/file/`, we will be redirected to `http://some-site.com/file`, which is incorrect.

By changing the RewriteRule to `RewriteRule ^(.*)/$ $1 [L,R=301]`, you now get redirected to the correct location: `http://some-site.com/newbase/file`

Granted, you could argue that since this issue would only come up if you edit the `.htaccess` file and add the `RewriteBase` rule, you should know to change the corresponding `RewriteRule`. However, if you know little about `.htaccess` files, you might spend an hour trying to figure out why you were redirected to the web root when an extra slash was added at the end.

It's important to note that this change does not appear adversely affect anything if we still use the web root.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -6,7 +6,7 @@
     RewriteEngine On
 
     # Redirect Trailing Slashes...
-    RewriteRule ^(.*)/$ /$1 [L,R=301]
+    RewriteRule ^(.*)/$ $1 [L,R=301]
 
     # Handle Front Controller...
     RewriteCond %{REQUEST_FILENAME} !-d
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -6,7 +6,8 @@
     RewriteEngine On
 
     # Redirect Trailing Slashes...
-    RewriteRule ^(.*)/$ $1 [L,R=301]
+    RewriteCond %{REQUEST_URI} (.*)/$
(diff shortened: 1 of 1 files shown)


stuarthannig:
(thumbsup)

vlakoff:
Why not?

franzliedke:
Probably because Taylor has said before Laravel should be run from a domain root.

It would be lovely if this could be reconsidered.

vlakoff:
This has revealed itself to be quite popular.

Considering I have developed for clear reference my [answer on SO](https://stackoverflow.com/questions/21417263/htaccess-add-remove-trailing-slash-from-url#27264788), could this issue be reconsidered?

(Ah, and… [mine is robuster than yours was](https://github.com/laravel/laravel/commit/dd036f559b3213279e4dfa8ab055aeac3dab2521) :p)

PR #3199, opened 2014-12-29
Warn that cache key prefix is ignored on cache flush

tremby:
As discussed in https://github.com/laravel/framework/issues/5034

If you would prefer this to be against a different branch I can rebase.

### DIFF
--- a/app/config/cache.php
+++ b/app/config/cache.php
@@ -82,6 +82,9 @@ return array(
 	| be other applications utilizing the same cache. So, we'll specify a
 	| value to get prefixed to all our keys so we can avoid collisions.
 	|
+	| Note that when flushing the cache, this cache key prefix will be ignored
+	| and the entire cache flushed.
+	|
 	*/
 
 	'prefix' => 'laravel',


JosephSilber:
FWIW, I agree with @tremby. Although we geeks like to pull out the RTFM stick whenever possible, truth is most people don't.

If you're setting a prefix, chances are you'd at least read what's on top of it (or at least more likely than go read the docs).

garygreen:
Wait, `Cache::flush()` get's rid of all cache and ignores the prefix for your app? Wtf, does anyone else find this really odd behavior? To me it should flush just your apps cache, i.e. within it's own namespace/prefix. Sounds more like a bug to be flushing all your other projects caches.
Is this the same behavior in 5.0? If so, I think it should change. Either way I'm +1 to this, was not obvious to me.

tremby:
As far as I know, yes. See https://github.com/laravel/framework/issues/5034
for my original bug report and some discussion.

taylorotwell:
This is more of a documentation concern.

PR #3207, opened 2015-01-02
[5.0] Move exception handler to Handlers directory

martinbean:
Move the exception handler to be under the *Handlers* directory.


GrahamCampbell:
This doesn't actually make sense because it's not a "handler" that's executed by the command bus.

martinbean:
@GrahamCampbell Fair enough.

PR #3208, opened 2015-01-03
[5.0] Modified default drivers to use env

johnathanmdell:
Based on some of the configuration files I changed the Auth, Database, Filesystem and Mail configurations files to use env for the selection of the default driver or falling back to the Laravel default values.

Signed-off-by: Johnathan Michael Dell <johnathanmdell@gmail.com>


config/database.php
```diff
@@ -26,7 +26,7 @@
 	|
 	*/
 
-	'default' => 'mysql',
+	'default' => env('DATABASE_DRIVER', 'mysql'),
```
ibrasho: This should be DEFAULT_CONNECTION or something similar...

johnathanmdell: Is your reasoning for this purely based on that this would allow the database default connection to be grouped with other database settings? As this breaks away from the "driver" convention.

As example from the cache configuration...

	/*
	| Default Cache Store
	|
	| This option controls the default cache connection that gets used while
	| using this caching library. This connection is used when another is
	| not explicitly specified when executing a given caching function.
	|
	*/

	'default' => env('CACHE_DRIVER', 'file'),

The above property is for the default cache connection but is named CACHE_DRIVER. Would it make more sense to name the default database connection something like DB_DRIVER, thus sticking to the already in use convention and allowing the grouping of database settings with the prefix DB_.

config/auth.php
```diff
@@ -15,7 +15,7 @@
 	|
 	*/
 
-	'driver' => 'eloquent',
+	'driver' => env('AUTH_DRIVER', 'eloquent'),
```
Stichoza: Everything else is ok, but are you sure about authentication driver?

config/database.php
joecohens: :+1: for `DB_DRIVER`

config/auth.php
johnathanmdell: @Stichoza not sure I'm following, could you elaborate please?

Stichoza: @johnathanmdell I'm not sure either, but as I think, In most cases (like 99.9%) authentication is code-specific. I mean the same app won't work on two environments if you change just an auth driver (and not code) on one environment. Tell me if I'm wrong.

johnathanmdell: @Stichoza should your authentication layer really depend that heavily on your application layer? I have had a instance where I created a SOAP based authentication driver and had to switch between the database and SOAP when the SOAP service went down.

crynobone: > I have had a instance where I created a SOAP based authentication driver and had to switch between the database and SOAP when the SOAP service went down.

Wouldn't you agree that fall under the 0.1% of the cases.

joecohens: I think this would work if the framework it self had by default other auth services but since eloquent it's the only one I don't think it's needed.

johnathanmdell: @joecohens Laravel has two default authentication drivers, database and eloquent.

@crynobone I completely agree it might be the 0.1%, however as stated it is completely optional and it will revert to eloquent if not set, all this is doing is extending configuration for the 0.1% without affecting the 99.99%

joecohens: @johnathanmdell yeah that's right sorry ;). I guess we could do this env config for most of the values on the configs, but I think this one is not something that you change depending on your env, I think is more project wise, so don't know if for the default app if this brings some kind of value. Hope I've explained my self.

johnathanmdell: @joecohens like I said it might not bring value to most, however it will bring value to some and is completely optional to use.

crynobone: >  however it will bring value to some and is completely optional to use.

In this case, it should only added to project that require them (there no reason to bloat .env with so many default pre-defined config), serving the 0.1% is a weak argument.

johnathanmdell: @crynobone you are being ridiculous.

The .env is not even touched by this change. If you do not use it (and for the umpteenth time you do not need to use it), it defaults to the string that was already in the configuration which in this case is 'eloquent'.

As an example, this is the queue driver configuration in Laravel 5.

```php
'default' => env('QUEUE_DRIVER', 'sync'),
```

Do you see it anywhere in the .env.example?

```
APP_ENV=local
APP_DEBUG=true
APP_KEY=SomeRandomString

DB_HOST=localhost
DB_DATABASE=homestead
DB_USERNAME=homestead
DB_PASSWORD=secret

CACHE_DRIVER=file
SESSION_DRIVER=file
```

If you are going to go around calling arguments weak then please read the entire commit log first.

crynobone: Yes it's not being touch, but you now bloating the config with an option that 99.9% is not going to be changed. The purpose of `.env` is to avoid storing sensitive data inside config such as L4 (e.g hardcoding database config in `app/config/production/database.php` etc. <http://12factor.net/config>) as well as adding options that are most of the type going to be changed between environment such as `APP_DEBUG`, `CACHE_DRIVER`, `SESSION_DRIVER`. 

Now we trying to force our 1% use cases in `.env`, then what the point of having `config` folder, why not just have everything under `.env` (which is my reason for calling it bloating).

crynobone: Same reason why <https://github.com/laravel/laravel/blob/develop/config/services.php> is empty instead having a predefined `.env` key "just in case you ever need it".

johnathanmdell: I'm trying really hard to see where this bloating is happening.

It is a an optional convenience for switching between authentication drivers that you are not forced to use. If you can show me exactly how it is bloating the configuration file or the .env file for that matter then I'll gladly remove it.

filipac: This isn't making the .ENV file to be bloated. If you want to use it, you add it to your .ENV file, otherwise it will use the default option. Simple as that.

johnathanmdell: @filipac :thumbsup:

johnathanmdell:
Changed the env database driver key from `DATABASE_DRIVER` to `DB_DRIVER` daf844e6c913b1e52a8ac352ae14670bc2baae87

config/auth.php
stuarthannig: I fail to see the need to have this setting as ENV variables rather than hardcoded.

It doesn't need to be dynamic based on environment or kept as sensitive information.

I could see DB_DRIVER, FILESYSTEM_DRIVER, and MAIL_DRIVER changing based on environment.

PR #3209, opened 2015-01-03
[5.0] Proposal - Install bower components in vendor dir

xsbeats:
Laravel Elixir's `publish` command expects bower components to be installed in `vendor/bower_components`


xsbeats:
Noticed that Elixir has added a configuration file.  It would likely be best if the bower_components configuration was added there, instead of here.

PR #3211, opened 2015-01-04
Environment calls in config use lowercase and dot notation.

Garbee:
Stemming from [issue 6733](https://github.com/laravel/framework/issues/6733) on the framework repo, this pull attempts to keep consistency between env naming and how config::get() pulls things.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,11 +1,11 @@
 APP_ENV=local
-APP_DEBUG=true
-APP_KEY=SomeRandomString
+app.debug=true
+app.key=SomeRandomString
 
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
I really don't think this matters. The dotenv is not meant to be a replacement for config as such. I worry this might actually be more confusing.

Garbee:
I really don't care myself. People will find it confusing either way it goes. Either it is inconsistent with how config::get works or it is not consistent with every other projects default way of assigning env variables. But, at least a good PR was made so they can't complain it was never considered.

PR #3212, opened 2015-01-08
[5.0][Proposal] Move csrf middleware to be optional

joecohens:
I was working on an API with L5 and you can't disable the csrf middleware so every request yo have to send it. I think csrf protection should be optional, what do you think?


rtablada:
The CSRF middleware is optional.
Just like you showed, it can be pulled out and made into a named middleware.

ibrasho:
Comment the Csrf middleware in `app/Http/Kernel.php`?

joecohens:
Yeah I know it's optional and that I can comment it. I just thought it doesn't need to be set default. Thank you all :)

ibrasho:
Look at the amount of damage that could happen if this is not enabled by default, I think I prefer to keep it as is. : )

PR #3215, opened 2015-01-17
Enable .env as default - app.php relies on this for csrf

jonoofficial:
Token mismatch occurs if .env isn't enabled. Allows laravel to be used out of the box.

### DIFF
--- a/.env
+++ b/.env
@@ -1,6 +1,6 @@
 APP_ENV=local
 APP_DEBUG=true
-APP_KEY=SomeRandomString
+APP_KEY=YourSecretKey!!!
 
 DB_HOST=localhost
 DB_DATABASE=homestead


TheShiftExchange:
I dont understand - this should be covered with 'key' => env('APP_KEY', 'YourSecretKey!!!'), that is already in the app config. If no APP_KEY is set in .env - it will already used YourSecretKey!!!?

TheShiftExchange:
I dont understand - this should be already taken care of by the `app.php` : `'key' => env('APP_KEY', 'YourSecretKey!!!'),`. - if no `.env` file is found, then it will already default to `YourSecretKey!!!`.

I just tried it on a new build of L5 - and I get no issues.

TheShiftExchange:
I did notice an issue with the environment keys as a result of this - but it is a different issue.

See https://github.com/laravel/laravel/pull/3216 . Its only half your PR - but it fixes the `key:generate` issue

jonoofficial:
When i run a default installation posting using the default auth registration method doesn't work until the .env file is in place. Weird!

TheShiftExchange:
Did you try a full new installation of the package? There were some changes to both `laravel/laravel` and `laravel/framework` - so if you are running an old `laravel/laravel` it wont work correctly.

PR #3217, opened 2015-01-18
added default mysql port config as well as environment variable

darryldecode:
When I was first using migrate command, it gives a port error and I need an extra step for it to work with my local homestead environment.
Homestead by default uses port 33060 and forwards it to 3306. So I thought it might be good to just add a default config for MySQL port for local environments to enable it to work out of the box.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -58,7 +58,7 @@ return [
 			'database'  => env('DB_DATABASE', 'forge'),
 			'username'  => env('DB_USERNAME', 'forge'),
 			'password'  => env('DB_PASSWORD', ''),
-			'port'  	=> env('DB_PORT', 3306),
+			'port'      => env('DB_PORT', 3306),
 			'charset'   => 'utf8',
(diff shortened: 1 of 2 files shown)


franzliedke:
Are these different numbers on purpose?

config/database.php
```diff
@@ -58,6 +58,7 @@
 			'database'  => env('DB_DATABASE', 'forge'),
 			'username'  => env('DB_USERNAME', 'forge'),
 			'password'  => env('DB_PASSWORD', ''),
+			'port'  	=> env('DB_PORT', 3306),
```
GrahamCampbell: align the array with spaces

darryldecode: done..

GrahamCampbell:
> Are these different numbers on purpose?

Ping @darryldecode.

darryldecode:
@franzliedke I believe they decided to use non-privileged ports for host machines to avoid conflicts.

PR #3218, opened 2015-01-18
copy .env file as after create project

hannesvdvreken:
Will work on any os.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
 			"php artisan optimize"
 		],
 		"post-create-project-cmd": [
+			"php -r \"copy('.env.example', '.env');\"",
 			"php artisan key:generate"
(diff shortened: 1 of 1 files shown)


rowanparker:
What if people don't want to use the environment file? I assume it's shipped as an example file deliberately, so that it isn't automatically used.

hannesvdvreken:
Oh, I thought it was shipped as an example file because you're not supposed to check your env file in your VCS.

GrahamCampbell:
Oh, it's shipped like that for the reason @hannesvdvreken said. You're meant to be using it.

rowanparker:
Ah ok, I understand. My apologies.

hannesvdvreken:
np

PR #3220, opened 2015-01-21
PHPSpec produces PSR-2 classes, modify template to produce Laravel-style Classes

chrispecoraro:
I would like to add in a template for the class file which is used by PHPSpec to generate the classes.  This formatting will make the class look like Laravel-style classes.   

PHPSpec produces:

```php
<?php

namespace MyCompany;

class Foo 
{
}
```

but to match the rest of the Laravel classes, it should be formatted like this:

```php
<?php namespace MyCompany;

 class Foo {
 }
```

N.B. *not* PSR-2 compatible

### DIFF
--- a/.phpspec/class.tpl
+++ b/.phpspec/class.tpl
@@ -1,0 +1,5 @@
+<?php namespace %namespace%;
+
+ class %name% {
+ 
+}


Anahkiasen:
Isn't Laravel 5 going to be PSR2 anyway?

chrispecoraro:
I'm not sure. I do want PHPSpec to produce classes formatted the same as the rest of the classes in Laravel.  When Taylor makes the final decision closer to launch, this "override" mechanism will be needed in any case IMHO.

GrahamCampbell:
Laravel 5 will be going PSR-2 very soon. I'll be doing it myself. It'll be done immediately after the next 4.2.x release.

chrispecoraro:
@GrahamCampbell will it be:

```php
<?php
namespace MyCompany;

class Foo 
{
}
```
 or 

```php
<?php namespace MyCompany;

class Foo 
{
}
```

GrahamCampbell:
Neither actually. It'll be:

```php
<?php

namespace MyCompany;

class Foo 
{
}
```

ie, the same way symfony do it.

GrahamCampbell:
And the same way it is now...

chrispecoraro:
That makes sense, thanks @GrahamCampbell for your hard work in getting Laravel 5 ready to go.

PR #3223, opened 2015-01-21
Automatic assignment of the command fields

mtvs:
# Automatic assignment of the command fields #

Currently in order to set the fields of a command object you need to write down the fields at least 4 times.

1. Attributes on the object
2. Constructor parameters
3. Passing the arguments for the constructor 
4. Actual assignment in the constructor

It is tedious especially when you need to make changes (adding or removing the fields).

I think if there was a way to declare them once it would be better.

### DIFF
--- a/app/Commands/Command.php
+++ b/app/Commands/Command.php
@@ -2,6 +2,44 @@
 
 abstract class Command {
 
-	//
 
+    /**
+     * list of the command data fields.
+     *
+     * @var array
+     */
+    protected $_fields=[];
+
+    /**
+     * an associative array of the command data fields and their values,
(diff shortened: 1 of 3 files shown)


mtvs:
It can also applied for the events too.

GrahamCampbell:
All your cs is wrong, and this defeats the object of the command classes. They are meant to be designed to explicitly require certain variables.

franzliedke:
@mtvs May the [`dispatchFrom()` method](https://github.com/laravel/framework/blob/bfd01c401f632d5701fee171c24d2b515527a176/src/Illuminate/Bus/Dispatcher.php#L71-82) be what you need?

PR #3230, opened 2015-01-25
[5.0] Set database.prefix based on env config

mtvs:
DB tables prefix should be an environment variable.

For example, consider you have a shared host with max 1 database limitation that you used for hosting your projects, so you need to set DB prefix while on your local development environment you don't need it.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -60,7 +60,7 @@ return [
 			'password'  => env('DB_PASSWORD', ''),
 			'charset'   => 'utf8',
 			'collation' => 'utf8_unicode_ci',
-			'prefix'    => '',
+			'prefix'    => env('DB_PREFIX', ''),
 			'strict'    => false,
 		],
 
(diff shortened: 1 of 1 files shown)


juukie14:
For me this would not be usefull. I think the most important things should be env()'ed. People can easily modify it themselfs if needed.

mtvs:
@juukie14 
I think *consistency* is important so when you set one environment variable using `env()` you should do the same about the rest too.

juukie14:
I don't agree with that because, in my opinion, there are many settings that are the same on every environment and some I wouldn't want to be controlled via the env().
But if more people feel like your wat, I'm fine and I'll adjust to what most users like.

juukie14:
wat = way

GrahamCampbell:
Just do this in your own app. I suspect this is a less common use case. We can't go through changing every config setting like this.

willishq:
Agreed, setting every config by .env is not ideal, and this Change would be heading that way.

PR #3233, opened 2015-01-27
Use stable dependencies

GrahamCampbell:


eduard44:
So does this mean that Composer would get stable dependencies for everything in the project except for laravel/framework dependencies?

Anahkiasen:
:+1:

GrahamCampbell:
> So does this mean that Composer would get stable dependencies for everything in the project except for laravel/framework dependencies?

No, this means composer will get stable for everything apart from laravel/framework itself. This means people will be running with stable symfony compoents etc.

eduard44:
Nice :+1:

deweller:
Doesn't adding ```@dev ``` mean that laravel will use UNstable dependencies by default?  

Forgive my misunderstanding here.

GrahamCampbell:
No. `'@dev` means get unstable laravel/framework only. The other meant get all dependencies dev.

PR #3236, opened 2015-01-30
Hash password in model

JosephSilber:
1. It feels safer this way. People won't accidentally store plain text passwords.
2. It's nice to introduce people to mutators out of the box.

### DIFF
--- a/app/Services/Registrar.php
+++ b/app/Services/Registrar.php
@@ -32,7 +32,7 @@ class Registrar implements RegistrarContract {
 		return User::create([
 			'name' => $data['name'],
 			'email' => $data['email'],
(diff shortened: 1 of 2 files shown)


nateajohnson:
Doing this does sound better. I actually changed my code to work exactly this way and then ran into problems with phpspec testing because it couldn't find the bcrypt function. I ended up reverting rather than mocking the function. Something to think about.

GrahamCampbell:
:+1:

rtablada:
Heads up that this will require a rewrite for the password reset functionality in framework (IMO the side effect seemed out of place to begin with though).

Thumbs up here!

PR #3234, opened 2015-01-28
Use no protocol to fetch Lato font

DavidLemayian:
When using https://example.com, the welcome page design looks off due to fetching the Lato font on http://.

Ref: https://github.com/laravel/framework/issues/7159

**Update:** Using no protocol i.e `//`

PR #3237, opened 2015-01-30
Use protocol relative link to fonts.googleapis.com

freekmurze:
Making this link protocol relative will make it work for both http and https


dwightwatson:
Already covered by #3234

PR #3238, opened 2015-01-31
Check environment for configuration defaults

croemmich:
I have updated the configuration files to first check the environment for items that are commonly environment dependent. I think most users on 12 factor providers would welcome this unless there is a significant impact on performance.


GrahamCampbell:
This is overkill. People can change this to what they want in their own app.

croemmich:
So why did you provide the env check for the database connection? Laravel is all about setting logical defaults. I hate having to change these in nearly every project I do and I don't really want to maintain a fork just to provide this. Since the removal of folder support from the config system, many of these things need to be set in the environment file.

juukie14:
STRIP_MODEL and STRIP_SECRET sound like things everone would like

croemmich:
@juukie14 Fixed my poor spelling

garygreen:
I think I would like to see one for `url` as this does change almost every time between local/production. The rest I would rather have changed in user's own app.

GrahamCampbell:
Way too much has changed here.

config/app.php
```diff
@@ -39,7 +39,7 @@
 	|
 	*/
 
-	'timezone' => 'UTC',
+	'timezone' => env('APP_TIMEZONE', 'UTC'),
```
GrahamCampbell: no

config/app.php
```diff
@@ -52,7 +52,7 @@
 	|
 	*/
 
-	'locale' => 'en',
+	'locale' => env('APP_LOCALE', 'en'),
```
GrahamCampbell: no

config/app.php
```diff
@@ -65,7 +65,7 @@
 	|
 	*/
 
-	'fallback_locale' => 'en',
+	'fallback_locale' => env('APP_FALLBACK_LOCALE', 'en'),
```
GrahamCampbell: no

config/app.php
```diff
@@ -95,7 +95,7 @@
 	|
 	*/
 
-	'log' => 'daily',
+	'log' => env('APP_LOG', 'daily'),
```
GrahamCampbell: no

config/cache.php
```diff
@@ -44,14 +44,14 @@
 
 		'file' => [
 			'driver' => 'file',
-			'path'   => storage_path().'/framework/cache',
+			'path'   => env('CACHE_FILE_PATH', storage_path().'/framework/cache'),
```
GrahamCampbell: no

RomainLanz:
if (false === @GrahamCampbell.say())
    return 'no';

GrahamCampbell:
Maybe send a much more stripped down version of this, and it'll get considered.

PR #3239, opened 2015-01-31
[5.0] Adding down file to gitignore

martiros:


baskan:
just curious; whats this file do?

martiros:
Laravel use this file(file always empty) to check is application in maintenance mode.

Source from Illuminate\Foundation\Application class

	/**
	 * Determine if the application is currently down for maintenance.
	 *
	 * @return bool
	 */
	public function isDownForMaintenance()
	{
		return file_exists($this->storagePath().'/framework/down');
	}

If you call "php artisan down", Laravel generate this empty file and will remove it, if you call "php artisan up".
I think no need store this file by default in VCS.

pespantelis:
I think it is not necessary to do this, however if it is, you have to add to [this gitignore file](https://github.com/laravel/laravel/blob/develop/storage/framework/.gitignore).

martiros:
I added in wrong place ?
https://github.com/martiros/laravel/blob/develop/storage/framework/.gitignore

martiros:
I think it must be here by default, because maintenance mode in general for production.
And in most cases my team members not needed in this file.

pespantelis:
Oups, I thought it was the gitignore of root. Sorry @martiros !

martindilling:
I can understand why you don't want that file in git. But I can't see how that should happen? Wouldn't that happen only if you commit and push from your production site?

martiros:
It can happen for example in case if developer implementing "maintenance mode" template, and in his development machine application in maintenance mode.

PR #3241, opened 2015-02-03
Make default middlewares implement contract

HipsterJazzbo:
See also laravel/framework#7246


GrahamCampbell:
This isn't required.

PR #3242, opened 2015-02-04
resolve phpmd static access warning

le6o:
PHPMD raises a [Static Access warning](https://github.com/phpmd/phpmd/blob/5d06695148d2203f56b27a5205ab87ce6a594a22/src/main/resources/rulesets/cleancode.xml#L69) for this line. It's only a problem for people using PHPMD, so appreciate the change could be made individually.

That said, if this is the only use case for the simpleMapping method then it could be changed to a non-static call within the framework.

### DIFF
--- a/app/Providers/BusServiceProvider.php
+++ b/app/Providers/BusServiceProvider.php
@@ -13,9 +13,9 @@ class BusServiceProvider extends ServiceProvider {
 	 */
 	public function boot(Dispatcher $dispatcher)
 	{
-		$dispatcher->mapUsing(function($command)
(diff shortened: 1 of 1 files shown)


taylorotwell:
It's a static method?

GrahamCampbell:
@taylorotwell You should not call static methods like that on a instance of a class. You should always use `->` even though the method is infact static. What's worse is this will totally break in php 7. Support for such function calls was totally removed, rather than just throwing a warning.

GrahamCampbell:
Wait, no. We're statically calling it, NOT ON AN INSTANCE, so this is ok.

PR #3244, opened 2015-02-04
Prevent committing down file

michellaurent:
Don't commit the down file created when the application is in maintenance mode (php artisan down)


GrahamCampbell:
Duplicate.

PR #3246, opened 2015-02-05
Remove executable status from queue config.

Garbee:
The filemode was changed to the queue config a [while back](https://github.com/laravel/laravel/commit/3e29d067087f48f395b7f41e2c35b6cf4175a088#diff-232407f37695671c6ce1016d182c8828) for no reason. A minor thing, but still should be consistent with every other file that doesn't need to be executed like artisan.


GrahamCampbell:
@Garbee Aren't the rest of the files in here 755 though? I brought this up a year ago, and Taylor seemed to want them all to be 755.

Garbee:
7 anywhere gives executable status which isn't needed. And the rest of the files are 644 so this is in alignment with every other config file.

Let me go dig around and see if I can find an old issue about this.

Garbee:
Oh issues are entirely gone... forgot about that. The only other thing I see about this is another PR doing the same thing about 20 days ago in which you said they are all 755 which isn't true at all. Files should be 644 as a good default and folders 755.

GrahamCampbell:
> they are all 755 which isn't true at all

They're all supposed to be. They've just all been accidentally changed over time.

Garbee:
All I see is *this one* being accidentally changed away from 644. I just asked Taylor in #laravel-dev about the permissions to see what he says since there is no record here of it afaik.

GrahamCampbell:
My personal opinion is we should use 644 for files unless we need them to be executable.

PR #3248, opened 2015-02-07
Load Glyphicons font from the BootstrapCDN

aFreshMelon:
This allows for out the box use of Glyphicons.

This is the cleanest approach in my opinion.


wilsonpinto360:
This will use only bootstrap version 3.3.2, i guess.

andriesinfoserv:
please don't load anything from anyone's cdn in laravel by default.

aFreshMelon:
@andriesinfoserv 

You realize that is already happening in the current version?

https://github.com/laravel/laravel/blob/master/resources/views/app.blade.php#L59

andriesinfoserv:
:) i was afraid of that. nevermind, no big deal.

PR #3249, opened 2015-02-07
Add url path on app.blade.php

tvalentius:


GrahamCampbell:
Duplicate. Already rejected, sorry.

tvalentius:
why not add the url path on the app.blade.php?

hfalucas:
Maybe because L5 doesn't ship anymore with the "illuminate/html" package?

GrahamCampbell:
@hfalucas The URL facade calls a class in the routing package under the hood. Nothing to do with html.

PR #3252, opened 2015-02-08
Use "for" attribute on labels in auth views

marcus-herrmann:
Added "for" attribute in auth related example views for both usability/accessibility reasons, and testability - regarding Codeception's `fillField` method, which will not work without the relation set by "for" (at least when the input is not nested within its label. Using "for" attribute is the cleaner method, imo).

### DIFF
--- a/resources/views/auth/login.blade.php
+++ b/resources/views/auth/login.blade.php
@@ -22,16 +22,16 @@
 						<input type="hidden" name="_token" value="{{ csrf_token() }}">
 
 						<div class="form-group">
-							<label class="col-md-4 control-label">E-Mail Address</label>
+							<label for="email" class="col-md-4 control-label">E-Mail Address</label>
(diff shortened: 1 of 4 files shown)


bbashy:
80% of people don't understand what it does or provides to accessibility, had issues with this before in Laravel. Someone even did a PR to link the "for" attribute to the name one :')

marcus-herrmann:
Even if you set the accessibility aside, there's still the testing factor left. Since you have to explicitly add `Illuminate\Html\HtmlServiceProvider` including the Form Facade now (and `Form::input` dealt with the "for" relation) I think it would be appropriate to include this to the Auth views - which can be understood by some users as examples. Imagine some of them not so frontend-savy copy pasting these into other places, and wondering why their functional test fail :)

taylorotwell:
Wrong branch. Would go on master branch.

PR #3259, opened 2015-02-12
Make pretty URLs and default views work when using an Apache alias

jotaelesalinas:
To make pretty URLS work when using an Apache alias:
 - Added a `RewriteBase` directive to `public/.htaccess` (commented out by default)

To make links and forms in the default views work:
- Changed hardcoded absolute URLs for their relative counterparts, using `url()`.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -5,8 +5,8 @@
 
     RewriteEngine On
 
-	# Uncomment and set your actual path if you run your application using an alias
-	# RewriteBase /your_aliased_path
+    # Uncomment and set your actual path if you run your application using an alias
+    # RewriteBase /your_aliased_path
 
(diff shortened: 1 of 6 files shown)


franzliedke:
+1 for using `url()`.

public/.htaccess
```diff
@@ -5,6 +5,9 @@
 
     RewriteEngine On
 
+	# Uncomment and set your actual path if you run your application using an alias
```
GrahamCampbell: incorrect indentation

jotaelesalinas: Fixed.

GrahamCampbell:
Needs to be sent to the master.

jotaelesalinas:
Ok, thanks. Pull request #3263 created in master.

PR #3263, opened 2015-02-13
Fixes for apps in subfolders

jotaelesalinas:
To make pretty URLS work when using an Apache alias:
- Added a `RewriteBase` directive to `public/.htaccess` (commented out
by default)

To make links and forms in the default views work:
- Changed hardcoded absolute URLs for their relative counterparts, using
`url()`.

### DIFF
--- a/resources/views/app.blade.php
+++ b/resources/views/app.blade.php
@@ -6,7 +6,7 @@
 	<meta name="viewport" content="width=device-width, initial-scale=1">
 	<title>Laravel</title>
 
-	<link href="/css/app.css" rel="stylesheet">
+	<link href="{{ URL::asset('css/app.css') }}" rel="stylesheet">
 
 	<!-- Fonts -->
 	<link href='//fonts.googleapis.com/css?family=Roboto:400,300' rel='stylesheet' type='text/css'>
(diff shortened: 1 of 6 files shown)


AtroxDev:
:+1:

resources/views/app.blade.php
```diff
@@ -6,7 +6,7 @@
 	<meta name="viewport" content="width=device-width, initial-scale=1">
 	<title>Laravel</title>
 
-	<link href="/css/app.css" rel="stylesheet">
+	<link href="{{ URL::asset('css/app.css') }}" rel="stylesheet">
```
AtroxDev: shouldn't this be: ```{{ asset('css/app.css') }}```

GrahamCampbell: Doesn't matter.

AtroxDev: Yeah but isn't it cleaner without URL::? I'm just asking :)

taylorotwell:
Laravel is meant to run in your document root as specified by your web server. I've never really liked the whole sub-folder business.

franzliedke:
@taylorotwell But why? It's really easy to do, and having an app in a subfolder is a legitimate, and wide-spread use case.

PR #3265, opened 2015-02-14
Have logical default for Queue Driver when testing

TheShiftExchange:
We should set the queue driver to be `sync` when testing by default. Makes sense along side the other testing defaults like cache and session etc

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -18,5 +18,6 @@
         <env name="APP_ENV" value="testing"/>
         <env name="CACHE_DRIVER" value="array"/>
         <env name="SESSION_DRIVER" value="array"/>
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I assume this wasn't done because the default is already sync?

TheShiftExchange:
Mmmm...maybe - but if people change the default (because they want Redis on all environments for example) - then they need to add this line.

Just a tweak either way - no biggie...

GrahamCampbell:
People should set their queue driver in the env file though?

TheShiftExchange:
Not necessarily.

I only use `.env` for configs that are *different* between my environments. If I plan on using a Redis queue on all my environments, including development - then I set it in the config itself.

GrahamCampbell:
:+1:

PR #3266, opened 2015-02-15
Update composer.json

mod3st:
Min stability "dev" is required according to the dev version of Symfony 2.7 (ex symfony/console)


GrahamCampbell:
Better fix: #3267.

NB, you shouldn't be using the develop branch right now. Laravel 5.1 is far from ready.

PR #3268, opened 2015-02-17
Make Trailing Slashes work with subdirectory.

rentalhost:
If you are working in a subdirectory like `http://localhost/projects/laravelhere/route` and put a Trailing Slashes, it redirect to `http://localhost/laravelhere/route`, excluding the subdirectory.

This PR solve this, allowing you work inside subdirectory, if you want.


taylorotwell:
Would prefer to maybe put this in the documentation instead.

PR #3269, opened 2015-02-17
[5.0] Set mail config values in .env file.

martinbean:
Make the mail configuration values look in **.env** file by default.


.env.example
```diff
@@ -8,4 +8,5 @@ DB_USERNAME=homestead
 DB_PASSWORD=secret
 
 CACHE_DRIVER=file
+MAIL_DRIVER=file
```
crynobone: file mail driver?

GrahamCampbell: lol

martinbean: @crynobone @GrahamCampbell Yeah, that’s a typo :blush:

Fixed! Thanks for pointing that out.

TheShiftExchange:
I like this PR. But I think the naming convention is wrong.

Instead of names like `SMTP_USERNAME` - it should be `MAIL_USERNAME` - because you might not be using an SMTP driver.

martinbean:
@theshiftexchange Thanks for the feedback. I’m named them as such as the comment blocks above each variable explicitly mention they’re the SMTP host/port/username/password configuration values.

Happy to rename them if these credentials are used in a non-SMTP instances, but can’t think of any of those instances off the top of my head.

TheShiftExchange:
No problems - its just a thought.

What about since you do not have a default username+password in the config `env()` - these should also probably be included in the default `.env` file since you are expected to set them?

PR #3273, opened 2015-02-19
Glyphicons are absent from Bootstrap

theravel:
Laravel uses Bootstrap styles, which use font-family "Glyphicons Halflings".
This font is supposed to be loaded from "../fonts" (relatively to css), see links [here](https://github.com/laravel/laravel/blob/master/public/css/app.css#L253). Actually the whole fonts directory is absent, that's why I cannot do fresh Laravel installation and use Glyphicons.

Are Glyphicons allowed to be included to Laravel by [license](http://getbootstrap.com/components/#glyphicons)?


rossedman:
I may be wrong and @GrahamCampbell may correct me, but I believe they were trying to keep the styles included as minimal as possible, this would include extra bloat like fonts. You can always setup a quick bower install and pull down the full Bootstrap and begin compiling.

To me it is easier if you want to Bootstrap to pull the whole thing down rather than have thousands of people who don't want it having to delete it everytime they do an install.

theravel:
@rossedman maybe you are right, but currently CSS file contains broken links. Is this good?

juukie14:
Sorry but I really don't like the font files to be added. Imo, it has enough boilerplate files already.

hydrarulz:
It shouldn't be broken. Either provide the font files or remove the CSS. Modifying the CSS is a hackish solution, I prefer adding the font files to complete the application.

PR #3275, opened 2015-02-21
[Fix] Css file not loading when extending app template

RajeebTheGreat:
Extending app.blade.php didn't fetch the CSS files. It's a small fix for that.


taylorotwell:
Hmm, that doesn't really make sense. The CSS path will be relative to public directory.

franzliedke:
But it breaks when the app is installed in a subfolder.

PR #3277, opened 2015-02-23
Use "for" attribute on labels in auth views (reopened on master)

marcus-herrmann:
Like #3252, but on master.

> Added "for" attribute in auth related example views for both usability/accessibility reasons, and testability - regarding Codeception's fillField method, which will not work without the relation set by "for" (at least when the input is not nested within its label. Using "for" attribute is the cleaner method, imo).


taylorotwell:
Really really sorry to do this, but can you send this back to develop. I screwed up.

PR #3284, opened 2015-02-26
Update 503.blade.php

sebdesign:
Added missing DOCTYPE and charset declarations.
The html element as well as the body element need to fill the viewport so the .container can be vertically centered.


GrahamCampbell:
Send to the master.

PR #3287, opened 2015-02-26
Update 503.blade.php

sebdesign:
Added missing DOCTYPE and charset declarations.
Added missing title element.
The html element as well as the body element need to fill the viewport
so the .container can be vertically centered.


sebdesign:
Anything wrong with this?

PR #3289, opened 2015-02-27
Abstract base command class should extend Illuminate core for OOB artisan support

j42:


crynobone:
App\Commands isn't exclusive for console, unlike App\Console

PR #3291, opened 2015-03-01
add RESTfulController

secnelis:
This is a RESTful style controller I use in my own projects.

This controller is similar to the resource controller provided by Laravel framework, which supports 7 actions: 

	index/create/store/show/edit/update/destroy

However, this controller is more flexible in defining/extending actions. Some business-specific actions can be defined easily, especially for PUT requests, such as:

	PUT /books/352/rename
	PUT /books/352/star

Usage:

1. Create a controller class extending RESTfulController
2. Define methods for actions, for example:

`
protected function putIDRename($id) { ... }
`

`
protected function putIDStar($id) { ... }
`

`
protected function deleteID($id) { ... }
`

The `ID` part in the method name works as a placeholder, and can be defined to your own pattern, see `getMethodNameIDPlaceHolder` method for more details.

The method name looks similar to the request uri and is thus easy to remember and find.

You can also define custom pattern of the resource ID, see `getIDPattern` and `decodeID` methods for more details.

There are some limitions of this controller.

1. It can work with the route filters but only the "only" type of before/after filters are supported at the moment.
2. The middleware introduced in Laravel 5 is not supported yet.
3. The implementation depends on some implementation details of the route subsystem, which may change in future Laravel versions.

References: 

1. http://codeplanet.io/principles-good-restful-api-design/
2. https://bourgeois.me/rest/
3. http://www.ruanyifeng.com/blog/2014/05/restful_api.html

### DIFF
--- a/app/Http/Controllers/RESTfulController.php
+++ b/app/Http/Controllers/RESTfulController.php
@@ -1,0 +1,193 @@
+<?php namespace App\Http\Controllers;
+
+use App, Exception, Request, Route;
+use Illuminate\Http\Response as HttpResponse;
+use Symfony\Component\HttpFoundation\Response as SymfonyResponse;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+/**
+ * This is a RESTful style controller.
+ *
+ * This controller is similar to the resource controller provided by 
+ * Laravel framework, which supports 7 actions: 
+ *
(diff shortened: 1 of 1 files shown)


overtrue:
FIRST: http://laravel.com/docs/5.0/contributions

crynobone:
Probably more suitable as external package.

PR #3292, opened 2015-03-01
Add resources/lang/en/auth.php to localize 'These credentials do not match our records.'

mattrandallbecker:
Adding a auth.php localization file so the string for failed authentication isn't hard coded.

Related to: https://github.com/laravel/framework/pull/7664


resources/lang/en/auth.php
```diff
@@ -0,0 +1,15 @@
...
+
+    /*
+    |--------------------------------------------------------------------------
+    | Auth Language Lines
+    |--------------------------------------------------------------------------
+    |
+    |
+    */
+
+    'failedLoginMessage' => 'These credentials do not match our records.'
```
overtrue: key should be snake case.

GrahamCampbell:
Send to develop please regarding my comment on your pull to laravel/framework.

PR #3295, opened 2015-03-02
route model bindings should be defined before calling parent boot method

lifesign:


taylorotwell:
I tested it and it doesn't matter which order it is.

PR #3296, opened 2015-03-02
remove https protocols from html5shiv and respond.js

Landish:


taylorotwell:
It's fine as is.

Landish:
Every external asset is loaded without specified protocols. I believe, those two should load in the same way, as others.

PR #3297, opened 2015-03-02
Remove unnecessary return

antonioribeiro:


taylorotwell:
Sometimes I keep them just in case I change something.

PR #3299, opened 2015-03-03
Added DB_CONNECTION to .env and config/database.php, MySQL default

jamieshepherd:
Feel this is a reasonable change as I have had to deploy on a few different drivers with the same software. Had originally named this DB_DRIVER but DB_CONNECTION seems more reasonable considering it pulls from an array of connection types rather than the actual driver (looks like you can have multiple connections with the same driver).


conorsmith:
Is this a common enough use case to be in the .env file for every install of Laravel? Especially given how trivial it is to change your own application to work this way.

I would worry that this could encourage developers to use the .env file for general application configuration instead of just per-environment config.

jamieshepherd:
You could easily say the same about https://github.com/laravel/laravel/commit/4310e0b3eaac71ae998c8a646b40608c6a5465f4 this commit. You can change it in config/mail if you wanted to, but defining the mail driver and settings) here makes sense. In something I'm building right now I've had to switch between SQlite and MySQL for different testing / production environments.

I'd rather throw a new .env file on to each one rather than updating the config - especially considering the .env file is not version controlled - one of the main reasons to use .env in the first place.

conorsmith:
Mail settings should certainly be part of a per-environment configuration. The `MAIL_DRIVER` variable would be analogous with your proposed `DB_CONNECTION` variable.

My gut feeling is that using different emailing setups on different environments would be a more common use case than switching up database eg using `log` for local development instead of the `smtp` server used in production.

However, your point about testing environments makes me reconsider my initial thoughts. I imagine using a different database (eg `sqlite`) for testing is a common enough use case.

Perhaps [phpunit.xml](https://github.com/laravel/laravel/blob/master/phpunit.xml) should also be changed to have a different default (probably `sqlite`)?

crynobone:
> However, your point about testing environments makes me reconsider my initial thoughts. I imagine using a different database (eg `sqlite`) for testing is a common enough use case.

More common for packages development than app development. Beside for unit testing you probably end up using `:memory:` option instead of a `database.sqlite` file.

jamieshepherd:
>More common for packages development than app development. Beside for unit testing you probably end up using :memory: option instead of a database.sqlite file.

That's not true, "testing" isn't just unit testing. Testing can be running a full application with a client, test users, real people, you want them to be running the database from memory?

I also have had a database driver change in the middle of development a few times between sqlsrv, sqlite and mysql. I have never had to change my mail driver, yet this is an environment variable.

Adding this single line harms nobody, but helps a small to moderate amount of people, and has absolutely no negatives.

crynobone:
> I also have had a database driver change in the middle of development a few times between sqlsrv, sqlite and mysql.

Some of not all app would have requirement to use database specific query, e.g `enum` with MySQL. To them, there no use case at all to swap between any database driver. Beside this has been PRed before and rejected.

conorsmith:
> I also have had a database driver change in the middle of development a few times between sqlsrv, sqlite and mysql.

Is this a common situation though? Would you not agree that a large majority of Laravel applications use the same database driver across all environments?

> Adding this single line harms nobody, but helps a small to moderate amount of people, and has absolutely no negatives.

As I said above, I think there is potential harm here. I think this has the potential, especially given that it is database-related, to confuse developers about what the .env file should be configuring.

It will save a trivial amount of effort for the "small to moderate amount" of developers who do need to change database driver across different environments. They simply have to change the two lines this PR changes.

I argue that this use case doesn't warrant the trade off of clarity of the .env file. The testing use case may or may not be common enough to warrant it, I don't know.

jamieshepherd:
I really don't care as much as apparently you guys do about a 1 line change to an environment *configuration* file, but you keep saying your reason for not wanting it is because the driver might not change throughout development. That's fine, in that case why did you not throw a tantrum about https://github.com/laravel/laravel/commit/4310e0b3eaac71ae998c8a646b40608c6a5465f4 this commit, it does literally the same thing, if I want to change the mail driver I can do that in 1 line too. It's hypocritical, and it's ridiculous that you consider it 'potentially harmful' to make configuring your environment easier.

crynobone:
No,  for email it more common to set mailtrap.io with smtp for local, or log, or dummy smtp with mailcatcher, for production you probably using mandrill, mailgun or amazon ses. This including it credential should never be part of `mail.php` or `services.php`. In fact most technical person (programmer/designer) shouldn't know the credential of this value.

What you are asking is database connection, which doesn't commonly change between environment because you most likely want to ensure all dev, staging and production works the same (on the same db driver). Unless you are building an off-sheif app or packages that supporting every database option.

conorsmith:
I didn't comment on the PR for the `MAIL_DRIVER` variable simply because I didn't see it while it was open. However, I disagree that it does the same thing. I think the use cases for having an application with different email drivers across environments are more common and, in fact, should be encouraged.

Local environments could use `mail` or `log` drivers, staging environments could use `smtp` for Mailtrap and production could use `mailgun` or `mandrill`.

taylorotwell:
Send it to develop branch.

PR #3301, opened 2015-03-03
Update .htaccess

AlanRezende:
Enable the Authorization Header to be used as it is, because the Symfony component looks for HTTP_AUTHORIZATION


taylorotwell:
It's already documented that you may need to add this depending on your server configuration.

PR #3302, opened 2015-03-04
Update server.php

EliuFlorez:
Replace and by &&


Batmandakh:
https://stackoverflow.com/questions/4511240/php-difference-between-and-and
is it the cause of this replacement, right?

thehamsterofgod:
sensiolabs insight?

PR #3308, opened 2015-03-16
Ignore the temp file created when in maintenance mode

moleculezz:
I've noticed that when I set a Laravel 5 site in maintenance mode. The file `storage/framework/down` gets created. Seems like this should be ignored.

### DIFF
--- a/storage/framework/.gitignore
+++ b/storage/framework/.gitignore
@@ -4,3 +4,4 @@ compiled.php
 services.json
 events.scanned.php
 routes.scanned.php
+down


taylorotwell:
It's removed anyways when maintenance mode is turned off.

EvgeniyN:
@taylorotwell - Yes this file will be removed after app up, ! But:
if Your deployment process based on GIT (like rocketeer) this file will be marked as New file or remote GIT, and git-pull on remote server will be broken (git write something like: commit Your changes before merge). So You put app down, try to fix bugs and deploy again and got an error in this case.
PS: Maybe I incorrectly set up my Rocketeer.
Sorry for my eng :(

vitor-faiante:
@EvgeniyN I believe that if you try and add storage/framework/down to your .gitignore file you should be fine because git would just ignore it and hopefully sort your error.

EvgeniyN:
@vitor-faiante I got it. But why I should search this file in this case (on git base deployment), and what about next Laravel update in which this file will be stored in other place? In other words I don't know where this file will be stored after another one update and it's hard to suggest what Your deployment doesn't work because of temp (phantom) framework file. So I agree with @moleculezz - add this stuff to default .gitignore

vitor-faiante:
@EvgeniyN Exactly I do agree as well and agree more with his approach of just leaving the down file alone in the .gitignore instead of what I said which was the full path, exactly due to the fact that it could change the place it is created in a next update. Therefore makes more sense to have the file alone in the default .gitignore

EvgeniyN:
@vitor-faiante Now I understand Your idea. Put this ignore to framework root .gitignore with full path to 'down' file, right? Maybe I miss something - as I wrote, my English is not very good :(

vitor-faiante:
@EvgeniyN Yes that is exactly the idea. But if you put only the file down instead of the full path to the file down in the root .gitignore then in theory it should protect you in the future in case the down file changes place in a next update.

EvgeniyN:
@vitor-faiante It's great what we understand each other, now will hope @taylorotwell will implement this stuff :)

vitor-faiante:
@EvgeniyN Yes, that would be good :)
Did you try adding down to your .gitignore file to see if the error you normally get is solved?

EvgeniyN:
@vitor-faiante Yep. In other way my deployer doesn't work when Laravel under maintenance mode :)
This problem also occurs with any other files which create my application, if they not in .gitignore, but it's a completely different story and is it not related to Laravel

paulrose:
Maintenance flag control should be down to the developer or repository, not the framework. Ignoring the flag shouldn't be at framework level. People are welcome to change their own .gitignore's as required.

PR #3309, opened 2015-03-18
Fix infinite redirect when accessing a folder

sorora:
Without the RewriteCond, accessing folders causes an infinite redirect instead of the relevant page to be displayed (e.g. a 403 server page).

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -6,6 +6,7 @@
     RewriteEngine On
 
     # Redirect Trailing Slashes...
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Duplicate.

sorora:
Duplicate of one denied? I've looked through previous requests and one said it had been "merged" but the that was only the tabs to spaces part, not the issue with the directories. This stops directories being able to be listed (if that's wanted) or the correct error pages displaying.

PR #3310, opened 2015-03-18
Converted $compiledPath var to snake_case for App consistency.

maparker21:


taiyaei:
Just a question, but how is this for consistency?  Do you have some examples of other places where it's snake_case?  I can't seem to find any.

maparker21:
After further investigation I can see camelCase is used in most if not all variable instances. I believe I had this in mind because I was just reviewing the config areas of the app, where the array key names are all snake case.

RomainLanz:
@maparker21 You can just close this PR. :)

PR #3314, opened 2015-03-21
Fix indent format.

taiyaei:
This converts AppServiceProvider to spaces since it wasn't and removes
a few empty lines after class declarations.

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -2,26 +2,25 @@
 
 use Illuminate\Support\ServiceProvider;
 
-class AppServiceProvider extends ServiceProvider {
+class AppServiceProvider extends ServiceProvider
+{
+    /**
+     * Bootstrap any application services.
+     *
+     * @return void
+     */
+    public function boot()
+    {
+        //
+    }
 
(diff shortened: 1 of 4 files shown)


OzanKurt:
Doesn't laravel use tabs?

yasinkocak:
2.4. Indenting
Code MUST use an indent of 4 spaces, and MUST NOT use tabs for indenting.

N.b.: Using only spaces, and not mixing spaces with tabs, helps to avoid problems with diffs, patches, history, and annotations. The use of spaces also makes it easy to insert fine-grained sub-indentation for inter-line alignment.

http://www.php-fig.org/psr/psr-2/

GrahamCampbell:
We are not psr2 yet.

pespantelis:
>Laravel follows the `PSR-4` and `PSR-1` coding standards. In addition to these standards, the following coding standards should be followed:

>- The class namespace declaration must be on the same line as `<?php`.
- A class opening `{` must be on the same line as the class name.
- Functions and control structures must use Allman style braces.
- Indent with tabs, align with spaces.

http://laravel.com/docs/master/contributions#coding-style

PR #3322, opened 2015-03-28
add terminate (exit)

ytake:
In my case,  
```bash
$ siege -b -c 10 -t 3S http://example.com
``` 
150%~ performance increase(Use artisan config:cache, route:cache, optimize --force commands.)
is there a better way?

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -55,3 +55,5 @@ $response = $kernel->handle(
 $response->send();
 
 $kernel->terminate($request, $response);
+
+exit(0);


GrahamCampbell:
Hmmm, is zero always the correct exit code?

GrahamCampbell:
Hmmm, Symfony doesn't exit like this: https://github.com/symfony/symfony-standard/blob/2.7/web/app.php. There must be a reason?

taiyaei:
I haven't ever really looked into what `exit` really does vs. just letting PHP end itself, but for some reason I have a feeling this would lead to some sort of issue/s or another.

I find it strange though that this would lead to such a performance increase. Anyone have an explanation for why that would be? Can't seem to find anything explaining it.

GrahamCampbell:
php docs say:

``
It should be noted that if building a site that runs on FastCGI, calling exit will generate an error in the server's log file. This can quickly fill up.
```

HiroKws:
I'll help ytaka said, the pointed out some situation in Japanese chats.

ytake had load tested a lot of PHP framewokrs in his environmets. (Mac and Ubuntu)
And he found only Laravel's responses were bumping in speed, like 0.06 - 0.42sec. Then totally it became slower than other frameworks. While other freamworks showed equally responses.

So, he thought this came from something unreleased memory problem. 
Then he tried some ways to resolve this issue, and found this solution.
In the result this, Laravel became faster like Taylor said. (Maybe "a bit faster then cake")

As GrahamCampbell pointed out, this will make another problem, so it is OK to close. However I think it is better to notice, because sometime Laravel user said 'slow', we answered 'optimize it', but heard an answer 'still slow', then we thought 'aha, maybe he didn't optimize correctly...'. However, there are possible to exist another hidding slowness reason he suspected.

PR #3324, opened 2015-03-28
Ported from LESS to SASS.

verheesj:
Bootstrap has ported over from LESS to SASS.


RomainLanz:
Why?

Bootstrap has been ported to SASS 2 years ago.

verheesj:
As of 3.3.2/3.3.4 - Bootstrap natively ships with SASS rather than LESS.

verheesj:
Totally mis-read the updates. my apologies.

PR #3326, opened 2015-03-29
add config parameter for mysql unix socket

yangqi:
spent quite some time to figure out how to connect to mysql through unix socket in laravel, hope this would help

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,6 +3,7 @@ APP_DEBUG=true
 APP_KEY=SomeRandomString
 
 DB_HOST=localhost
+DB_SOCKET=/path/to/mysql.sock
 DB_DATABASE=homestead
 DB_USERNAME=homestead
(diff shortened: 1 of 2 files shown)


verheesj:
Is this config param implemented in Laravel core?

GrahamCampbell:
This should just be added to our docs, but I think it already is?

yangqi:
can you point me where in the docs, I couldn't find it.

GrahamCampbell:
> can you point me where in the docs, I couldn't find it.

It's likely it needs adding then. Could you send a pull please.

yangqi:
Sure, will do.

lury:
actually the same is true for 'port' - it doesn't exist in default configuration and .env file but it should
```			
'port'		=> env('DB_PORT', 3306),
```

PR #3327, opened 2015-03-29
Add .navbar-static-top class to navbar

sw-double:
Navbar should have `.navbar-static-top` class applied

### DIFF
--- a/resources/views/app.blade.php
+++ b/resources/views/app.blade.php
@@ -19,7 +19,7 @@
 	<![endif]-->
 </head>
 <body>
-	<nav class="navbar navbar-default">
+	<nav class="navbar navbar-default navbar-static-top">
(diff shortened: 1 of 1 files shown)


taylorotwell:
What does this do?

sw-double:
Just removes some borders and border-radius.
Intended to be applied for navbars 100% wide positioned at the very top of the page.
https://github.com/twbs/bootstrap/blob/v3.3.4/less/navbar.less#L128

The difference is subtle on default navbar, but quite noticeable on inverted one.

navbar

danielboendergaard:
:+1:

taylorotwell:
Since all of this stuff is opt-in in 5.1 which releases in just about 6 weeks, let's just fix it there.

PR #3330, opened 2015-04-01
Change @endsection to @stop

natzim:
The documentation says to use `@stop` rather than `@endsection`, so you might as well reflect that in the in-built views.


natzim:
Closed because I didn't follow the contribution guide

GrahamCampbell:
?

PR #3331, opened 2015-04-01
Change @endsection to @stop

natzim:
Closed previous request because I forgot to read the contribution guide.

Just a small fix, as the documentation says to use `@stop` rather than `@endsection`.

### DIFF
--- a/resources/views/auth/login.blade.php
+++ b/resources/views/auth/login.blade.php
@@ -58,4 +58,4 @@
 		</div>
 	</div>
 </div>
-@endsection
+@stop
--- a/resources/views/auth/password.blade.php
+++ b/resources/views/auth/password.blade.php
@@ -47,4 +47,4 @@
 		</div>
 	</div>
 </div>
-@endsection
+@stop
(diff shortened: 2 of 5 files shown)


taiyaei:
What's different between this and #3330?

natzim:
Different branch name.

I'm not used to contributing to open source, so feel free to tell me if I'm doing something wrong.

GrahamCampbell:
If you read the source code, you'll see that these two statements are in fact identical.

GrahamCampbell:
> Different branch name.

Both pulls are targeted at the master.

natzim:
I get that, but why not stick to what is in the documentation?

taiyaei:
@GrahamCampbell I think he meant that he wanted his branch name different than `master`. It's not in the contribution documentation explicitly from what I can find though

@natzim It's true that `@endsection` is not in the documentation (at least not the templating documentation), it might be a good idea to PR to laravel/docs to add it there though

natzim:
@taiyaei It was my understanding that `@stop` had replaced `@endsection` a while ago, so surely the inbuilt views should be using the more up-to-date syntax?

PR #3332, opened 2015-04-02
[5.0] Use $errors->any() instead of count($errors) > 0

Landish:


GrahamCampbell:
If it ain't broke, don't fix it. The current code is just fine.

franzliedke:
I agree with @Landish that the default skeleton should promote best practices.

GrahamCampbell:
Depends what you consider a best practice.

franzliedke:
Well, since Laravel boasts how expressive it is right on its homepage, I know what I would choose. ;)

Landish:
@GrahamCampbell  So, you're saying that, this `count($errors) > 0` is better than this `$errors->any()`?

> If it ain't broke, don't fix it.

I don't agree with you on this, but whatever. I tried to make my favorite php framework a *little* more better, but...

GrahamCampbell:
> So, you're saying that, this count($errors) > 0 is better than this 

No, I'm saying it's not worth arguing over.

PR #3334, opened 2015-04-02
Correctly deal with the compiled file

GrahamCampbell:


bootstrap/autoload.php
```diff
@@ -27,9 +27,11 @@
 |
 */
 
-$compiledPath = __DIR__.'/../vendor/compiled.php';
-
-if (file_exists($compiledPath))
+if (file_exists($compiledPath = __DIR__.'/../vendor/compiled.php';))
```
taiyaei: This shouldn't compile, extra `;`

bootstrap/autoload.php
```diff
@@ -27,9 +27,11 @@
...
+{
+	require $compiledPath;
+}
+elseif (file_exists($compiledPath = __DIR__.'/../storage/framework/compiled.php';))
```
taiyaei: same

bootstrap/autoload.php
GrahamCampbell: thanks :)

taiyaei:
:+1: for this :)

PR #3337, opened 2015-04-08
run clear-compiled prior to install/update

le6o:
The clear-compiled script should be run prior to install/update to ensure it is run using the same codebase that probably compiled it.

See https://github.com/laravel/framework/issues/8124 and https://github.com/laravel/framework/issues/8127

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,12 +25,16 @@
 		]
 	},
 	"scripts": {
+		"pre-install-cmd": [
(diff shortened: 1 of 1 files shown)


le6o:
You can ignore this as it will fail to run on a fresh install when no vendor directory exists.
Also, Composer's docs recommend not using pre hooks that depend on Composer managed dependencies.

PR #3338, opened 2015-04-10
Defining global patterns before parent boot function call

makque:
Kindly check if my assumption is correct? In relation to this issue, http://stackoverflow.com/questions/28251154/laravel-5-0-dev-defining-global-patterns-is-not-working/29567578#29567578 and https://laracasts.com/discuss/channels/general-discussion/route-global-pattern-in-routeserviceprovider-not-working-in-laravel-5?page=1

Not really an issue, it's just probably creating some confusions for others

or this could be a different issue/bug

thanks!

### DIFF
--- a/app/Providers/RouteServiceProvider.php
+++ b/app/Providers/RouteServiceProvider.php
@@ -22,9 +22,9 @@ class RouteServiceProvider extends ServiceProvider {
 	 */
 	public function boot(Router $router)
 	{
-		parent::boot($router);
-
 		//
+		
+		parent::boot($router);
 	}
 
 	/**


app/Providers/RouteServiceProvider.php
```diff
@@ -22,9 +22,9 @@ class RouteServiceProvider extends ServiceProvider {
 	 */
 	public function boot(Router $router)
 	{
-		parent::boot($router);
-
 		//
+		
+		parent::boot($router);
```
mickaelandrieu: what does this commit change ?

It's almost the *same* code with an useless comment strip and a new empty line. I don't see the relation between this and the pull request description.

RomainLanz: The comment strip is where the dev should put his code ;)

mickaelandrieu: This was not obvious...  thank you for your answer @RomainLanz :)

PR #3339, opened 2015-04-12
Updated middleware comments with the full path to guard contract

scrubmx:
Updated the Authenticate and RedirectIfAuthenticated middleware comments to reflect the full path to the Guard contract.


GrahamCampbell:
Should be sent to master.

PR #3342, opened 2015-04-14
Minor fixes to make phpstorm happy

martinlindhe:


barryvdh:
I think Laravel actually explicitly wants those `@return void` on constructors..

martinlindhe:
@barryvdh: how come? constructors cant do anything except return void

barryvdh:
Don't ask me, ask @GrahamCampbell ;)
I don't get it either and phpStorm is complaining when I need to add them..

martinlindhe:
Ah. Well if it's there for a reason it would be nice to understand it. I keep removing these @return void:s on each new project i create.

GrahamCampbell:
phpstorm is wrong, sorry.

w2pc:
I think we understand that PhpStorm is wrong. I think people are asking why Laravel wants the @return void on constructors.

GrahamCampbell:
Because return void is the correct annotation. A constructor does not return an instance of the object.

PR #3347, opened 2015-04-22
Constructor has not return value

morilog:


brunogaspar:
Hence the usage of `return void`.

This was discussed many times already.

GrahamCampbell:
Please close this.

PR #3349, opened 2015-04-23
Use UTF-8 charset in form explicitly

proshanto:
FormBuilder from [illuminate\html](https://github.com/illuminate/html/blob/v5.0.0/FormBuilder.php#L106) automatically sets accept-charset="UTF-8".


taylorotwell:
Gonna hold off on this because these views are opt-in in 5.1... you would send this to laravel/framework master branch.

PR #3355, opened 2015-05-03
[Proposal] Set `pretend` via env as well

alnutile:
This would allow setting to change on different servers as needed eg CI server, dev, prod, local etc.

Maybe `driver` makes this not a concern?

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -119,6 +119,6 @@ return [
 	|
 	*/
 
-	'pretend' => false,
+	'pretend' => env('MAIL_PRETEND'),
 
 ];


GrahamCampbell:
Seems a bit overkill to add everything. It's pretty easy for people to add this themselves if they want it. It's up to Taylor though.

taylorotwell:
I kinda think it's better to use the `log` driver actually for these situations. I'm kind of wondering if we can deprecate `pretend` mode.

GrahamCampbell:
> I'm kind of wondering if we can deprecate pretend mode.

:+1: for doing that in 5.1. :)

alnutile:
Makes a lot of sense. Pretend seemed kind of redundant to me.

ibrasho:
:+1: for doing that in 5.1

PR #3356, opened 2015-05-04
Fix controller constructors, they must call parent's method

theravel:
I want to be able to add any logic to base controller's constructor, so that it must be executed


lukasgeiter:
Then just do that in your application!? No need to put it in the skeleton app in my opinion.

GrahamCampbell:
@lukasgeiter I disagree. This looks like a good default to me.

GrahamCampbell:
:+1:

GrahamCampbell:
This might need to go into develop instead though. Ping @taylorotwell.

taylorotwell:
You can't include these unless the base controller actually includes a constructor I believe. But, yes, would also go to develop branch.

PR #3358, opened 2015-05-05
Change package.json so changed to elixir will be non-breaking changes

lukeb:
Previous default version was "*"

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,6 @@
   "private": true,
   "devDependencies": {
     "gulp": "^3.8.8",
-    "laravel-elixir": "~1.0.0"
+    "laravel-elixir": "^1.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -2,6 +2,6 @@
   "private": true,
   "devDependencies": {
     "gulp": "^3.8.8",
-    "laravel-elixir": "*"
+    "laravel-elixir": "~1.0.0"
   }
 }


brunowego:
+1

package.json
```diff
@@ -2,6 +2,6 @@
   "private": true,
   "devDependencies": {
     "gulp": "^3.8.8",
-    "laravel-elixir": "*"
+    "laravel-elixir": "~1.0.0"
```
GrahamCampbell: Why not `^1.0.0`?

lukeb: Per http://browsenpm.org/package.json 

^1.0.0 would  >= 1.0.0  and < 2.0.0    (could include breaking changes)
whereas
~1.0.0 would be >= 1.0.0 and < 1.1   (should only be patches, not breaking changes)

"The hope here, is that the package author has followed best-practices and patch versions of the package will NOT be breaking." - taken from the above referenced guide.

GrahamCampbell: "Best practices" would mean minor versions arn't breaking either...

lukeb: This is true. Are you recommending I modify my pull request, then?

GrahamCampbell: Not sure. Ping @taylorotwell and @JeffreyWay regarding how elixer will be versioned.

JeffreyWay: ^1.0.0.

Thanks for the PR.

lukeb: Changed. Ready for merge.

rafaell-lycan:
+1

GrahamCampbell:
:+1:

PR #3359, opened 2015-05-06
Don't return in @return void method

claar:


taiyaei:
Already closed here https://github.com/laravel/laravel/pull/3166 and here https://github.com/laravel/laravel/pull/3276

Taylor's explanation here: https://github.com/laravel/laravel/pull/3297

claar:
@taiyaei Ah, thanks. The reason I submitted this, and I'm guessing the reason this keeps getting submitted over and over, is because PhpStorm highlights returning from a @return void method as a code smell inspection.

Personally, @taylorotwell's explanation makes sense for framework code, but I think it's less defensible in this scaffolding repository, since this code becomes part of the user's app.

GrahamCampbell:
> The reason I submitted this, and I'm guessing the reason this keeps getting submitted over and over, is because PhpStorm highlights returning from a @return void method as a code smell inspection.

PhpStorm is wrong then. `void` just means unspecified, NOT `null`.

lukasgeiter:
@GrahamCampbell I don't think you should have any (public) method with an unspecified return type.

GrahamCampbell:
> I don't think you should have any (public) method with an unspecified return type.

And why should that be the case?

lukasgeiter:
Because I'd like to know what I get from a method. That's what `@return` is for, isn't it?

lukasgeiter:
Oh wait I think I misunderstood you. I thought by `unspecified` you meant that the developer didn't care to specify the return type (or couldn't decide which it should be)
Now I realize you were probably just talking about the fact that `void` is basically "nothing" or undefined...

GrahamCampbell:
I was meaning, it doesn't matter what it returns. `void` means, "don't reply on the returned value", so developers should just discard anything returned from statements marked as void.

lukasgeiter:
Ah I see... That makes sense. It also matches the official definition from [php.net](http://php.net/manual/en/language.pseudo-types.php#language.types.void):

> *void* as a return type means that the return value is useless.

GrahamCampbell:
Yeh, I didn't just make it up for a laugh, lol. :)

lukasgeiter:
Of course you didn't :clap:

engma:
Perhaps this issue is because PHPStorm was developed by Java Developers, cause in Java there is no void method with a return. and its false to specify '@return void' on a method that returns a value.

shin1x1:
@taiyaei Oops! I couldn't find here. I send just same PR #3371. Thanks for your suggestion.

PR #3361, opened 2015-05-06
No “From” env causing the failure to send with Gmail

enhao:
Added env in .env.example and config/mail.php.


brunowego:
+1

tshafer:
Why would this need to be in the environment file, couldn't you just go in and change it in the mail config?

taylorotwell:
You would just change them in the file directly.

PR #3362, opened 2015-05-07
add Sensio security checker

Tjoosten:


composer.json
```diff
@@ -26,10 +27,12 @@
 	},
 	"scripts": {
 		"post-install-cmd": [
+      "php vendor/bin/security-checker security:check --format=simple",
```
overtrue: Indentation

composer.json
```diff
@@ -26,10 +27,12 @@
...
 			"php artisan clear-compiled",
 			"php artisan optimize"
 		],
 		"post-update-cmd": [
+      "php vendor/bin/security-checker security:check --format=simple",
```
overtrue: Indentation

barryvdh:
We could also require it in the framework and add it to the commands, as it's just a Symfony command. And then run it here as `php artisan security:check`. (related: https://github.com/laravel/framework/issues/440)

franzliedke:
What about this one? https://github.com/Roave/SecurityAdvisories

GrahamCampbell:
> https://github.com/Roave/SecurityAdvisories

Would make more sense to me.

PR #3363, opened 2015-05-07
add `subscribe` property

browner12:
the `subscribe` property is yet another way to register subscribers,
but it is not apparent to most users unless they dig through the source
code a little bit. this update to the default `EventServiceProvider`
will help make this option more apparent to the user


GrahamCampbell:
Please send to develop.

PR #3370, opened 2015-05-10
Fixed docblocks type hints.

paulofreitas:
- Also tweaked some docblocks alignment.

### DIFF
--- a/app/Http/Controllers/WelcomeController.php
+++ b/app/Http/Controllers/WelcomeController.php
@@ -26,7 +26,7 @@ class WelcomeController extends Controller {
 	/**
 	 * Show the application welcome screen to the user.
 	 *
-	 * @return Response
+	 * @return \Illuminate\Http\Response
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/WelcomeController.php
```diff
@@ -26,7 +26,7 @@ public function __construct()
 	/**
 	 * Show the application welcome screen to the user.
 	 *
-	 * @return Response
+	 * @return \Illuminate\Http\Response
```
crynobone: This is still misleading, if you do `return View::make('foo');`. it still an instance of `Illuminate\View\View`, if you return `return User::find(1);`, then it an instance of `Illuminate\Database\Eloquent\Model`. The dispatcher is the one that prepare the response to `Illuminate\Http\Response`.

taylorotwell:
I don't typically align doc blocks.

PR #3374, opened 2015-05-12
Add Fanout event broadcasting

cwt137:
Fanout is a real-time, cloud based event system like Pusher. http://fanout.io . It would be nice if Laravel supported this system as well. This pull request goes with laravel/framework#8705


taylorotwell:
You can add support for it in a package using the extend method on the Broadcast manager.

PR #3375, opened 2015-05-14
Added <title> tag

proshanto:
When app is in maintenance mode, without title tag it just shows the URL of the requested page.


GrahamCampbell:
"Laravel" wouldn't be correct though. This is meant to be a starter project for people. We wouldn't want this left in people's apps because they've missed it.

proshanto:
How about title being "Be right back."? Without title it looks ugly to me and it is easy to spot and replace than no title tag.

PR #3378, opened 2015-05-18
Add validation message for string rule

TheShiftExchange:
Turns out there is no validation message for the `string` rule.

Now there is :)


lukasgeiter:
The title is a bit misleading... You added a the message for the string validation rule not the rule itself :wink:

TheShiftExchange:
Thanks - updated :)

PR #3381, opened 2015-05-22
Eloquent Facade Suggestion

antonybudianto:
Since we use "php artisan make:model ModelName" every time we create a model, and it always includes the:
I think the facade is only useful for those who migrating models from L4. 
I suggest we either use Model or Eloquent for the standard.
Just suggestion. Thanks.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -170,7 +170,6 @@ return [
 		'Cookie'    => 'Illuminate\Support\Facades\Cookie',
 		'Crypt'     => 'Illuminate\Support\Facades\Crypt',
 		'DB'        => 'Illuminate\Support\Facades\DB',
-		'Eloquent'  => 'Illuminate\Database\Eloquent\Model',
 		'Event'     => 'Illuminate\Support\Facades\Event',
 		'File'      => 'Illuminate\Support\Facades\File',
 		'Hash'      => 'Illuminate\Support\Facades\Hash',


shehi:
You shouldn't generalize like that. I never use Artisan to create models or anything for that matter (unless I don't know how to of course), mostly because in many cases Laravel's coding-style doesn't work for me, in other cases Artisan creates bunch of useless code for me... not to mention cloning and renaming existing file is much easier in general. Removing an existing, non-breaking feature from the framework never is a clever move - you may choose not to use that facade, but you can't decide for the remaining folks and relieve them off that feature. 

Disclaimer: No pun or offense intended to PR author - my English can sound a bit rough at times. Apologies in advance.

Anahkiasen:
:+1: for that, extending the facade only gives a harder time to autocompletion mechanisms and it doesn't provide any benefit besides saving you three characters.

shehi:
True, and for that we have @barryvdh 's awesome IDE helper. Get some intelligent IDEA-compliant tool and everything work like a charm. And those three characters tremendously increase code readability and hence, product's survival, cost-effectiveness and value as an IT asset.

Anahkiasen:
> True, and for that we have @barryvdh 's awesome IDE helper. Get some intelligent IDEA-compliant tool and everything work like a charm. And those three characters tremendously increase code readability and hence, product's survival, cost-effectiveness and value as an IT asset.

I do use the IDE helper, but while facade makes sense for other contexts (like Route, etc) I just don't think it does in this case. My point is this at least would provide autocompletion *natively* without needing extra tools

shehi:
Then don't use it. Simple. Don't use it! :) You want your project in some other way, by all means, feel free to stick to classes themselves. 

Anyway, I believe I made my point. No need to elaborate any more.

antonybudianto:
Well it's just suggestion, and it's true that I can simply don't use the facade (removing it from app.php), Thanks for the comment.

antonybudianto:
After some considerations, I'll close this PR since the Facade will actually needed by L4 users who migrate to L5 so they don't need to change the Eloquent to Model.

PR #3382, opened 2015-05-22
Add title to 503 error page.

codybutz:
This is just a small (and maybe dumb) change, I know, but just an idea.

### DIFF
--- a/resources/views/errors/503.blade.php
+++ b/resources/views/errors/503.blade.php
@@ -1,5 +1,7 @@
 <html>
 	<head>
+        <title>Be right back.</title>
+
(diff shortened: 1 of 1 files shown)


resources/views/errors/503.blade.php
```diff
@@ -1,5 +1,7 @@
 <html>
 	<head>
+        <title>Be right back.</title>
```
taiyaei: Tabs

taiyaei:
:+1:

resources/views/errors/503.blade.php
mickaelandrieu: "Service unavailable" sounds better (see RFC)

taylorotwell:
Can this be sent to develop branch?

PR #3384, opened 2015-05-22
Added title tag

rafaelfs17:


resources/views/errors/503.blade.php
```diff
@@ -1,5 +1,7 @@
 <html>
     <head>
+        <title>Be right back</title>
```
taylorotwell: Add a period after back please?

rafaelfs17: Something like "Be right back soon"?

PR #3389, opened 2015-05-25
Use of const instead of literal string

jaschweder:
A simple change of use of the literal string for others platforms instead of *nix and mac


GrahamCampbell:
This already works fine on windows.

PR #3392, opened 2015-05-26
Fix occasional clear-compiled failure after update

Fuxy22:
Clear compiled can fail after update it should be run before update.


GrahamCampbell:
We've been meaning to do this for a while. I'll give Taylor a buzz. :)

PR #3393, opened 2015-05-26
Add composer.phar to .gitignore

jovib:
composer.phar be editing on the production server when run "composer selfupdate"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 /vendor
 /node_modules
+composer
+composer.phar
+composer.lock
 .env


RDelorier:
Composer lock should stay under version control to make sure all devs have same dependency versions

jovib:
Hi!! :)
But generate problem when run "git pull" on server...

w2pc:
@RDelorier  is correct composer.lock should not be in gitignore file. That should alway be checked into the repository. 

I agree composer.phar should not be in the repository.

shehi:
composer.phar shouldn't be in any repo. It's a tool, not related to any project. Otherwise, let's also put git binaries into our projects as well, right? :) 

And I don't get why people claim having composer.lock helps devs to have the same dependency versions: why on Earth would someone define wide version ranges in composer.json if they are this paranoiac about versions?! You want version ```1.2.3```, define it so, instead of ```1.*```! I understand ```composer.lock``` is rather a security measure to make sure everyone uses same package versions with invocation of ```composer install``` command (```composer update``` will override the lock file anyway). But that isn't real security either: if someone by mistake ```composer update```s and overwrites the lock file and then commits it into the project, you are done. All package security is gone. In this case, ```composer.lock``` actually should be FORCE committed into a project and ignored with ```.gitignore``` file, so that noone would coincidentally ```git add --update``` and push. Just ```composer install``` once, force add, and commit that - everyone else won't be able to add lock file ever (unless they choose to do so by force adding), which serves the initial purpose of package security.

yasinkocak:
This can be set on your global .gitignore

https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

PR #3395, opened 2015-05-28
Using the path parameter in the path method.

jimmypuckett:
Some places the path was passed in & some it was being concatenated, so just making it all use the path parameter.


ibrasho:
Finally. :heart:

Anahkiasen:
This had been bugging me for longer than I care to admit

proshanto:
:+1:

PR #3396, opened 2015-05-31
[5.1] Reguard models after doing seeds

phaza:
If someone uses seeds in their tests they can be caught by models not being guarded. (I know, I just experienced this myself)

I think it's a good default to reguard after doing the seeding.


GrahamCampbell:
:+1:

lukasgeiter:
Makes sense :+1:

taylorotwell:
Can you send this to develop branch please.

PR #3400, opened 2015-06-04
Disable CSRF protection for auth/*

constb:
# TL;DR

You don't really need CSRF protection on authentication endpoint. It only breaks user experience sometimes and provides no real security.

## Why?

Obviously, upon logging in your request is already authorized using login-password pair, so any extra parameters are redundant. In fact CSRF protection here is even a little harmful. My client, while beta-testing a new project, has been bitten by it multiple times. Consider the following scenario:

1. User logs in into an app, uses it for a while and logs off.
2. Then the work day ends and he goes home, leaving PC on and browser tab with authorization page open.
3. While he's away his session expires and CSRF token becomes invalid.
4. He comes back in the morning, unlocks PC and enters his credentials on auth page.
5. He gets a "Whoops" and calls to tell me that "the login page is buggy" and I need to fix it.

You'd think it would happen so rarely it doesn't need to be fixed, well actually it happened almost every morning back then. It was really annoying.

## Ideas and questions

Obviously, we don't need a CSRF check here. What we actually need is to throttle login attempts to prevent bruteforcing a password. I doubt whether throttling needs to be in a project boilerplate, although maybe it does. Maybe one could use some ``RouteThrottleMiddleware`` available out-of-the-box in the framework.

Another question is whether only login endpoint should skip token checks or if registration and email confirmation endpoints should behave the same way. I think they should but I'm not sure yet.

Suggestions?

### DIFF
--- a/app/Http/Middleware/VerifyCsrfToken.php
+++ b/app/Http/Middleware/VerifyCsrfToken.php
@@ -12,6 +12,7 @@ class VerifyCsrfToken extends BaseVerifier
      * @var array
      */
     protected $except = [
+        'auth/*',
         //
     ];
 }


GrahamCampbell:
:-1:

rtablada:
IMO this isn't a unique problem to login. Any page could be left on a users computer overnight.

Maybe this is my front end bias, but I would say if this is a common problem for your users, make a small js script that knows the session TTL and refreshes the page after that long.

mvdstam:
Agreed with @rtablada. The issue is clear, but the solution isn't correct IMHO, since the underlying problem has nothing to do with CRSF as-is.

constb:
@rtablada right. but for pages that require authorization it would redirect to ``auth/login`` if session is gone instead, I think. because ``auth`` middleware is triggered before CSRF-check.

also working this around by adding extra js-code feels wrong.

the question is really why whould you want to CSRF-protect auth page anyway? what for?

GrahamCampbell:
People can add this if they want it.

rtablada:
The auto page is no different than other forms on the site when looking from the perspective of a server only site.

In terms of actions when a user TTLs while logged in: better UX would state that being greeted with a login screen that has an alert saying "you have been automatically logged out for your security" is preferable to filling out a huge form and then losing your work because your session is invalid and you now have to go and login before repeating what you did.

constb:
@rtablada that makes total sense. the thing is that login page is different. you expect it to always work if login and password are right. of course, you can always refresh the token right before submitting it to server. but then why have a token there at all?

@GrahamCampbell of course they can. but I believe project boilerplate is there to provide good defaults for new apps. this seems like one to me...

rtablada:
@constb I understand your use case in thinking the login to be different however. It is different, in a different way that makes dropping CSRF a very real issue. Behind sweepstakes and possibly newsletter signups, login pages are the most likely page on your application to be attacked via auto posters or CSRF attack. I realize that you recommended this to be handled by throttling, but that is only a partial and more complex solution than CSRF.

In terms of "adding extra js" this really boils down to a single LOC in JS and a two LOC in your main layout.

In blade you can do this:

```
@inject('conf', 'config')
<meta name="js:session_ttl" value="{{$config->get('session.lifetime')}}">
```

Then in js things are as simple as this:

```
(function() { window.setTimeout(function() { window.location.href=window.location.href }, document.querySelector("meta[name='js:session_ttl']").getAttribute('value') * 60 * 1000);})();
```

Broken out into a few lines, that js looks like this (with comments).

```js
(function() {
    // Grabs meta tag based on specific name attribute value
    var metaTag = document.querySelector("meta[name='js:session_ttl']");
    // Grabs ttl in mins from meta tag
    var ttlMins = metaTag.getAttribute('value');
    // Set timout uses miliseconds so this is basic conversion
    var ttlMs = ttlMins * 60 * 1000;

    // Create a new timeout to run
    window.setTimeout(function() {
        // Refresh the window by updating the window.location.href to self
        window.location.href=window.location.href 
    }, ttlMs);
})();
```

PR #3401, opened 2015-06-05

EspadaV8:
This is a RP based on claars suggestion on reddit here http://redd.it/38j8a8. By using the PHP 5.5 ::class property instead of a string it allows easier refactoring in e.g. PHPStorm when namespaces for providers/aliases change.


EspadaV8:
Opened against wrong branch

PR #3402, opened 2015-06-05

EspadaV8:
This is a RP based on claars suggestion on reddit here http://redd.it/38j8a8. By using the PHP 5.5 ::class property instead of a string it allows easier refactoring in e.g. PHPStorm when namespaces for providers/aliases change.


etcinit:
I've been considering doing this on some of my projects. Does this affect the startup time of the application in any significant way?

EspadaV8:
Doing some _very_ simple tests on the homepage of the project I'm working on the differences are either nothing or _very_ minor. Over 15 runs the average time difference was about 5ms according to Clockwork. Both the min and max time were in the same ballpack (50ms min 450ms max, average of around 185ms. It seemed that every 3rd request would spike to 430+ms and then the next 2 would be around 50ms).

franzliedke:
:+1: for doing this at least in the skeleton. That way we can encourage best practices for users of Laravel, and slowly transition the core as we go.

GrahamCampbell:
How. That's totally ridiculous. Could you do it for everything except the app config file please.

nWidart:
Nice as always Mr. Graham. :+1:

GrahamCampbell:
> Nice as always Mr. Graham.

Having like 100 imports in config/app.php is never what we intended. It's fine to use ::class in your app/ code, but not in that config file.

crynobone:
> By using the PHP 5.5 ::class property instead of a string it allows easier refactoring.

Personally I feel this argument is misleading, sure it might be easier to refactor for your app related services/facades/alias if you have any but for Illuminate you still need to physically update the classname to something else (or have the risk of renaming the Illuminate class etc). 

Pretty much a rare use case which you can do in your own application.

GrahamCampbell:
I totally agree with @crynobone.

franzliedke:
I always though this was about IDEs etc. being able to warn you that a given class does not exist (typos etc.), whereas with strings they could not.

crynobone:
> I always though this was about IDEs etc. being able to warn you that a given class does not exist

Yeh, pretty much suitable for **app** or **packages** related class, but for Illuminate when you include the whole framework?

crynobone:
```php
use App\Providers\AppServiceProvider;
use App\Providers\EventServiceProvider;
use App\Providers\RouteServiceProvider;
```

I feel only this is fine, if you want to make it easier to change the app structure later on. We all know `php artisan app:name` is suitable only for right after the first installation.

nWidart:
If the list of use statements at the top is what's annoying, and I do get that argument, an option would be to use the full classname.

``` php
'View'      => Illuminate\Support\Facades\View::class,
```

GrahamCampbell:
@nWidart That might be a better idea for the app config file. :)

patrickbrouwers:
I think @nWidart suggestion is pretty neat. I would be able to easily click on the ServiceProvider and open the class right away. (Without having all those imports in top of the file.)

And I think that we should be consistent in the use and not do it as ::class on one place and as string on another place.

rtablada:
I use ::class for alot of our work (mostly in routes, factories, etc) and that is super helpful.

I also use them in the config files but as discussed a ton of use statements makes it ugly. But it also makes it hard to parse. It's like "I'm booting ViewServiceProvider, ok from where"?
IMO when configuring things you should know the full class path in one LOC rather than import statements.

That rant over... I do like @nWidart's recommendation. Code intel can use it, fewer sneaky little stringses, and it still maintains readability.

hobitses

franzliedke:
Awesome! I like the combination of the fully-qualified class names with `::class`, too.

mvdstam:
+1 for @nWidart's suggestion! :+1:

JulienTant:
Import everything is aweful.
Using string to represent class is aweful.

So, i guess it's a :+1: for @nWidart's solution !

shehi:
I like the idea, but be advised, the package minimum PHP version dependency will increase from 5.4 to 5.5.

@nWidart : are you going to send in an updated PR?

RomainLanz:
@shehi Already merged https://github.com/laravel/laravel/pull/3403.

nWidart:
And #3405 too :smiley:

shehi:
Ok. Just as a side note, for devs around the globe: as Igor Pellegrini commented in official docs:

> It should be noted that the new keyword "::class", added in 5.5, WILL NOT PERFORM ANY CHECK on > the existence of the targeted class, either in the current namespace or another.

> If the targeted class has not an explicitly defined namespace the keyword will JUST PROVIDE a fully qualified name corresponding to:

> CURRENT_NAMESPACE + TARGET_CLASS_NAME

> This could NOT correspond to any existing class!

etc etc... :)

shehi:
Oh and tiny but important question:  are we going to do the same for framework itself as well? :)

PR #3403, opened 2015-06-05

RomainLanz:
See https://github.com/laravel/laravel/pull/3402


GrahamCampbell:
This looks much better to me. :)

RomainLanz:
Do we also change `auth.php` file for `App\User`?

nWidart:
Hehe, I was about to submit same PR. :smile: 

I'd do the same for `auth.php`.

RomainLanz:
@nWidart You can also do the same for `services.php`.

Send a PR to my fork. I'll squash it.

EspadaV8:
Wow, gone for a day and miss out on all the action. Thanks for this @RomainLanz this does look a lot better (don't know why I didn't do this in the first place).

PR #3404, opened 2015-06-05
Consistency between files.

RomainLanz:
Align correctly all lines to be consistent with other files.

### DIFF
--- a/config/auth.php
+++ b/config/auth.php
@@ -59,8 +59,8 @@ return [
     */
 
     'password' => [
-        'email' => 'emails.password',
-        'table' => 'password_resets',
+        'email'  => 'emails.password',
+        'table'  => 'password_resets',
         'expire' => 60,
     ],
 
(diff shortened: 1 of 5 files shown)


lukasgeiter:
:-1:

RomainLanz:
@lukasgeiter Maybe better to argument why is this bad instead of just write a thumb down...

Other files align their lines, we need to be consistent between them, so we have 2 options:
 1. Remove space in every files.
 2. Align in every files.

lukasgeiter:
@RomainLanz I don't think it's bad but I personally prefer not to align them. So the thumbs down is just my opinion. However the majority of config files are not aligned so I'd rather remove those spaces. Except maybe for the long lists in `app.php`, for those it kind of makes sense...

RomainLanz:
@taylorotwell What do you think about removing alignement in other files except `app.php`?

PR #3405, opened 2015-06-05
Use the ::class notation

nWidart:
Use the ::class notation for class references in app/ folder.
Similar to #3403

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -13,7 +13,7 @@ class Kernel extends ConsoleKernel
      * @var array
      */
     protected $commands = [
-        'App\Console\Commands\Inspire',
+        \App\Console\Commands\Inspire::class,
     ];
 
     /**
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -13,7 +13,7 @@ class Handler extends ExceptionHandler
      * @var array
      */
     protected $dontReport = [
-        'Symfony\Component\HttpKernel\Exception\HttpException',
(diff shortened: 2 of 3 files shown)


RomainLanz:
Do we need the slash at the beginning of each line?

taylorotwell:
Yes since we're in a namespace.

RomainLanz:
Correct. Thanks for this answer.

app/Console/Kernel.php
```diff
@@ -13,7 +13,7 @@ class Kernel extends ConsoleKernel
      * @var array
      */
     protected $commands = [
-        'App\Console\Commands\Inspire',
+        \App\Console\Commands\Inspire::class,
```
GrahamCampbell: Prefer if we imported this.

app/Exceptions/Handler.php
```diff
@@ -13,7 +13,7 @@ class Handler extends ExceptionHandler
      * @var array
      */
     protected $dontReport = [
-        'Symfony\Component\HttpKernel\Exception\HttpException',
+        \Symfony\Component\HttpKernel\Exception\HttpException::class,
```
GrahamCampbell: Prefer if we imported this.

app/Console/Kernel.php
nWidart: What if we have 30 commands ? :smiley: 
That was your argument on the other PR, not having tons of imports at top of file.

rtablada:
Local app ::class will be relative or you could import the namespace (I know namespace imports in PHP aren't ideal). So that brings it down to a shorter length and a single import.

PR #3408, opened 2015-06-07
[5.1] Updated the sample encryption key

bertramtruong:
Hi everyone,

I recently tried to initialize a new project from the develop branch but after installing all dependencies, I would get the following exception:

```
RuntimeException in EncryptionServiceProvider.php line 29:
No supported encrypter found. The cipher and / or key length are invalid.
```

After investigation and drilling down into the files, I notice the issue is that it's not possible to use Laravel straight out of the box due to the following line in Encrypter::supported

```
return ($cipher === 'AES-128-CBC' && ($length === 16)) || ($cipher === 'AES-256-CBC' && $length === 32);
```

Since the default encryption is AES-256-CBC, it's length must be 32 characters long. Therefore, I updated both the config/app.php sample string to 'SomeSuperIncrediblyRandomString!' as well as the .env.example sample string.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,6 @@
 APP_ENV=local
 APP_DEBUG=true
-APP_KEY=SomeRandomString
+APP_KEY=SomeSuperIncrediblyRandomString!
 
 DB_HOST=localhost
 DB_DATABASE=homestead
--- a/config/app.php
+++ b/config/app.php
@@ -78,7 +78,7 @@ return [
     |
     */
 
-    'key' => env('APP_KEY', 'SomeRandomString'),
+    'key' => env('APP_KEY', 'SomeSuperIncrediblyRandomString!'),
 
     'cipher' => 'AES-256-CBC',
 


GrahamCampbell:
That's because you're meant to change the key before you use it.

bertramtruong:
Graham,

So you're pretty much saying that Laravel can't be used straight out of the box, as it used to be able to, when the cipher was AES-128-CBC and the encryption key is 'SomeRandomString'?

RomainLanz:
It can be used out of the box with the installer https://github.com/laravel/installer.

GrahamCampbell:
I'm saying that anyone that even thinks of using that default encryption key is asking for trouble, and we never should have allowed it through in the first place.

bertramtruong:
@RomainLanz, the installer will give you version 5.0.

GrahamCampbell:
@bertramtruong You must use our key generate command after install.

GrahamCampbell:
Or set your own secure key.

GrahamCampbell:
If you try to use our default key, you might as well not bother with the encryption in the first place.

bertramtruong:
@GrahamCampbell so you're encouraging that Laravel should not be allowed to be used out of the box, really.

I understand that it's a security concern, but I don't see how it would be of any concern if I was to spin up an instance in development. I don't need a secure key in a development VM.

RomainLanz:
@bertramtruong Yes because Laravel 5.1 is not released yet!

GrahamCampbell:
> so you're encouraging that Laravel should not be functional out of the box, really.

It is functional out of the box. That's like saying, oh, i can't be arsed to create a database.

PR #3410, opened 2015-06-08
Use the ::class notation

pespantelis:


shehi:
Hmm... Looks nice: I have seen this notation in PHPStorm, but couldn't find anything about its function in official documents or web.

pespantelis:
@shehi I found this http://php.net/manual/tr/migration55.new-features.php#migration55.new-features.class-name

shehi:
Yea, I realized after seeing #3402 and PRs derived from there on.

pespantelis:
:+1:

GrahamCampbell:
:-1: from me

pespantelis:
@GrahamCampbell IMO the reference of User's model at ModelFactory.php needs to be changed.
About the other changes, maybe you are right.

shehi:
@GrahamCampbell : what is different here from #3402 and its derivatives?

shehi:
@taylorotwell : Quick question - having non-FQNS here, won't break the code, will it? I mean, leading-back-slashes are missing. :)

taylorotwell:
None of these files are namespaced so it will be fine.

PR #3411, opened 2015-06-08
Fix for badges

shehi:
Apparently badges belonging to Framework have been used for this Package as well. Especially version numbers are misleading there.


GrahamCampbell:
This is intended. The version numbers are intended to be for the framework repo.

PR #3413, opened 2015-06-08
Update composer.json to require php version 5.5 or greater

scrubmx:
Given the recent changes on the framework. I think is better if to require 5.5 explicitly in composer.json.


taylorotwell:
Should be 5.5.9.

taylorotwell:
Done.

PR #3414, opened 2015-06-10
minimum stability

KluVerKamp:
minimum stability wasn't "dev" in the previous releases
was it left on purpose this time?


GrahamCampbell:
Yeh, this is fine, because of the prefer stable.

KluVerKamp:
All right then.

barryvdh:
Imho, minimum stability should be stable..

PR #3417, opened 2015-06-11
Add dotenv support for several database directives

baspeters:
For single database instance applications, dotenv directives like DB_HOST and DB_USERNAME can be used. However, not all directives are supported. This pull request adds support for charset, collation, prefix and schema

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -58,8 +58,8 @@ return [
             'database'  => env('DB_DATABASE', 'forge'),
             'username'  => env('DB_USERNAME', 'forge'),
             'password'  => env('DB_PASSWORD', ''),
-            'charset'   => 'utf8',
-            'collation' => 'utf8_unicode_ci',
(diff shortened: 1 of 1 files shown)


baspeters:
For schema and prefix I would say yes, charset and collation probably makes no sense. However I have encountered badly configured servers that would default the charset on an MSSQL dblib connection to something other than UTF8. If changing the system configuration of such a server is not an option, there is no easy way to set the charset other than creating a custom database configuration and overriding it there. Having a dotenv option to set this for a specific environment is a lot easier.

taylorotwell:
You can set these yourself.

baspeters:
These settings can indeed be set with a custom configuration, however a dotenv directive would avoid having to create this custom configuration in the first place to cater for differently behaving environments.

Can at least be considered to hardcode 'utf8' charset in the 'sqlsrv' configuration, just as it is present in the mysql and pgsql configuration?

PR #3420, opened 2015-06-12
Add support for catchmail and mhsendmail

wienczny:
Add support for catchmail and [mhsendmail](https://github.com/mailhog/mhsendmail) using MAIL_SENDMAIL in environment

Allow pretend to be configured using environment

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -106,7 +106,7 @@ return [
     |
     */
 
-    'sendmail' => '/usr/sbin/sendmail -bs',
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Configure this in your own app.

wienczny:
I did that for my projects and think it is useful for other people. But both options don't make sense to be configured project wide.
Is there any use case where you enable pretend and this is not your development environment? You either don't use mail in your app or the value differs between development and production.
Changing the value for sendmail is only useful if it can differ between environments, too.
What reasons are there for not reading these values from environment?

PR #3427, opened 2015-06-17
Classmap whole /tests folder, like /database.

JerseyMilker:
Classmap whole /tests folder, like the classmap for /database?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     },
     "autoload-dev": {
         "classmap": [
-            "tests/TestCase.php"
+            "tests"
         ]
     },
     "scripts": {


taylorotwell:
This has never been necessary for me.

JerseyMilker:
@taylorotwell I was thinking it'd be helpful for those who might want to add their own traits with helper methods to their tests cases.  This would make autoloading within /tests less opinionated, much like how autoloading of /database/seeders is not opinionated on how you structure.  Could it be useful to classmap everything within /tests and leave test structure up to user?

GrahamCampbell:
I prefer to psr4 autoloader them.

JerseyMilker:
@GrahamCampbell By "them", you prefer original classmap as is, or classmaps as per my suggestion?

shehi:
I bet he means tests in general. Thing is, tests (at least in my case) don't really need namespacing, so PSR-4 is out of question. And them (i.e. tests) not really being a part of the project codebase (in many CI environments they are even excluded from reviewing and grading), I don't believe they should even be a part of project's namespace. PHPUnit doesn't need autoloading unit-test classes, it just runs everything it finds in default tests/ folder. So question is: why would you need these tests being mentioned in composer.json? For traits etc: if you really want to go that complex with unit-tests, then you will need to create a full-blown test-suite fully namespaced and all, and it should use PSR-4, which will be auto-resolved anyway. No need to put them in classmaps of composer.json. Classmaps are good for the cases where you don't have namespaced or PSR-x compatible code and you just need to give a hint of folder location for autoloading: in PHPUnit's case, it's just not required. There is phpunit.xml file for that purpose which should have white- or black-list of everything test-related.

PR #3429, opened 2015-06-18
Adding PHP Coding Standards Fixer

PSmolic:
I am suggesting we add [php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer). It would allow us to add CS rules and enforce them (Tracis CI can check it).

By running ```$ php-cs-fixer fix .``` we would fix everything that isn't following included rules.

For now I enabled ```PSR2_LEVEL``` with disabled ```psr0``` which was changing namespace ```App``` to ```app``` and enabled ```align_double_arrow``` which aligns arrows in array, example: https://github.com/laravel/laravel/compare/master...PSmolic:implementing-php-cs-fixer#diff-b3af398d5ac2670049a1f17f4683c03eR29

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /vendor
 /node_modules
 .env
+php-cs-fixer
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,13 @@
+<?php
+
+$finder = Symfony\CS\Finder\DefaultFinder::create()
+    ->exclude(['vendor'])
+    ->in(__DIR__);
+
+return Symfony\CS\Config\Config::create()
+    ->level(Symfony\CS\FixerInterface::PSR2_LEVEL)
+    ->fixers([
+        'align_double_arrow',
+        '-psr0'
+        ])
+    ->finder($finder);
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 /vendor
 /node_modules
 .env
+php-cs-fixer
```
GrahamCampbell: no

.php_cs
```diff
@@ -0,0 +1,13 @@
+<?php
+
+$finder = Symfony\CS\Finder\DefaultFinder::create()
+    ->exclude(['vendor'])
+    ->in(__DIR__);
+
+return Symfony\CS\Config\Config::create()
+    ->level(Symfony\CS\FixerInterface::PSR2_LEVEL)
+    ->fixers([
+        'align_double_arrow',
```
GrahamCampbell: we already have our own coding style, and this isn't included

composer.json
```diff
@@ -9,6 +9,7 @@
         "laravel/framework": "5.1.*"
     },
     "require-dev": {
+        "fabpot/php-cs-fixer": "^1.9",
```
GrahamCampbell: wrong!

GrahamCampbell: this would belong as a global install anyway

PR #3432, opened 2015-06-21
can you add this

obb12:
public function withCookies(Array $cookies)
{
    foreach($cookies as $cookie)
        $this->headers->setCookie($cookie);

    return $this;
}


obb12:
Just want to return two or more cookies to browser

iPad$B$+$iAw?.(B

2015/06/21 21:20$B!"(BTaylor Otwell <notifications@github.com> $B$N%a%C%;!<%8(B:

PR #3433, opened 2015-06-23
removed an unnecessary return statement

schmanz:
Hallo,

just a small optimization as the docblock explains that the function does only return void.

Best regards,
Jochen


GrahamCampbell:
`void` means undefined NOT null, so we can return anything we want there.

PR #3435, opened 2015-06-23
Use double quotes for html attributes

tonglil:
Just to keep things consistent.


tonglil:
If you want me to create a PR for the 5.0 branch I can do so as well.

PR #3437, opened 2015-06-24
having a named route rather than a static route

rizerzero:
having a named route rather than a static route makes it easier to change change login uri.


taylorotwell:
you can do this in your own app

PR #3439, opened 2015-06-27
Ignore Vagrantfile in .gitignore

endyjasmi:
I was using Homestead in per project setup when I noticed that Vagrantfile is not ignored. It might be accidentally committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@
 /node_modules
 Homestead.yaml
 Vagrantfile
+.vagrant
 .env
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /vendor
 /node_modules
 Homestead.yaml
+Vagrantfile
 .env


taylorotwell:
It's OK if this is committed.

endyjasmi:
Hello again, I understand that ignoring Homestead.yml and regenerate it every time helps auto set the project path in Homestead.yml (even though there's need of manual tweak on windows).

Commiting Vagrantfile seems okay too.

I was just wondering what is the rational behind the commit of the .vagrant directory. Cause, when we commit the directory, for each member who pull request, there's a need for them to destroy current instance of the virtual machine which is quite troublesome.

command

Anahkiasen:
You can just add things to the gitignore, you don't have to stick exactly with what comes with Laravel

shehi:
Agreed with @Anahkiasen and I hope this kinda PRs end one day.

endyjasmi:
Been doing that with my project. Just thought of it as a friendly suggestion and was wondering the reasoning behind the decision made out of the box.

No hostile meaning intended.

Peave =.=V

Anahkiasen:
The reason is most likely that a lot of people have `.vagrant` in their global gitignore already as this isn't Laravel specific and Laravel doesn't come with a Vagrant install by default (Homstead being optional)

endyjasmi:
Okay. Maybe not the Vagrantfile, what about the .vagrant directory which store virtual machine specific configuration.

crynobone:
@endyjasmi carefully read what @Anahkiasen reply, or read on "global .gitignore".

endyjasmi:
Correct me if I'm wrong, so basically it just that you guy prefer to add to .gitignore manually over something come out of the box?

crynobone:
>  you guy prefer to add to .gitignore manually over something come out of the box?

It a one time process instead of having it done per project. Did you even search "global .gitignore", cause I found these:

* http://makandracards.com/makandra/15947-use-a-global-gitignore-file-to-ignore-stuff-from-your-machine
* http://islegend.com/development/setting-global-gitignore-mac-windows/
* http://devoh.com/blog/2013/01/global-gitignore
* https://robots.thoughtbot.com/global-gitignore

> Using a .gitignore file can be very useful. It is used to ignore certain files on specific projects. But what should you do if you find yourself consistently creating a .gitignore for the same files on every project? Create a global .gitignore!

endyjasmi:
I do know about "global .gitignore" just thought that ignoring `.vagrant` directory would be a plus especially for new user using homestead. Just a thought.

w2pc:
We should teach new users about global .gitignore as it is the proper way
to ignore this type of file along with several other system files like
ds_storw and .thumbnail.  let's properly educate on best practices, not
promote ignorance.

endyjasmi:
Okay, got it. It just that I thought framework should ease this kind of task.

Anyway, thanks for the heads up.

PR #3440, opened 2015-06-27
Change comment to reflect the change of the default css preprocessor.

mikaelmattsson:


Mevrael:
run "php artisan vendor:publish"

http://laravel.com/docs/5.1/packages#public-assets

PR #3441, opened 2015-06-28
Removing length requirement for email validation

ssfinney:
PHP's ```FILTER_VALIDATE_EMAIL``` already checks for email lengths, and enforces a limit of 254 characters on emails, per the [RFC 2821](https://www.ietf.org/rfc/rfc2821.txt) standard. Therefore, enforcing a length here is redundant.

PHP 5.5's source code for the ```FILTER_VALIDATE_EMAIL``` constant can be found [here](http://lxr.php.net/xref/PHP_5_5/ext/filter/logical_filters.c#503).


taylorotwell:
Since that isn't entirely obvious, going to leave it.

PR #3446, opened 2015-07-05
remove @return annotations from constructor doc blocks

petercoles:
Scrutinzer objects to @return annotations for doc blocks even when they are "void" on the grounds that "a constructor does not have a meaningful return value."

Removing them from the code will make it easier for Laravel-based applications to be "scrutinized".

### DIFF
--- a/app/Http/Controllers/Auth/AuthController.php
+++ b/app/Http/Controllers/Auth/AuthController.php
@@ -25,8 +25,6 @@ class AuthController extends Controller
 
     /**
      * Create a new authentication controller instance.
-     *
-     * @return void
      */
     public function __construct()
     {
--- a/app/Http/Controllers/Auth/PasswordController.php
+++ b/app/Http/Controllers/Auth/PasswordController.php
@@ -22,8 +22,6 @@ class PasswordController extends Controller
 
     /**
      * Create a new password controller instance.
-     *
-     * @return void
      */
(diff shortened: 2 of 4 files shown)


GrahamCampbell:
It is meaningful.

petercoles:
I agree, it is meaningful. Scrutinizer's comment misses the point slightly in this case, as the problem isn't that "void" isn't meaningful in this context, it's that it's a bit wrong. I'll raise a new pull request with corrected value that complies with the phpDocumentor docs and satisfies Scrutinizer.

GrahamCampbell:
Please don't send any PRs. Our phpdoc is just fine as it is.

shehi:
Actually it is not meaningful: Scrutinizer is just another PHP-sniffer + PHP-mess-detector wannabe with tons of checks, some of which are turned on by default. Those rules are not even a convention, consisting of mostly developer preferences. There is no need to mess with void returns - instead just add an ignore instruction into ```scrutinizer.yml``` file and Scrutinizer won't bother you.

GrahamCampbell:
It IS meaningful. `void` means "do not rely on the return value of this method".

shehi:
```void``` is, Scrutinizer's whinings are not :)

GrahamCampbell:
Yeh, scrutinizer has a fair due bits of advice which are wrong.

PR #3447, opened 2015-07-05
correct @return annotations in constructor doc blocks

petercoles:
"void" isn't quite accurate as the value for the @return annotation on a constructor. phpDocumentor is expecting "self" or no annotation.

This correction has the added advantage of satisfying Scrutinizer, which treats "void" as a documentation issue.

### DIFF
--- a/app/Http/Controllers/Auth/AuthController.php
+++ b/app/Http/Controllers/Auth/AuthController.php
@@ -26,7 +26,7 @@ class AuthController extends Controller
     /**
      * Create a new authentication controller instance.
      *
-     * @return void
+     * @return self
      */
     public function __construct()
     {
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
This is wrong. Constructors do not return anything.

petercoles:
They don't have a "return" statement, but they implicitly return an instance of the class, which is why "self" is correct and "void" is treated as issue.

GrahamCampbell:
> but they implicitly return an instance of the class

No they don't. If I put `return 1` in the constructor, then it returns 1, not the class. If I don't return anything, the nothing is returned...

GrahamCampbell:
PHP's constructors don't behave in the same way as perhaps some other languages.

PR #3449, opened 2015-07-07
type hint faker generator

browner12:
allow IDE to help with autocompletion

### DIFF
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -11,7 +11,7 @@
 |
 */
 
-$factory->define(App\User::class, function (\Faker\Generator $faker) {
+$factory->define(App\User::class, function (Faker\Generator $faker) {
     return [
         'name' => $faker->name,
         'email' => $faker->email,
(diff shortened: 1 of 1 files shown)


database/factories/ModelFactory.php
```diff
@@ -11,7 +11,7 @@
 |
 */
 
-$factory->define(App\User::class, function ($faker) {
+$factory->define(App\User::class, function (\Faker\Generator $faker) {
```
GrahamCampbell: No leading slash please.

browner12:
sure thing.

is that a PSR styling thing? or some other reason?

GrahamCampbell:
The leading slash is actually wrong - that's why it needs removing. PHP just supports it anyway because people commonly use it by mistake.

GrahamCampbell:
It's only use is to tell php to look to the global namespace while inside a namespace. All fully qualified class names are defined without a leading slash. :)

browner12:
good to know!

should be updated

GrahamCampbell:
:)

GrahamCampbell:
:+1: from me

PR #3450, opened 2015-07-08
add DB_CONNECTION variable in .env.example file

emmanuelgautier:


GrahamCampbell:
:+1:

shehi:
As https://github.com/laravel/laravel/blob/master/config/database.php says, it can be optional. Why bother? I think we shouldn't clutter ```.env``` file any more. The bare minimum is the best.

lukasgeiter:
I think instead of adding nearly everything to the example file, it would make sense to have a list of all used environment variables in the docs. @taylorotwell what are your thoughts on that?

GrahamCampbell:
Why do we bother with the mail driver but not the db driver?

GrahamCampbell:
Surely the db driver is more important?

shehi:
Mail driver should only have "log" as default driver and be removed from
example env file.

emmanuelgautier:
I think so @lukasgeiter . Have, at least, env variables used in configuration files is important to know what is possible to configure fastly.

adaojunior:
I agree with @lukasgeiter

PR #3451, opened 2015-07-09
default APP_KEY to a 32 characters length string

KluVerKamp:
So  fresh installation of laravel won't throw the cipher related exception anymore.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,6 @@
 APP_ENV=local
 APP_DEBUG=true
-APP_KEY=SomeRandomString
+APP_KEY=SomeRandomStringSomeRandomString
 
 DB_HOST=localhost
 DB_DATABASE=homestead
--- a/config/app.php
+++ b/config/app.php
@@ -78,7 +78,7 @@ return [
     |
     */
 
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
This is intentionally wrong to make people run the key generate command.

alexleonard:
How many times... 

We should add a comment in config/app.php which says the default is intentionally not 32 characters to force key generation per install.

lukasgeiter:
Or how about `ThisIsIntentionallyNot32Chars` as default? :stuck_out_tongue_winking_eye:

fideloper:
heh

KluVerKamp:
I see.
@lukasgeiter sounds better! :stuck_out_tongue_closed_eyes:

alexleonard:
Yes! That's a great idea, no unnecessary comment needed.

ThisIsIntentionallyNot32CharactersSoYouHaveToSetItManuallyDontSubmitAGitPullPlease....

;)

KluVerKamp:
@alexleonard well that has escalated quickly!

PR #3452, opened 2015-07-11
Update queue.php

andr3s2:
Is important to maintain the consistency in the coding style.


andr3s2:
It is the only place in the framework where you can find a one-line array.
In the other files each key in an array is located in a different line.

lukasgeiter:
:+1:

PR #3454, opened 2015-07-11
Environment variable to toggle HTTPS only session cookies

kezkankrayon:


GrahamCampbell:
Just do this in your own app.

kezkankrayon:
This is worth noting if your SSL setup differs between environments. If you have `'secure' => true` in an environment that doesn't have SSL setup, you will receive a TokenMismatchException. In my case, SSL is not setup in my local or development environment, but is setup in my production environment.

LKDevelopment:
i think this environment variable would be useful!

mateusjatenee:
I think it would be useful too.

vinicius73:
It's just a parameter to more that can help in the development.

I do not see why not have it in the settings, even if not in the case of .env.exemple

kezkankrayon:
I think it does more than help development; it allows a setting to be secure by default. Also, exposing the variable in .env.exemple not only assists with development, it will raise awareness around security.

devsalman:
Imho, i think this env variable doesn't have big impact because its too environment specific. For me, i always reconfigure my application config / env variable when im deploying my application for production

PR #3455, opened 2015-07-13
Move terminate process before sending the response

yoramdelangen:
Moving terminate process before sending the response back to the browser.

Within the `send()` method from `Symfony\Component\HttpFoundation\Response`, it checks if the 'fastcgi_finish_request()` function exists if it is will directly flush all response data to the browser. There will be no more code executed.

I guess it will only appear for fastcgi users and when using terminable middleware.

Using PHP 5.5.9

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -53,6 +53,6 @@ $response = $kernel->handle(
     $request = Illuminate\Http\Request::capture()
 );
 
-$response->send();
-
 $kernel->terminate($request, $response);
+
+$response->send();


GrahamCampbell:
It is after on purpose.

yoramdelangen:
Whats the purpose? 
For servers running fastcgi it is not working atm?

GrahamCampbell:
The point of this method is to do things after the response has been sent.

yoramdelangen:
But when the function fastcgi_finish_request() is called the connection get closed so `$kernel->terminate()` is never called.

GrahamCampbell:
How come nobody else is having this issue? For example, sessions are dependent on the terminate method? Are you using HHVM or something?

GrahamCampbell:
FPM is recommended over CGI. Any reason why you can't use that?

yoramdelangen:
Because homestead is doing that by default. And fastcgi is used to pass all php files through the php-fpm socket. 

```
location ~ \.php$ {
        fastcgi_split_path_info ^(.+\.php)(/.+)$;
        fastcgi_pass unix:/var/run/php5-fpm.sock;
        fastcgi_index index.php;
        include fastcgi_params;
        fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
        fastcgi_intercept_errors off;
        fastcgi_buffer_size 16k;
        fastcgi_buffers 4 16k;
    }
```

yoramdelangen:
My bad i had the wrong idea/interpretation about fastcgi_finish_request! After calling fastcgi_finish_request the output is already flushed to the browser but continues to executing the terminate process. 

Case closed

userscapechris:
FPM is a FastCGI implementation, which that function seems to be
specifically for. However since I (and no one else that I know of) have
experienced issues (that I know of), I'm definitely also curious as to if
there's actually an issue there.

However, this comment seems relevant:
https://github.com/symfony/symfony/issues/5921#issuecomment-10109511

Essentially that function running first would be normal behavior if
expecting something to run after it!

GrahamCampbell:
Just because you send a response, it doesn't mean the process should exit?

PR #3458, opened 2015-07-17
Load Logging configuration via Env, defaults to single

marcorivm:
Allow the Logging configuration to be set via enviroment, for example, for development I want a single log, in forge I would prefer to log to the syslog so integrating papertrail is easier.


GrahamCampbell:
Just define this in your own app.

marcorivm:
@GrahamCampbell If Laravel is already providing .env abstraction for some variables, I do think that logging config would be good to be included by default.

Laravel should provide a consistent interface for using Config and .env, I think that changing the logging configuration depending on the environment is a good use case, and something that makes sense, so it should be available by default as an .env option.

shehi:
@marcorivm You maybe right, but bombarding issue management with micro-"desires" (yes, they are not features) like this which mostly are subjective ideas, never helps. Please do this 2-second-long task in your apps.

crynobone:
There no need to argue, a similar PR has been accepted and merged by Taylor #3551.

PR #3460, opened 2015-07-18
Update Authenticate middleware to correctly respond to Json requests

pespantelis:


GrahamCampbell:
The response returned is html though not json.

GrahamCampbell:
Dadly formatted html, but the content type is html.

pespantelis:
:+1:

PR #3461, opened 2015-07-18
Remove clear-compiled from pre-install

antonioribeiro:
If your application is somehow broken (you have a missing class, for instance), you won't be able to `composer update` because, well, your application is broken and composer is trying to call your application...

http://i.imgur.com/v2sUx5w.png

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,9 +32,6 @@
             "php artisan clear-compiled",
             "php artisan optimize"
         ],
-        "pre-update-cmd": [
-            "php artisan clear-compiled"
-        ],
         "post-update-cmd": [
             "php artisan optimize"
         ],


GrahamCampbell:
As discussed in https://github.com/laravel/framework/issues/9660, we really need a better way to do this.

garygreen:
@GrahamCampbell you are closing these issues but agreeing that something needs to change to improve it... May I suggest than we at least re-open #9660 until Taylor has a chance to see the proposed standalone command and discuss improvements/possibilities in fixing the issue? Closing it is only giving the impression that 'it won't be fixed'

GrahamCampbell:
I'll talk to Taylor about this outside of github. Easier that way.

mvdstam:
@GrahamCampbell Also very much not transparant and not giving the impression that this is taken seriously.

Also: Simply saying "Taylor doesn't like the idea" and then closing the issue without any further explanation is simply stupid and rude. Do you even _want_ people to contribute?

GrahamCampbell:
> Also very much not transparant

This is just how we deal with issues.

GrahamCampbell:
Just be patient. Taylor is actually current on holiday.

gmsantosxl:
In this kind of situations I simply run `composer update --no-scripts`. 

Usually `artisan clear-compiled` works again after that.

GrahamCampbell:
> In this kind of situations I simply run composer update --no-scripts. 

Same here, but that situation isn't ideal for automated deployments.

GrahamCampbell:
I tell you what, I'll make a new issue for this.

GrahamCampbell:
See https://github.com/laravel/framework/issues/9678.

progmars:
I too was using `--no-scripts` workaround as a last resort, but it does not work for `require` command:
`"--no-scripts" option does not exist`

So no way to disable scripts when you want to add a new package.

PR #3466, opened 2015-07-25
I don't find Solution

miadzadfalah:
i set two primary key and then migrate (By mistake),i get this error
 SQLSTATE[42000]: Syntax error or access violation: 1068 Multiple primary key defined
then is i run this:
php artisan migrate:reset
and i get this :
Table 'users' already exists

i know this table is hidden and i did not set database engine type 
have any Solution in php artisan or
Always i should remove (recreate) database Handy ?

or some questions and answer
http://stackoverflow.com/questions/26077458/laravel-migration-table-already-exists-but-i-want-to-add-new-not-the-older
{The reason this happens is that you ran a rollback previously and it had some error in the code or did not drop the table. This still however messes up the laravel migration table and as far as it's concerned you now have no record of pushing the user table up. The user table does already exist however and this error is throw.}

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -2,7 +2,6 @@
 
 namespace App\Exceptions;
 
-use Exception;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Please try our forums. :)

miadzadfalah:
thank you very much,
i search that before ask , 
nothing!
 :)
maybe answer is Handy!

PR #3468, opened 2015-07-27
Bump Laravel Elixir version

laracasts:


package.json
```diff
@@ -4,7 +4,7 @@
     "gulp": "^3.8.8"
   },
   "dependencies": {
-    "laravel-elixir": "^2.0.0",
+    "laravel-elixir": "^3.0.0",
```
GrahamCampbell: `^3.0.3` might be better because technically 3.0.1 and lower behave differently, so the version constraint should be updated to be more accurate.

PR #3469, opened 2015-07-30
app.scss: corrected path for bootstrap comment

AndreasHeiberg:
I assume that this used to work, but at least for me using the latest version of homestead with everything freshly installed you have to reference it relatively to load bootstrap.

### DIFF
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -1,2 +1,2 @@
-// @import "node_modules/bootstrap-sass/assets/stylesheets/bootstrap";
+// @import "../../../node_modules/bootstrap-sass/assets/stylesheets/bootstrap";
 


RomainLanz:
The default path works fine with me in OSX (`php artisan serve`).

taylorotwell:
Works for me as well.

AndreasHeiberg:
Well I would assume that the whole point of vm's is that you run everything including gulp from within the vm removing the need to setup your env on every machine. Eliminating any "works for me on OSX" comments :-)

I setup a new homestead vm for a new laravel 5.1 project. Went inside the vm, did npm install (which doesn't work by the way, you have to do `npm install -g npm@latest` first) and then ran gulp. Doing this it did not work for me.

PR #3470, opened 2015-07-31
Adding sass precision option of 8 as required by Bootstrap

votemike:
[Bootstrap requires a sass precision of 8](https://github.com/twbs/bootstrap-sass#sass-number-precision). The default sass file that is included with Laravel already has the line for including Bootstrap sass (although it is commented out)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -12,5 +12,5 @@ var elixir = require('laravel-elixir');
  */
 
 elixir(function(mix) {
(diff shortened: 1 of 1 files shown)


lukasgeiter:
The I personally would rather completely remove the commented out bootstrap import.

taylorotwell:
I believe @JeffreyWay is going to make this the default in Elixir.

PR #3471, opened 2015-07-31
Change URIs to routes (URIs is a bit misleading)

mikefrancis:
`App\Http\Middleware\VerifyCsrfToken::shouldPassThrough()` is checking routes rather than URIs - had the misfortune of trying to match a URI when I should have been trying to match a route - so amended the note/comment.


taylorotwell:
No it's not, it's checking URIs.

PR #3472, opened 2015-07-31
Bug Report: Model Observer Testing

Mulkave:
When testing models and their observers, in the first test run in a testing class the observer methods are called as usual, but are never called for the rest of the testing methods.

### Run The Tests
This pull request contains reproduces the issue, all you have to do is:
- clone this PR
- install dependencies with `composer install`
- run the tests with `./vendor/bin/phpunit`

The result will look as follows:

screen shot 2015-07-31 at 6 04 03 pm

The constraint error occurs because the `observed` key was not set by the observer, though worked correctly in the first test.

> This is not the first time I encounter this issue and have worked around it in the past by stuffing all the test methods in the first test method and calling them manually with something like:

```php
public function test_all_tests()
{
        $this->caching_articles();
        $this->caching_article_without_author();
        $this->caching_article_without_cover();
        $this->caching_article_without_photos();
}
```

### DIFF
--- a/app/User.php
+++ b/app/User.php
@@ -33,3 +33,5 @@ class User extends Model implements AuthenticatableContract, CanResetPasswordCon
      */
     protected $hidden = ['password', 'remember_token'];
 }
+
+User::observe(new UserObserver());
--- a/app/UserObserver.php
+++ b/app/UserObserver.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace App;
+
+class UserObserver
+{
+    public function saving($model)
+    {
+        $model->observed = true;
+    }
+}
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,2934 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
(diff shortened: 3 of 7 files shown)


GrahamCampbell:
This was fixed in 5.1.8.

app/User.php
```diff
@@ -33,3 +33,5 @@ class User extends Model implements AuthenticatableContract, CanResetPasswordCon
      */
     protected $hidden = ['password', 'remember_token'];
 }
+
+User::observe(new UserObserver());
```
GrahamCampbell: you can't put that there

GrahamCampbell: You MUST put that in a service provider.

GrahamCampbell: It's VERY bad to have this registered on class load, which is what you're doing,

GrahamCampbell: It also violates PSR0/PSR4, which is what you are using for autoloading.

Mulkave:
@GrahamCampbell thanks for the indication.

If I'm understanding you correctly, you're saying that this observer setup should live in a service provider as such:

```php
public function register()
{
    App\ User::observe(new UserObserver());
}
```

GrahamCampbell:
Boot method would be better, yes.

GrahamCampbell:
If you put tha tin the boot method of a service provider, you should be good.

PR #3473, opened 2015-07-31
32 Character key that won't throw a cipher error

dolbex:
When deploying from Forge on an initial deployment it breaks because there is not a valid key from the sample .env or in the app.php. It also causes general deployment issues and isn't very clear on what is happening.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -78,7 +78,7 @@ return [
     |
     */
 
-    'key' => env('APP_KEY', 'SomeRandomString'),
+    'key' => env('APP_KEY', 'DonotForgetYour32CharacterString'),
 
     'cipher' => 'AES-256-CBC',
 


GrahamCampbell:
The key is intentionally wrong to force people to set a secure one.

dolbex:
@GrahamCampbell - I get that - but the problem is it breaks forge... there is no way to do the initial deployment b/c the composer update breaks. Also, the error message isn't clear on what is even happening.

dolbex:
So, just to point out... in the composer:

```
"post-create-project-cmd": [
    "php artisan key:generate"
]
```

But even key:generate fries on the cipher error... Bit of a cyclical issue here.

GrahamCampbell:
> "php artisan key:generate"

Running that command shouldn't try to create a new encrypter instance though?

GrahamCampbell:
Anyway, the best way you can generate a secure random key is to type one in yourself, lol.

GrahamCampbell:
Nothing beats a bit of human randomness.

franzliedke:
> Nothing beats a bit of human randomness.

I hope you weren't too serious, when you wrote that. :)

GrahamCampbell:
lol :P

PR #3474, opened 2015-08-01
Added StyleCI config

GrahamCampbell:
With this config, we're already passing unlike laravel/framework was.

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,0 +1,9 @@
+preset: laravel
+
+disabled:
+  - unused_use
+
+finder:
+  not-name:
+    - index.php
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Going to hold off on this as we can configure this from StyleCI's gui.

shehi:
I'd strongly recommend CircleCI, as they also fully support Docker compositions, configuration for which I could help with. I use it for my own Laravel 5 based project, and test against MySQL and PgSQL versions using various PHP versions including HHVM.

shehi:
Ah, scratch it, it's rather CS related :)

PR #3479, opened 2015-08-15
Fix phpDoc.

devisualy:


taylorotwell:
This wouldn't really be consistent with the rest of the framework.

PR #3481, opened 2015-08-16
Fix slash redirect, dealing with non-ASCII chars

hudsontavares:
Some Apache servers, like the one where I keep the Laravel application I'm working, apply double url encoding when dealing with strings containing national characters, so a query string like:

/myurl/?q=cana-de-a%C3%A7%C3%BAcar
(cana-de-açúcar)

becomes:

/myurl?=cana-de-a%25C3%25A7%25C3%25BAcar
(cana-de-a%ç%úcar)

This PR solves this issue, while keep the same behavior when dealing with A-Z characters.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -7,7 +7,7 @@
 
     # Redirect Trailing Slashes If Not A Folder...
     RewriteCond %{REQUEST_FILENAME} !-d
-    RewriteRule ^(.*)/$ /$1 [L,R=301]
+    RewriteRule ^(.*)/$ /$1 [L,R=301,NE]
 
     # Handle Front Controller...
     RewriteCond %{REQUEST_FILENAME} !-d


taylorotwell:
Can you provide some documentation on what adding `NE` does?

hudsontavares:
@taylorotwell 

First of all, I want to say Laravel did a revolution in the way I code.

This NE addresses what seems to be a bug on Apache itself (as I found on [Server Fault](http://serverfault.com/questions/331899/apache-mod-rewrite-double-encodes-query-string-on-redirect), per example), where some versions, probably prior to Apache 2.4.1, double-escape query strings.

If you're dealing with non-ASCII characters on the query string (the sample I provided was found by an user from my application, which did a search using ç / ú chars), you can workaround this problem by urldecode() the values (maybe twice), with the burden of having to treat every GET variable before using it.

My testing machine, which runs Apache 2.4.12 / Ubuntu 12.04 / PHP 5.6.10, handles perfectly the slash redirection with non-ASCII characters (both with and without NE), while the staging area where I'm presenting the work in-progress to my client only works with NE.

As the pagination classes all generates the links with the trailing slash,I tought it was worthy to provide this easier workaround to the whole community, if you think its necessary.

taylorotwell:
I'm going to let people add this to their own file as needed.

MangoMM:
Hello!

 

Check it out http://www.horoscopes-love.eu/repeated.php?l63ck

 

ahmet@mangomm.co.uk

PR #3483, opened 2015-08-17
Use https rather than // for google fonts in default views.

Zarthus:
Ensure the default views always use HTTPS for the loading of fonts.

https://github.com/konklone/cdns-to-https#background

This change, due to its size and what it does - breaks nothing as far as compatibility goes. It just enforces a slightly better/newer practice. As such, I'm merging it to the 5.0 branch.

Sorry about the previous PR, I accidentally added the commits of `master` on top. I was expecting git to only merge my commit, not everyone elses.


taylorotwell:
5.0 is basically closed for changes.

PR #3486, opened 2015-08-24
Password Reset Token expiration date field

razamehdi-vteams:
Added Password Reset Token expiration date field to the default passwords_reset table.


zippex:
But why not use the created_at value and calculate the expiration date via Carbon with something like $obj->addHours(2); There is no need to store that in db imho.

razamehdi-vteams:
@zippex So should i close the pull request then?

PR #3488, opened 2015-08-27
Removed "return" statement

Rogerkael:
I think this return statement should be removed due the parent is void .


GrahamCampbell:
`void` IS NOT `null`. `void` means "we can return anything we feel like, and it should be considered useless"

PR #3489, opened 2015-08-28
Add default database ports to config

aejnsn:
Add default ports to their respective RDBMS in config/database.php. Add
an environment configuration for Homestead's default MySQL port in .env.example. I think this may speed the process of starting a Laravel project for new users of Homestead/Vagrant.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,6 +3,7 @@ APP_DEBUG=true
 APP_KEY=SomeRandomString
 
 DB_HOST=localhost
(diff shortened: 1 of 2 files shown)


lukasgeiter:
`33060` is only when you access the MySQL server from outside the VM. From the Laravel point of view it's the normal `3306`.

aejnsn:
Gah! I don't know what I was thinking.

PR #3493, opened 2015-09-04
Make App URL configurable with env

LKDevelopment:
Make APP URL configurable like DB Connection.
Sample:

If you have an DEV environment with URL dev.example.com, an TEST environment with URL test.example.com and production environment with URL example.com. Actually you need to change this with every deploy. With this PR you are allowed to configure this once in the .env file for every environment.


GrahamCampbell:
Already proposed and rejected, sorry. Just make this change on your own app.

PR #3499, opened 2015-09-15
Ignores .vagrant directory in case of per-project homestead.

systemovich:


GrahamCampbell:
1. 5.0 is not maintained.
2. Actually, it's not always the best idea to ignore it, see https://github.com/laravel/laravel/pull/3439#issuecomment-115937818.

systemovich:
1. According to the contribution guide, minor features may be sent to the latest stable branch. One would expect there to be a 5.1 branch. I went for 5.0, because it is the closest.
2. OK

PR #3371, opened 2015-05-10
Fix unnecessary return in App\Exceptions\Handler#report()

shin1x1:
Doc comment in App\Exceptions\Handler#report() said return type is void.
This method seems to me that return no value.
This PR removed return statement.

PR #3502, opened 2015-09-22
Report method has nothing to return

papa-smurf:
Nothing to return in report method


bonzai:
See: https://github.com/laravel/laravel/pull/3371

papa-smurf:
I see, thanks for the heads-up! Since parent::report has no return statement at all however I still feel like this doesn't make sense.

PR #3505, opened 2015-09-24
Change db conf to allow running of migrations from outside of Homestead

mikebronner:


GrahamCampbell:
Just set the "port" in the config.

mikebronner:
Setting the port fixes it for external access, but breaks internal access -- or am I misunderstanding?

PR #3506, opened 2015-09-24
File created for SQLite database

anteriovieira:
File created for those who want to use SQLite database. Thus we need only set DB_CONNECTION as environment variable.


leandrorosendo:
:+1:

jorgemurta:
That file existed in Laravel 4.
The community surelly had a reason to delete it I think

leandrorosendo:
I used here and it worked , please inform the reason of this file delete in Laravel 4 . thank's

taylorotwell:
It's just simple to create one if needed.

anteriovieira:
It makes sense, thank you.

PR #3507, opened 2015-09-27
php artisan serve now allows CORS requests

rcbytes:
This is important because we can now use `php artisan serve` to create SPA's and use Laravel as a remote endpoint.

e.g.: Your front-end application runs on `http://localhost:3000` and Laravel at `http://localhost:8000`, now Laravel can be requested by our front-end app.

### DIFF
--- a/server.php
+++ b/server.php
@@ -18,4 +18,7 @@ if ($uri !== '/' && file_exists(__DIR__.'/public'.$uri)) {
     return false;
 }
 
+// Allowing CORS Requests
+header("Access-Control-Allow-Origin: *");
+
 require_once __DIR__.'/public/index.php';


mvdstam:
Just use https://github.com/barryvdh/laravel-cors

rcbytes:
Since `php artisar serve` is supposed to be our development environment, it makes sense to allow everyone to fetch it, plus do we really need a third-party component just for CORS?

crynobone:
This is wrong, header should be handled via Symfony/Laravel Response object. We shouldn't add `header()` directly.

GrahamCampbell:
Yeh, this is totally wrong. Sorry.

rcbytes:
Hum.. I get it thought it could be handled there since `mod_rewrite` is "emulated" on this file also. 
Anyways this is an issue, I'll open it so you can solve in the `Laravel-way` hehe 
Ty for your time guys :smile:

GrahamCampbell:
Please don't open another PR. There is no bug.

GrahamCampbell:
If you want to set this header, I'd suggest you just add yourself a middleware.

rcbytes:
Hum... but `php artisan serve` should always accept CORS requests if we want a dev environment that is not specific. Not accepting them is a malfunction, therefore a bug, of course this is  my point of view.

Anyways, I'll add a middleware. (:

PR #3508, opened 2015-10-01
get timezone from .env if defined

faisaladlin:
For teams working in different timezones.
Can use their own .env TIME_ZONE setting without touching the project config/app.php


taylorotwell:
Can be added to your own app as needed.

PR #3509, opened 2015-10-02
You can set the database directory on file .env

anteriovieira:
It enables me to do something like:

```
DB_CONNECTION=sqlite
DB_DATABASE=/home/user/database/project-name.sqlite
```


GrahamCampbell:
This is not normally something useful to configure on a per environment basis. You should just set this file in the config file itself.

anteriovieira:
Ok, thank you!

PR #3513, opened 2015-10-06
Maybe it's time to remove it.

RryLee:


muhghazaliakbar:
Cool

mateusjatenee:
I don't see a problem in that. It may even look more of a "fluent API" for some people.

GrahamCampbell:
I think this is maybe best going to 5.2, though, actually, i think Route::get is way more expressive.

anteriovieira:
@GrahamCampbell +1

PR #3514, opened 2015-10-06
[Maybe BUG] Laravel 5 softDeletes Model cause blank white page response if table don't have deleted_at column

darkcyber:
Hi,

My `Options` table contain `id, name, value, autoload` columns.

My `routes.php`

```
<?php
Route::get('/settings', 'SettingsController@index');
```

Here is my `SettingsController.php`

```
<?php namespace App\Http\Controllers\Admin;

use App\Http\Requests;
use App\Http\Controllers\Controller;
use App\Option;

class SettingsController extends Controller {

	public function index()
	{
		$options = Option::all();
		return dd($options);
	}
```

Here is my `Option.php` model

```
<?php namespace App;

use Illuminate\Database\Eloquent\SoftDeletes;

/**
* 
*/
class Option extends Model
{
	use SoftDeletes;
	protected $fillable = ['id', 'name', 'value', 'autoload'];
	protected $table = 'options';
}
```

If I visit `myproject.dev/settings` it will give me blank white page without any error, try to delete views cache at storage folder didn't work.

Try with artisan tinker give me some idea

```
use App\Option;
dd(Option::all());
```

It give me `Illuminate\Database\QueryException with message 'SQLSTATE[42S22]: Column not found: 1054 Unknown column 'coeg_options.deleted_at' in 'where clause' (SQL: select * from `coeg_options` where `coeg_options`.`deleted_at` is null)`

Then I add `deleted_at` column at my table, finally it work and I didn't get blank white pages anymore.

Maybe will be good if Laravel check `deleted_at` column for model which use `SoftDeletes` and throw an error if `deleted_at` column not found, just like artisan tinker error above.


MangoMM:
Don't use softDeletes in your model, and Laravel will know not to look 
at the deleted_at column

PR #3516, opened 2015-10-06
Revert to store sqlite in /storage by default

JoelESvensson:
Revert https://github.com/laravel/laravel/commit/879bc146504462ef273b893010edf004026aa021

Deployments with envoyer will have to use storage_path anyway (or any other linked folder that's not under source control)

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -48,7 +48,7 @@ return [
 
         'sqlite' => [
             'driver'   => 'sqlite',
-            'database' => database_path('database.sqlite'),
+            'database' => storage_path('database.sqlite'),
             'prefix'   => '',
         ],
 


GrahamCampbell:
Envoyer deployments should not be using sqlite as I already explained.

GrahamCampbell:
Envoyer is designed to take the pain out of production deployments at scale.
Sqlite is designed to make local development easy or for use on low traffic production systems.

The two do not overlap.

GrahamCampbell:
If you really do want to use this on envoyer though, you can just edit this file on your app. I'm not forcing anything on anyone.

JoelESvensson:
Sure. It's a very small thing that. I just thought it wouldn't hurt to make the service easier to use for whoever who wants to use it that way. I have some low-traffic deployments myself

PR #3518, opened 2015-10-07
Removed unused option

martinssipenko:


GrahamCampbell:
Thanks. :heart:

vlakoff:
Wow, I can't even find out when this option was part of PHPUnit. Maybe it was never?

GrahamCampbell:
It was there in like 3.2 or something. :)

PR #3523, opened 2015-10-14
ignore some IDE files

Jheysoon:


bkintanar:
ignoring IDE files are best left to the developer since I don't think laravel will be ignoring all of the IDE files that exists.

jahvi:
IDE-specific files should be ignored in a global .gitignore file

mvdstam:
Seriously don't get these kind of PR's, which keep coming back for some reason. Here's why it's a bad idea:

[*]Appending rules for IDE's to the `.gitignore` file is a slippery slope. There are a lot of IDE's out there; are we going to add them all?
[*]The

PR #3524, opened 2015-10-15
Update 2014_10_12_000000_create_users_table.php

TheEVIL:


GrahamCampbell:
This isn't right is it?

StelianAndrei:
Maybe **username ** would be a better idea? Many websites this day offer the user the ability to login either by username or email

lucasmichot:
:+1: for `username`

pespantelis:
@StelianAndrei :+1:

youchenlee:
:+1: for  `username`

overtrue:
:+1: for `username`

vlsoprun:
:+1: for username

arrilot:
Is there any code in laravel/framework repo that depends on this `name` field? If so this change can lead to some annoying problems.

shehi:
-1 for username.

anteriovieira:
Username make sense, but rather keep name. This does not prevent implement the rules to save the name of User in this column since the name is of user.

ChadTaljaardt:
@shesi +1

PR #3525, opened 2015-10-15
Update database.php

TheEVIL:


mydnic:
Should be set in .env anyway

shehi:
+1 VM environment shouldn't be a priority IMO.

PR #3532, opened 2015-10-20
Use env('DB_PREFIX', '') instead of empty string literals ''

andela-cijeomah:
Create an environment variable for a database prefix


GrahamCampbell:
5.0 is closed for changes anyway.

PR #3535, opened 2015-10-26
Added configuration settings for default values for Global "From" Address & Name.

razamehdi-vteams:


GrahamCampbell:
This is not normally something that will be environment specific. Just set application specific config in the actual config files.

PR #3539, opened 2015-10-27
Swap order of boot() and register() in service providers

davestewart:
No functional changes in this PR, so what's the point?

As a newcomer to Laravel (but not programming or PHP) there's a lot to wrap your head around. Service providers are one of the big topics, what with setting up your DI, and specifically what's the difference between the `register()` and `boot()` methods?

To help, all I've done here is put them in the order they are called, which just helps a bit when your head is swimming with a dozen new concepts, and you're looking at your own code and doing your best to "get" it.

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -5,16 +5,6 @@ use Illuminate\Support\ServiceProvider;
 class AppServiceProvider extends ServiceProvider {
 
 	/**
-	 * Bootstrap any application services.
-	 *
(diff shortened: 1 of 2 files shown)


davestewart:
Not sure what is up with those style checks there... all I did was move the methods!

GrahamCampbell:
Laravel 5.0 is not accepting pull requests.

davestewart:
No probs. Might want to consider that one if you find yourself in these classes in future tho.

Cheers.

PR #3540, opened 2015-10-28
Added namesapce for helper functions

peeto:
Taylor,

As I said I don't mind doing the work to wrap these helper functions in a namespace. The codebase I'm using already has a global function called "redirect" and I expect to see such things in other places on the internet. I'd be happy to get some feedback / remarks / abuse from you in regards to this.


taylorotwell:
No plans to make this change.

PR #3544, opened 2015-10-30
Move global middleware to RouteServiceProvider

rkgrep:
This is related to https://github.com/laravel/framework/pull/10758

Prevents probable performance problems and unexpected responses, in case middleware applied to non-existent routes.

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -9,15 +9,12 @@ class Kernel extends HttpKernel
     /**
      * The application's global HTTP middleware stack.
      *
+     * This middlewares are executed before route handling.
+     *
      * @var array
      */
     protected $middleware = [
         \Illuminate\Foundation\Http\Middleware\CheckForMaintenanceMode::class,
-        \App\Http\Middleware\EncryptCookies::class,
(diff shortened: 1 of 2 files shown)


crynobone:
This approach also create potential headache when say a package middleware that depends on `StartSession` middleware to be executed before it.

Laravel maintainer has been advertising that anything that depends on session should be converted to middleware and append it to http kernel contract via `appendMiddleware()`.

GrahamCampbell:
:-1

GrahamCampbell:
This behaviour is wrong imo.

GrahamCampbell:
I'd be :+1: to moving just the csrf middleware though.

rkgrep:
@crynobone I'm not sure if this will affect packages. Although, I haven't seen packages modifying kernel middlewares yet (only route ones).

rkgrep:
@GrahamCampbell You have this behavior in StyleCI :smile:

GrahamCampbell:
> You have this behavior in StyleCI

StyleCI isn't an app structure that is suitable for the majority of Laravel users though. :)

PR #3545, opened 2015-10-30
Move CSRF check to RouteServiceProvider

rkgrep:
However I don't like session middlewares in Kernel, I PR this as mentioned in #3544


GrahamCampbell:
I kind of wonder if this is something that should only be changed on the develop branch?

rkgrep:
This repository is only used for initial installation, so this shouldn't break anything.

GrahamCampbell:
Yeh, but i'm saying maybe we shouldn't have as a default until 5.2.

razamehdi-vteams:
:+1: for this. I believe this should be in the RouteServiceProvider

taylorotwell:
Going to just hold off on this entirely.

PR #3550, opened 2015-11-04
Added port to database connection options

codenhagen:
Being able to specify the port when connecting to a database is sometimes handy. This adds a port option to the PostgreSQL/MySQL/SQLServer connection options.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,6 +3,7 @@ APP_DEBUG=true
 APP_KEY=SomeRandomString
 
 DB_HOST=localhost
+DB_PORT=null
 DB_DATABASE=homestead
 DB_USERNAME=homestead
 DB_PASSWORD=secret
--- a/config/database.php
+++ b/config/database.php
@@ -55,6 +55,7 @@ return [
         'mysql' => [
             'driver'    => 'mysql',
             'host'      => env('DB_HOST', 'localhost'),
+            'port'      => env('DB_PORT', null),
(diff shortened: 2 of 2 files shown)


ClausMunch:
+1
Great idea.. ☺

shehi:
+1

RomainLanz:
Why set them to null by default?

codenhagen:
They are set to null by default because oftentimes you don't connect via TCP/IP, and in those cases it doesn't make sense to set a port number.

arrilot:
The vast majority uses default ports. 
If you happen to need to change it you can do what you do it in your app so I don't see any reason to bloat congifs and .env file with it.

GrahamCampbell:
Any reason why this went to 5.2 and not to 5.1?

anteriovieira:
It is a good alternative, but we can keep the default port instead of null.

crynobone:
> Any reason why this went to 5.2 and not to 5.1?

Doesn't think change require user to manually set the port, or does setting it to null automatically make it use the default port for every connection?

codenhagen:
crynobone: You're not required to manually set the port.

codenhagen:
anteriovieira: The default port varies according to your RDBMS, but setting it to null ensures that the default port is used regardless of RDBMS (when connecting over TCP/IP).

taylorotwell:
Just going to let people make this changing on their own if needed.

PR #3551, opened 2015-11-05
Updating the log system with env variable

arcanedev-maroc:
The `single` log system is only useful in dev/local environment.


GrahamCampbell:
The default shouldn't be changed from single.

GrahamCampbell:
This PR does two things.

GrahamCampbell:
> The single log system is only useful in dev/local environment.

Not true actually.

arcanedev-maroc:
@GrahamCampbell 
What are the benefits of using `single` log system in production env ?

GrahamCampbell:
> What are the benefits of using single log system in production env ?

You might be using a system outside of laravel's control to process that single log file, like logrotate.

arcanedev-maroc:
Why using a system outside of laravel's control ? 

I'm sure a lot of artisans like to see their log files inside their apps by using log viewer packages.

GrahamCampbell:
I'm sure they do. That's why it's configurable.

rkgrep:
:+1: for .env variable, :-1: for daily log by default

Anahkiasen:
Yeah same here, I don't like the default but I don't think there _is_ a default that would please everyone, on the other hand I would like it to be an env variable

arcanedev-maroc:
Is there any other good reason of using `single` (or not using `daily`) log system on production mode (inside laravel's control) ?

This is the benefits of using `daily` logs in my opinion:
 * Keeping log files small/readable/manageable after months/years of use.
 * Archives-able/downloadable logs files.
 * Avoiding the performance issues.
 * Friendly with log viewer packages.

brunogaspar:
Having it on the `.env` file would be nice, but i also prefer `single`.

GrahamCampbell:
Yeh, you shouldn't change two things in a single PR. Please just make the env change here.

arcanedev-maroc:
@GrahamCampbell 
OK, but first i want to be sure about the default log system before i rebase this PR.

It's debatable and good to know how other think about it.

@brunogaspar @rkgrep Please, be more specific about your choice.

brunogaspar:
> OK, but first i want to be sure about the default log system before i rebase this PR.

It's easy, keep the `single` as @GrahamCampbell mentioned but use the env variable. Then send a different PR as proposal to change from `single` to `daily`.

> Please, be more specific about your choice.

- I have only one file to delete instead of multiple.
- I have only one file to read instead of multiple.
- I have only one file to download instead of multiple if i ever have the need to.
- I find it easier to find errors on one file, even if the file is big, we can search.
- One file for development is more easier that multiple IMHO.

Don't take me wrong and i do see your point, but you need to see ours as well.

Laravel makes it easy to start development with a minimum configuration required, once you go to production, just do your preparations like everyone should be doing, like set the env to production, set debug to false, in your case set the log to daily instead of single.

Just because you fancy for daily during development, not everyone likes that, i'm included.
Since `single` is default at the moment, you should adapt your application(s) to use `daily`, it's impossible to please everyone :)

shehi:
@arcanedev-maroc do you even know what logrotate is? if so, could you be so kind not to waste our time? :(

shehi:
Why the community is wasting time on not important details is beyond me. Who cares if its single or not?! We can change it with one single word edit. Simple.

If you guys want to save everyone's times - yes, everyone's, as we all receive notifications for these PRs - please create Contribution Guidelines and ban all these nonsensical contributions to the package which would take 2-3 word replacements from any developer using it. Really tired of these nonsensical PRs. And please don't start another nonsense with notions like "Then unsubscribe-from/unwatch the project."

arcanedev-maroc:
@shehi I'm here to contribute and share my ideas/opinions. Do you think this is a waste of time ?

shehi:
@arcanedev-maroc : I do mate. Nothing personal. You are not the first, and won't be the last anyway :(

arcanedev-maroc:
@shehi This may helps you : https://speakerdeck.com/mattstauffer/empathy-gives-you-superpowers

:+1:

mateusjatenee:
(y) for .env.

PR #3552, opened 2015-11-06
clear-compiled fix by short circuiting the app and running before creation

leemason:
A fix to address the issue discussed here: https://github.com/laravel/framework/issues/9678

The root of the problem is the app relying on the application to run such a small task, and the app relying the composer packages being installed to register the service providers.

If the codebase isn't in sync with the installed packages the app fails, and then cant clear-compiled file, and then composer cant update to install the correct packages because the clear-compiled failed.

So a nice little circular problem here fixed with a few lines in the bootstrap/autoload.php.

Its put in this place as this is the only place before the app gets created we can intercept.

The code simply checks for cli use and the artisan clear-compiled in the argv. Then it does esencially what the clear-compiled command does without the application and then exits.


GrahamCampbell:
Looks like you targeted the wrong branch here.

leemason:
i checked the contributing docs and it recommended bug fixed be sent to the latest stable branch. is 5.0 not it? It also said only future features should go to master, which one would you suggest?

GrahamCampbell:
Nope, 5.1 is the latest stable. That is the master branch on laravel/laravel.

leemason:
ok cool, ill sort that now, maybe the docs should be updated to reflect that, or the docs be adhered to?

Is this isolated to laravel/laravel or does laravel/framework follow the convention set in the docs for future reference?

PR #3553, opened 2015-11-06
clear-compiled fix by short circuiting the app and running before creation

leemason:
A fix to address the issue discussed here: https://github.com/laravel/framework/issues/9678

The root of the problem is the app relying on the application to run such a small task, and the app relying the composer packages being installed to register the service providers.

If the codebase isn't in sync with the installed packages the app fails, and then cant clear-compiled file, and then composer cant update to install the correct packages because the clear-compiled failed.

So a nice little circular problem here fixed with a few lines in the bootstrap/autoload.php.

Its put in this place as this is the only place before the app gets created we can intercept.

The code simply checks for cli use and the artisan clear-compiled in the argv. Then it does esencially what the clear-compiled command does without the application and then exits.

### DIFF
--- a/artisan
+++ b/artisan
@@ -20,7 +20,7 @@
 */
 if (php_sapi_name() == 'cli' && isset($_SERVER['argv'])) {
     $argv = $_SERVER['argv'];
-    if (in_array('artisan', $argv) && in_array('clear-compiled', $argv)) {
+    if (count($argv) == 2 && 'artisan' == $argv[0] && 'clear-compiled' == $argv[1]) {
         $files = glob('./bootstrap/cache/*');
         foreach ($files as $file) {
             if (file_exists($file)) {
--- a/artisan
+++ b/artisan
@@ -3,6 +3,36 @@
 
 /*
 |--------------------------------------------------------------------------
+| Run the artisan clear-compiled command without loading the framework
+|--------------------------------------------------------------------------
+|
+| When running composer update, laravel runs the clear-compiled command.
+| Under most circumstances this works fine, however if your package
+| requirements change, and your updating via version control there is
+| potential for this to fail as your providers config array could differ
+| from whats installed causing the application to fail during provider 
+| registration. This snippet simply short circuits artisan and runs the
+| same functionality as the command, but without Application creation.
+| This doesnt remove the actual clear-compiled command from artisan, as
+| using the Artisan::call() method would still result in the command being 
+| used and not this.
+|
+*/
+if (php_sapi_name() == 'cli' && isset($_SERVER['argv'])) {
+    $argv = $_SERVER['argv'];
+    if (in_array('artisan', $argv) && in_array('clear-compiled', $argv)) {
+        $files = glob('./bootstrap/cache/*');
+        foreach ($files as $file) {
+            if (file_exists($file)) {
+                @unlink($file);
+            }
+        }
+        exit();
+    }
+}
+
+/*
+|--------------------------------------------------------------------------
 | Register The Auto Loader
 |--------------------------------------------------------------------------
 |
--- a/artisan
+++ b/artisan
@@ -19,8 +19,10 @@
 |
 */
 if (php_sapi_name() == 'cli' && isset($_SERVER['argv'])) {
+
     $argv = $_SERVER['argv'];
-    if (count($argv) == 2 && 'artisan' == $argv[0] && 'clear-compiled' == $argv[1]) {
+
+    if (count($argv) == 2 && 'artisan' == substr($argv[0], -7) && 'clear-compiled' == $argv[1]) {
         $files = glob('./bootstrap/cache/*');
         foreach ($files as $file) {
             if (file_exists($file)) {
@@ -29,6 +31,7 @@ if (php_sapi_name() == 'cli' && isset($_SERVER['argv'])) {
         }
         exit();
     }
+    
 }
 
 /*
--- a/bootstrap/autoload.php
+++ b/bootstrap/autoload.php
@@ -18,6 +18,36 @@ require __DIR__.'/../vendor/autoload.php';
 
 /*
 |--------------------------------------------------------------------------
+| Run the artisan clear-compiled command without loading the framework
+|--------------------------------------------------------------------------
+|
+| When running composer update, laravel runs the clear-compiled command.
+| Under most circumstances this works fine, however if your package
+| requirements change, and your updating via version control there is
+| potential for this to fail as your providers config array could differ
+| from whats installed causing the application to fail during provider 
+| registration. This snippet simply short circuits artisan and runs the
+| same functionality as the command, but without Application creation.
+| This doesnt remove the actual clear-compiled command from artisan, as
+| using the Artisan::call() method would still result in the command being 
+| used and not this.
+|
+*/
+if(php_sapi_name() == "cli" && isset($_SERVER['argv'])) {
+    $argv = $_SERVER['argv'];
+    if(in_array('artisan', $argv) && in_array('clear-compiled', $argv)){
+        $files = glob("./bootstrap/cache/*");
+        foreach($files as $file) {
+            if (file_exists($file)) {
+                @unlink($file);
+            }
+        }
+        exit();
+    }
+}
+
+/*
+|--------------------------------------------------------------------------
 | Include The Compiled Class File
 |--------------------------------------------------------------------------
 |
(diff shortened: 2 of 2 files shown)


bootstrap/autoload.php
```diff
@@ -18,6 +18,36 @@
...
+    if (in_array('artisan', $argv) && in_array('clear-compiled', $argv)) {
+        $files = glob('./bootstrap/cache/*');
+        foreach ($files as $file) {
+            if (file_exists($file)) {
+                @unlink($file);
+            }
+        }
+        exit();
+    }
+}
```
crynobone: If this is the way to go, why isn't this in `artisan` file instead of `bootstrap/autoload.php`.

leemason:
its detailed in the description:

"Its put in this place as this is the only place before the app gets created we can intercept."

if you put in it in artisan the boostrap/autoload.php file is required, BUT the app is created by that time, so the problem will already have happened.

crynobone:
Isn't it possible to have that executed before requiring `bootstrap/autoload.php`? This piece of code doesn't require anything from laravel and exited immediately.

leemason:
good point, it could just go in artisan before the autoload, didn't even think of that, ill update it now.

artisan
```diff
@@ -3,6 +3,36 @@
...
+| registration. This snippet simply short circuits artisan and runs the
+| same functionality as the command, but without Application creation.
+| This doesnt remove the actual clear-compiled command from artisan, as
+| using the Artisan::call() method would still result in the command being 
+| used and not this.
+|
+*/
+if (php_sapi_name() == 'cli' && isset($_SERVER['argv'])) {
+    $argv = $_SERVER['argv'];
+    if (in_array('artisan', $argv) && in_array('clear-compiled', $argv)) {
```
rkgrep: Wrong condition. `php artisan something something clear-compiled something` will pass this condition.

leemason: Fixed in the last commit, will only work when the command arguments are explicitly ```php artisan clear-compiled```

GrahamCampbell:
Still doesn't work properly. eg: `php artisan -vv clear-compiled`.

GrahamCampbell:
or `hhvm --php artisan clear-compiled`.

leemason:
In what situation would:

1. anyone try and run the command like that?
2. want verbose output of a command that returns nothing?

At the end of the day it just removes some cache files, nothing else, as many have said this bug is a problematic issue and has been a problem for a long time.

If you want to add verbose options to a command that offers no verbose output that too is possible, but were back to loading the autoload file and framework for no reason again.

the way the command is documented to work, and what its documented to do are both achieved by this change?

rkgrep:
It may also be called as `./artisan clear-compiled` or with full path `php /home/foo/web/site/artisan clear-compiled` - all of this won't work as intended in your commit.

rkgrep:
P. S. I'm :+1: for this change. But it should work as expected in any use cases.

leemason:
The latest commit accounts for that by fetching the last 7 chars from the first argument, so ./artisan or /path/to/artisan will work as well.

dan-har:
I liked the way you solved this, but there might be some problems with this change -

In all artisan commands you can use the cli, or use call the 
commands from the code using ```Artisan::call('...')```.

Either way the ```Console\Application::call()``` is called.

But this change make the ``` artisan clear-compile ``` run 
different code then ```Artisan::call('clear-complied);```. 

So thats not the artisan command expected behavior and it
is an attempt to get around the console application contract.

I think that in a case where clear-command needs to be removed from being handled by the application @garygreen code is better

leemason:
yes the actual code run would be different, but the point is it would still run as it does now without the possible bug. and stick with the whole laravel artisan = console methodology.

If it was to go down the route of a seperate file to clear the compiled we might as well just run "rm bootstrap/cache/compiled.php" directly in the composer.json file.

Plus its then another breaking change which means it will get put off till 5.2.

I think there is are pros cons to multiple different methods people have contributed (this, seperate file, etc).

I thinks its probably best to get the opinion of @taylorotwell himself to see what he thinks as we just seem to going round in circles without any clear direction from the guys who can commit either approach to the framework.

What say you taylor?

Im happy to get this right if it needs amending, but cant commit anymore time to it without knowing if its worth it or not.

leemason:
PS. ```\Artisan::call(....)``` and ```php artisan ....``` dont actually perform the same lifecycle either.

the ```::call()``` method is never called during an artisan request, its a shortcut method which runs directly on the command class.

The ```::call()``` method runs the command directly, without putting it through the artisan application at all, but thats a seperate issue already noted on the framework repo.

garygreen:
> If it was to go down the route of a seperate file to clear the compiled we might as well just run "rm bootstrap/cache/compiled.php" directly in the composer.json file.

That wouldn't be cross OS compatible so its best to rely on a php script. Also it's not just a case of removing compiled.php but a few other things like services.json.

leemason:
thats why the current version ive supplied removes everything but hidden files from the folder, i was mearly illustrating the point of keeping it in artisan is a good way to go.

taylorotwell:
Why not just export a file from laravel/framework into the Composer `bin` directory called `clear-compiled`? Then we could do php vendor/bin/clear-compiled?

GrahamCampbell:
@taylorotwell's idea is much better imo.

garygreen:
Sounds like a decent idea. Only thing I'd say is it might be confusing having just the one command in vendor/bin as nothing else in laravel is like that. Would `php artisan` still list the command / how would people be informed about the clear-compiled command?

Personally I think it would just be easier to have the `artisan` file check to see if your calling `clear-compiled` and if so define an env to not include the compiled file, similar to how @leemason is proposing but without all the weird CLI checks. That way nothing changes for people currently using Laravel, no composer.json changes or publishing bin commands.

Also @taylorotwell any reason why artisan command was setup to include a lot of bootstrap code? I think it would be possible to just have a few tiny lines that bootstrap the artisan tool from within Laravel itself, that way the `artisan` file can be updated simply by a composer update if needs be. For instance the current artisan comand can go from:

```php
#!/usr/bin/env php
<?php
/*
| Register The Auto Loader
|
| Composer provides a convenient, automatically generated class loader
| for our application. We just need to utilize it! We'll require it
| into the script here so that we do not have to worry about the
| loading of any our classes "manually". Feels great to relax.
|
*/
require __DIR__.'/bootstrap/autoload.php';
$app = require_once __DIR__.'/bootstrap/app.php';
/*
| Run The Artisan Application
|
| When we run the console application, the current CLI command will be
| executed in this console and the response sent back to a terminal
| or another output device for the developers. Here goes nothing!
|
*/
$kernel = $app->make(Illuminate\Contracts\Console\Kernel::class);
$status = $kernel->handle(
    $input = new Symfony\Component\Console\Input\ArgvInput,
    new Symfony\Component\Console\Output\ConsoleOutput
);
/*
| Shutdown The Application
|
| Once Artisan has finished running. We will fire off the shutdown events
| so that any final work may be done by the application before we shut
| down the process. This is the last thing to happen to the request.
|
*/
$kernel->terminate($input, $status);
exit($status);
```

To simply:

```php
#!/usr/bin/env php
<?php

$artisan = require __DIR__.'/vendor/laravel/framework/artisan.php';
$artisan->run(__DIR__);
```

leemason:
@garygreen the problem is the cached file is called in the bootstrap autoload, so theres no way to *not* include it before artisan runs, that exactly why the code i added is at the top of the artisan file, before that file is included.

And even if we could prevent the cached file being loaded before artisan runs, the only way to do that is by using *weird cli checks*

i agree with taylor that a vendor/bin file would be the cleanest/simplist approach, but as mentioned before, and as you just mentioned gary, this means braking changes for everyone using laravel as of now and brings us back to no consistency.

Currently cli == artisan, web == Application.

with a vendor/bin command we are effectivley saying:

cli == artisan**, web == Application

**artisan cant remove 2 simple files so we have a seperate command for this ......

To me that sounds a bit embarrassing, means updating docs, and also means the bug will still exist for people who dont read the upgrade docs (admittedly thats there fault, but wont stop the issues being filed).

garygreen:
I honestly don't mind too much how it's fixed as long as it has no impact on performance of the app (which your suggestion probably does, and even mine slightly). Taylors suggestion is the best one atm, though I'd think that would go in 5.2 as its not too backwards compatible.

At the end of the day the main usage for the command is in production purely for deployment, so it's just a simple case of php artisan vendor/bin/clear-compiled vs php artisan clear-compiled. I really don't care about the 'embarrassing' factor of artisan not being able to delete some files, it's really no big deal. Once composer.json is setup right you shouldn't need to run that command manually anyway, though it would be worth noting it on the docs for some kind of deployment / optimisation notes.

taylorotwell:
Closing this particular PR since the vendor/bin route is definitely the correct way to implement something like this. Will let someone else open a new PR.

PR #3556, opened 2015-11-12
Added namespaces for the sample tests

hudsontavares:
People probably create their own tests using the built-in Laravel structure as base. For me, the existing Laravel classes are lacking the improvements on this PR.

The existing tests, aside the fact of not having a namespace defined (a fact that goes against [PSR-1](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-1-basic-coding-standard.md), as Laravel expects PHP 5.5.9 at minimum), aren't automatically allowing the user instantiate other classes contained on the same folder, as the autoloading features aren't aware of the tests/ folder.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
             "database"
         ],
         "psr-4": {
-            "App\\": "app/"
+            "App\\": "app/",
+            "Tests\\": "tests/"
         }
     },
     "autoload-dev": {
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
             "database"
         ],
         "psr-4": {
-            "App\\": "app/",
+            "App\\": "app/"
         }
     },
     "autoload-dev": {
--- a/composer.json
+++ b/composer.json
@@ -23,9 +23,6 @@
         }
     },
     "autoload-dev": {
-        "classmap": [
-            "tests/TestCase.php"
-        ],
         "psr-4": {
             "Tests\\": "tests/"
         }
--- a/composer.json
+++ b/composer.json
@@ -20,13 +20,15 @@
         ],
         "psr-4": {
             "App\\": "app/",
-            "Tests\\": "tests/"
         }
     },
     "autoload-dev": {
         "classmap": [
             "tests/TestCase.php"
-        ]
+        ],
+        "psr-4": {
+            "Tests\\": "tests/"
+        }
     },
     "scripts": {
         "post-install-cmd": [
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -1,5 +1,7 @@
 <?php
 
+namespace Tests;
+
 class TestCase extends Illuminate\Foundation\Testing\TestCase
 {
     /**
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -2,7 +2,7 @@
 
 namespace Tests;
 
-class TestCase extends Illuminate\Foundation\Testing\TestCase
+class TestCase extends \Illuminate\Foundation\Testing\TestCase
 {
     /**
      * The base URL to use while testing the application.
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -19,7 +19,8 @@
             "database"
         ],
         "psr-4": {
-            "App\\": "app/"
+            "App\\": "app/",
+            "Tests\\": "tests/"
```
Anahkiasen: Shouldn't this be in `autoload-dev`?

hudsontavares: @Anahkiasen Yea, you're right. Moved it, thank you.

shehi:
IMO, you could've just got away with ```App\Tests``` namespace, instead of creating additional one. But overall, I support this move: I also have tests namespaced - they look clearer in --debug mode as well.

composer.json
```diff
@@ -25,7 +25,10 @@
     "autoload-dev": {
         "classmap": [
```
lucasmichot: So `classmap` shall then be removed

GrahamCampbell:
:-1: for "Tests"

GrahamCampbell:
Maybe "App\Test" or something?

tests/TestCase.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+namespace Tests;
+
 class TestCase extends Illuminate\Foundation\Testing\TestCase
```
lucasmichot: This shoudl be `class TestCase extends \Illuminate\Foundation\Testing\TestCase` with a backslash

lucasmichot:
Some comments, but :+1:

lucasmichot:
Same, I would use `\\App\\Tests` instead. In this case,  `\Illuminate\Foundation\Console\AppNameCommand` might need to be updated

hudsontavares:
@GrahamCampbell

The fact the tests/ folder is located outside the app structure turns things clearer for me, specially if a newcomer on Laravel looks on the folder structure. The app folder already contains almost 10 namespaces (8, to be more precise) on master, isn't too easy to figure out what each one holds IMHO.

Furthermore, the tests aren't a direct part of application (in the sense of something required or called by the application itself), they only exists to test the application. You can create tests prior to write application code or even create no tests at all. So, why should they be located inside the application / App namespace?

@lucasmichot Thank you for your review, I'm updating these points.

tests/TestCase.php
```diff
@@ -1,6 +1,8 @@
 <?php
 
-class TestCase extends Illuminate\Foundation\Testing\TestCase
+namespace Tests;
+
+class TestCase extends \Illuminate\Foundation\Testing\TestCase
```
lucasmichot: Your could import ` \Illuminate\Foundation\Testing\TestCase` and alias it as `AbstractTestCase`

tests/TestCase.php
```diff
@@ -18,7 +20,7 @@ public function createApplication()
     {
         $app = require __DIR__.'/../bootstrap/app.php';
 
-        $app->make(Illuminate\Contracts\Console\Kernel::class)->bootstrap();
+        $app->make(\Illuminate\Contracts\Console\Kernel::class)->bootstrap();
```
lucasmichot: You could also import `\Illuminate\Contracts\Console\Kernel`

shehi:
@hudsontavares : Your points re tests folder not being inside app folder is valid. But I disagree with fact that tests are not part of application - they may not be for production, but they are when it comes to CI/CD and QA. Adding a different root namespace to overall codebase which resides inside the same repository isn't really intuitive. You can consider this as an exception scenario to overall PSR logic and exceptions don't break rules.

hudsontavares:
@shehi Everyone on comments prefer the App\Tests namespace, so, I'll update it accordingly.

Do you guys think it's an issue to keep the /tests/ folder as the holder of the App\Tests namespace, or it would be better move it to app/Tests instead?

lucasmichot:
@hudsontavares it should stay in `/tests`. Make sure to check to update `\Illuminate\Foundation\Console\AppNameCommand` too

taylorotwell:
Can you send this to `develop` branch? Thanks.

PR #3557, opened 2015-11-13
.gitignore Simplified

anteriovieira:
Hometead files are usually managed by the homestead. And if they are here could be ignored by the developer.


taylorotwell:
Homesteads can be local installed.

PR #3558, opened 2015-11-13
[5.2] Added two symfony deps to require-dev

GrahamCampbell:
Goes with https://github.com/laravel/framework/pull/10938.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,9 @@
     "require-dev": {
         "fzaninotto/faker": "~1.4",
         "mockery/mockery": "0.9.*",
-        "phpunit/phpunit": "~4.0"
+        "phpunit/phpunit": "~4.0",
+        "symfony/css-selector": "2.8.*|3.0.*",
+        "symfony/dom-crawler": "2.8.*|3.0.*"
     },
     "autoload": {
         "classmap": [


shehi:
@GrahamCampbell : If the dependency library (framework) has these in its composer.json, what is the point of adding them again to this package? It will fetch framework's dependencies anyway.

GrahamCampbell:
The framework doesn't depend on them. They're only needed to use the integration testing crawler.

GrahamCampbell:
Thus, it makes sense to move them out of laravel/framework so they can only be loaded as dev-dependencies.

shehi:
I actually was thinking the other way around: removing them from this package. That way, we won't need to update our apps.

GrahamCampbell:
We can't do that? We have no way to install on dev only.

shehi:
I don't get you: both repos have those requirements as ```"require-dev"```, so removing that from one of these repos won't hurt. So instead of removing from framework, I suggest doing the same here, remove it from ```laravel/laravel```. Because we all will be updating our requirements via composer anyway, including this repo as well (it will fetch it's dependencies and run tests afterwards), meaning those two symfony packages will come into play as ```"require-dev"``` eventually. That way, all of us won't have to update our composer.json files in our apps.

Am I missing something here?

GrahamCampbell:
Require-dev does not inherit.

shehi:
Ah ok. Thanks.

PR #3559, opened 2015-11-14
Updated PHPUnit version to ~5.0

bkintanar:


GrahamCampbell:
That requires php 5.6, and we need to ensure we support php 5.5.

shehi:
@bkintanar And there is PHPUnit 5 related bug in framework.

PR #3560, opened 2015-11-16
Add PSR4 autoload for tests

lucasmichot:
Add PSR4 autoload for tests

---

Replace #3556

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -1,6 +1,8 @@
 <?php
 
-class TestCase extends Illuminate\Foundation\Testing\TestCase
+namespace Tests;
+
+class TestCase extends \Illuminate\Foundation\Testing\TestCase
 {
     /**
      * The base URL to use while testing the application.
@@ -18,7 +20,7 @@ class TestCase extends Illuminate\Foundation\Testing\TestCase
     {
         $app = require __DIR__.'/../bootstrap/app.php';
 
-        $app->make(Illuminate\Contracts\Console\Kernel::class)->bootstrap();
+        $app->make(\Illuminate\Contracts\Console\Kernel::class)->bootstrap();
 
         return $app;
     }
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -1,8 +1,11 @@
 <?php
 
-namespace Tests;
+namespace App\Tests;
 
-class TestCase extends \Illuminate\Foundation\Testing\TestCase
+use Illuminate\Contracts\Console\Kernel;
+use Illuminate\Foundation\Testing\TestCase as AbstractTestCase;
+
+class TestCase extends AbstractTestCase
 {
     /**
      * The base URL to use while testing the application.
@@ -20,7 +23,7 @@ class TestCase extends \Illuminate\Foundation\Testing\TestCase
     {
         $app = require __DIR__.'/../bootstrap/app.php';
 
-        $app->make(\Illuminate\Contracts\Console\Kernel::class)->bootstrap();
+        $app->make(Kernel::class)->bootstrap();
 
         return $app;
     }
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Tests\\": "tests/"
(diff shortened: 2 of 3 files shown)


tests/TestCase.php
```diff
@@ -1,6 +1,11 @@
 <?php
 
-class TestCase extends Illuminate\Foundation\Testing\TestCase
+namespace App\Tests;
+
+use Illuminate\Contracts\Console\Kernel;
```
shehi: Add alias of ```as ConsoleKernel``` maybe? There are so many "kernels" in this package... Just a thought.

lucasmichot: True, but there's only one in this class

skyrpex:
I've done this always in my Laravel projects... Would love to see it merged! :+1:

vrkansagara:
:+1:  good thought.

hernandev:
Not sure about that. i like to keep the tests on the same namespace than the actual code. it helps a lot with keeping less code (less use statements)

By Default, PHPUnit would recognise the files recursively, following the name structure would be nice but autoloading the files is completely unnecessary.

For having a default namespace on the tests that follows the application namespace: :+1: 
For having tests on the autoload: :-1:

shehi:
@hernandev : You can't remove tests from autoload: ```\App``` namespace is looking for ```/app``` folder by default. ```/tests``` folder however is outside that folder, but as a namespace is inside ```\App``` namespace. It's duality so we have to declare a new autoloader resolution implicitly for new namespace of ```\App\Tests```. 

And PHPUnit is fed an autoloader file from application bootstrap, which technically works via composer's autoloader. Not to mention many IDE's, including PHPStorm need an autoload file location to perform in-IDE-tests as well. So overall, all this needs full compliance with composer. If ```composer dump``` command doesn't recognize all test files, we will have a problem.

What you desire is impossible to achieve I am afraid.

hernandev:
@shehi 

1) I'm against using App\Tests, just App is easier to import classes.
2) Any namespace you put your tests under, will be recognised by PHPUnit. The bootstrap is for PHPUnit to load files OUTSIDE the tests directory.
it's the <testsuites> that loads tests. not PSR-4 Loader.
3) About IDE, no idea, i always run stuff on the command line.

You have some valid points. but i don't think this way would cover most of developers.

I'll keep my vote to same tests and app namespace, no PSR-4. It would not hurt me, but i don't honestly see a point of autoloading tests since you're not instantiating them directly.

Let's see what the other folks have to say.

hernandev:
P.S. About the "duality", if you don't declare the tests under PSR-4, it works fine on same namespace and makes tests shorter, easier. That's my point on using the same namespace.

hernandev:
Sent PR #3563 as PoC that PSR-4 is not necessary and that Tests and App can, indeed, have the same namespace.

Zizaco:
`App\Tests` namsepace seems wierd. I don't like having separate test namespaces.

@lucasmichot see http://stackoverflow.com/a/12141610/1063081

rkgrep:
I think, this discussion leads to nowhere. There are different pros and cons for both versions, and framework does not force you to use specific method. You can override it as you want just as you override most of initial project configuration.

lucasmichot:
@taylorotwell @GrahamCampbell @crynobone @JosephSilber thoughts?

crynobone:
> I think, this discussion leads to nowhere. There are different pros and cons for both versions, and framework does not force you to use specific method. You can override it as you want just as you override most of initial project configuration.

Totally agree with this, regardless which option we choose there will always be a number of users that would prefer the alternative options. So why bother changing the default.

GrahamCampbell:
> So why bother changing the default.

I totally agree. The default we have is fine, and changing it is very trivial.

shehi:
@Zizaco : I disagree. Just another unnecessary ideology from perfectionists. 

And I tend to agree with @rkgrep , some guys here appear to talk for the sake of talking, instead of moving these discussions forward. We are not developing nuclear reactor monitoring tool for God's sake.

Overall, either way is fine: both default and the current PR state. Guys up top, please either decline or merge it and let's be done with it. Thanks @lucasmichot again for this PR and your time.

PR #3561, opened 2015-11-18
Removed reference to nonexistant Policy

tuxd00d:
'App\Policies\ModelPolicy' does not exist on this branch (or any other at this time).

### DIFF
--- a/app/Providers/AuthServiceProvider.php
+++ b/app/Providers/AuthServiceProvider.php
@@ -13,7 +13,7 @@ class AuthServiceProvider extends ServiceProvider
      * @var array
      */
     protected $policies = [
-        'App\Model' => 'App\Policies\ModelPolicy',
+        //
     ];
 
     /**


GrahamCampbell:
:-1:

GrahamCampbell:
Maybe just comment that out rather than removing it, so people can still see how to add policies from the code.

shehi:
-1: Agreed with @GrahamCampbell

rkgrep:
:-1: This does not cause any errors

taylorotwell:
It's just an example to show you what the syntax looks like.

tuxd00d:
Leaving it as an example is also a good idea.  @GrahamCampbell 's suggestion of commenting may be helpful.
```php
     protected $policies = [
       //  'App\Model' => 'App\Policies\ModelPolicy',
       //  Post::class => PostPolicy::class,
     ];
```

I felt the pull request was beneficial as the fresh app skeleton does an excellent job ( @taylorotwell ) of being slim and functional out-of-the-box.  Although it doesn't produce an error ( @rkgrep ), this particular line caused me to consider that I botched the upgrade to 5.1.11 because I couldn't find the referenced policy file.  I was hoping to help prevent the same confusion for others.

RomainLanz:
@tuxd00d This is just an example, like the event/listener one.

https://github.com/laravel/laravel/blob/master/app/Providers/EventServiceProvider.php#L15

PR #3563, opened 2015-11-19
Tests namespacing

hernandev:
Alternative approach to PR #3560 .

PSR-4 for tests is unnecessary, since the files are loaded by PHPUnit, not Composer autoload.

Keeping the same namespace between code and tests makes tests easier and shorter.

### DIFF
--- a/tests/Example/ExampleTest.php
+++ b/tests/Example/ExampleTest.php
@@ -1,5 +1,8 @@
 <?php
 
+namespace App\Example;
+
+use App\TestCase;
 use Illuminate\Foundation\Testing\WithoutMiddleware;
 use Illuminate\Foundation\Testing\DatabaseMigrations;
 use Illuminate\Foundation\Testing\DatabaseTransactions;
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -1,6 +1,11 @@
 <?php
 
-class TestCase extends Illuminate\Foundation\Testing\TestCase
+namespace App;
+
+use Illuminate\Contracts\Console\Kernel;
+use Illuminate\Foundation\Testing\TestCase as FrameworkTestCase;
+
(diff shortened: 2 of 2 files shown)


lukzgois:
I agree with this approach :+1:

vinicius73:
:+1:  I like that.

dlimars:
:+1: I agree

GrahamCampbell:
:-1:

GrahamCampbell:
Tests should 100% get their own namespace.

rkgrep:
:-1: I agree with Graham.

ChadTaljaardt:
:-1:  I agree with Graham on this one aswell, it makes more sence to me if they had their own namespace.

GrahamCampbell:
The tests folder should definitely be autoloaded too if we want to go down this route.

hernandev:
@GrahamCampbell  "Tests should 100% get their own namespace." I don't really see the point, but i want to learn why.

GrahamCampbell:
In the same way you'd want to namespace anything?

hernandev:
But we are namespacing, just using the same namespace as the code being tested.
Anyway, except the easy importing (mapping) of the classes, Namespaces on tests still useless. since you can create multiple subfolders as the same you have on your code without any problems.

Using the same is the only benefit i see on it. and PSR-4 still useless.

For Reference: https://github.com/laravel/framework/blob/5.1/tests/

I'm not gonna fight this. It's a really simple change that i would have to do in my projects to use it my own way, we are just discussing what is gonna be the standard for the app.

Using the same namespace save me some time while writing, but of course it has it's disadvantages.

The only thing i would suggest, if you're going the PSR-4 route, use require-dev so the tests don't grow the autoloader in production environments.

hernandev:
oh the original pull request already have the autoload-dev, so i'm good with it

PR #3564, opened 2015-11-19
Change redirect when authenticated

martinbean:
The route **/home** doesn’t exist in a default Laravel application, whereas **/** does.

### DIFF
--- a/app/Http/Middleware/RedirectIfAuthenticated.php
+++ b/app/Http/Middleware/RedirectIfAuthenticated.php
@@ -35,7 +35,7 @@ class RedirectIfAuthenticated
     public function handle($request, Closure $next)
     {
         if ($this->auth->check()) {
-            return redirect('/home');
+            return redirect('/');
         }
 
         return $next($request);


surfer190:
If you check the [docs](http://laravel.com/docs/5.1/authentication#authenticating), you can put this in the AuthController:

```
protected $redirectPath = '/dashboard';
```

that will change the location after a successful login (The default being `/home`).
However, it doesn't actually use this `$redirectPath` value. It is hard-coded to `/home`:

```
public function handle($request, Closure $next)
    {
        if ($this->auth->check()) {
            return redirect('/home');
        }

        return $next($request);
    }
````

arcanedev-maroc:
:+1:  for @martinbean

martinbean:
@surfer190 I know this, but in a new, clean Laravel installation, the route **/home** doesn’t exist, therefore it doesn’t make sense to redirect there on successful authentication.

tux-world:
:100: for @martinbean

RomainLanz:
:+1:

I do that every time I begin a new project.

mydnic:
:+1:

surfer190:
Yes but that is the point...you need to create the `home` route. Laravel assumes that the authenticated home screen, is different from the public root of the site. That is why `$redirectPath` exists, so you can easily override that path. But that does not work...

shawnlindstrom:
It's a built in penalty (I would call it a feature honestly)  for not RTFM at: http://laravel.com/docs/5.1/authentication#included-authenticating

When you forget on a new project it's really easy to identify the issue because you get immediate, obvious feedback. If you are new to Laravel, the docs or a very simple Google search of 'laravel redirect home on login' will straighten you out quickly.

The suggested change wouldn't provide any immediate feedback which would actually be more confusing. Additionally, some will want the redirect to / others will want something else. Most, if not all, will not route to /home on login.

martinbean:
:tada:

PR #3565, opened 2015-11-20
Fix PSR-2 standard on User model

josebailo:
According to PSR-2 standard, point 4.1 about extends and implements:

> "Lists of implements MAY be split across multiple lines, where each
> subsequent line is indented once. When doing so, the first item in the
> list MUST be on the next line, and there MUST be only one interface per
> line.", so User model implements are wrong.

This change should also be made in Laravel documentation pages.


crynobone:
It's already fixed for 5.2 release. <https://github.com/laravel/laravel/blob/develop/app/User.php>

PR #3566, opened 2015-11-21
Update `config.key` default value

n1c:
The `SomeRandomString` value is too short for `AES-256-CBC` so it errors out when no `APP_KEY` is available.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -78,7 +78,7 @@ return [
     |
     */
 
-    'key' => env('APP_KEY', 'SomeRandomString'),
+    'key' => env('APP_KEY', 'SomeRandomStringSomeRandomString'),
 
     'cipher' => 'AES-256-CBC',
 


rkgrep:
It is supposed to be short, so users have to replace it. Otherwise there will be a plenty of projects with this default not replaced.

GrahamCampbell:
It is intentionally the wrong length to force you to set a random one.

alexleonard:
This again! We really should change this to 

YouMustGenerateAUnique32CharacterStringHere

Or something. I'm losing track of the number of times people have submitted a pull request to change this.

GrahamCampbell:
> I'm losing track of the number of times people have submitted a pull request to change this.

So? Doesn't meant we should.

GrahamCampbell:
> This again! We really should change this to

No we shouldn't! Doing that would be a massive security issue. We 100% require people to set their own key. Failing to do so is like not using encryption at all.

alexleonard:
@GrahamCampbell Hey, sorry I didn't mean that we should submit to the incorrect change request, no matter how many times its made :)

I mean we should still have the default key not 32 characters long so that it throws an error if not changed, but make the text clearer so that people might think before submitting a pull request (unless a key more than 32 characters wouldn't throw the error).

That or put a comment in above the key like:

// This key is purposefully set to less than 32 characters by default because, for security, you must generate your own unique key

Of course that comment would be meaningless for people that use composer create-project or the laravel installer as you get a key generated for you, but at least it might stop people submitting the same pull request over and over ;)

GrahamCampbell:
> I mean we should still have the default key not 32 characters long so that it throws an error if not changed, but make the text clearer so that people might think before submitting a pull request (unless a key more than 32 characters wouldn't throw the error).

Short answer. No.

alexleonard:
Ok, cool. I'll avoid contributing to the email noise if it crops up again :)

n1c:
And what about leaving it empty? At least then a new install works the `php artisan key:generate` command that composer fires. And then we aren't stuck with a non-sane default?

rkgrep:
`php artisan key:generate` puts the key into `.env` file. It may not work if you have `putenv()` function disabled in php.

JulienTant:
@GrahamCampbell strikes again :D

- John doe : "Blabla with some real arguments like having an error message which will help users"
- Graham : No i don't give a shit and i won't argue cause i can't.

Ifnot:
@JulienTant :+1:

shehi:
Laravel can't bother with disabled functions or safe mode really. And if developer uses this framework, they should read installation guide in docs - it's written there clearly why that key needs updating. I agree there should be a comment saying why it is short by default etc, but that can't be a requirement in any way. Overall, I am with @GrahamCampbell in this one.

JulienTant:
This ain't about disabling anything, it's about having a descriptive error message for this case, so the guy know directly what's the issue.

Well, this is not the topic of this issue, and that particular issue deserved to be closed ;-)

GrahamCampbell:
I tried to have this done a while ago, but it turned out to be a stupid idea, and we ended up reverting the changes.

GrahamCampbell:
We're happy with the current situation, and don't intend to change it. It's secure out of the box.

PR #3570, opened 2015-11-25
Remove unnecessary fully quantified name

anik786:
The class is unnessarily fully quantified and is an warning that is also flagged up in popular IDE's such as PHPStorm. It is also arguably bad practice as it can lead to problems if directories are moved in the future.

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -13,7 +13,7 @@ class Kernel extends ConsoleKernel
      * @var array
      */
     protected $commands = [
-        \App\Console\Commands\Inspire::class,
+        Commands\Inspire::class,
     ];
 
     /**


vlakoff:
I don't really like the "split namespace",
would PhpStorm choke on `use App\Console\Commands\Inspire`?

Also, I would rename the class to `InspireCommand`,
so we have `InspireCommand` in the code which is more explicit,
and consistency with the [core commands](https://github.com/laravel/framework/tree/5.1/src/Illuminate/Foundation/Console).

anik786:
I agree with the renaming Inspire to InspireCommand, but I believe there is talk to remove it altogether

anik786:
I added a pull request based on our suggestion: https://github.com/laravel/laravel/pull/3574

anteriovieira:
I was wondering if it is not the case of passing the contents of the ```Inspiring::quote()``` function to the inspiring command. And completely remove the inspiring file ``` Illuminate/Foundation/Inspiring.php ```.

PR #3574, opened 2015-11-26
Rename Inspire to InspireCommand to be consistent with core commands

anik786:
Make Inspire consistent with core commands: https://github.com/laravel/framework/tree/5.1/src/Illuminate/Foundation/Console


anik786:
Why was this merge closed? I thought it was a fair change that made sense and consistent with the rest of the framework.

taylorotwell:
The command suffix is redundant since we're in the commands namespace.

anik786:
Although I agree with you, this is inconsistent with the rest of the framework. 

For example consider the controllers: "AuthController" and "PasswordController" which sits in the "Controller" directory. 

Or consider the "AppServiceProvider", "AuthServiceProvider", "EventServiceProvider", "RouteServiceProvider" in the "Providers" directory.

Therefore I think consistency is more important.

However, with that said I think you have a point and I would suggest removing the words "Provider" and "Controller" from these classes.

anik786:
But consistency beats shorthand

taylorotwell:
No. I am not obligated to name it consistently with the framework core commands, whose class names aren't even exposed to the user, if I no longer feel like that is the best way to name commands.

anik786:
fair enough

PR #3575, opened 2015-11-26
Removed the hard-coded paths from the Auth Middlewares

shiruken1:
Hey y'all!

First PR :)

I believe these few changes make the system more flexible, as the login and redirect paths can now be set in the AuthController, along with all the other paths.

Please let me know if I missed something obvious in the docs, or if there's a more elegant way to do this.

Happy thanksgiving!

### DIFF
--- a/app/Http/Controllers/Auth/AuthController.php
+++ b/app/Http/Controllers/Auth/AuthController.php
@@ -62,4 +62,23 @@ class AuthController extends Controller
             'password' => bcrypt($data['password']),
         ]);
     }
+    /**
+     * Returns the login path declared, or 'auth/login'
+     *
+     * @return string
+     **/
+    public function getLoginPath()
+    {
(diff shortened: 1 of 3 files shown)


lukepolo:
Can't you just do this 

From the auth docs 

When a user is successfully authenticated, they will be redirected to the /home URI, which you will need to register a route to handle. You can customize the post-authentication redirect location by defining a redirectPath property on the AuthController:

protected $redirectPath = '/dashboard';

GrahamCampbell:
Hi @shiruken1. Thanks for the PR, but I think this is something best suited for your own app. Also, in future, please review our code style guide. StyleCI provided a diff for the things wrong in this PR for reference: https://styleci.io/analyses/qBldg8.

shiruken1:
Because the redirect past never worked and this way you also get to
redirect when they are not authorized.

PR #3576, opened 2015-12-01
added queue prefix to match the framework

pulkitjalan:
Updated queue config file to match sqs prefix changes in the framework

### DIFF
--- a/config/queue.php
+++ b/config/queue.php
@@ -53,7 +53,8 @@ return [
             'driver' => 'sqs',
             'key'    => 'your-public-key',
             'secret' => 'your-secret-key',
-            'queue'  => 'your-queue-url',
+            'prefix' => 'https://sqs.us-east-1.amazonaws.com/your-account-id',
+            'queue'  => 'your-queue-name',
             'region' => 'us-east-1',
         ],
 


taylorotwell:
Can you send this to `develop` branch instead?

fideloper:
That URL also assumes `us-east-1`, which is probably OK, but not necessarily obvious to users new to aws (maybe use a placeholder one there too?).

---

Other random note: SQS provides a way to get the URL of the queue by queue name (http://docs.aws.amazon.com/aws-sdk-php/v3/api/api-sqs-2012-11-05.html#getqueueurl), which I mention just in case that's a preferred way to drop the need for the 'prefix'. 

Then perhaps the code could detect if a url is used (`preg_match("/(https?\:\/\/)/", $input_line, $output_array);`) and decide to make that call if not.

(although since that makes extra api calls, hard-coding that is probably "better", so ignore that dump of bile, since it's just making extra work).

pulkitjalan:
@fideloper I mostly copied the iron config, it looks like that also has the same `us-east-1` region in the url. Yea the extra api call was the reason the function is not used.

PR #3579, opened 2015-12-01
Windows support & redis config via env

SyntaxC4:
## Windows Support
* Run `.htaccess` file through [htaccess to web.config converter](http://htaccesstowebconfig.com/) to get `web.config`.

## Redis Config
* `.env`: REDIS_HOST, REDIS_KEY, REDIS_PASSWORD
* `Config/Database.php`: env() of the above values.


GrahamCampbell:
:-1: for adding the extra config file

SyntaxC4:
@GrahamCampbell It's important for people that want to run Laravel on Windows. Do you have any suggestions on how we could enable people to deploy to Windows without the extra config?

GrahamCampbell:
> It's important for people that want to run Laravel on Windows.

Not really?

SyntaxC4:
@GrahamCampbell Care to elaborate?

taylorotwell:
One thing I'm a little concerned about is that that Redis password setting. Predis will check `isset` on that and see an empty string instead of null and attempt to authenticate with Redis with no password.

SyntaxC4:
@taylorotwell Ok - I'll work on getting that set to null. Thanks!

RomainLanz:
Maybe it would be better to only put the `web.config` file into the documentation and not add it to the core (or create a special branch for it)?

.env.example
```diff
@@ -17,3 +17,7 @@ MAIL_PORT=2525
 MAIL_USERNAME=null
 MAIL_PASSWORD=null
 MAIL_ENCRYPTION=null
+
+REDIS_HOST=localhost
+REDiS_KEY=
```
RomainLanz: case

SyntaxC4:
@RomainLanz Thanks for the spot on the case :+1:

public/web.config
```diff
@@ -0,0 +1,23 @@
+<configuration>
```
lucasmichot: `phpunit.xml` is using 4 spaces for indentation. The same should apply here I guess

JulienTant:
// unconstructive comment
@SyntaxC4 Look's like you've been GrahamCampbelled

vlakoff:
Strong :-1:. 

If I understand correctly, this file is not for Windows, but only for IIS.

IIS is the crappiest web server ever, from far, but let's put this aside.
If people have to work on IIS, I assume they should already know how to reuse an existing `.htaccess`.

This PR just adds some very confusing clutter.
Aaaand possibly exposing a config file on many servers, doesn't sounds fine to me.

vlakoff:
@taylorotwell May I ask for your opinion on this?

crynobone:
`web.config` could be useful for those working with Microsoft Azure https://ntotten.com/2012/06/07/10-things-about-windows-azure-web-sites/

public/web.config
```diff
@@ -0,0 +1,23 @@
...
+          </conditions>
+          <action type="Redirect" redirectType="Permanent" url="/{R:1}" />
+        </rule>
+        <rule name="Imported Rule 2" stopProcessing="true">
+          <match url="^" ignoreCase="false" />
+          <conditions>
+            <add input="{REQUEST_FILENAME}" matchType="IsDirectory" ignoreCase="false" negate="true" />
+            <add input="{REQUEST_FILENAME}" matchType="IsFile" ignoreCase="false" negate="true" />
+          </conditions>
+	  <action type="Rewrite" url="index.php" />
```
vlakoff: Line 18 has an unforeseen tab.

marktopper:
:-1: This might give a problem when Redis does not have a password, since `predis` uses `isset` when setting connection attributes, so having `password` set to `null` will still be sent as a password for `redis`.
I think Windows Azure users might have to do those settings manually.

public/web.config
vlakoff: Corrected in "5.2" branch by #3613.

PR #3580, opened 2015-12-02
Create a new middlware instance instead of filter.

jspekken:
Shouldn't this be middleware?

### DIFF
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -15,7 +15,7 @@ class Authenticate
     protected $auth;
 
     /**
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Yes. Thank you. :)

app/Http/Middleware/Authenticate.php
```diff
@@ -15,7 +15,7 @@ class Authenticate
     protected $auth;
 
     /**
-     * Create a new filter instance.
+     * Create a new middelware instance.
```
caouecs: Hello,
middleware

PR #3581, opened 2015-12-03
Make it possible to configure the log level

kielabokkie:
Added a setting to the app.php config file for configuring the log level. That way you can easily prevent your production logs from filling up with debug or info logs if you don't want to.


taylorotwell:
Take a look at the `configureMonologUsing` method on the app instance.

PR #3586, opened 2015-12-06
Update .env.example

NogaevPN:
Add DB_CONNECTION setting, useful for another database (not mysql)  users


arrilot:
:-1: 
It's not really environment specific.
Just change it config, or tweak your own .env if you like it.

PR #3587, opened 2015-12-06
update default `APP_KEY` to 32 character string

browner12:
the cipher needs a 32 character string, which the documentation clearly states.

The default is still 16 characters, which was used for the old cipher.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -78,7 +78,7 @@ return [
     |
     */
 
-    'key' => env('APP_KEY', 'SomeRandomString'),
+    'key' => env('APP_KEY', 'SomeRandomStringWith32Characters'),
 
     'cipher' => 'AES-256-CBC',
 


arrilot:
That's intended and was proposed to change many-many times before.

Please use search for more info.

GrahamCampbell:
Maybe we should just totally remove the example?

browner12:
sorry about that. 

IMO the string should either be valid (32 characters), or null, which **forces** the user to utilize the environment variable then. putting in an invalid string seems to add a lot of confusion.

here's my issue though, with either an invalid or null string, when you do an initial deploy, on Forge for example, you don't have a valid `.env` file, so it fails because of the invalid key.  so then i'll switch the default to a 32 character string, and then run my `key:generate` after my initial deploy.

it's fine on any subsequent deploys, since now I have a valid `.env`, but that first one is an issue.

is there another solution that people use for this?

GrahamCampbell:
> which forces the user to utilize the environment variable then. putting in an invalid string seems to add a lot of confusion.

Yes. We don't want people using the default encryption key. Doing that, they might as well just make their root password "password" and email their ip to everyone on github.

:trollface:

browner12:
if that's the intention, I would agree with removing it completely.

how do you solve this issue then on an initial deploy, when your `.env` doesn't exist yet?

we *could* suggest generating a default key to replace `SomeRandomString` (which would then live in source control), and then **also** generate a new key in your `.env` file.  but that seems a little like kicking the bucket, and could/would cause issues when users forgot to regenerate a new key on their servers.

browner12:
2 possible solutions, let me know your thoughts.

1. add a `pre-install-cmd` to composer to `php artisan key:generate`. this would occur after the `clone` or `pull`, but before anything that relied on a valid key.  one side effect would be that a new key is generated for every deploy, although I'm not sure if there is a down side to that.

2. Add another 'Installation Option' when installing a new repository on Forge. Currently the only option is 'Install Composer Dependencies'. We could add another one called 'Generate Encryption Key'. Then the deploy script would simply run `php artisan key:generate` at the appropriate time if checked.  It's great that we have the ability to edit our deploy script afterwards, we just have less control on this initial one.

Obviously option 2 requires some outside input and doesn't directly affect this repository.

Thanks.

browner12:
any more thoughts on this @GrahamCampbell ?

PR #3588, opened 2015-12-09
[5.1] Used dependencies in an alphabetical order

tyloo:
Hey folks!

I thought it would be nice to have our `use` statements ordered alphabetically. What do you guys think?

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -4,9 +4,9 @@ namespace App\Exceptions;
 
 use Exception;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
+use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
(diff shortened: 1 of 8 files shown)


mydnic:
:+1: My IDE automatically does that on every project files when applying PSR-2 rules

GrahamCampbell:
Our code style is to sort by length.

GrahamCampbell:
> My IDE automatically does that on every project files when applying PSR-2 rules

This has nothing to do with PSR-2.

tyloo:
https://github.com/php-fig-rectified/fig-rectified-standards/blob/master/PSR-2-R-coding-style-guide-additions.md#use-declarations

This is a lie then ?

eko:
:+1: for this PSR-2 rule, I think @tyloo is right.

mydnic:
@tyloo This : https://github.com/php-fig-rectified/fig-rectified-standards is not the official PSR styleguide. It's a "rectified" version.

The alphabetical order has indeed nothing to do with PSR-2, it's just the phpfmt package for Sublime Text that applies this rule.

GrahamCampbell:
This is NOT PSR-2.

PR #3589, opened 2015-12-09
Allow defined routes/ips to bypass the down command

leemason:
When the app is down no requests get through (by design).

However there are cases where it would be ideal to keep certain urls/ips/etc open during this period, for example:

* Webhook/IPN callbacks from payment providers
* Admin users when upgrading
* App setup when installed

Plus im sure there are many other use cases.

This middleware simply extends the original one (keeping backwards compatibility), and allows you to provide an array of excluded routes and ips.

There is also a helpful comment to aid users in adding their own logic for situations like admin users.

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -2,6 +2,7 @@
 
 namespace App\Http;
 
+use App\Http\Middleware\CheckForMaintenanceMode;
 use Illuminate\Foundation\Http\Kernel as HttpKernel;
 
 class Kernel extends HttpKernel
@@ -12,7 +13,7 @@ class Kernel extends HttpKernel
      * @var array
      */
     protected $middleware = [
-        \Illuminate\Foundation\Http\Middleware\CheckForMaintenanceMode::class,
+        CheckForMaintenanceMode::class,
     ];
 
     /**
--- a/app/Http/Middleware/CheckForMaintenanceMode.php
+++ b/app/Http/Middleware/CheckForMaintenanceMode.php
@@ -46,7 +46,7 @@ class CheckForMaintenanceMode extends Original
 
             $route = $request->route();
 
-            if($route instanceof Route){
+            if ($route instanceof Route) {
 
                 if (in_array($route->getName(), $this->excludedRoutes)) {
                     return $response;
--- a/app/Http/Middleware/CheckForMaintenanceMode.php
+++ b/app/Http/Middleware/CheckForMaintenanceMode.php
@@ -35,9 +35,7 @@ class CheckForMaintenanceMode extends Original
      */
     public function handle($request, Closure $next)
     {
-
         if ($this->app->isDownForMaintenance()) {
-
             $response = $next($request);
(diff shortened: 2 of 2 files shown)


lucasmichot:
@leemason 
:+1: this feature is indeed missing and we should give the possibility to bypass/override this maintenance mode
:-1: the rules in your code are based on ips or routes, but many other custom bypass rules could be set by the user, ips ranges, header token and so on. A package already exists : https://github.com/MisterPhilip/maintenance-mode

anteriovieira:
:-1: This feature can be left to the programmer to implement.

GrahamCampbell:
:-1: from me too, for the same reason as @anteriovieira.

leemason:
"We sweat the small stuff to help you deliver amazing applications."

Isnt the point of the framework to make it as rapid as possible?

Everything can be implemented by the programmer but the beauty of laravel is majority of boilerplate stuff is handled for you in a really nice way.

granted there are more advanced versions of this as @lucasmichot has referenced (additional checks can be added as per the code comments), but at the cost of 2kb to provide more features with no backwards compatibility problems im struggling to see why this wouldn't be a good addition considering the many use cases?

djallits:
@leemason Yes it is to make development rapid, but not at the cost of bloat. A feature to one developer is just a debt to another developer.

leemason:
@djallits totally agree with bloat, but i dont see how theres any bloat, or how this would effect anyone not using its features, just like the current middlware it requires no interaction if you dont want to?

lucasmichot:
Don't misunderstand me/us @leemason it's a valuable PR, but the question I would wonder for Middleware PR on laravel/laravel is:
- would at least 50% of the user would use this middleware with overriding : not sure in this case
- would at least 50% of the user would use this middleware without overriding : not sure either

I definitely thing you could create a package or even submit some improvements to https://github.com/MisterPhilip/maintenance-mode (it does not seem to accept IP ranges for instance, or header token)

ed-fruty:
It's very specific cases for maintenance mode for every application.
In your case you only throw the exception to show a view about maintenance... but actions will work (all database queries will execute, etc).

```php
if ($this->app->isDownForMaintenance()) {
    $response = $next($request);
    //...
}
```

And what about if my database will be down or other scenarios?
For example I don't want execute any actions code at all in my maintenance mode.

And as it was said above, the rules for maintenance mode can be very different.

GrahamCampbell:
I think you;re best controlling this at a level higher up than your laravel app anyway,

imanghafoori1:
This nice feature is a missing one.
Whether it is going to be built into the core or live as a separate package.
@leemason 
But better to start as a package and after a while it shows it's own value to be included into the core or not...

imanghafoori1:
Laravel should have a style guide about how to contribute to the core and what the core team is interesting to the and what is not.
The main goals of the project should be clear for people before they submit their code as pull requests.

taylorotwell:
Holding off on this.

PR #3590, opened 2015-12-11
Differentiate between `exists` and `in` validation messages

tdhsmith:
The `in`/`not_in` rules have the exact same failure message as `exists` when they generally have quite distinct usage. It's unlikely many people would ever use both rules on one field, but as it stands quick identification is still more challenging than necessary.

I'm not too attached to any particular message, just that it's distinct, so I tried to keep the new one neutral and general. (I would've suggested something about "references" because I use `exists` exclusively for foreign keys, but I know others use it differently...)

Obviously there are probably project terminology guidelines I don't know about. :stuck_out_tongue:


taylorotwell:
I don't think this would always make sense as an error message. You can change it for your application if it does.

PR #3592, opened 2015-12-18
Use single line implements for PSR2

w0ng:
Use single line implements for <120 character class definition for PSR2.

```
phpcs --standard=PSR2 app/User.php

13 | ERROR | [x] The first item in a multi-line implements list must
   |       |     be on the line following the implements keyword
14 | ERROR | [x] Expected 4 spaces before interface name; 36 found
15 | ERROR | [x] Expected 4 spaces before interface name; 36 found
```

### DIFF
--- a/app/User.php
+++ b/app/User.php
@@ -10,9 +10,7 @@ use Illuminate\Contracts\Auth\Authenticatable as AuthenticatableContract;
 use Illuminate\Contracts\Auth\Access\Authorizable as AuthorizableContract;
 use Illuminate\Contracts\Auth\CanResetPassword as CanResetPasswordContract;
 
-class User extends Model implements AuthenticatableContract,
-                                    AuthorizableContract,
-                                    CanResetPasswordContract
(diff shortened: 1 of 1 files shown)


crynobone:
https://github.com/laravel/laravel/pull/3565#issuecomment-158398201

GrahamCampbell:
Still a :+1: from me to have this in 5.1, since 5.1 is LTS.

crynobone:
And I thought I just saw you say Laravel doesn't follow ALL the PSR-2 rules :smile_cat:

GrahamCampbell:
I said there was one exception regarding multiine function calls. There are no other exceptions.

GrahamCampbell:
Can I just point out that PSR-2 DOES NOT say that they have to all go on oneline. It says that if they start on the top line, they must end on it, otherwise, you must start the first implements on the next line.

GrahamCampbell:
Both of these are valid for example:

```php
class Foo implements Bar, Baz
{
}
```

```php
class foo implements
    Bar,
    Baz
{
}
```

PR #3593, opened 2015-12-19
Authenticate middleware's login path should be configurable

sepehr:
The PR improves the `App\Http\Middleware\Authenticate` middleware class:  

- Removes the non-necessary `elseif`; a cleaner more eye-comforting code.  
- Makes the middleware's login path configurable through `$loginPath` property; just like the default  `AuthController` class.

### DIFF
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -22,7 +22,9 @@ class Authenticate
                 return response('Unauthorized.', 401);
             }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
It is configurable. Just change the strings value :smile:

sepehr:
Hah :)) right. I rethought this and there's no need to make this a property.  

So here's just a codestyle PR; I think it cleans the code:  
https://github.com/laravel/laravel/pull/3594

PR #3595, opened 2015-12-19
Move clear-complied from pre-update-cmd to post-update-cmd

sumityadav:
This change will help in the situations when other dev already included a new package in the providers array and running `clear-complied` on `pre-update-cmd` triggers an exception. With this change the latest dependencies are downloaded first and then the complied files are removed and new one's are created.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,10 +32,8 @@
             "php artisan clear-compiled",
             "php artisan optimize"
         ],
-        "pre-update-cmd": [
-            "php artisan clear-compiled"
-        ],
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:-1:

GrahamCampbell:
Important to clear compiled before we update.

GrahamCampbell:
NB, this is already known to be an issue with no good solution yet.

GrahamCampbell:
https://github.com/laravel/framework/issues/9678

PR #3597, opened 2015-12-21
[5.2] Remove unused import

KennedyTedesco:


GrahamCampbell:
StyleCI didn't pick this up because we've disabled the unused import checking because of the stub classes we have in here with intentional unused imports. ;)

GrahamCampbell:
image

PR #3598, opened 2015-12-22
env to from too

emtudo:
Being using env for everything except from.
I believe it is good to use, since you can configure everything via env, if this is the will.


GrahamCampbell:
Please revert the changes to the comments, and use single quotes.

PR #3600, opened 2015-12-22
[5.2] config/mail from using env

emtudo:
Being using env for everything except from.
I believe it is good to use, since you can configure everything via env, if this is the will.

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -54,7 +54,7 @@ return [
     |
     */
 
-    'from' => ['address' => null, 'name' => null],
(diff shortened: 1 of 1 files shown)


emtudo:
Why was it rejected?
All other parameters are using "env" just from No.
Gmail for example among others do not support sending e-mail without the "from" header.
Could you tell me the only reason for me to understand?

GrahamCampbell:
Well, because it's not something environment specific. It's application specific.

emtudo:
thanks!

PR #3601, opened 2015-12-22
Disable demonstration command

vinicius73:
If the purpose of the command is to demonstrate, does not become nescessário leave it enabled by default.


mro95:
I agree. This should be at least commented.

PR #3604, opened 2015-12-23
Optimized public/index.php

kazi-mainuddin-ahmed:
Optimized public/index.php

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -19,7 +19,7 @@
 |
 */
 
-require __DIR__.'/../bootstrap/autoload.php';
+require dirname(__DIR__).'/bootstrap/autoload.php';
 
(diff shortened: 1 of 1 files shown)


public/index.php
```diff
@@ -19,7 +19,7 @@
 |
 */
 
-require __DIR__.'/../bootstrap/autoload.php';
+require dirname(__DIR__).'/bootstrap/autoload.php';
```
phecho: Why add dirname?
It does't work without change?

kazi-mainuddin-ahmed: Well... I generally trust php to do the work for backing up one directory... Let it control the One Level Jump. And also it looks more nice with using a function than /../

PR #3605, opened 2015-12-23
Remove erroneous return

nhowell:
We should not use return here because it would implicitly mean "return null;" which will not be allowed by PHP 7.1 (https://wiki.php.net/rfc/void_return_type).

Also, PhpStorm produces a warning about this return statement.


nhowell:
Never mind. @taylorotwell just pushed this same change.

PR #3607, opened 2015-12-25
Update welcome.blade.php

robennelson:
Add meta tag viewport and charset.


robennelson:
Why the request was closed?

GrahamCampbell:
The language and charset stuff is unneeded and should be omitted since we already set that stuff using http headers. The html hacks are just there if you can't do it the proper way with headers, and we can, and do.

GrahamCampbell:
Thanks anyway though.

PR #3608, opened 2015-12-26
Do not use compiled file for artisan

CurosMJ:
Use `autoload-without-compiled.php` 

This PR fixes https://github.com/laravel/framework/issues/9678

This should also be merged with master also.

### DIFF
--- a/artisan
+++ b/artisan
@@ -13,7 +13,7 @@
 |
 */
 
-require __DIR__.'/bootstrap/autoload.php';
+require __DIR__.'/bootstrap/autoload-without-compiled.php';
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Why would we not just check if we are running in the console in the autoload file?

CurosMJ:
Because the process would be complex I presume.

CurosMJ:
Another way would be to not depend on `bootstrap/app.php` in the `artisan` file.

taylorotwell:
Here's how you check for CLI usage:

```
return php_sapi_name() == 'cli';
```

taylorotwell:
Closing this for now.

PR #3609, opened 2015-12-26
we should move the default route to web group

efrontsa:
we should move the default route to web group because it's one of them!

### DIFF
--- a/app/Http/routes.php
+++ b/app/Http/routes.php
@@ -11,10 +11,6 @@
 |
 */
 
-Route::get('/', function () {
-    return view('welcome');
-});
-
 /*
 |--------------------------------------------------------------------------
 | Application Routes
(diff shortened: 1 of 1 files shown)


lagbox:
From a clean install that route does not require the use of any of the middlewares in the web group, which is why that group of middleware isn't applied.

GrahamCampbell:
We don't do this out of the box so laravel performs well in third party benchmarks.

vlakoff:
Lol. This being said, I found out the current (5.2) `routes.php` to be quite confusing.

JeffreyWay:
Yeah, I'd expect every newcomer to be confused by the routes.php file.

lagbox:
I am in favor of changing the wording that is currently in the routes.php file. Though the route file tells you via comments what is going on, it seems a bit much for newer people to infer the direct link between auth and sessions and what they do. I see this come up in support constantly.

vlakoff:
See https://github.com/laravel/laravel/commit/5c30c98db96459b4cc878d085490e4677b0b67ed (and https://github.com/laravel/laravel/commit/bcc54357bdb80c5ffcf41bdda222706b7aee5637).

PR #3611, opened 2015-12-28
Update routes.php

jpmurray:
To be honest, I'm not sure if I should or should not do this, but it's the only way I see to fix the welcome view did that did not get the session variables, thus making the authenticated option (like logged in menu) broken if I ever came back to welcome after being logged in.

### DIFF
--- a/app/Http/routes.php
+++ b/app/Http/routes.php
@@ -11,10 +11,6 @@
 |
 */
 
-Route::get('/', function () {
-    return view('welcome');
-});
-
 /*
 |--------------------------------------------------------------------------
 | Application Routes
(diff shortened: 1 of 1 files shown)


taylorotwell:
You should move this in your own app if you need the session on the main page of your app. Many front / marketing pages do not need session on the front page.

lagbox:
If you want your welcome page to be in that group, just modify it in your own routes.php.
That route by default does not require any of the middleware that is offered by the `web` middleware group.

jpmurray:
@taylorotwell Perfect, that's what I did. I just found it weird that the auth views would still offer the "login/register" menu when we're logged in, but I totally get your point!

GrahamCampbell:
It also serves as a good example of a route ourside of that group for newcomers.

PR #3612, opened 2015-12-29
Update routes.php

bradbernard:
Change from middleware array of single element 'web' to the string 'web' on Line 29.


bradbernard:
Nevermind, I just saw that middleware only accepts an array.

PR #3616, opened 2015-12-31
Flexbox update

bart:
Replaced table cell alignment by flexbox

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -11,35 +11,16 @@
             }
 
             body {
-                margin: 0;
-                padding: 0;
-                width: 100%;
-                display: table;
+                display: flex;
+                align-items: center;
+                justify-content: center;
+                font-size: 96px;
                 font-weight: 100;
                 font-family: 'Lato';
-            }
-
-            .container {
-                text-align: center;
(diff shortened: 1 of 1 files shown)


ludo237:
Are flexbox well supported?

pavankumarkatakam:
Checkout this url http://caniuse.com/#search=flexbox

mikerockett:
I don't see the purpose of this. Who keeps the welcome view anyway? Does it really matter if it uses flexbox?

bart:
In my opinion it would be a good start for some people who don't know it, doesn't it?

bart:
Why did you close my PR @taylorotwell. That's IMO not what community driven development should look like!

jonagoldman:
@bart flexbox is not supported in IE. Besides that, the welcome page is just an example you should be deleting as soon as you start working on your app. Its not a place to show the latest bleeding edge css feaures.

ludo237:
@bart So a community driven development should accept everything beside its real effectiveness only because someone has decide that is the right thing to do? Lmao

bart:
Nope @ludo237 but a bit of discussion would be great instead of instantly closing a pull request!

PR #3617, opened 2015-12-31
use class notation

ChrisEsser:


GrahamCampbell:
That code is actually broken. It's resolving as App\Providers\App\Events\SomeEvent

PR #3618, opened 2016-01-02
Fix auth/logout uri for logging out

jastew:
The middleware for the auth/logout uri is set to guest, meaning you can't logout.

Logout for only authorized users is down to personal preference.


taylorotwell:
I think it's fine. The middleware for all routes EXCEPT logout is set to guest.

PR #3623, opened 2016-01-04
Update .env.example

geekcom:
It is interesting to keep it that way is more intuitive for those who wish to make multiple connections in laravel.


taylorotwell:
This default value isn't valid.

marktopper:
Instead of `database_driver` it should properly have the default value `mysql` instead.

Just like the standard is now at the [config](https://github.com/laravel/laravel/blob/master/config/database.php#L29).

PR #3625, opened 2016-01-05
Check laravel requirements

aleixfabra:
Check laravel requirements at welcome.blade.php view.

I'm new in laravel so I don't know if you already discussed this in the past but I think it's a great help when you stard a new project.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,3013 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "This file is @generated automatically"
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
These requirements are incorrect.

aleixfabra:
Can you say me why @GrahamCampbell? 
I read it here: https://laravel.com/docs/5.2#installation

GrahamCampbell:
Those docs are incorrect. Not sure how that happened. `PHP >= 5.6.4` is only needed starting from L5.3.

aleixfabra:
So it's php 5.5.9?

GrahamCampbell:
> So it's php 5.5.9?

Yes, laravel 5.1 and 5.2 are both php 5.5.9+.

PR #3626, opened 2016-01-05
Check laravel requirements (fix https://github.com/laravel/laravel/pull/3625)

aleixfabra:
Check laravel requirements at welcome.blade.php view.

I'm new in laravel so I don't know if you already discussed this in the past but I think it's a great help when you stard a new project.

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -33,12 +33,52 @@
             .title {
                 font-size: 96px;
             }
+
+            .problem {
+                color: red;
+            }
         </style>
     </head>
     <body>
         <div class="container">
             <div class="content">
                 <div class="title">Laravel 5</div>
+
+                @if (version_compare(PHP_VERSION, '5.6.4', '>='))
+                    <p>Your version of PHP is 5.6.4 or higher.</p>
+                @else
+                    <p class="problem">Your version of PHP is too low. You need PHP 5.6.4 or higher to use CakePHP.</p>
+                @endif
+
+                @if (extension_loaded('openssl'))
+                    <p>Your version of PHP has the openssl extension loaded.</p>
+                @else
+                    <p class="problem">Your version of PHP does NOT have the openssl extension loaded.</p>
+                @endif
+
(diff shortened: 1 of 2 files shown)


taylorotwell:
The requirements should just be documented in the documentation.

resources/views/welcome.blade.php
```diff
@@ -33,12 +33,52 @@
...
     </head>
     <body>
         <div class="container">
             <div class="content">
                 <div class="title">Laravel 5</div>
+
+                @if (version_compare(PHP_VERSION, '5.5.9', '>='))
+                    <p>Your version of PHP is 5.5.9 or higher.</p>
+                @else
+                    <p class="problem">Your version of PHP is too low. You need PHP 5.5.9 or higher to use CakePHP.</p>
```
GrahamCampbell: > or higher to use CakePHP

:trollface:

aleixfabra:
:+1:

AhmetS:
Forgetting the Cake part... The fact that composer should never allow the install if the wrong PHP version was used makes the version check redundant.

resources/views/welcome.blade.php
Amrit01: :laughing:

GrahamCampbell:
> The fact that composer should never allow the install if the wrong PHP version was used makes the version check redundant.

Not true, you can pass a flag to ignore platform dependencies. Also, there's no reason to assume that this was even installed using composer. People can use the laravel installer.

AhmetS:
I thought the laravel installer used composer? 

Also if you know how to ignore platform dependencies with composer - one would assume that you are also aware that this could cause breaking changes and that you would know how to read the composer.json file which would specify the minimum required PHP version?

djallits:
My humble opinion is this is bloat, respectfully. The requirements are specified in the documentation.

resources/views/welcome.blade.php
jessiesiat: troll  :laughing:

PR #3627, opened 2016-01-07
added config item session.ignore_previous_url

hamedmehryar:
This option Must be checked in the StartSession Middleware: storeCurrentUrl
function.


GrahamCampbell:
This doesn't really belong here.

hamedmehryar:
Hi,
So where is the right place for adding this confic item?

PR #3628, opened 2016-01-12
Fix elixir installation error.

aramirez92:


GrahamCampbell:
The existing version constraint already resolves to the newest 4.x version.

PR #3629, opened 2016-01-12
Fix elixir installation error.

aramirez92:
I've fixed with this change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "gulp": "^3.8.8"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
The existing version constraint already resolves to the newest 4.x version.

LukasMa:
@GrahamCampbell  Does the same apply to bootstrap-sass? Will "^3.0.0" get 3.3.6?

GrahamCampbell:
Yes.

PR #3632, opened 2016-01-17
Fixed the AuthController.php not allowing the logout function

Quezler:
When using logout instead of getLogout, the browser just redirects, thinking it was a page a logged in user was not supposed to see, by changing it to getLogout it works as expected again.


lagbox:
`Route::auth()` sets up the auth route for logout to use `Auth\AuthController@logout`.
If your routes are using a different method you can adjust your AuthController as you have.

Quezler:
my apologies, appears i didn't notice the major auth changes from 5.1 to 5.2

PR #3634, opened 2016-01-19
Latest of 176 releases.

jovib:


LukasMa:
Take a look at #3628 and #3629

PR #3635, opened 2016-01-22
[5.2] Don't return a login page to a JSON request

JacobBennett:
Currently, any unauthorized API requests that pass through the `auth` middleware get a redirect to the login page. Adding the `wantsJson` flag to the conditional corrects this behavior.


pespantelis:
:+1: It has reported again #3460

vlakoff:
(bump) Should this be backported to 5.1?

PR #3636, opened 2016-01-23
Handle authorization header

paulvl:
Added conds and rules to .htaccess file to handle Authorization Header that are necesary to use Token Guard Authorization.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -13,4 +13,8 @@
     RewriteCond %{REQUEST_FILENAME} !-d
     RewriteCond %{REQUEST_FILENAME} !-f
     RewriteRule ^ index.php [L]
+
+    # Handle Authorization Header
+    RewriteCond %{HTTP:Authorization} .
+    RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
 </IfModule>


vlakoff:
See this question on Stack Overflow: [What does this this HTTP Authorization RewriteRule do?](http://stackoverflow.com/questions/4947548/what-does-this-this-http-authorization-rewriterule-do)

and this comment in particular:
> If there are other rewrite rules `RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]` should be added **before** them, **without the "L" option**.

Maybe this needs more investigation.

vlakoff:
ping @paulvl
Is this rule really effective, despite being after a [L] rule?

paulvl:
@vlakoff this rule is working without "L" option regardless if, like in this case, is added after any other rewrite rule because is not an URL rewriting rule, it is just assigning values to the Header. I checked this post http://webmasters.stackexchange.com/questions/71225/htaccess-rewriterule-premature-ending-of-rewrite-iteration
As I say it is working, but if I'm wrong, please confirm a use case when it fails, to ask @taylorotwell  a merging revert.

vlakoff:
Thanks for these clarifications. If it works for you, then I guess it's ok.

mikerockett:
That question on StackExchange is not actually related to @vlakoff's question as the OP there asked about ending rewriting due to the use of `-` ... I find it strange that this works, and still think it should come before the previous rule...

vlakoff:
Indeed, the above StackExchange topic doesn't directly answer my question, it gives some hints though.

If it works, that's probably because the [L] flag applies only to URL rewriting rules, the [E] flag being a separate feature of mod_rewrite.

As the code does work, it should be fine as is. Maybe confirming on apache 2.2 and 2.4, to make sure?

mikerockett:
Right, I see the difference now. So the assumption is that the rule is read anyway, but if the rule dictated that a rewrite were to occur, it would be skipped due to the previous flag.

Doubtful that the behavior has changed between 2.2 and 2.4, but a test would still be necessary.

Sent from my Windows Phone

PR #3640, opened 2016-01-28
[5.2] Make Memcached options configurable.

martinbean:
Had an issue where my employer has an application that they deploy to different targets. Memcached needed to be configurable on a per-environment basis (i.e. staging, beta, production) so having hard-coded options wasn’t ideal. Would be great if these were configurable via environment variables out of the box.

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -51,7 +51,9 @@ return [
             'driver'  => 'memcached',
             'servers' => [
                 [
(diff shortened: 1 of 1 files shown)


vlakoff:
Maybe these entries should be added to [.env.example](https://github.com/laravel/laravel/blob/master/.env.example) as well?

PR #3641, opened 2016-01-28
.env.example - Changed localhost to 127.0.0.1

jspringe:
This change was made for those developing on Windows and noticed ridiculous response times (1 second) even locally. Windows performs a DNS lookup for 'localhost' which always takes 1 second. I thought I would save the few Windows people (like myself) some time with troubleshooting slow installs of redis (this is where I first noticed it).

[http://stackoverflow.com/questions/14076403/redis-connection-slow-from-php](http://stackoverflow.com/questions/14076403/redis-connection-slow-from-php)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,7 +2,7 @@ APP_ENV=local
 APP_DEBUG=true
 APP_KEY=SomeRandomString
 
-DB_HOST=localhost
+DB_HOST=127.0.0.1
 DB_DATABASE=homestead
 DB_USERNAME=homestead
 DB_PASSWORD=secret
@@ -11,7 +11,7 @@ CACHE_DRIVER=file
 SESSION_DRIVER=file
 QUEUE_DRIVER=sync
 
-REDIS_HOST=localhost
+REDIS_HOST=127.0.0.1
 REDIS_PASSWORD=null
 REDIS_PORT=6379
 


fideloper:
I believe the root cause of this is that windows will default to trying the IPv6 network first (if it's enabled). When it fails to connect, it then goes to the IPv4 network, which is actually in use by redis/mysql/sqlserver/etc.

(Actually reading the linked SO article, they basically say that).

Knowing that fact, I don't have a strong opinion on this one way or another (not that you care, this is just to say that I'm doing a little more than "+1", based on my experience with this issue on Windows).

I'm thinking `127.0.0.1` will just about always be safe to use (vs `localhost`) for Linux people, until IPv6 finally is the default used over IPv4 (no clue how many years in the future that will be).

jspringe:
I honestly don't have a strong opinion about it either. I figured I'd just throw it out there to see if it would stick. At the very least this may one day show up in google for some other poor sap having similar issues.

lucasmichot:
:+1: too

TommyC81:
Been there, done that. Sensible default that will avoid some frustration. +1

nhowell:
Please take into consideration that MySQL on Unix will use TCP/IP instead of a socket due to this change.

userscapechris:
That's interesting, could affect  default settings on Redhat/CentOS servers
with SELinux enabled. Would need to test if default is to only allow unix
socket connection instead of allowing tcp socket connection.

jspringe:
I guess it's a measurement of the lesser of 2 evils. On one hand *nix has much more documentation and support than Windows when it comes to things like Redis. On the other *nix has a much larger user base.

vlakoff:
@nhowell @userscapechris It would be nice of you to share here if you find out any negative effect caused by this change.

PR #3642, opened 2016-01-29
Replaced "@return void" in constructor phpdocs with "@return self"

deframe:
The phpDocumentor documentation states that the **@return** tag for a constructor must either be omitted or have a description of **self**. Laravel incorrectly uses **@return void**, which causes syntax warnings in IDEs such as PHPStorm.

http://www.phpdoc.org/docs/latest/references/phpdoc/tags/return.html

I have also submitted a PR for the _laravel/framework_ repository to patch up the stub files used by artisan's scaffolding.

Hopefully I'm right in branching master for this (there doesn't appear to be a 5.2 branch?)

### DIFF
--- a/app/Http/Controllers/Auth/AuthController.php
+++ b/app/Http/Controllers/Auth/AuthController.php
@@ -33,7 +33,7 @@ class AuthController extends Controller
     /**
      * Create a new authentication controller instance.
      *
-     * @return void
+     * @return self
      */
     public function __construct()
     {
--- a/app/Http/Controllers/Auth/PasswordController.php
+++ b/app/Http/Controllers/Auth/PasswordController.php
@@ -23,7 +23,7 @@ class PasswordController extends Controller
     /**
      * Create a new password controller instance.
      *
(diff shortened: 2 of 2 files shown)


w2pc:
you might as well close this yourself because it's going to get closed anyways. This has been attempted no less than 30 times.

pietdevries94:
@w2pc What is the reason for this?

deframe:
@w2pc I see what you mean. I'll leave it open for a short period as I don't think the actual phpDocumentor documentation has been referenced in this way before. If the implied return type of a constructor is 'self' then that would seem to be what the creators expect, regardless of whether it makes sense or not.

GrahamCampbell:
Just because phpdoc.org says something, it doesn't make it true.

GrahamCampbell:
https://3v4l.org/PDpdQ

As we can see, the return type clearly is not an object...

w2pc:
@GrahamCampbell It's on the internet it has to be true!!!

PR #3643, opened 2016-01-30
create `pagination` configuration file

browner12:
hey @taylorotwell .  I was the one who submitted the PR for the Bootstrap 4 pagination presenter recently.  It sounds like you were in agreement with me that there needs to be a more flexible solution to this.

I created this branch to start thinking about how this would work ideally from an end user perspective, and came up with this config file. I started toying around with how this would work on the `laravel/framework` side, but I wanted to get this up to start the discussion, and see if this is a direction you would want to go.

thanks!

### DIFF
--- a/config/pagination.php
+++ b/config/pagination.php
@@ -1,0 +1,71 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Pagination Wrapper
+    |--------------------------------------------------------------------------
+    |
+    | Define the HTML for your pagination wrapper.
+    |
+    */
+
+    'pagination_wrapper' => '<ul class="pagination">%s</ul>',
+
+    /*
+    |--------------------------------------------------------------------------
+    | Simple Pagination Wrapper
+    |--------------------------------------------------------------------------
+    |
+    | Define the HTML for your simple pagination wrapper.
+    |
+    */
+
+    'simple_pagination_wrapper' => '<ul class="pager">%s</ul>',
+
+    /*
+    |--------------------------------------------------------------------------
+    | Available Page Wrapper
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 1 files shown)


crynobone:
Shouldn't you just be doing something likes...

```
$user->links(Illuminate\Pagination\BootstrapFourPresenter::class);
```

browner12:
i believe that would be possible, but the issue becomes that this is not a very customizable or scalable solution.  I had to go in and create the PR for the `BootstrapFourPresenter`, because it didn't previously exist.  now imagine a year down the road and BS5 comes out, and we need to make a new one.  not only are we making more work for ourselves, but we're also polluting the `Illuminate/Pagination` repo because we can't remove old Presenters in case people still rely on them.

now imagine someone wants a Jquery UI presenter, or maybe they want to write their own CSS for pagination.   The current solution just isn't flexible enough.

crynobone:
Now imagine if we have one part of the application using Bootstrap 3, and another part of the application upgraded to Bootstrap 4. With a global configuration you are now facing the difficulty of having to swap the configuration manually on runtime based on the template that the request are loading, instead of toggling via `->links()`. 

Configuration based pagination style was once used in Laravel 3, and probably changes in Laravel 4 due to this.

GrahamCampbell:
This should not be "config" because there is not situation where it would make sense to change it on per install of an app.

browner12:
i feel like someone having 2 different version of BS in their site is much more of an edge case than having different users of laravel using different frameworks or styling their own. maybe I'm wrong, but that's my experience on it. 

if you look at this [PR](https://github.com/laravel/framework/pull/11985) from the framework repo, you can see Taylor agrees that we need to make configuration of pagination easier.  This was my initial thought, but I'm open to hearing other suggestions.

PR #3647, opened 2016-02-05
Remove unnecessary "index.php/" from URLs

swvjeff:
Laravel allows duplicate URLs if the internally-rewritten URL (containing "index.php/") is accessed via the browser (e.g. https://laravel.com/index.php/docs/5.2/validation). This PR implements an htaccess redirect to remove "index.php/" from all requests. For example, https://laravel.com/index.php/docs/5.2/validation would redirect to https://laravel.com/docs/5.2/validation.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -5,6 +5,10 @@
 
     RewriteEngine On
 
+    # Remove Unnecessary "index.php/" From URLs
+    RewriteCond %{THE_REQUEST} ^GET.*index\.php [NC]
+    RewriteRule (.*?)index\.php/*(.*) /$1$2 [R=301,NE,L]
+
     # Redirect Trailing Slashes If Not A Folder...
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This doesn't nessessarly work. What if someone has `index.php` as part of their actual url.

swvjeff:
How about a RewriteRule that only matches URLs starting with "index.php/"

`RewriteRule ^index\.php/(.*) /$1 [R=301,NE,L]`

swvjeff:
@taylorotwell Any reason why you closed it? Curious if it's because of technical reasons or you just don't think the redirect is necessary?

w2pc:
I think he mentioned the reason in the first comment. What if people have
index.php in the URL on purpose?

swvjeff:
@w2pc I responded to that. It will only redirect URLs that start with *exactly* "index.php/" (including the trailing slash).

PR #3652, opened 2016-02-08
Move welcome style to a Sass file

let-aurn:
This change will the drive the user to :
- use CSS Preprocessor (modern web development best practice)
- be familiar with Elixir

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -12,5 +12,5 @@ var elixir = require('laravel-elixir');
  */
 
 elixir(function(mix) {
-    mix.sass('app.scss');
+    mix.sass('**/*.scss');
 });
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -1,2 +1,0 @@
-// @import "node_modules/bootstrap-sass/assets/stylesheets/bootstrap";
-
--- a/resources/assets/sass/welcome.scss
+++ b/resources/assets/sass/welcome.scss
@@ -1,0 +1,29 @@
+// @import "node_modules/bootstrap-sass/assets/stylesheets/bootstrap";
+
+html, body {
+    height: 100%;
+}
+
+body {
+    margin: 0;
+    padding: 0;
+    width: 100%;
+    display: table;
+    font-weight: 100;
+    font-family: 'Lato';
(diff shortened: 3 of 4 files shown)


gulpfile.js
```diff
@@ -12,5 +12,5 @@ var elixir = require('laravel-elixir');
  */
 
 elixir(function(mix) {
-    mix.sass('app.scss');
+    mix.sass('**/*.scss');
```
bbashy: This isn't best practice. At least use welcome.scss and add an @import for other partials/modules within that file.

daylerees:
Is this useful for a static page? My thought being that the majority of framework users will delete the view/route for this resource upon the start of a new project. That is, at least, in my opinion.

let-aurn:
@daylerees Yep, I did it for all my Laravel projects. But this is a kind of template showing how Sass/Less styles can be easily managed Elixir. With this, chances are that developers will use CSS preprocessors instead of plain CSS, which is the best choice. (Should show how to use CoffeeScript too?)

gulpfile.js
let-aurn: Humm. It depends on how sass files are organized. If each of your sass file stands for a style for a page, this can be better.. But yep, if you have another structure (with files containing only variables for instance) `@import` would do it better.

daylerees:
I'd argue that using a pre-processor isn't always the best choice. On the welcome page, it's 5 or so selectors. The only preprocessor feature being used is the nesting, which is more stylistic. Same with coffeescript, sometimes a preprocessor isn't required. They are definitely not bad tools, but you pick your tools for your project.

Worth noting that if this is merged, you'll need to add a `gulp` post-install command to composer, otherwise the first hitting of the welcome page will be missing the CSS.

let-aurn:
I totally agree the style for the welcome page of Laravel is not a good example to demonstrate how preprocessors work. This is just to encourage the user to find more about them and Elixir. They are syntactically amazing and should be used even on "small" projects.

A fresh installation of Laravel includes an example of route, of Eloquent class (`UserController`), of database migration, of tests and more. Let's add some for front end too? Anyway, they don't do anything else than driving developers to adopt best practices.

PR #3653, opened 2016-02-09
Add dependency injection to test classes

clarkeash:
Not sure what you think of this but it would allow users to inject their dependencies into their tests and have them resolved. Also by adding these new methods, it avoids the issue of forgetting to call `parent::setUp()` when overriding that method.

e.g.

````php
class ExampleTest extends TestCase
{
    public function before(User $user)
    {
        $this->user = $user;
    }
    
    /**
     * @test
     */
    public function it_works()
    {
        $this->assertInstanceOf(User::class, $this->user);
    }
}
````

This maybe better suited to the Illuminate TestCase class.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -22,4 +22,20 @@ class TestCase extends Illuminate\Foundation\Testing\TestCase
 
         return $app;
     }
+    
+    public function setUp()
+    {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This belongs in laravel/framework surely?

PR #3655, opened 2016-02-09
Make queue name configurable

gizburdt:
We deploy multiple sites on 1 server. When using `default` queue for all projects, the queues become dodgy. We had to change the `queue` for all projects. This is what we did in all our projects. Maybe a handy addition for other people too :)

### DIFF
--- a/config/queue.php
+++ b/config/queue.php
@@ -38,14 +38,14 @@ return [
         'database' => [
             'driver' => 'database',
             'table'  => 'jobs',
-            'queue'  => 'default',
+            'queue'  => env('QUEUE_NAME', 'default'),
             'expire' => 60,
         ],
 
         'beanstalkd' => [
             'driver' => 'beanstalkd',
             'host'   => 'localhost',
-            'queue'  => 'default',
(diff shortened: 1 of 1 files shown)


RomainLanz:
> We had to change the queue for all projects.

gizburdt:
Sorry, I don't think we understand each other correctly. What we did was we changed the `config/queue.php` like I did in this PR. And we added QUEUE_NAME to our .env. When I changed it in our projects I thought maybe more people can use this, so I created this PR.

spyric:
@RomainLanz  means that queue_name is not depending from environment. You use the same queue name on dev and prod environments

gizburdt:
Ah ok! Let's see what Taylor has to say about it.

taylorotwell:
Agree this doesn't really need to be environment specific.

gizburdt:
Thanks for your answer. In our setup it is environment specific, because we need it on our test server, but don't necessarily need it on the production server. I now understand this is very specific to our setup. :) Thanks again, keep the work coming! +1

PR #3663, opened 2016-02-18
Patch 1 - 	allow a specific guard login URL

amosmos:
Since Laravel already supports calling auth middleware with a specific guard (auth:admin), this will allow customizing each guard login url

### DIFF
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -22,7 +22,7 @@ class Authenticate
                 return response('Unauthorized.', 401);
             } else {
                 $loginUrl = config("auth.guards.{$guard}.loginUrl", 'login');
-                
+
                 return redirect()->guest($loginUrl);
             }
         }
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -21,7 +21,8 @@ class Authenticate
             if ($request->ajax() || $request->wantsJson()) {
                 return response('Unauthorized.', 401);
             } else {
-                return redirect()->guest('login');
+                $loginUrl = config("auth.guards.{$guard}.loginUrl", 'login');
+                return redirect()->guest($loginUrl);
             }
         }
 
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -22,6 +22,7 @@ class Authenticate
                 return response('Unauthorized.', 401);
             } else {
                 $loginUrl = config("auth.guards.{$guard}.loginUrl", 'login');
+                
                 return redirect()->guest($loginUrl);
             }
         }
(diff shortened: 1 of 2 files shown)


amosmos:
Sorry I don't understand the StyleCI error...

lagbox:
looks like you left unneeded whitespace on that line

amosmos:
ok I think you were right :-)

scrubmx:
https://github.com/amosmos/laravel/blob/patch-1/config/auth.php#L42

I think `loginUrl` should be in snake case to be more consistent with other configuration files.

amosmos:
To be honest I thought so too but then I noticed all customizing auth variables are written in cameCase like loginView, registerView and linkRequestView, so I thought this is how Taylor meant for it to be...

But of course I wouldn't mind changing it to whatever case it needs to be...

taylorotwell:
You can make this change in your own application if you want.

amosmos:
Hey Taylor,

Sure I can make it in my own app but wouldn't this assist a true multi-auth feature?

Right now if you really want multi auth you must make separate middlewares for each guard, just because of the different login url, even though you do already allow specifying a guard parameter.

So just this little change will support it - while of course keeping the default 'login' so everything works the same out of the box even if no specific URL was set in the config.

I guess I just wonder why you think it's not needed in laravel out of the box.

Thanks,
Amos

amosmos:
What I mean is that this change keeps loginUrl optional, if none is specified the 'login' will be used just like it is today.

danieliancu:
I think this is great but needs more thinking.

Named routes should be catered for as well and config may not be the best place for defining guard routes, but I can't come up with something better now.

danieliancu:
I think you can discuss it depth on Slack http://larachat.co

amosmos:
Maybe you could let the guard controller handle guests redirect? Than you could set the loginUrl in the auth controller, much like the redirectTo variable is set today (and others: loginView, registerView etc).

And of course this would be optional so if someone wants their own redirect handler logic they can just write it in the middleware like today.

amosmos:
something like:

Auth::guard($guard)->redirectGuest()

amosmos:
@danieliancu about named routes - I just went with the convention Taylor did on customizing auth routes - for example $redirectTo variable expects a URL and not a named route.

danieliancu:
@amosmos that is intentional because you can easily set that with `route()`, but in config you don't have access to it because the routes have not been registered yet.

amosmos:
oh great idea! i didn't think of it at all. yes you are correct about that

amosmos:
Although for some reason I am getting an error when trying:
`protected $redirectTo = route('user.favorites');`

danieliancu:
@amosmos set the value of the property to the named route and then call `route($this->redirectTo)` from the constructor to update its value.

amosmos:
Oh, in that case I still have to manually set the controller to work with named routes.

So if it's my choice and needs a manual action by me to work, you can say the same thing about my suggestion for loginUrl...

Just set the config to the named route, and then call `route(loginUrl)` right in the next line in the middleware...

PR #3664, opened 2016-02-19
Guest middleware can be called with specific guard

amosmos:
And so shouldn't it be called with the same guard this controller is using?

Not sure if my syntax is the best though...

### DIFF
--- a/app/Http/Controllers/Auth/AuthController.php
+++ b/app/Http/Controllers/Auth/AuthController.php
@@ -37,7 +37,7 @@ class AuthController extends Controller
      */
     public function __construct()
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
What is guard is null?

amosmos:
I thought I saw somewhere in the code that if no guard is specified the than the default guard (from the config) is used.

But maybe this isn't being done at that point in the controller.

I'm sure though there is a solution for this no?

Maybe:
$this->middleware('guest'.(($this->guard) ? ":{$this->guard}" ; ""), ['except' => 'logout']);

Amos

PR #3665, opened 2016-02-19
Guest middleware can be called with specific guard

amosmos:
But of course, it is optional and if no guard is specified then it will work just like today


taylorotwell:
I implemented a better way to do this in a recent commit. Will fix it.

amosmos:
Thanks for updating!

PR #3672, opened 2016-02-26
Add language line for the "present" validation rule.

themsaid:
Adding a language line for the newly added "present" validation rule. Refer to https://github.com/laravel/framework/pull/12487


resources/lang/en/validation.php
```diff
@@ -58,6 +58,7 @@
     'not_in'               => 'The selected :attribute is invalid.',
     'numeric'              => 'The :attribute must be a number.',
     'regex'                => 'The :attribute format is invalid.',
+    'present'              => 'The :attribute field must be present.',
```
GrahamCampbell: Not sure this is the most user friendly message.

themsaid: Any suggestions?

GrahamCampbell: D'no. Taylor is normally good at this sort of thing.

lagbox: Maybe `The :attribute field is not present.`

gaomd:
I think without *field* word should be enough, check the `numeric` line 2 lines above the added line.

taylorotwell:
I think these should also be in alphabetical order.

themsaid:
Sorted. Didn't notice the sorting in the docs and in here, will take care from now on.

PR #3673, opened 2016-02-28
Add Vietnamese default language  to Laravel

dinhquochan:
Add Vietnamese default language  to Laravel.


taylorotwell:
We don't keep community language files in the main repo. We encourage them to be shared in teh community.

PR #3676, opened 2016-03-01
Added DB_SOCKET as a default environment variable

NiclasHedam:
It should be easier to define database sockets as they are used by a lot of localhost applications.
It should be falling back to null unless it is defined.

I've added the variable in the .env.example file and changed database.php accordingly.


GrahamCampbell:
I don't think this needs changing in the core.

PR #3677, opened 2016-03-01
 Added DB_PORT as a default enviroment variable 

TGM:
Hello,

The attached paramater is required when working with MySQL homestead on a Windows enviroment.

Please take this pull request into consideration.

Regards,
Dan C.


GrahamCampbell:
What about all the other drivers?

TGM:
Looks like Taylor beat me to it.

PR #3682, opened 2016-03-02
Fixed order of password validation in registration

jadjoubran:
When you're filling the registration form and you leave the password empty:
you're going to get an error that both passwords do not match
whereas it makes sense to show the "your password must be at least 6 characters long" error first
or else a user might have a `confirmed` 5 characters password first and then he sees the `min:6` error


GrahamCampbell:
:+1:

PR #3683, opened 2016-03-10
Add DB_CONNECTION to environment

marktopper:


jadjoubran:
I guess most people are running on MYSQL that's why this remains hidden in config/database.php file.. It can be easily extracted to the `.env` file
I think it makes sense to keep it there so that the `.env` file doesn't become cluttered

robclancy:
Not sure how this is hidden in any way or form.

PR #3685, opened 2016-03-15
Allow passing multiple gaurds to the auth middleware

JosephSilber:
Using this, you'd be able to allow access to a single route for multiple guards:

```php
Route::get('foo', 'Foo@foo')->middleware('auth:web,api');
```

### DIFF
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -12,19 +12,40 @@ class Authenticate
      *
      * @param  \Illuminate\Http\Request  $request
      * @param  \Closure  $next
-     * @param  string|null  $guard
+     * @param  string  ...$guards
      * @return mixed
      */
-    public function handle($request, Closure $next, $guard = null)
+    public function handle($request, Closure $next, ...$guards)
     {
-        if (Auth::guard($guard)->guest()) {
-            if ($request->ajax() || $request->wantsJson()) {
-                return response('Unauthorized.', 401);
-            } else {
-                return redirect()->guest('login');
+        if ($this->check($guards)) {
+            return $next($request);
+        }
+
+        if ($request->ajax() || $request->wantsJson()) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think the check might should be changed to `if (! $this->authenticated())` because the `guest` method reads weird how it returns false, etc.

taylorotwell:
Thoughts?

JosephSilber:
@taylorotwell updated.

app/Http/Middleware/Authenticate.php
```diff
@@ -12,19 +12,40 @@ class Authenticate
...
      */
-    public function handle($request, Closure $next, $guard = null)
+    public function handle($request, Closure $next, ...$guards)
     {
-        if (Auth::guard($guard)->guest()) {
-            if ($request->ajax() || $request->wantsJson()) {
-                return response('Unauthorized.', 401);
-            } else {
-                return redirect()->guest('login');
+        if ($this->check($guards)) {
```
gaomd: Looks weird to put success condition before failed condition, in Middleware.

JosephSilber: It's not weird at all. You just maybe haven't seen it be done that way.

I actually like it quite a bit. It decreases the level of indentation, which is always a good thing.

taylorotwell:
Would anyone ever want this to be an "and" check instead of an "or" check?

JosephSilber:
For an "and" check, you can just run the middleware twice:

```php
->middleware('auth|auth:api');
```

shehi:
Make it OR please. Not every JSON request is an AJAX request.

shehi:
And why "allow" has become synonymous with "make it mandatory"?! Why ```$guards``` parameter is mandatory?

taylorotwell:
`$guards` isn't mandatory.

PR #3687, opened 2016-03-16
Remove pre-update-cmd

barryvdh:
Can't rely on being able to run php artisan, before updating. See https://github.com/composer/composer/issues/5066

> Before, running install without a lock file did an install that was almost an update, but still ran the pre/post-install-cmd scripts. Now we normalized it to actually run an update and fire pre/post-update-cmd, because that's what it is doing. If you have a lock file though it will run install as before.

When installing without lockfile, Composer now behaves as upgrade. It also executes the `pre-upgrade-cmd` instead of `pre-install-cmd`. Behaviour with composer.lock available is not changed.

With the latest composer version, doing a fresh install will fail because the vendor files are not yet present.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,10 +39,8 @@
             "php artisan clear-compiled",
             "php artisan optimize"
         ],
-        "pre-update-cmd": [
-            "php artisan clear-compiled"
-        ],
         "post-update-cmd": [
+            "php artisan clear-compiled",
             "php artisan optimize"
         ]
     },


vlakoff:
You should make the change to laravel/laravel 5.1 as well, otherwise I see the support nightmare coming.

Fuxy22:
Running `php artisan clear-complied` after update will use the old compiled artisan code which can cause errors when updating.

That is why i changed it in commit c334e472196e2457c3cca127645d4b77d0d952d7.

barryvdh:
Yes I know, but we can't rely on having a working install before..
Ideally, it would be done with running `php artisan` at all. So for example:

`"php -r \"@unlink('bootstrap/cache/compiled.php');@unlink('bootstrap/cache/services.json');\""`

But that would have the downside of not being able to change it.

Or we could check if the autoload exists in the artisan file. If not, assume we're on a fresh install and just skip, without producing an error/stopping composer.

Fuxy22:
Your suggestion for the check seems to be the best option.

We can't rely on composer and the location of the files can change for instance I have `bootstrap/cache/compiled.php` and `bootstrap/cache/services.php`.

mcampa:
Is this needed for Laravel 4.2?

GrahamCampbell:
4.2 is not supported

mcampa:
I just need to know if I need to apply this change to my 4.2 projects

vlakoff:
As people encountering the issue may land on this page, note an improved fix has been implemented since then. See https://github.com/laravel/framework/pull/12827 and https://github.com/laravel/laravel/pull/3699.

omarjebari:
What about people with old 5.1 projects that they are checking out from source control, ie not installing Laravel from scratch?

Promixius:
@omarjebari  Well what i've found useful was first running
composer install --no-script
the run a normal composer install after to execute the scripts as a workaround not the best but will have to do for now for me.

ramseyjiang:
The latest version which can be installed on command line, it still has the same problem. And "composer install --no-script" does not work on this version. How can I fix it?

vlakoff:
You should apply #3699 to your existing projects.

ramseyjiang:
I have viewed that, but I think I cannot fix the problem.

I always use "composer create-project --prefer-dist laravel/laravel blog" to install laravel
but now it does not work.
After install, if success, I run : php artisan -V
     it will show "Laravel Framework version 5.2.32"

But now, it just show error. bootstrap/autoload.php on line 17

Please help me.

ramseyjiang:
I fixed it

dalenguyen:
@ramseyjiang I got the same problem. What did you do with it?

ito2307:
@ramseyjiang  i have the same issue but i downloaded a project in my company and i'm trying to do composer install but display this http://stackoverflow.com/questions/39665219/why-does-composer-install-show-this-error-message

PR #3693, opened 2016-03-18
feat: add editorconfig file

RomainLanz:
This file force consistency between developper.

More information can be found on http://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_size = 4
+indent_style = space
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false


.editorconfig
```diff
@@ -0,0 +1,13 @@
...
+[*]
+indent_size = 4
+indent_style = space
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
```
GrahamCampbell: ?

RomainLanz: Your question is about `trim_trailing_whitespace = false` on `[*.md]`?

To create a line break two spaces is used at the end of line. So you shouldn't trim trailing whitespace. It'll remove your line break.

GrahamCampbell: i'd say that's a bad idea though - just add a physical line break

RomainLanz: But that doesn't work. You couldn't create a (small) line break without writing two spaces.

Example: https://github.com/Slynova-Org/yoyo/blob/master/README.md

taylorotwell:
The laravel/laravel repo is more where companies should add their own editorconfig IMO.

.editorconfig
vlakoff: I couldn't agree more with GrahamCampbell, meaningful trailing spaces are a goddamn wrong thing, and *always* source of problems. Better play with `<br>` if you really want to insert linebreaks.

PR #3694, opened 2016-03-19
[5.1+] Fallback for "clear-compiled" command when vendors not yet installed

vlakoff:
Follow-up to #3687. Just if it may be of inspiration.
It's so much very perfectible, but at least it doesn't even need modification of composer.json.

Note that Laravel 5.2 uses `services.php` instead of `services.json`.

<br>"- 3 chars" compliant. :information_desk_person:

### DIFF
--- a/artisan
+++ b/artisan
@@ -4,6 +4,17 @@
 $bootstrapPath = __DIR__.'/bootstrap';
 
 /*
+    Composer may call "clear-compiled" before installing the autoloader for
+    the first time. In this particular case, we'll try to do the cleanup
+    here, then exit as artisan won't be able to continue nevertheless.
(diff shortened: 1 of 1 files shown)


vlakoff:
ping @barryvdh :gift:

vlakoff:
Another solution may be to rather put this in a new script `bootstrap/clear-compiled.php` and call it in composer, `php bootstrap/clear-compiled.php`.

vlakoff:
I targeted the wrong branch, this PR was intended for 5.1 branch.

vlakoff:
I just submitted #3695 which is simpler, and in my opinion a better solution.

PR #3695, opened 2016-03-19
[5.1] Standalone "clear-compiled" script for use in Composer

vlakoff:
Alternative to #3694. Another way of solving #3687.

<br>Ref: [ClearCompiledCommand.php
](https://github.com/laravel/framework/blob/5.2/src/Illuminate/Foundation/Console/ClearCompiledCommand.php)

### DIFF
--- a/bootstrap/clear-compiled.php
+++ b/bootstrap/clear-compiled.php
@@ -1,4 +1,5 @@
 <?php
 
 @unlink(__DIR__.'/cache/compiled.php');
-@unlink(__DIR__.'/cache/services.json');
+@unlink(__DIR__.'/cache/services.json'); // Laravel 5.1
+@unlink(__DIR__.'/cache/services.php'); // Laravel 5.2
--- a/bootstrap/clear-compiled.php
+++ b/bootstrap/clear-compiled.php
@@ -1,5 +1,13 @@
 <?php
 
-@unlink(__DIR__.'/cache/compiled.php');
-@unlink(__DIR__.'/cache/services.json'); // Laravel 5.1
-@unlink(__DIR__.'/cache/services.php'); // Laravel 5.2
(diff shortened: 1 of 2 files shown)


vlakoff:
Reminder: Laravel 5.2 uses `services.php` instead of `services.json`.

vlakoff:
Added a commit so that the script is compatible with both Laravel 5.1 and 5.2.
Less hassle for everybody. :relieved:

taylorotwell:
The idea is good but we probably want this file somewhere in the framework itself in case we have to change it mid release. Change this PR to just be the script change and then PR to laravel/framework with the script that deletes the files.

vlakoff:
Yeah but that's precisely what we can't do and is the source of the issues...
vendor/laravel/framework may not be installed yet.

taylorotwell:
Hmm, true.

crynobone:
Shoulder still be safe to do it on post update cmd since it doesn't load the whole app.

GrahamCampbell:
@crynobone Yeh. It's fine to put this in laravel/framework, because we have no need to run commands before composer update.

taylorotwell:
Yeah true we don't need to be doing anything pre-update anyways. So this can go into framework.

barryvdh:
:+1: for doing it post-update/post-install with a simple script in laravel/framework

vlakoff:
But putting it in laravel/framework makes it hard to resolve `basePath`, thus `bootstrapPath`.

jmarcher:
I like this solution better than moving the command to the post-update-cmd

barryvdh:
So like this? https://github.com/laravel/framework/pull/12827

barryvdh:
Above is merged. Related: #3699

PR #3697, opened 2016-03-21
Set HttpOnly flag

mzaalan:


mzaalan:
In class : Illuminate\Foundation\Http\Middleware\VerifyCsrfToken
funtion : addCookieToResponse
$response->headers->setCookie(
            new Cookie(
                'XSRF-TOKEN', $request->session()->token(), time() + 60 * 120,
                $config['path'], $config['domain'], $config['secure'], false
            )
        );
we need to set this cookie as http only, so can not read by javascript, so we need to add httponly flag to session configuration file

shehi:
+1 for security reasons.

PR #3698, opened 2016-03-21
[5.2] Allow for setting sqlite database via env

dakira:
Having this configurable by ENV allows for easily running tests in memory by adding this to the phpunit.xml:

```
<env name="DB_CONNECTION" value="sqlite"/>
<env name="DB_DATABASE" value=":memory:"/>

```

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -48,7 +48,7 @@ return [
 
         'sqlite' => [
             'driver' => 'sqlite',
-            'database' => database_path('database.sqlite'),
+            'database' => env('DB_DATABASE', database_path('database.sqlite')),
             'prefix' => '',
         ],
 


MadMikeyB:
This is causing breaks in 5.2.27. To reproduce, grab a fresh laravel install.

`nano .env`

Change DB_CONNECTION to sqlite in .env only. Save file.

`touch database/database.sqlite ` (the `database_path('database.sqlite')` set in database/config.php before this change)
`php artisan make:auth`
`php artisan migrate`

```
  [InvalidArgumentException]            
  Database (homestead) does not exist.  
```

Because it's looking for homestead.

Change DB_DATABASE to this:
`DB_DATABASE=/database/database.sqlite`

and we still receive an error

```                                 
  [InvalidArgumentException]                            
  Database (/database/database.sqlite) does not exist.                                                     
```

Additionally, even if I put a full path in and it works - do we really want a path as a DB_DATABASE? Could that not be abstracted to a DB_PATH environment variable which is only then used by the SQLite Driver?

dakira:
@jamesfairhurst This change makes it actually easier to do the local testing in sqlite. You only need the two settings above to your phpunit.xml and use the *DatabaseMigrations* trait in your tests. No manual migrations necessary.

PR #3700, opened 2016-03-23
Remove Mandrill from list of supported email providers

ssfinney:
Considering that Mandrill's entry was removed from `config/services.php` in d998b5bd0392f76dcaa461fdec919658947c2e65, shouldn't we go ahead and remove Mandrill from the list?

P.S. I tried to get the formatting consistent with d998b5bd0392f76dcaa461fdec919658947c2e65, but please correct me if you feel the list should be formatted differently.

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -11,8 +11,8 @@ return [
     | sending of e-mail. You may specify which one you're using throughout
     | your application here. By default, Laravel is setup for SMTP mail.
     |
-    | Supported: "smtp", "mail", "sendmail", "mailgun", "mandrill",
-    |            "ses", "sparkpost", "log"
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
No, since we've not removed the mandrill driver?

GrahamCampbell:
@taylorotwell why did you remove it from the services anyway?

nacr-dev:
Yes why? The service has chanced but mailchimp paid users can still use mandrill.

ssfinney:
I'm just advocating for consistency. I have no problem keeping Mandrill in the services file.

GrahamCampbell:
Mandrill is not at all dead. I guess they just want to improve their deliverability and lower the spam going through their service.

PR #3703, opened 2016-03-24
Fix docblock in the authenticate middleware

scrubmx:


GrahamCampbell:
No, we want the user to pass strings as parameters.

PR #3707, opened 2016-03-24
Update .gitignore

bradbernard:
When using the Homestead per project installation, the virtual machine is created within the project directory. https://laravel.com/docs/5.2/homestead#per-project-installation

The .vagrant directory exposes the user's private key at: 
.vagrant/machines/default/virtualbox/private_key

I think adding the directory to .gitignore would be a safe measure to ensure future users do not push their private key to a publicly available site like GitHub.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/.vagrant
 /vendor
 /node_modules
 /public/storage


crynobone:
More suitable for global .gitignore

https://robots.thoughtbot.com/global-gitignore

bradbernard:
Agreed @crynobone but no harm in adding it here.

gaomd:
Do not agree with global option, node_modules exists in .gitignore (which may be useless if I don't use Elixir), especially Homestead.yml (we already assume user will use Homestead), then I think ignore .vagrant would be a huge positive.

bradbernard:
I think global option is good if you know what you are doing, but if you just wanna get things up and running I don't see the harm in adding it to the official repo's gitignore. As you have more and more laravel projects, you want to separate them into per project VMs instead of running all sites in one large VM, so I could see more people switching to the per project in the future and not realizing it publishes private key.

PR #3709, opened 2016-03-25
Update .env.example

MadMikeyB:
add DB_CONNECTION as env variable - it is already checked for in config/database.php


vlakoff:
Refs #3390

PR #3712, opened 2016-03-25
Return json response from controller

ahmedash95:
instead of using Response::json() on each method when i want to return json data let's make it simple with $this->json($data)


GrahamCampbell:
You can add this in your own app as you please. Everyone has different preferences.

jenky:
You can do it on your own app. BTW just return the array, it should be cast to json.

GrahamCampbell:
You can't set headers at the same time though.

PR #3713, opened 2016-03-25
add APP_TIMEZONE to env config

mkwia:
No reason why not, many people make apps specific to their country/area.


jadjoubran:
I'd always keep the timezone to UTC even if I'm serving another country just because it's a default for almost everything (database, server, php, etc..)

PR #3714, opened 2016-03-27
Remove directory check for Front

let-aurn:
Let's say we have a this 

    - public/
        - js/
            - app.js
            - app.js.map

When I request the page `http://example.com/js/app.js`, response is 200. That's ok.
When I request the page `http://example.com/js/doesntexist.js`, response is 404 and 404.blade.php is displayed. That's ok.
When I request the page `http://example.com/js/` (for some reason), response is 404 but the displayed page is not 404.blade.php.

This PR is a fix for that.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -10,7 +10,6 @@
     RewriteRule ^(.*)/$ /$1 [L,R=301]
 
     # Handle Front Controller...
-    RewriteCond %{REQUEST_FILENAME} !-d
     RewriteCond %{REQUEST_FILENAME} !-f
     RewriteRule ^ index.php [L]
 


GrahamCampbell:
> response is 404 but the displayed page is not 404.blade.php.

You mean, it's not routed through Laravel?

let-aurn:
The 404 page dislayed is a *default* error page. Laravel doesn't serve 404.blade.php (that's because the request isn't redirected to `index.php`)

It displays something like this.

    Not Found
    The requested URL /js/ was not found on this server.
    Apache/2.4.17 (Win64) PHP/5.6.16 Server at project.dev Port 80

GrahamCampbell:
So it's not routed through laravel then?

taylorotwell:
Holding off on this. You can tweak your own file as needed.

let-aurn:
@GrahamCampbell Nope, it is not routed through Laravel because the request can't reach index.php due to the htaccess

vlakoff:
But why is the server giving a 404 though the directory exists?

let-aurn:
@vlakoff Because there is no index.html or such file in it.

vlakoff:
But other users could use automatic directory listing ([mod_autoindex](http://httpd.apache.org/docs/2.4/mod/mod_autoindex.html)). There isn't really a solution which fits for everyone.

sivpro:
Because http-server by default when no index file
27 марта 2016 г. 22:14 пользователь "Let Aurn IV" <notifications@github.com>
написал:

PR #3715, opened 2016-03-27
one more update .env.example

jips:
add APP_LOG env variable, already checked for in config/app.php


mg-freelancer:
What is importance of that ? , it depends on your preference in the app

jips:
If we have env tag `'log' => env('APP_LOG', 'single'),` in config/app.php it should also be in the .env file like other tags...

PR #3719, opened 2016-03-29
To use web group by default, the array must be passed as an array

drbyte:
If it's not passed as an array, none of the Validator errors get passed to the session.

Ref: https://github.com/laravel/laravel/commit/5c30c98db96459b4cc878d085490e4677b0b67ed

### DIFF
--- a/app/Providers/RouteServiceProvider.php
+++ b/app/Providers/RouteServiceProvider.php
@@ -53,7 +53,7 @@ class RouteServiceProvider extends ServiceProvider
     protected function mapWebRoutes(Router $router)
     {
         $router->group([
-            'namespace' => $this->namespace, 'middleware' => 'web',
+            'namespace' => $this->namespace, ['middleware' => 'web'],
         ], function ($router) {
             require app_path('Http/routes.php');
         });


it-can:
Your change does not work for me...

app/Providers/RouteServiceProvider.php
```diff
@@ -53,7 +53,7 @@ public function map(Router $router)
     protected function mapWebRoutes(Router $router)
     {
         $router->group([
-            'namespace' => $this->namespace, 'middleware' => 'web',
+            'namespace' => $this->namespace, ['middleware' => 'web'],
```
mark86092: Actually, 'middleware' => 'web' can work

However, your syntax is wrong, it should be 'middleware' => ['web']

it-can: yes that works :+1:

drbyte:
With all due respect, I believe this needs to be reopened.

In 5.2.25, 5.2.26 if you try to use `$errors` or `old()` in the view to display Validation issues or redisplay previous inputs, it's broken.

Proof of concept:
- `laravel new` (which as of 5 minutes ago gave me a fresh v5.2.26)
- `artisan make:auth`
- visit `homestead.app/login`
- click Login without entering any data

Notice that you get NO ERRORs.  

But apply the code change as in this PR, and suddenly the errors show properly.

Using `'middleware' => ['web']` doesn't work either. Just `['middleware' => 'web']`

taylorotwell:
Try `laravel new` now...

drbyte:
Okay, ran `laravel new` again, got 5.2.27 this time.

**The issue appears to be fixed in 5.2.27** via https://github.com/laravel/framework/pull/12903 and https://github.com/laravel/framework/pull/12911 which were just merged today.

Unfortunately people still have to manually change their routes file (see https://github.com/laravel/framework/pull/12903) to benefit from the fix. 

I tested on this new 5.2.27 and confirm that someone who still has `Route::auth()` inside another (or at least `web`) middleware group the way the old routes.stub did will still have the old symptoms, even after running `composer update` to get the rest of the framework updates.

PR #3720, opened 2016-03-29
Add a maximum length password rule

deringer:
This addresses a change proposed in laravel/framework#12905 due to a length limitation in Bcrypt causing strings longer than 72 characters to be truncated when computing their hash.

Adding a validation rule when creating a new password will provide a better user experience when creating new passwords, avoiding providing users with false expectation around the length of their password.


vlakoff:
I'm -1 on this one, because we shouldn't forbid the user to use 128 character passwords if he wishes so. The 72 character limit from bcrypt is a technical limitation the user shouldn't care about.

taylorotwell:
Yeah agree with @vlakoff, also when some people see a character limit they assume people are storing passwords in plain text and then tweet about it, etc. heh

PR #3721, opened 2016-03-29
Client Ip Address

egorogl:
```php
Illuminate\Http\Request.php

    /**
     * Returns the client IP address.
     *
     * @return string
     */
    public function ip()
    {
        return $this->getClientIp();
    }
```
It returns the address of the server, not the client
Because of this throttle it is not working properly


barryvdh:
- This doesn't belong on laravel/laravel
- You're probably behind a proxy and didn't setup your trusted headers for this, so in that case it's a support question and belongs to the forum.

PR #3722, opened 2016-03-29
Create an Authorize middleware

JosephSilber:
Can be used with a simple ability:

```php
Route::group(['middleware' => 'can:view-dashboard'], function () {
    // Dashboard routes...
});
```

Can be used on a single model:

```php
Route::get('users/{user}', 'UserController@show')->middleware('can:view,user');
```

It will pull the `$user` model from the route.

Finally, it can be used on a model type:

```php
Route::get('users', 'UserController@index')->middleware('can:view,'.User::class);
```

### DIFF
--- a/app/Http/Middleware/Authorize.php
+++ b/app/Http/Middleware/Authorize.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Gate;
+use Closure;
+
+class Authorize
+{
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @param  string  $ability
+     * @param  string|null  $model
+     * @return mixed
+     *
+     * @throws \Illuminate\Auth\Access\AuthorizationException
+     */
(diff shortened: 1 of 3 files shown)


app/Http/Middleware/Authorize.php
```diff
@@ -0,0 +1,50 @@
...
+    protected function getGateArguments($request, $model)
+    {
+        // If there's no model, we'll pass an empty array to the gate. If it
+        // looks like a FQCN of a model, we'll send it to the gate as is.
+        // Otherwise, we'll resolve the Eloquent model from the route.
+        if (is_null($model)) {
+            return [];
+        }
+
+        if (strpos($model, '\\') !== false) {
```
sileence: this could use a comment (if this is a class name...)

JosephSilber: I think the first comment is enough.

sileence: maybe leaving the first comment but using else if below? To make this one block.

JosephSilber:
Closing this in favor of [having the middleware in core](https://github.com/laravel/framework/pull/12913).

PR #3726, opened 2016-03-30
Bump package versions

JeffreyWay:
Just giving a bump to the Elixir and Gulp dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "private": true,
   "devDependencies": {
-    "gulp": "^3.8.8"
+    "gulp": "^3.9.1"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,10 +1,10 @@
 {
   "private": true,
   "devDependencies": {
-    "gulp": "^3.8.8"
+    "gulp": "^3.9.1"
```
GrahamCampbell: this change was unnecessary since this already resolves to the latest 3.x

JeffreyWay: Oh well. Might as well be explicit about the current version.

GrahamCampbell: Yeh, it's not harming anyone.

PR #3727, opened 2016-04-01
Allow passing multiple gaurds to the auth middleware

JosephSilber:
Essentially #3685 for master.

### DIFF
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -17,14 +17,37 @@ class Authenticate
      */
     public function handle($request, Closure $next, $guard = null)
     {
-        if (Auth::guard($guard)->guest()) {
-            if ($request->ajax() || $request->wantsJson()) {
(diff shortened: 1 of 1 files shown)


app/Http/Middleware/Authenticate.php
```diff
@@ -17,14 +17,37 @@ class Authenticate
...
-        if (Auth::guard($guard)->guest()) {
-            if ($request->ajax() || $request->wantsJson()) {
-                return response('Unauthorized.', 401);
-            } else {
-                return redirect()->guest('login');
+        if ($this->check(array_slice(func_get_args(), 2))) {
+            return $next($request);
+        }
+
+        if ($request->ajax() || $request->wantsJson()) {
```
GrahamCampbell: elseif would be more consistent with syntax here

taylorotwell:
Gonna hold off on this because dat array_slice is nasty. Will just wait for PHP 5.6 on 5.3.

PR #3729, opened 2016-04-01
Introduce DB_PATH for sqlite issue in .env

MadMikeyB:
Per my comment on PR https://github.com/laravel/laravel/pull/3698

This PR resolves that issue - though I do have a strong feeling that this could be done a lot better, I simply didn't want to venture into laravel/framework to change the database_path() helper to check for an env('DB_PATH') - though I would imagine that would be somewhat cleaner.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,7 +7,7 @@ DB_CONNECTION=mysql
 DB_HOST=127.0.0.1
 DB_PORT=3306
 DB_DATABASE=homestead
-DB_PATH=/database/
+DB_PATH=/var/www/html/laravel/database/
 DB_USERNAME=homestead
 DB_PASSWORD=secret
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
We already have a way to set this.

MadMikeyB:
@GrahamCampbell I know that we have `database_path()` but there are errors for certain scenarios..

Please see my linked comment for the example?

GrahamCampbell:
Right, but the env allows us to bypass calling the database_path function.

GrahamCampbell:
We also can't make assumptions about the filename.

GrahamCampbell:
Thanks for the PR, but I'd suggest you just tweak your own apps to get them how you like them. Everyone likes things setup slightly differently.

PR #3731, opened 2016-04-01
Update composer.json

FintanK:


GrahamCampbell:
Thanks, but it doesn't make sense to update laravel/laravel with this, since this is meant to be the boilerplate app for people, and people's applications themselves don't have that homepage.

PR #3735, opened 2016-04-05
create .env file only if it do not exists already.

SachinAgarwal1337:
On server we do `composer install` to update the dependencies which will reset the `.env` file every time. This will prevent from resetting it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
     },
     "scripts": {
         "post-root-package-install": [
-            "php -r \"copy('.env.example', '.env');\""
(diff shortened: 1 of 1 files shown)


shehi:
Why did you close this PR?

crynobone:
`composer install` doesn't execute `post-root-package-install` script, that only happen on `composer create-project`.

SachinAgarwal1337:
I created a new one https://github.com/laravel/laravel/pull/3736. This was showing error in StyleCI check.

SachinAgarwal1337:
@crynobone Oh I dint know that. Thank for info :D

PR #3736, opened 2016-04-05
create .env file only if it do not exists already.

SachinAgarwal1337:
On server we do `composer install` to update the dependencies which will reset the `.env` file every time. This will prevent from resetting it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
     },
     "scripts": {
         "post-root-package-install": [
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks, but, composer doesn't actually run that command when you run `composer install`.

SachinAgarwal1337:
@GrahamCampbell I dint know that :P Thanks for the info :D

PR #3743, opened 2016-04-05
[5.2] Bring back SQL Server settings

larschinkel:
Finally


sroutier:
:thumbsup:

taylorotwell:
It's not super common to use SQL Server with Laravel I don't think and they are easily copied and pasted from documentation.

PR #3745, opened 2016-04-07
'key'' is not null at Users

thusk:
migrations declare it as not null, so try to seed maybe gets:
  SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "key" violates not-null constraint


GrahamCampbell:
I wasn't aware our migrations have that field?

thusk:
o.0 what did i clone to see that key ._.

thusk:
sorry idk what project migration came without it.

was a mistake.

PR #3753, opened 2016-04-13
Run gulp after installing npm dependencies

martinbean:
Automatically run `gulp` after installing npm dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,5 +6,8 @@
   "dependencies": {
     "laravel-elixir": "^5.0.0",
     "bootstrap-sass": "^3.0.0"
+  },
+  "scripts": {
+    "postinstall": "gulp"
   }
 }


GrahamCampbell:
:-1:

GrahamCampbell:
This won't actually work unless you've aliased gulp.

GrahamCampbell:
You'll need `./node_modules/.bin/gulp`.

GrahamCampbell:
People commonly want to run it with the `--production` flag too. I don't see this change as being needed, or indeed wanted by everyone.

GrahamCampbell:
People also commonly use bower too.

martinbean:
> You'll need ./node_modules/.bin/gulp.

Not true. npm copies binaries to *node_modules/.bin*, which is part of the path npm looks in.

GrahamCampbell:
> which is part of the path npm looks in.

It doesn't on every system. For example, it doesn't for me on my PC.

PR #3758, opened 2016-04-17
Adds support for custom app url's in your test cases.

Cayan:


crynobone:
Not required, anyone who need this can pretty much customise it per application.

Cayan:
2 months ago a new configuration was added to the env file (278c41887ce73914dc0ee645eb13ae4e5b60f1b0). I opened this PR to prevent inconsistencies within the project. 

I wasn't sure if the best approach would be through env('APP_URL') or config('app.url')

thecrypticace:
I do something similar to this. I got into a situation where the difference between APP_URL (in `.env`) and `$this->baseUrl` (in `TestCase`) would make tests fail. I solved it by overriding the environment variable at the beginning of `createApplication()` like so: `putenv("APP_URL={$this->baseUrl}");`

But, like @crynobone said, it's not completely necessary (Well, unless you run into the testing situation above — but that's only because I actually customize `APP_URL` in the environment file).

crynobone:
I really don't like this idea since it enforce use of a single `$baseUrl` value from `env('APP_URL')`. `config('app.url')` is not possible with this PR since you're doing it under `__construct()` when all other configuration is done under `setUp()` method (when `$app` is created).

Why single `$baseUrl` is bad? Well it disallow you to swap `$baseUrl` between test cases, one test case maybe checking the global URL but there might be certain tests that routed to subdomain. 

Before, to do this is just as simple as defining/overriding `$baseUrl` property, with this changes, you have to override the `__construct` method.

thecrypticace:
^ This is a valid point. The setup in this PR, if done at all, should be in `createApplication` since it is run before every test.

shehi:
As a side note: once there was this file in ```app/Providers/ConfigServiceProvider.php``` which was later removed from laravel base, I still keep that file in my project: 
https://github.com/AudithSoftworks/Basis/blob/master/app/Providers/ConfigServiceProvider.php 
(please ignore the ```if``` clause inside ```register()``` method, that's my addition). As you can see, at the bottom, there is ```config()``` invokation, through which you can push custom config settings to the app environment.

PR #3759, opened 2016-04-21
Rename `phpunit.xml` to `phpunit.xml.dist`

shulard:
From PHPUnit documentation :

> If phpunit.xml or phpunit.xml.dist (in that order) exist in the current working directory and --configuration is not used, the configuration will be automatically read from that file.

Changing the name to `phpunit.xml.dist` allow users to easily create specific local configuration when necessary.


GrahamCampbell:
Maybe this would be best introduced in L5.3 instead?

shulard:
This update change nothing on the PHPUnit behaviour. The file is just used as the default one like before... For me it doesn't require a minor version to be introduced.

shulard:
Just depends on the versioning politic 😄

onbjerg:
Why was this closed?

shulard:
@taylorotwell, can you just explain why you closed that PR ?

mg-freelancer:
+1

shehi:
-1: It's pointless to have .dist file in a new project. You will have .xml file which should be edited for new project's needs. Otherwise this will become a nuisance everytime you create a project, making you rename this file so that IDE didn't freak out.

onbjerg:
@shehi How would your IDE "freak out"? Also, the idea is to _not_ rename the file, but create your own config that overrides the defaults shipped with Laravel.. Pretty much like the phpunit docs say.

shulard:
Hum, there isn't something to change to have the project work with the `dist` file. PHPUnit handle that file automatically.

When you create a new project, you just run `vendor/bin/phpunit` and it's ok, just like before that PR. It's juste that sometimes you need to have local settings to run tests without the need to put them on Git (check with a specific session driver for example). PHPUnit handle that can case with the `phpunit.xml.dist` and the `phpunit.xml` files.

For me it's just a kind of good practice to allow users have the possibility to use a tweaked configuration. As I already said, it's local only, the CI or other checks use the dist file as default.

shehi:
I understand what you guys mean, but why would I care for default Laravel settings once I start working on my project?! What is the point of keeping original settings? A year or two after project is finished, when someone else re-visits the code, it will be like "Huh?" situation for that person who doesn't know this small fact about PhpUnit. Most probably file will get deleted if not merged into a single file.

I always failed to understand why you people hassle about these conventions-"sugars" that don't do anything vital actually. Sorry, but I am still against this change. One should do that in their own projects if it's that important.

GrahamCampbell:
I personally am fine with this change, but it's a personal preference, and is trivial to change in your project. I don't see any need for further discussion here.

PR #3761, opened 2016-04-25
Adds database_path to env DB_DATABASE too

mateusjatenee:
For some reason there's a need to pass the absolute path to an SQLite database whenever we're setting that through the .env file. I think that wrapping that up inside the database_path method is a less-confusing way of setting that db configuration as it removes the need of passing the absolute path of the sqlite file. 

Thanks!

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -48,7 +48,7 @@ return [
 
         'sqlite' => [
             'driver' => 'sqlite',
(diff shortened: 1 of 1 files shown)


taylorotwell:
Going to hold off on this in case people are used to current behavior.

PR #3764, opened 2016-04-27
There are chances to deploy same application into different time zone…

sasureshkumar:
…s, and it will provide the option to configure TIMEZONE via .env


GrahamCampbell:
I don't think this needs to go into the core.

PR #3765, opened 2016-04-28
Update server.php

thelastblack:
Sometimes, using `server.php` for PHP local server gives 404 error for static files in `public` folder. I think the problem lies in `return false` when using that.

I am using NetBeans 8.1 for development, and configuring to run `server.php` with PHP local server, and debugging with XDebug, static files fail to load. It just gives 404, generated by PHP local server itself.

Although running `php -S localhost:8000 server.php` is fine on command-line, and then running debug in NetBeans works, but it does not work when run with NetBeans and in debug mode.

This patch fixes the issue, although it would be better to serve files based on mime-type. In my tests, using `mime_content_type` or `finfo_*` can not correctly detect type of minified files. As this file is to be used with local development, it is not a great issue to use simple extension-based detection.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,11 @@
 {
   "private": true,
-  "devDependencies": {
-    "gulp": "^3.9.1"
+  "scripts": {
+    "prod": "gulp --production",
+    "dev": "gulp watch"
   },
-  "dependencies": {
+  "devDependencies": {
+    "gulp": "^3.9.1",
     "laravel-elixir": "^5.0.0",
     "bootstrap-sass": "^3.0.0"
   }
--- a/server.php
+++ b/server.php
@@ -14,8 +14,15 @@ $uri = urldecode(
 // This file allows us to emulate Apache's "mod_rewrite" functionality from the
 // built-in PHP web server. This provides a convenient way to test a Laravel
 // application without having installed a "real" web server software here.
-if ($uri !== '/' && file_exists(__DIR__.'/public'.$uri)) {
-    return false;
+$file = __DIR__.'/public'.$uri;
+if ($uri !== '/' && file_exists($file)) {
+    $ext = pathinfo($file, PATHINFO_EXTENSION);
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
I'm sorry, but this doesn't look like a correct fix.

thelastblack:
You have any clue regarding problem?

thelastblack:
I just checked PHP doc for [built-in web server](http://php.net/manual/en/features.commandline.webserver.php), section `Example #5 Handling Unsupported File Types`.
It is basically doing what I have done, so what is wrong with it?

GrahamCampbell:
> so what is wrong with it?

Because there are not unsupported. They're listed as supported.

GrahamCampbell:
> Standard MIME types are returned for files with extensions: .3gp, .apk, .avi, .bmp, .css, .csv, .doc, .docx, .flac, .gif, .gz, .gzip, .htm, .html, .ics, .jpe, .jpeg, .jpg, .js, .kml, .kmz, .m4a, .mov, .mp3, .mp4, .mpeg, .mpg, .odp, .ods, .odt, .oga, .ogg, .ogv, .pdf, .pdf, .png, .pps, .pptx, .qt, .svg, .swf, .tar, .text, .tif, .txt, .wav, .webm, .wmv, .xls, .xlsx, .xml, .xsl, .xsd, and .zip.

shehi:
@thelastblack : Using ```readfile()``` can be dangerous for users who might choose utilizing output buffer. When OB is used, it will fill up the memory and "drown" the app.

thelastblack:
@shehi True, but note that this is only a "local development server". It is by no mean used in production and all.
Besides, it is easy to change it to read files chunked.
Anyway, I have found that maybe NetBeans is using project root for file finding, whereas files should be resolved relative to `public` folder. I did not find a way to tell NetBeans to use `-t project-root/public` when running PHP yet.

PR #3768, opened 2016-04-30
[5.3] Add config for new Memcached features

tomcastleman:
Adds config for persistent connections, SASL authentication, and custom options

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -49,6 +49,14 @@ return [
 
         'memcached' => [
             'driver' => 'memcached',
+            'persistent_id' => env('MEMCACHED_PERSISTENT_ID'),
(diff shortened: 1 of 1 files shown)


taylorotwell:
In this case the credentials count would always be 2 and it would always try to add credentials even though the `env` function will return `null` if the variables doesn't exist. I think the SASL section should be commented out.

tomcastleman:
That was my reasoning for using `array_filter` in [CacheManager](https://github.com/laravel/framework/blob/master/src/Illuminate/Cache/CacheManager.php#L168) which would mean `sasl` would be an empty array unless both `env` variables were set.

taylorotwell:
Ah OK.

PR #3769, opened 2016-05-01
Add attempts_table to log job exceptions.

Edgpaez:
I also added a little comment to explain how to disable exception and job failing logging. 
I always ended up having the `no such failed_jobs table` error in my logs, I didn't know I could disable it.

I'd like to add a specific config to enable/disable. Something like:

```php
log_failed_jobs => true,
log_failed_job_attempts => true,
```

What do you think ?

### DIFF
--- a/config/queue.php
+++ b/config/queue.php
@@ -74,12 +74,15 @@ return [
     | These options configure the behavior of failed queue job logging so you
     | can control which database and table are used to store the jobs that
     | have failed. You may change them to any database / table you wish.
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Your framework PR was to 5.3, but this PR is to 5.2?

Edgpaez:
Both PRs are to master. 
I'm guessing 5.3 is where it'd be merged to.

Please, let me know if there's something I need to change

GrahamCampbell:
Please target the develop branch here.

PR #3770, opened 2016-05-01
Use Bootstrap 3.3

tillkruss:
Unless there is a reason why not?


GrahamCampbell:
we already were using it...

GrahamCampbell:
`^3.0.0` means 3.x

tillkruss:
@GrahamCampbell Ah, thanks for the clarification.

PR #3772, opened 2016-05-03
Rename routes.php to indicate the name of the middleware group of the routes it contains

philbates35:
Following on from #3724, I'm opening this here to continue the discussion on `develop` as requested by Taylor.

If all the routes in `routes.php` are in the `web` middleware group, then that immediately begs the question what do I call the routes file that that will contain the `api` middleware group routes? Probably either `routes-api.php` or `api.php`, so if you went with that ideally you would want to be consistent and would rename `routes.php` to `routes-web.php` or `web.php` to match. Not doing so would result in having `routes.php` and `routes-api.php` or `api.php`, and that lack of consistency seems a little sloppy.

Personally I think keeping all routes files in a separate directory to avoid the `routes-` prefix of the routes files is preferable, (one file per middleware group) so that's what I've done here, but I'm happy to change it if others prefer.

The main issue is the inconsistency of having a specific middleware group routes file being given a general, non-specific name.


alexleonard:
@philbates35 I've taken the foldered approach on a recent project and it definitely feels neatest to me. In my own case I've gone for 

* Routes/public.php
* Routes/auth.php
* Routes/api.php
* Routes/admin.php

But I think having a default of Routes/web.php makes perfect sense for people new to the framework.

eskrano:
You can add api routes in your RouteServiceProvider and include your routes file .
https://github.com/laravel/laravel/blob/master/app/Providers/RouteServiceProvider.php#L55

phpunit.xml
```diff
@@ -17,7 +17,7 @@
         <whitelist processUncoveredFilesFromWhitelist="true">
             <directory suffix=".php">./app</directory>
             <exclude>
-                <file>./app/Http/routes.php</file>
+                <file>./app/Http/Routes/web.php</file>
```
GrahamCampbell: Just exclude the entire Routes folder please.

philbates35:
@GrahamCampbell I've updated `phpunit.xml` to exclude the whole `Routes` directory as requested.

phpunit.xml
```diff
@@ -17,7 +17,7 @@
         <whitelist processUncoveredFilesFromWhitelist="true">
             <directory suffix=".php">./app</directory>
             <exclude>
-                <file>./app/Http/routes.php</file>
+                <file>./app/Http/Routes</file>
```
GrahamCampbell: This doesn't work. It's not a `file` your excluding.

GrahamCampbell: You need `<directory>./app/Http/Routes</directory>`.

philbates35:
@GrahamCampbell Changed `<file>` to `<directory>`, sorry about that!

robclancy:
`Routes` feels weird to me with all the files inside it being `api.php` etc instead of `Api.php`. The studly case implies classes to me so what about the folder being `routes` instead? But then that is different to all other folders.

naming things is hard

GrahamCampbell:
@robclancy I just do this in my apps: https://github.com/CachetHQ/Cachet/tree/2.3/app/Http/Routes, https://github.com/CachetHQ/Cachet/blob/2.3/app/Foundation/Providers/RouteServiceProvider.php#L68-L74.

robclancy:
Looks like overkill to me but could actually clean up our stuff quite a bit since we have a massive routes.php file.

taylorotwell:
Holding off on this. Obviously you are free to change it to whatever you want in your own apps.

PR #3774, opened 2016-05-06
.env.example: APP_KEY instruction to call key:generate

amenk:
I was getting 'No supported encrypter found. The cipher and / or key length are invalid.' with some bogus key, so it is better to recommend key:generate

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,6 @@
 APP_ENV=local
 APP_DEBUG=true
-APP_KEY=SomeRandomString
+APP_KEY=Please execute php artisan key:generate to generate
 APP_URL=http://localhost
 
 DB_CONNECTION=mysql


GrahamCampbell:
This is invalid and would crash the whole app due to the spaces and no double quotes.

dashingsad:
@GrahamCampbell it looks like you are writing code all the time. :) Life is not just about coding. Visit some exotic countries i.e United Arab Emirates , China  and enjoy :))

amenk:
But typing sind random key can also crash the APP. That is why I thought it's better to give this instruction.

GrahamCampbell:
Not true. This one will crash the app because the .env file has invalid syntax, vs crashing only the encrypter.

dashingsad:
App will not run anyway whether whole app is crash or only encrypter.

GrahamCampbell:
> App will not run anyway whether whole app is crash or only encrypter.

Not true. The key generate command will not crash before the PR, but it actually will crash after the PR...

amenk:
Interesting. I actually did not check that. What I did is was typing something like `dahdsjdhködasjh` and it did crash. Maybe because of the `ö`. But `SomeRandomString` is just a bit misleading there...

would

    APP_KEY="Please execute php artisan key:generate to generate"

do?

zorx:
Or you can just put a comment, dotenv support comments

`APP_KEY=SomeRandomString #Please execute php artisan key:generate to generate`

But I think that this kind of information have to be in the documentation and not in the file it self

GrahamCampbell:
This change is still unnecessary. Running that command is not the only way to set a key anyway.

amenk:
Okay, let's leave it as it is.
I was just thinking it is a usability improvement for those not reading the docs. SomeRandomString implies, that I can just type whatever I want - but that's not true.
Nevermind.

GrahamCampbell:
> implies, that I can just type whatever I want - but that's not true.

You can. Just so long as it's a valid key.

rkukuh:
I wonder that those encryption exception provides an instruction how to fix this issue, like so:

encryption-exception-instruction

PR #3777, opened 2016-05-10
[5.3] Enabled MySQL "strict" mode by default

Adam14Four:
When "strict" mode is set to false it allows the database to ignore significant errors rather than abort and return the error to PHP. An example would be trying to save a string in an INT field...instead of throwing an error, the save completes successfully and simply inserts a 0 in the field.

This means that the default setting allows for data loss with no notification, which seems like a really bad thing. The default should be to the safest setting, which could then be reconfigured by the developer if necessary.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -62,7 +62,7 @@ return [
             'charset' => 'utf8',
             'collation' => 'utf8_unicode_ci',
             'prefix' => '',
-            'strict' => false,
+            'strict' => true,
             'engine' => null,
         ],
 


w2pc:
YES!!!

shehi:
I hope ORM isn't depending on something non-strict.

chelseau:
I feel like this could create potential issues. Strict mode also (at least, off the top of my head):

- disallows division by zero, where the developer may be counting on it being set to NULL (non-strict behaviour)

- disallows inserts if any column is non-nullable, doesn't have an explicit default value, and isn't included in the field list of said insert.

- disallows the date 0000-00-00 (admittedly NULL makes more sense then 0000-00-00 but allowing NULL requires an extra byte of storage). It also disallows dates such as 2016-00-00 but that is just a stupid date so I'm on board for that being disallowed.

I feel like it should just be up to the developer to validate their data ;) but I guess it's an easy enough config change to reverse if preferences necessitate it.

taylorotwell:
I think this makes sense and we'll document it. Timestamps in Laravel have defaulted to nullable for a while now to avoid errors with strict mode.

taylorotwell:
And  as @chelseau notes, easy to reverse.

PR #3782, opened 2016-05-16
[5.2] Make SES region dynamic

mahony0:
Also preserving with default value, us-east-1


taylorotwell:
You can define this in your own application.

PR #3783, opened 2016-05-22
[5.3] Make the Authenticate middleware throw an AuthenticationException

JosephSilber:
This PR is part of [this change](https://github.com/laravel/framework/pull/13651), so that we can [unify unauthenticated responses](https://github.com/laravel/framework/pull/13113).

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -34,7 +34,11 @@ class Handler extends ExceptionHandler
      */
     protected function unauthenticated($request, AuthenticationException $e)
     {
-        parent::unauthenticated($request, $e);
+        if ($request->ajax() || $request->wantsJson()) {
+            return response('Unauthorized.', 401);
+        } else {
+            return redirect()->guest('login');
+        }
     }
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
This can't be merged without solution for redirect URI.

JosephSilber:
@taylor currently, we ship [the full response logic](https://github.com/laravel/laravel/blob/master/app/Http/Middleware/Authenticate.php#L24) in laravel/laravel, and allow people to tweak it there.

Can't we do the same here? Ship the whole `unauthenticated` method in laravel/laravel, and allow people to tweak it inline.

taylorotwell:
Sure, change the PR to do that

JosephSilber:
@taylorotwell should we make the base exception handler an abstract class, and create an abstract `unauthenticated` method on it?

taylorotwell:
I don't see any need for that really.

taylorotwell:
Just do what we said a few comments ago.

JosephSilber:
@taylorotwell done.

PR #3789, opened 2016-05-29
Add missing Mandrill secret keys in config.services

luqmanrom:
Add the Mandrill array in config.services to access the secret key from the `.env`.

### DIFF
--- a/config/services.php
+++ b/config/services.php
@@ -19,6 +19,10 @@ return [
         'secret' => env('MAILGUN_SECRET'),
     ],
 
+    'mandrill' => [
+        'secret' => env('MANDRILL_SECRET'),
+    ],
+
     'ses' => [
         'key' => env('SES_KEY'),
         'secret' => env('SES_SECRET'),


luqmanrom:
In Transport manager, it is using this value in config to send email. It will be better if the secret key is stored in `.env` so that it will not be shared with others 

```   
protected function createMandrillDriver()
    {
        $config = $this->app['config']->get('services.mandrill', []);

        return new MandrillTransport(
            new HttpClient(Arr::get($config, 'guzzle', [])), $config['secret']
        );
    }
```

GrahamCampbell:
I think they were intentionally removed since mandrill is not used as much as it was before due to it no longer being free.

luqmanrom:
Ahh. I see. @taylorotwell  actually clarified in [here](https://github.com/laravel/laravel/commit/d998b5bd0392f76dcaa461fdec919658947c2e65).  Since Laravel is supporting it and most tutorials out there assumes the config.services is still accessing the secret key from the `.env` that only requires user to add secret key in `.env` , don't you think it is better to keep it for now to avoid error/confusions? Or maybe from the `TransportManager#createMandrillDriver`, get the `secret` key value directly from `.env`

PR #3790, opened 2016-05-29
Default welcome.blade.php

Pellmellism:
animated gradient text -- demo @ http://codepen.io/pellmellism/pen/OXJvBE


GrahamCampbell:
It looks like this went a bit wrong. ;)

PR #3792, opened 2016-06-01
Add APP_TIMEZONE to environment configuration

ElfSundae:
Many websites need to change the value of `app.timezone`, so we can add `APP_TIMEZONE` to the environment configuration, it could be configurated in `.env` file to make it out of box when update & merge Laravel into our app.


RomainLanz:
Is `app.timezone` specific to the environment?

GrahamCampbell:

PR #3793, opened 2016-06-01
Add DB_CHARSET and DB_COLLATION to env configuration

ElfSundae:
To store Emoji, the charset of MySQL should be set to  `utf8mb4`.


RomainLanz:
Is the charset of MySQL specific to the environment?

GrahamCampbell:

ElfSundae:
No, it is specified to the app, but I think it could be configurated without modifying `config/app.php`

RomainLanz:
It could be configured without changing the configuration file? This is no sense.

`.env` file is only for specific environment things.

GrahamCampbell:
> No, it is specified to the app

If it's app specific, you should not be using an env variable. They are for non-app specific things.

ElfSundae:
OK, thank you to point it.

PR #3795, opened 2016-06-01
Add Models folder

davex69:
Added Models folder to app root, because they are one of the main components in Laravel applications.


GrahamCampbell:
I think laravel's default is to not put Models in their own namespace. Feel free to do this in your own app though.

PR #3796, opened 2016-06-01
Create an app ThrottleRequests middleware

lucasmichot:
The purpose of this PR is to allow a quicker extension of `\Illuminate\Routing\Middleware\ThrottleRequests`

The original middleware is good for most standard cases but does not bring some useful features.

- skip hit on an internal request
- skip hit on 304 not-modified responses
- change header names
- define a max attempt value, per user or group of user

Use cases can be so numerous that it would not make sense to put all of them in the original middleware.
Instead this PR allows the user to quickly override `\Illuminate\Routing\Middleware\ThrottleRequests`

It tends to mimic what has been done for `\App\Exceptions\Handler::report`


GrahamCampbell:
I don't think it's necessary to include this in the default app.

lucasmichot:
I didn't feel too that it was necessary before, but it turns out that this is the only middleware that I have to override on each projects

hadifarnoud:
I think this is very much needed to be in default app.

PR #3805, opened 2016-06-05
Ignore /public/build

whoan:
Since it's the deafult for versioned files, and also could be for any other files generated for Elixir.


GrahamCampbell:
Not sure this should be a default. Some people lock minified files in their source control, like they do a composer.lock.

PR #3808, opened 2016-06-10
[5.2] Fix redirections to inexistent paths

xxnoobmanxx:
This redirection was causing "Not found" errors every time the URL ends up with a slash, simply because the path does not exists, 

In my case, when I called "http://svrdev001/carloscarucce/gpdseguranca/register/teste", it was redirecting to "http://svrdev001/var/www/carloscarucce/gpdseguranca/public/register/teste", causing a 404 - not found

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -5,10 +5,6 @@
 
     RewriteEngine On
 
-    # Redirect Trailing Slashes If Not A Folder...
-    RewriteCond %{REQUEST_FILENAME} !-d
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I'm not entirely sure this is correct?

xxnoobmanxx:
I checked and its working fine, as far as  `RewriteCond %{REQUEST_FILENAME} !-d` is covered by 
_# Handle Front Controller..._ rules

PR #3809, opened 2016-06-10
[5.3] Add configuration option to disable caching of compiled Blade templates

michaeldyrynda:
In combination with laravel/framework#13938, this change targets 5.3, adding the necessary key to the `view.php` configuration file, in order to enable or disable caching of blade templates.


taylorotwell:
See comment on other PR.

PR #3794, opened 2016-06-01
[5.2] Add MAIL_FROM_ADDRESS, MAIL_FROM_NAME to env configuration

ElfSundae:

PR #3810, opened 2016-06-10
Mail Name and Address Environment Variables

DanRovito:
Added environment variables for mail name and mail address.  Defaults to null if no variable is present.


GrahamCampbell:
Dupe of https://github.com/laravel/laravel/pull/3794.

PR #3814, opened 2016-06-18
[5.2] Use proper PDO fetch style

vlakoff:
PDO fetch style `FETCH_CLASS` is expected to be used along a class name.

<br>It currently works in Laravel because the latter does the following, which happens to work only by chance:
```php
$results = $stmt->fetchAll(PDO::FETCH_CLASS);
```

<br>However, the following would produce an error:
```php
$stmt->setFetchMode(PDO::FETCH_CLASS);
$results = $stmt->fetchAll();
```

> PDOStatement::setFetchMode(): SQLSTATE[HY000]: General error: fetch mode requires the classname argument

<br>This PR switches to the proper fetch style, which is robuster, and even more importantly, less confusing for the user. It's fully backward compatible.

<br>Refs:
* http://php.net/manual/en/pdostatement.fetch.php
* http://php.net/manual/en/pdostatement.setfetchmode.php


GrahamCampbell:
Thanks. Probably best to do this on develop.

taylorotwell:
Yeah can we go ahead and send this to 5.3 when you get a chance just to be safe? Thanks

GrahamCampbell:
:+1:

PR #3815, opened 2016-06-19
[5.3] Use proper PDO fetch style

vlakoff:
Resubmitting #3814 for 5.3 as requested.


Fridz5:
How it use in 5.4 version after deleted this proper?

PR #3816, opened 2016-06-22
Use router instance over facade, to promote good behavior.

dannyvankooten:
The default `app/Http/routes.php` uses the `Route` facade to access the Router instance. 

I think it would be a good idea to show people that they have access to the raw instance straight away, to promote good behaviour and to prevent the service look-up for each route people declare.

### DIFF
--- a/app/Http/routes.php
+++ b/app/Http/routes.php
@@ -11,6 +11,7 @@
 |
 */
 
-Route::get('/', function () {
+/** @var Illuminate\Routing\Router $router */
+$router->get('/', function () {
     return view('welcome');
 });


GrahamCampbell:
:+1:

shehi:
:-1: I'd rather use ```app('router')->get()``` and similars... That'd make more sense and teach people the real way of coding, without Facades.

shehi:
On second thought... Yea, routes.php is used inside RouteServiceProvider and variable is already in there (i.e. ```$routes```), so it makes sense to re-use it from performance standpoint. Argh!

dannyvankooten:
@shehi I don't think coding gets any realer than accessing an instance by its variable name. 😉

shehi:
@dannyvankooten : It's not the case here. routes.php already inherits the variable ```$router``` from its service provider, so your solution makes sense. Otherwise, I am always against facades, I always prefer direct "makes" (i.e. ```app('x')``` which is shortform of ```app()->make('x')```) and bringing-in resources via DI/IoC cycle.

GrahamCampbell:
> I'd rather use app('router')->get() and similars... That'd make more sense and teach people the real way of coding, without Facades.

That is actually the same as using the facade, but is not the same as using the variable.

GrahamCampbell:
This file might want to use a different router instance.

shehi:
@GrahamCampbell : Why a different instance? In my app for example, I don't even use the file, it's all inside the service provider. You are directly require-ing the file into the provider. So why a different instance?

GrahamCampbell:
You require the file, and also pass through a router instance to the file.

shehi:
@GrahamCampbell , yes I know what is there? Instance is there. Why a new one would be needed, when you already have one? :D

DontBeMadBro:
We will need to indicate variable name in group callbacks, otherwise we will get an error with unknown variable name, like that:
```php
$router->group(['as' => 'public'], function ($router) {
    $router->get('/', ['uses' => 'controller@action']);
});
```
instead of 
```php
Route::group(['as' => 'public'], function () {
    Route::get('/', ['uses' => 'controller@action']);
});
```
personally i don't like facades..

taylorotwell:
Prefer the facade. If your routes are cached (as they should be in production) you won't be hitting the facade anyways.

PR #3817, opened 2016-06-22
[5.3] Return JSON for unauthenticated AJAX requests

JosephSilber:


barryvdh:
Perhaps add a 'not PJAX' call here also?

taylorotwell:
You can use the new `probablyWantsJson` method 😆

JosephSilber:
For the record, the method has been renamed to `expectsJson` :+1:

PR #3821, opened 2016-06-27
Moved Usermodel to App\Models, change namespace and calls to it.

chilion:
I moved the User model to a newly created folder App\Models and made sure all namespaces and uses where set right again.

Why did I do this? I have never seen an application with only one Model and so for me it breaks the complete structure and overview of my files when I put the Models straight in the app folder. I noticed this a few times on wishlists and decided it was time for me to go and create a PR.

### DIFF
--- a/app/Http/Controllers/Auth/AuthController.php
+++ b/app/Http/Controllers/Auth/AuthController.php
@@ -2,7 +2,7 @@
 
 namespace App\Http\Controllers\Auth;
 
-use App\User;
+use App\Models\User;
 use Validator;
 use App\Http\Controllers\Controller;
 use Illuminate\Foundation\Auth\ThrottlesLogins;
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace App;
+namespace App\Models;
 
 use Illuminate\Foundation\Auth\User as Authenticatable;
 
--- a/config/auth.php
+++ b/config/auth.php
@@ -67,7 +67,7 @@ return [
     'providers' => [
         'users' => [
             'driver' => 'eloquent',
-            'model' => App\User::class,
+            'model' => App\Models\User::class,
         ],
 
         // 'users' => [
--- a/config/services.php
+++ b/config/services.php
@@ -34,7 +34,7 @@ return [
     ],
 
     'stripe' => [
-        'model' => App\User::class,
+        'model' => App\Models\User::class,
         'key' => env('STRIPE_KEY'),
         'secret' => env('STRIPE_SECRET'),
     ],


zaknes:
Honestly for how much I do this in every Laravel project, I seriously think that it should be added.

Even though it has been suggested time after time.

GrahamCampbell:
This should go to develop really, if at all.

chilion:
@GrahamCampbell I can do that :-) Can you give me a reason why it shouldn't be merged?

jadjoubran:
@chilion what about the generators as well?

chilion:
@jadjoubran Good point, if there is any change at all it might gets merged I will changed them as well. I'm quite curious about @taylorotwell's opinion... :-)

jaschweder:
@chilion for generators you can pass the namespace has parameter and it put your models in correct place, but, this will force all users to put their models inside "models" folders, I'dunno if is this the purpose.

chilion:
@jaschweder At this moment it "forces" all users to put them in the App folder which seems even less the purpose....

jaschweder:
@chilion true

GrahamCampbell:
>  can do that :-) Can you give me a reason why it shouldn't be merged?

Taylor will be in touch soon. It's 100% his decision.

taylorotwell:
There is one main reason I don't like this change, and that is because it forces on people the false idea that Model = Database, and I intentionally wanted to get away from that in the 5.x series because it creates a lot of confusion down the line.

chilion:
@taylorotwell I don't see where it forces people into that idea, why doesn't it do that in the App folder? 
It's a Model right, a folder called Models, not a folder called DatabaseModels or Database.

I get your point about Models !== Database, but I don't see where my PR forces that idea onto people. Could you please elaborate so I can sleep silently tonight? ;-)

taylorotwell:
Because I maintained Laravel for years when there was a "models" folder. Here's what happens: people will put all of their database models in that directory. Then, at some point they will need to write a class that isn't a database model, and they will ask on the forums where to put it, and they will want to create a folder called "libraries" or "classes" or something in the core.

chilion:
@taylorotwell I get that yes, thanks for explaining!
_Can we get a console command then that will do this hassle of a work for every new project for the once willing to use it? Don't document it, people who want it will find it 👼_

GrahamCampbell:
I don't think a console command is needed. It takes one second to move it.

Hoosk:
@chilion +1😸

DanRovito:
@chilion this might help if you want to generate your models in a different folder https://github.com/laracademy/generators

taylorotwell:
`php artisan make:model Models/User`

rk-development:
is anything wrong with 'libraries' directory in app core?

PR #3825, opened 2016-07-02
Add meta auto-refresh to 503 view

alariva:
Happy users not having to hit F5.

Since this view is expected to be displayed for hopefully really
short interruptions, we expect the user to be back as soon as
possible.

This change sets a 10 seconds time window to retry the page load,
aiming to reduce the chance of losing visitors.

---
This change is [Reviewable](https://reviewable.io/reviews/laravel/laravel/3825)


taylorotwell:
Probably won't include this in core by default.

PR #3827, opened 2016-07-06
cleanup

c9n:
remove empty lines


philbates35:
The additional lines are here to guide the developer where to put additional boot and map logic.

PR #3828, opened 2016-07-09
Removed unnecessary else

kamaroly:


jonagoldman:
don't think they will accept this pull request. the else statement improves readability.

PR #3829, opened 2016-07-12
configuration to allow local development pgsql/mysql db access

vmadev:
I do this on every Laravel app I spin up. I never have the need to change it even in production so I figure I submit a pr.

This is particularly convenient when using a Homestead VM since the ports are pre-configured, however, they are pretty standard. 

If you have installed and configured php to also run locally on your machine then this is definitely a big convenience. This would allow you to run migrations and complete composer updates directly from the terminal without needing to login to the VM. 

Of course if your VM environment does not have an APP_ENV declared then it would inherit from the .env, but for the majority of us on Homestead that is simply not the case. 

Dont know about valet since I am on Windows but i assume that you would need a local install of SQL and assign the port manually anyways. 

I fail to see the downside overall, but perhaps i can be enlightened...

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -55,7 +55,7 @@ return [
         'mysql' => [
             'driver' => 'mysql',
             'host' => env('DB_HOST', 'localhost'),
-            'port' => env('DB_PORT', '3306'),
+            'port' => 'local' === env('APP_ENV') ? 33060 : env('DB_PORT', '3306'),
             'database' => env('DB_DATABASE', 'forge'),
             'username' => env('DB_USERNAME', 'forge'),
             'password' => env('DB_PASSWORD', ''),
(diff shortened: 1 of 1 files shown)


Gompje:
Hi @vmadev, 

Personally I don't see the upside of this; why don't you just set the port inside the .env file as well? That way no extra lookup is needed each run. No extra VAR is needed as well.

My dev scenario: I don't use Homestead, I use my own vagrant setup based on an existing Debian box on Parallels. I've configured this box to be as close as the production box; all the ports are therefore exactly the same etc. I use a shared folder for my Laravel app, everything I do in this folder locally is instantly changed on the server - eg composer runs and the like. I hardly need to log into it - not in a way it causes a real nuisance. 

Experience taught me the closer your dev env is to production the happier you will be. :-) I realize everyone's setup is different but for me this pull would mean I need to remember to erase this line each time.

GrahamCampbell:
I don't think this is necessary.

crynobone:
> This is particularly convenient when using a Homestead VM since the ports are pre-configured, however, they are pretty standard.

```yaml
variables:
    - key: DB_PORT
      value: 33060
```

It pretty easy to customize `Homestead.yaml` based on your requirement.

> Dont know about valet since I am on Windows but i assume that you would need a local install of SQL and assign the port manually anyways.

This means everyone else who not using your setup would have to change. One sided convenient?

scrwdnet:
IMO it feels like it clutters the config file making it less easily readable, probably takes you a couple of seconds to amend your .env file. Kind of the point of the .env files?

taylorotwell:
Holding off on this by default.

PR #3832, opened 2016-07-16
[5.3] Remove unused imports

KennedyTedesco:


taylorotwell:
The reason Gate is here is because you would commonly use it in this place, even though we are not using it by default.

PR #3836, opened 2016-07-17
[5.3] Add new password validation rule

driesvints:
For https://github.com/laravel/framework/pull/14345

### DIFF
--- a/resources/lang/en/validation.php
+++ b/resources/lang/en/validation.php
@@ -61,6 +61,7 @@ return [
     ],
     'not_in'               => 'The selected :attribute is invalid.',
     'numeric'              => 'The :attribute must be a number.',
+    'password'             => 'The password is incorrect.',
     'present'              => 'The :attribute field must be present.',
     'regex'                => 'The :attribute format is invalid.',
     'required'             => 'The :attribute field is required.',


resources/lang/en/validation.php
```diff
@@ -61,6 +61,7 @@
     ],
     'not_in'               => 'The selected :attribute is invalid.',
     'numeric'              => 'The :attribute must be a number.',
+    'password'             => 'The filled in password is incorrect.',
```
michaeldyrynda: Would it make sense to use `The :attribute field is incorrect.` here?

driesvints: I guess so. Wasn't really sure because it's a pretty specific use-case. Either way is fine I guess. Let's see if there are other opinions on this.

masterdeduction:
It is much better if you make it, ```Password does not match```

zaknes:
@mastereducation

"The password password is incorrect?"

resources/lang/en/validation.php
michaeldyrynda: I can't think of an instance where it wouldn't be password, given it's very specific - but maybe somebody is using the non-english naming of password? I don't think it's a showstopper either way, though.

jacurtis: Why can't it just say `The password is incorrect.` or `The password you entered is incorrect.`?

The "filled in" portion of the current message seems clunky and unnatural.

HoangTuanXS: https://us-mg5.mail.yahoo.com/neo/launch?.rand=4e8mdk2sqag9n#

driesvints: @jacurtis hmm, yes that indeed seems better. I'll update my PR when I get home.

nyedidikeke: @michaeldyrynda: I think `The :attribute field is incorrect.` makes sense and shouldn't disrupt; it can be used for *verification, activation code and coupon(s)* among others.

driesvints: @nyedidikeke except that this is specifically tied to the user's password. It'll never be used for the things you listed above.

driesvints:
@jacurtis updated it with your suggestion. Thanks 👍

AliFarhoudi:
This should not be merged. We should never tell the user that his entered username is correct and password is incorrect.
The only thing that user should know is that combination of username and password is incorrect.
If we tell the password is incorrect, it means that username is correct. This is not so good.

Promixius:
I have to agree with @AliFarhoudi that would be a really bad idea.

driesvints:
@AliFarhoudi @Promixius I can't see why this is a problem. You'd mostly use this inside the application on actions where you'd require a user to fill in his password before he can execute that action. You'd never use this to login.

AliFarhoudi:
@driesvints Yes, you're right. I had not considered that. In situations like changing account settings only password is needed.
This should be merged ;)

Promixius:
I think I just assumed login, instead of thinking of other applications my bad.

PR #3839, opened 2016-07-26
[5.3] Make use of Bootstrap 3.3.7 with jQuery 3 support

PascaleBeier:
Bootstrap very recently added official jQuery 3 support and I guess it would be great to make use of it :+1:


taylorotwell:
Well that's convenient :)

PR #3841, opened 2016-07-30
[5.3] fix wrong spelling

ac1982:
applications


GrahamCampbell:
Thanks for the PR. Is this broken on earlier versions too?

ac1982:
I don't think earlier version has this typo. I am really expecting the new 5.3, great work, thanks all for that.

PR #3842, opened 2016-07-30
Remove removed routes file

spyric:
Because routes file now outside app folder by default. We can remove that section


KluVerKamp:
👍

GrahamCampbell:
Thanks. :)

PR #3843, opened 2016-07-30
[5.3] Removed unnecessary directives

mayoz:


rk-development:
this else statement might be there for readability purpose

taylorotwell:
I prefer it the first way because the indentation makes it clear they are two distinct paths through the functions.

PR #3844, opened 2016-07-30
[5.3] Unused things removed

mayoz:
Gate is not used in this provider. By default, Laravel doesn't have Policy. Laravel documentation is really simple. I think, developers need Policy/Gate definitions when they can add them.


GrahamCampbell:
I think this was included on purpose to be clear that these things exist?

taylorotwell:
They are there as examples.

PR #3846, opened 2016-07-30
[5.3] Update autoload.php

mayoz:
See https://github.com/laravel/framework/pull/14564

### DIFF
--- a/bootstrap/autoload.php
+++ b/bootstrap/autoload.php
@@ -27,8 +27,4 @@ require __DIR__.'/../vendor/autoload.php';
 |
 */
 
-$compiledPath = __DIR__.'/cache/compiled.php';
-
-if (file_exists($compiledPath)) {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks for the PR. I'm not too sure this is a good idea though.

GrahamCampbell:
How does it compare performance wise?

GrahamCampbell:
Doesn't PHP cache file_exists calls?

mayoz:
@GrahamCampbell frankly, don't tested it. Perhaps, no effect on performance. Yes, PHP cache yet it but just think it's unnecessary. The solution is very simple :)

vlakoff:
Probably breaking as the `compiled.php` can be missing during dev, and I'm pretty sure there is no performance gain.

taylorotwell:
Would be breaking.

PR #3847, opened 2016-07-31
[5.3] Add target blank to links in welcome view

SteveAzz:
Just so that the user doesn't have to go black or ctrl+click


zaknes:
I know there are entire articles on why this generally isn't a good idea but I will say this. This is just a temporary screen that is rarely (never) used in production.

For me, it's one of the first things that I remove.

SteveAzz:
That is a valid point as well, up to internals if they want it or not.

PR #3849, opened 2016-08-01
Wrap .env.example variable values in quotes

philbates35:
When the developer copies `.env.example` to `.env` they are unlikely to add quotes to the values in `.env`. When the developer needs to set the value of an environment variable to a value containing a
space, as none of the existing values in `.env` are quoted, the developer is unlikely to wrap this new value with spaces in quotes. This will result in an error, as the vlucas/phpdotenv library throws
an error when setting an environment variable to a value with a space.

Quote all default environment variables by default, reducing the likelihood of the developer ever receiving the error relating to environment variables needing to be quoted when they contain spaces.

The current default values in `.env.example` don't contain any spaces, so this commit will make no difference out of the box; its only purpose is to help out the developer further down the line on the day
when they need to set an environment variable to a value containing spaces.

Also, the vlucas/phpdotenv library docs currently uses quoted variables in every example, so quoting `.env.example`'s values will lead to more consistency with those docs. This will likely stop any
confusion for the developer when reading the vlucas/phpdotenv docs.

See laravel/framework#14586, laravel/docs#2223

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,27 +1,27 @@
-APP_ENV=local
-APP_KEY=SomeRandomString
+APP_ENV="local"
+APP_KEY="SomeRandomString"
 APP_DEBUG=true
-APP_LOG_LEVEL=debug
-APP_URL=http://localhost
+APP_LOG_LEVEL="debug"
+APP_URL="http://localhost"
 
-DB_CONNECTION=mysql
-DB_HOST=127.0.0.1
-DB_PORT=3306
-DB_DATABASE=homestead
-DB_USERNAME=homestead
-DB_PASSWORD=secret
+DB_CONNECTION="mysql"
+DB_HOST="127.0.0.1"
+DB_PORT="3306"
+DB_DATABASE="homestead"
+DB_USERNAME="homestead"
+DB_PASSWORD="secret"
 
-CACHE_DRIVER=file
-SESSION_DRIVER=file
-QUEUE_DRIVER=sync
+CACHE_DRIVER="file"
+SESSION_DRIVER="file"
+QUEUE_DRIVER="sync"
 
-REDIS_HOST=127.0.0.1
+REDIS_HOST="127.0.0.1"
 REDIS_PASSWORD=null
-REDIS_PORT=6379
+REDIS_PORT="6379"
 
-MAIL_DRIVER=smtp
-MAIL_HOST=mailtrap.io
-MAIL_PORT=2525
+MAIL_DRIVER="smtp"
+MAIL_HOST="mailtrap.io"
+MAIL_PORT="2525"
 MAIL_USERNAME=null
 MAIL_PASSWORD=null
 MAIL_ENCRYPTION=null


.env.example
```diff
@@ -1,27 +1,27 @@
-APP_ENV=local
-APP_KEY=SomeRandomString
-APP_DEBUG=true
-APP_LOG_LEVEL=debug
-APP_URL=http://localhost
+APP_ENV="local"
+APP_KEY="SomeRandomString"
+APP_DEBUG="true"
```
crynobone: should `true` actually valid. Since we're expecting a `boolean` and not a `string`.

.env.example
```diff
@@ -1,27 +1,27 @@
...
-QUEUE_DRIVER=sync
+CACHE_DRIVER="file"
+SESSION_DRIVER="file"
+QUEUE_DRIVER="sync"
 
-REDIS_HOST=127.0.0.1
-REDIS_PASSWORD=null
-REDIS_PORT=6379
+REDIS_HOST="127.0.0.1"
+REDIS_PASSWORD="null"
```
crynobone: Same here.

.env.example
philbates35: I thought the dotenv library didn't parse anything to boolean (or any other types). Laravel's `env()` helper function does that (e.g converts `"true"` string into `true` boolean)? I can check later.

philbates35: From https://github.com/vlucas/phpdotenv/issues/104#issuecomment-121990200:

> These are converted in Laravel, but not in phpdotenv itself. The only reason this is not done in phpdotenv is because I originally tried to keep it as close to shell as possible, and in PHP ENV variables, everything is a string.

The phpdotenv library doesn't do any casting, meaning that `APP_DEBUG="true"` is correct, because `getenv('APP_DEBUG')` and `$_ENV['APP_DEBUG']` will return string `"true"`. Only Laravel's helper `env('APP_DEBUG')` will return boolean `true`.

crynobone: But It wouldn't because you going to pass in `"true"` and not `true` or `(true)` as expected under `env()` helper https://github.com/laravel/framework/blob/5.2/src/Illuminate/Foundation/helpers.php#L385.

Once `''`, `null`, `true` and `false` has been parse. It would strip `""` from the string as explain in  https://github.com/laravel/framework/blob/5.2/src/Illuminate/Foundation/helpers.php#L400

So you are returning a string `true`, not a boolean.

philbates35: >  you going to pass in "true" and not true or (true)

Actually you would just pass in the string `true`, resulting in boolean true being returned:
```php
// .env
APP_DEBUG="true"

// Retrieve the value...
$appDebug = env('APP_DEBUG');

// ... results in the following execution: 
function env($key, $default = null)
{
    // $key === 'APP_DEBUG'

    $value = getenv($key);
    // $value === 'true'; (string)

    // Conditional fails
    if ($value === false) {
        return value($default);
    }

    switch (strtolower($value)) {
        // strtolower('true') is also 'true' (string), so code execution goes in here
        case 'true':
        case '(true)':
            // And therefore in here, so boolean true is returned
            return true;
        case 'false':
        case '(false)':
            return false;
        case 'empty':
        case '(empty)':
            return '';
        case 'null':
        case '(null)':
            return;
    }
}
```

crynobone: If that's the case, why do we even have this commit? <https://github.com/laravel/framework/commit/9a8ebf9b60417db75de31c2e8844b7dd92781e69>

crynobone: https://github.com/laravel/framework/pull/7913

philbates35: Hm, reading through those links, I think that's relevant when the value of the environment variable actual contains the double quotes in the value e.g. `getenv('APP_DEBUG') === '"true"'`.

To set a string value like that containing the wrapped quotes using the phpdotenv library, you would have to escape the quotes when setting the value, e.g. `APP_DEBUG=\"true\"` (or `APP_DEBUG="\"true\""` if you wanted to wrap the entire string in quotes too).

taylorotwell:
True and null should not be wrapped.

philbates35:
@taylorotwell As requested, I've updated the commit so that `true` and `null` aren't wrapped in quotes.

Just to note though, as discussed with @crynobone above, I think wrapping them in quotes would still result in them being cast to their actual boolean and null types when using the `env()` helper function so i don't think it technically makes a difference. That said, its possible the developer might expect setting an environment variable to `"true"` (wrapped) would return the string rather than boolean, which wouldn't happen, and this could be misleading.

personnage:
Wrap .env.example variable values in quotes.

```
APP_ENV="local"
APP_KEY=base64:baNO+NKgzhIptGRg8ka0SD1Ny78V4o2aluOIMWxy130=
APP_DEBUG=true
APP_LOG_LEVEL="debug"
APP_URL="http://localhost"

DB_CONNECTION="pgsql"
DB_HOST="db"
DB_PORT=5432
DB_DATABASE="app_main"
DB_USERNAME="username"
DB_PASSWORD="password"

CACHE_DRIVER="redis"
SESSION_DRIVER="redis"
QUEUE_DRIVER="sync"

REDIS_HOST="redis"
REDIS_PASSWORD=null
REDIS_PORT=6379

MAIL_DRIVER="log"
MAIL_HOST="mailtrap.io"
MAIL_PORT=2525
MAIL_USERNAME=null
MAIL_PASSWORD=null
MAIL_ENCRYPTION=null

PUSHER_KEY=
PUSHER_SECRET=
PUSHER_APP_ID=
```

```
postgres_1  | LOG:  MultiXact member wraparound protections are now enabled
postgres_1  | LOG:  database system is ready to accept connections
postgres_1  | LOG:  autovacuum launcher started
postgres_1  |  done
postgres_1  | server started
postgres_1  | ERROR:  zero-length delimited identifier at or near """" at character 17
postgres_1  | STATEMENT:  CREATE DATABASE ""app_main"" ;
postgres_1  | ERROR:  zero-length delimited identifier at or near """"
postgres_1  | LINE 1: CREATE DATABASE ""app_main"" ;
postgres_1  |                         ^
laravel_postgres_1 exited with code 3
```

GrahamCampbell:
This PR needs reverting IMO. To be frank, it was a stupid idea.

GrahamCampbell:
Thanks for pointing that out @personnage. 👍

personnage:
I am glad to help!!!

taylorotwell:
Yeah let's revert this.

PR #3850, opened 2016-08-02
[5.2] Remove default APP_KEY value

crynobone:
Since we now have a default config as:

```php
'key' => env('APP_KEY'),
```

And during the `artisan key:generate` command as following:

```php
file_put_contents($this->laravel->environmentFilePath(), str_replace(
    'APP_KEY='.$this->laravel['config']['app.key'],
    'APP_KEY='.$key,
    file_get_contents($this->laravel->environmentFilePath())
));
```

The generated app key would be generated with suffix of `SomeRandomString` because it's not the original key, the original key is either `null` (or `''`).

Signed-off-by: crynobone <crynobone@gmail.com>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,5 @@
 APP_ENV=local
-APP_KEY=SomeRandomString
+APP_KEY=
 APP_DEBUG=true
 APP_LOG_LEVEL=debug
 APP_URL=http://localhost


alexbowers:
This has been discussed quite a few times, and always comes out as it doesn't want changing, since the length of the key is not the required 32 chars, it throws a security exception.

How is no key better than the default key?

alexbowers:
https://github.com/laravel/laravel/pull/3587
https://github.com/laravel/laravel/pull/3566
https://github.com/laravel/laravel/pull/3477
https://github.com/laravel/laravel/pull/3451
https://github.com/laravel/laravel/pull/3431

etc

crynobone:
> How is no key better than the default key?

Making a default of 32-char is a lazy solution to be honest. 

> This has been discussed quite a few times, and always comes out as it doesn't want changing

This is not changing anything, the config has been updated to default to `null` since December 2015 https://github.com/laravel/laravel/commit/b81af30b93ef08a630e0123bc633212df1bf5175

PR #3851, opened 2016-08-02
Created generic validator middleware

narsariamanu:
As we know in MVC Pattern:
**A model is more than just data, it enforces all the business rules that apply to that data.**

Thus writing a validator in the model itself gets the Validator much closer to the Model.
Transferring the validator from app/Http/Controllers/Auth/AuthController.php to app/User.php

    public static function validator(array $data)
    {
        $rules = array(
            'name' => 'required|max:255',
            'email' => 'required|email|max:255|unique:users',
            'password' => 'required|min:6|confirmed',
        );
        return Validator::make($data, $rules);
    }

Also updating the validator function in AutController.php

    protected function validator(array $data)
    {
        return User::validator($data);
    }

instead of previous 

    protected function validator(array $data)
    {
        return Validator::make($data, [
            'name' => 'required|max:255',
            'email' => 'required|email|max:255|unique:users',
            'password' => 'required|min:6|confirmed',
        ]);
    }

Now, if all the validator can be declared in models, a generic middleware can be designed and used instead of writing validation action in controllers. As mentioned in Laravel Middleware Docs 

> HTTP middleware provide a convenient mechanism for filtering HTTP requests entering your application.

This Validator Middleware will validate data before allowing the request to proceed further into the application.
Simplifying validation procedure using the Validator Middleware:-
1. Declare the Validator in Model
2. Add the Middleware to the controller wherever needed with Model name passed as parameter
    validator:Model_Name

### DIFF
--- a/app/Http/Controllers/Auth/AuthController.php
+++ b/app/Http/Controllers/Auth/AuthController.php
@@ -41,18 +41,14 @@ class AuthController extends Controller
     }
 
     /**
-     * Get a validator for an incoming registration request.
+     * Get a validator from the respective model.
      *
      * @param  array  $data
      * @return \Illuminate\Contracts\Validation\Validator
      */
     protected function validator(array $data)
     {
-        return Validator::make($data, [
-            'name' => 'required|max:255',
-            'email' => 'required|email|max:255|unique:users',
-            'password' => 'required|min:6|confirmed',
-        ]);
+        return User::validator($data);
     }
 
     /**
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -49,5 +49,6 @@ class Kernel extends HttpKernel
         'can' => \Illuminate\Foundation\Http\Middleware\Authorize::class,
         'guest' => \App\Http\Middleware\RedirectIfAuthenticated::class,
         'throttle' => \Illuminate\Routing\Middleware\ThrottleRequests::class,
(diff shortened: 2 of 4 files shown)


scrwdnet:
Should an Eloquent model be concerned about validation? What happens when you have different validation for creating, updating and perhaps updating a small part of that model.

Isn't this what Form Requests are for? https://laravel.com/docs/5.2/validation#form-request-validation

This way the validation is done on request even **before** controller logic.

narsariamanu:
As per generic MVC pattern, a model can provide the validation for its objects, aligning it with its real world representation.
Coming to the case of different validation for creating, updating, etc. there are different rules defined - [Validation Rules](https://laravel.com/docs/5.2/validation#available-validation-rules).
But still, there might be cases for which rules are not yet available, which as mentioned by @scrwdnet can be resolved by Form Requests.

The Middleware suggested in this PR will help in implementing similar validation behaviour for different models and help the newbies or developers accustomed with Model Validation concept to adapt to Laravel more easily.

taylorotwell:
I don't like the idea of validation being tied to Eloquent. That is why it is purposefully separated.

ps92:
[https://pbs.twimg.com/media/BvyPpvrIYAERdf2.jpg](url)

You can have a look at this. I think it wouldn't be a bad idea to elaborate to include the validation of the incoming request payload data according to the data type of the respective models.

PR #3854, opened 2016-08-03
Compliance with commonly used phpmd rules

jhoff:
This is a totally extraneous PR that just fixes a few things that we commonly see with our usage of Laravel and phpmd.

Php Mess Detector ([phpmd](https://phpmd.org)) takes a given PHP source code base and looks for several potential problems within that source. Including possible bugs, suboptimal code, overcomplicated expressions, unused parameters, methods, properties, etc.

This PR fixes a few minor things that phpmd complains about with every new project we make. Specifically this replaces variable names that are too short ($e) and removes an extraneous `else` statement.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -26,39 +26,39 @@ class Handler extends ExceptionHandler
      *
      * This is a great spot to send exceptions to Sentry, Bugsnag, etc.
      *
-     * @param  \Exception  $e
+     * @param  \Exception  $exception
      * @return void
      */
-    public function report(Exception $e)
+    public function report(Exception $exception)
     {
-        parent::report($e);
+        parent::report($exception);
     }
 
     /**
      * Render an exception into an HTTP response.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @param  \Exception  $e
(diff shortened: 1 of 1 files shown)


morloderex:
You should really instruct phpmd to use $e instead of this pr. 

👍  for the else statement tho :)

jhoff:
This is the only file in the app folder that uses a short variable name. Personally, I feel like $exception is better as it's more descriptive, but I guess thats a decision for the approver. I'm happy to pull those out and just leave the else fix.

scrwdnet:
I believe @taylorotwell wanted to leave that if/else in place as he preferred the indentation, showing a clearer path through that method. It was quashed in a previous PR, IIRC.

GrahamCampbell:
It's generally accepted and preferred to use `$e` as an exception. It's one of those special cases where a short variable name is already very clear.

bbashy:
Have to modify that file every time :P

GrahamCampbell:
phpmd needs fixing then

bbashy:
Thanks for the merge. Be good for some option to ignore $e when used for exceptions within phpmd, didn't see one myself.

PR #3855, opened 2016-08-04
give newbees a package service providers section, preventing adding after application

BartHuis:
when i started, i ended up putting all package specific service providers just at the bottom of the application service providers. But when making a route group like domain.com/{keyword}, the packages that are registering its own url's, not work anymore (like now when i installed the translations manager from barryvdh). When putting them above the app specific, everything works. So just give it to the user as a default place?

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -149,6 +149,11 @@ return [
         Illuminate\Translation\TranslationServiceProvider::class,
         Illuminate\Validation\ValidationServiceProvider::class,
         Illuminate\View\ViewServiceProvider::class,
+        
+        /*
+         * Package Service Providers...
+         */
+         
 
         /*
          * Application Service Providers...


BartHuis:
why is this check failed? :S i didn't change any code, only comments...

GrahamCampbell:
There's still cs rules there. Don't worry about it though. StyleCI will auto-fix after merge.

BartHuis:
Ah, thanks @GrahamCampbell :) will you be merging my suggestion? ;)

GrahamCampbell:
Up to Taylor to decide. :)

brayniverse:
I _always_ add this myself, though, I typically call it "Third Party Service Providers", nonetheless, it would be nice to see this merged.

BartHuis:
yeah, thats good for me too @brayniverse

GrahamCampbell:
On a side note, the first thing I do is delete these comments. :trollface:

BartHuis:
hahaha

BartHuis:
that's your choice ;)

barryvdh:
I do this also, and I leave the comments ;)

But yeah :+1: for this, especially prevents problems with a catch-all route in your own Provider.

brayniverse:
@GrahamCampbell haha, well I guess it's down to personal taste.

PR #3859, opened 2016-08-04
[5.2] Set hashing rounds to 4 when running tests

scrubmx:
This will greatly improve the performance of tests.

Since a bcrypt cost of 10 means 1024 rounds (2^10 = 1024). 
By passing 4 as the cost we only go through 16 rounds (2^4 = 16) 
this takes considerably less time achieving better performance.

The minimum cost accepted by password_hash is 4 for some reason.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -1,5 +1,7 @@
 <?php
 
+use Illuminate\Support\Facades\Hash;
+
 abstract class TestCase extends Illuminate\Foundation\Testing\TestCase
 {
     /**
@@ -19,6 +21,8 @@ abstract class TestCase extends Illuminate\Foundation\Testing\TestCase
         $app = require __DIR__.'/../bootstrap/app.php';
 
         $app->make(Illuminate\Contracts\Console\Kernel::class)->bootstrap();
+        
+        Hash::setRounds(4);
 
         return $app;
     }


alexleonard:
I find you're better off just pre-hashing a password and using that in your model factory for App\User.
eg
````
$factory->define(
    'App\User',
    function (Faker\Generator $faker) {
        return [
            'first_name' => $faker->firstName,
            'last_name' => $faker->lastName,
            'email' => $faker->safeEmail,
            // A pre-bcrypted 'testing12' password
            'password' => '$2y$10$wGpdHGdL524jVcZaZz.FzucPEgLMWm7NU09X3KAoaesJy5TpYJY2q',
            'phone' => $faker->phoneNumber,
            'status' => 'active',
        ];
    }
);
````

**Comment by martindilling** [comment]
I'm usually just doing the hashing once and using that :)

```php
$password = bcrypt('password');
$factory->define(App\User::class, function (Faker\Generator $faker) use ($password) {
    return [
        //...
        'password' => $password,
        //...
    ];
});
```

scrubmx:
@martindilling @alexleonard I did the same thing, but is very common to have a mutator in the user model so every time you call `factory(User::class)->create()` the mutator runs and the password gets hashed.

Another thing is that you **have to know** to put those functions you are talking about in your ModelFactory every time. This can save a lot of time for people that have a few dozen tests and dont know about this "hacks"

taylorotwell:
Going to hold off on this for now.

PR #3860, opened 2016-08-04
[5.3] Remove unused import

vinkla:
The `Event` facade doesn't seem to be used?


GrahamCampbell:
It's there incase you want to use it?

taylorotwell:
What Graham said.

vinkla:
Makes sense.

PR #3861, opened 2016-08-04
Revert "Removes padding on html/body to kill the scrollbar"

taylorotwell:
Reverts laravel/laravel#3853

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -19,7 +19,7 @@
                 font-weight: 100;
                 height: 100vh;
                 margin: 0;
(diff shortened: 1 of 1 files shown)


taylorotwell:
@jackmcdade actually this change looked weird to me. When you run php artisan make:auth it puts the links right against the top whereas they used to be more padded http://d.pr/i/1kFlK

jackmcdade:
Ah, okay moving it into a wrapper would make the most sense then.

vijaythecoder:
I have made fix for the login/register links here
https://github.com/laravel/laravel/pull/3862

PR #3863, opened 2016-08-05
[5.3] Remove some not necessary .gitkeep

diogoazevedos:
Remove not necessary `.gitkeep` file in `migrations` and `seeds` folder.


GrahamCampbell:
I don't think this is a good idea. We want those folders there on purpose.

PR #3865, opened 2016-08-06
Fix missing word in docblock

svenluijten:
the word "framework" was missing in the docblock above Vue in `resources/js/bootstrap.js`.

I managed to make the docblock fit Laravel's style by changing around some of the wording while keeping the general message the same.

### DIFF
--- a/resources/assets/js/bootstrap.js
+++ b/resources/assets/js/bootstrap.js
@@ -12,9 +12,9 @@ window.$ = window.jQuery = require('jquery');
 require('bootstrap-sass/assets/javascripts/bootstrap');
 
 /**
- * Vue is a modern JavaScript for building interactive web interfaces using
- * reacting data binding and reusable components. Vue's API is clean and
- * simple, leaving you to focus only on building your next great idea.
+ * Vue is a modern JavaScript framework for building interactive web interfaces
+ * using reactive data binding and reusable components. Vue has a simple and
+ * clean API, leaving you to focus only on building your next great idea.
  */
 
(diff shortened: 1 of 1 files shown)


brunocascio:
Vuejs is not a framework. Is a library. :)

`Simple yet powerful library for building modern web interfaces. http://vuejs.org`

Docs: https://github.com/vuejs/vue

svenluijten:
Damnit you're right. Let me fix that and change the wording around some
more.

svenluijten:
What do you think about it now, @brunocascio? :smile:

brunocascio:
Great @svenluijten :dancer:

resources/assets/js/bootstrap.js
```diff
@@ -12,9 +12,9 @@ window.$ = window.jQuery = require('jquery');
 require('bootstrap-sass/assets/javascripts/bootstrap');
 
 /**
- * Vue is a modern JavaScript for building interactive web interfaces using
- * reacting data binding and reusable components. Vue's API is clean and
- * simple, leaving you to focus only on building your next great idea.
+ * Vue is a modern JavaScript library for building interactive web interfaces
+ * using reactive data binding and reusable components. Vue's API is clean
+ * and simple, leaving you to focus on building your next great project.
```
GrahamCampbell: If we dropped the comma on this line, we'd have perfect alignment. 😆

svenluijten: I think that would decrease readability though, @GrahamCampbell :smile:

PR #3868, opened 2016-08-10
Use double quotes; Remove un-used fonts

tillkruss:


taylorotwell:
When you run make:auth I believe those other sizes are used.

taylorotwell:
Err nvm this is just the welcome page.

PR #3871, opened 2016-08-12
[5.3] Use const instead of var in require

diogoazevedos:


JeffreyWay:
If a person tries to reassign that elixir variable, then they deserve to see everything blow up.

PR #3874, opened 2016-08-12
[5.2] Add TokenMismatchException to the dontReport array

lancepioch:
Inspired by: https://twitter.com/reinink/status/764093723680452608

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -3,6 +3,7 @@
 namespace App\Exceptions;
 
 use Exception;
+use Illuminate\Session\TokenMismatchException;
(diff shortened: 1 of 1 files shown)


bkuhl:
I think this commit should be reverted.  When deploying Laravel applications to a highly available environment this exception can be thrown if [stickiness](http://stackoverflow.com/a/13641836/197606) is not configured correctly.  Adding this to the ignore list masks this error and can make it significantly more difficult for HA deployments to catch this problem.

lancepioch:
I feel like it'd be better to recommend against storing sessions in memory like that when you have multiple servers. This error is just so frequently thrown around that it's useless to see it the majority of the time.

PR #3877, opened 2016-08-14
GitIgnore PHPStorm's config directory

balping:
If you open a (laravel) project in phpstorm, it places an `.idea` folder in the project root. This contains project-specific ide settings. You usually don't want to include such files in a git repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 Homestead.yaml
 Homestead.json
 .env
+/.idea


diogoazevedos:
I think this ignore must be in your global ignore not in the project ignore.

philbates35:
Agree with @mul14 too, this really belongs in each developer's core.excludesfile

GrahamCampbell:
I'd assume Taylor was sick of getting PRs for this so just merged it.

brayniverse:
Surely this isn't a big deal?

If enough Netbeans or Visual Studios users feel the need for this, they could always submit a PR too.

I can't think of a good reason against allowing this.

g3n1us:
This shouldn't be in .gitignore. What if say, one wanted a file to jot notes, they would potentially call it .idea. Just a contrived example, but should illustrate why .gitignore in a core repo shouldn't suppress anything unrelated to content that doesn't exist there already.

balping:
But then following the same logic what if someone wants to collect something in Homestead.json? For example he/she's developing an app that keeps track of income and outgoings of a farm and wants to keep some initial data of egg production/day in that json file?

Ok, maybe this is a bit absurd but I hope you get my point. There are so many defaults in laravel that try to make developers' life easier but surely they can't fit for everyone. But it's never a big deal to remove some boilerplate code, rename a file or just delete *one single line* from gitignore.

Furthermore, I'm sure that more people are using phpstorm than people wanting to collect ideas in that folder.

scoutman57:
I don't like accepting this PR and I am an avid JetBrains users. What does
these people do when they work on other projects that are Laravel.. This
should be in the global gitignore along with some other system ignores.
Next, we will have all sorts of system/ide specific settings in gitignore.

brayniverse:
I'm a PHPStorm user myself and don't care either way. I've always added `./idea` to my `.gitignore` and it's never been a problem. I never knew about global gitignore settings, now I know about it, I'll be using it.

If PRs like this are a regular occurrence, and that's why Taylor accepted it. Maybe we should have an FAQ.md file which explains why `./idea` is or is not included in `.gitignore`, as well as responses for why other frequent PRs are turned down. We could make use of a [Pull Request Template](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/) to encourage contributors to read it before submitting. Just a thought.

PR #3882, opened 2016-08-16
updated: env for timezone

sudiptpa:


GrahamCampbell:
This is not something that needs to be changed per environment, but per app, so you should edit the config file.

PR #3884, opened 2016-08-17
Remove lone editor ignore rule.

Garbee:
Remove the Intellij project folder ignore rule.
Leave a comment on how to handle things like this.

This keeps the base focused on the project itself and not the tooling developers chose. However, educates them on how to handle their tooling properly instead of assuming every project they use should handle it for them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,9 @@
+# Your tools may add their own folders and files to the project.
+# Ignoring those within your global gitignore will make it so
+# they will never get commited within any project you have.
 /vendor
 /node_modules
 /public/storage
-/.idea
 Homestead.yaml
 Homestead.json
 .env


Gompje:
You're kidding right? Starting this whole discussion again? Just let it be. 

Taylor gets a gazillion of pull requests for this folder or some weird config change as it is - if having this ignore rule set in this repo officially stops a bunch, those so he/we can finally do stuff that actually matters: what's the harm?

I'm all for teaching others stuff that you think are best practices but in the end the decision lies with the developer/team/project. In this case the core team of Laravel.

GrahamCampbell:
Yeh, but now we're getting them to remove it. :/

GrahamCampbell:
Just add a comment that says don't bother sending a PR. :trollface:

Garbee:
@Gompje The problem as I see it is, projects shouldn't be ignoring tooling for people. If people wish to do so they can chose to do it on their own and in their own way. This only encourages more requests for things like `.vscode` or any other editor/tooling specific piece to get globally ignored.

This PR however instead of just removing it and "letting it be" to get requested *yet again*, adds value with a good comment to explain to developers that it is on them to handle their tooling ignores how they see fit.

scoutman57:
well since we already include atom, vim and emacs project files we should
include them all ;)

taylorotwell:
Tired of messing with this.

PR #3885, opened 2016-08-17
Create a shortcut for artisan command when using windows

xxnoobmanxx:
Just an alternative for windows users to execute 'artisan' command without 'php' prefix

### DIFF
--- a/artisan.bat
+++ b/artisan.bat
@@ -1,0 +1,2 @@
+@ECHO OFF
+php "%~dp0artisan" %*


jhoff:
FWIW, the actual artisan command isn't executable by default so mac / linux users have to type `php artisan` too...

zaknes:
Yeah I don't see many benefits of doing this unless we add one for Mac / Linux users, too.

Just create an alias.

xxnoobmanxx:
It is just something cool I've found in composer installation for windows... 
Not sure how it would be implemented for other OS

taylorotwell:
Holding off on this.

onbjerg:
Shouldn't ``chmod +x artisan`` make it executable? On macOS and Linux it would read the shebang to understand that it's a PHP file.

PR #3892, opened 2016-08-20
Remove default auth:api middleware

themsaid:
Having this middleware will enforce protection over all routes defined in `api.php`, however not all API routes needs to be protected.

I suggest we remove that middleware and allow developers to assign middleware as they see fit.

I know they can always remove it manually from the ServiceProvider, but I think it shouldn't exist by default from the beginning ¯\_(ツ)_/¯

### DIFF
--- a/app/Providers/RouteServiceProvider.php
+++ b/app/Providers/RouteServiceProvider.php
@@ -69,7 +69,7 @@ class RouteServiceProvider extends ServiceProvider
     protected function mapApiRoutes()
     {
         Route::group([
-            'middleware' => ['api'],
+            'middleware' => 'api',
             'namespace' => $this->namespace,
             'prefix' => 'api',
         ], function ($router) {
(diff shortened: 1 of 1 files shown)


app/Providers/RouteServiceProvider.php
```diff
@@ -69,7 +69,7 @@ protected function mapWebRoutes()
     protected function mapApiRoutes()
     {
         Route::group([
-            'middleware' => ['api', 'auth:api'],
+            'middleware' => ['api'],
```
diogoazevedos: The brackets can be omitted.

taylorotwell:
I've moved the middleware to the route itself.

shahidnajam:
@taylorotwell sorry As for me but you are little stupid. Think before making change than reverting the change later.

scrwdnet:
It's a shame Github doesn't have any expletive reactions...

PR #3893, opened 2016-08-21
add app name to env file

sudiptpa:
Added application name to env file


GrahamCampbell:
That's not environment specific, it's application specific.

spyric:
I think that a bad idea, because usually app name doesn't depend on environment

PR #3898, opened 2016-08-22
Laravel now supports emojis!

thomasruiz:
Use `utf8mb4` instead of `utf8` for mysql charset to be able to store emojis in the database.

This charset have been around since MySQL 5.5 and there is no loss in using it.


GrahamCampbell:
I don't think we should change this default.

taylorotwell:
We can't actually blindly change this because if people try to index the default string size generated by the schema builder they will get an error. We will have to consider it more and look at all of the things involved.

PR #3900, opened 2016-08-22
More readable report exceptions

mayoz:


taylorotwell:
I don't really like importing everything at the top like that.

mayoz:
Nope, but importing at the top in 5.1 and 5.2 versions.

PR #3902, opened 2016-08-22
firstOrNew returns unexpected results when using non mass-assignable fields

tjmckenzie:
When running Model::firstOrNew() with a non mass-assignable field (id), it always returns the very first table row, regardless if the row exists or not.

Seems to have been introduced with this commit:
https://github.com/laravel/framework/commit/df2761f0f3e24e884a9db0271044cff41238b564

### DIFF
--- a/tests/FirstOrNewTest.php
+++ b/tests/FirstOrNewTest.php
@@ -1,0 +1,90 @@
+<?php
+
+use Illuminate\Database\Eloquent\Model as Eloquent;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Foundation\Testing\WithoutMiddleware;
+use Illuminate\Foundation\Testing\DatabaseMigrations;
+use Illuminate\Foundation\Testing\DatabaseTransactions;
+
+class FirstOrNewTest extends TestCase
+{
+    public function setUp()
+    {
+        parent::setUp();
+
+        if (Schema::hasTable('first_or_new_demo')) {
+            Schema::drop('first_or_new_demo');
+        }
+
+        Schema::create('first_or_new_demo', function(Blueprint $table) {
+            $table->increments('id');
+            $table->string('name');
+            $table->timestamps();
+        });
+
+        $this->faker = Faker\Factory::create();
+
+    }
+
+    /**
+     * Works correctly when searched value is mass assignable
+     *
+     * @test
+     * @return void
+     */
+    public function it_passes_when_id_is_mass_assignable()
+    {
+        // Create three new test rows [1, 2, 3]
+        for ($j=0; $j < 3; $j++) {
+            \WorkingDemo::create(['name' => $this->faker->name]);
+        }
+
+        // This should be an existing row
+        $workingDemo = \WorkingDemo::firstOrNew(['id' => 3]);
+
+        $this->assertEquals(3, $workingDemo->id);
+    }
+
+    /**
+     * FirstOrNew should return a valid model but not when searching by a
+     * non-massasignable field
+     *
+     * @test
+     * @return void
+     */
+    public function it_fails_when_id_is_not_mass_assignable()
+    {
+        // Create three new test rows [1, 2, 3]
+    	for ($i=0; $i < 3; $i++) {
+            \Demo::create(['name' => $this->faker->name]);
+        }
+
+        // This should be an existing row
+        $failDemo = \Demo::firstOrNew(['id' => 3]);
+
+        $this->assertEquals(3, $failDemo->id);
+    }
+
+    public function tearDown()
+    {
+        Schema::drop('first_or_new_demo');
+    }
+}
+
+/**
+ * Model to show failure if ID is not mass-assignable
+ */
+class Demo extends Eloquent
+{
+    protected $table = 'first_or_new_demo';
+    protected $fillable = ['name'];
+}
+
+/**
+ * Model with mass-assignabale ID
+ */
+class WorkingDemo extends Eloquent
+{
+    protected $table = 'first_or_new_demo';
+    protected $fillable = ['id', 'name'];
+}


taylorotwell:
This PR seems to include a bunch of other stuff? @miscbits can you comment on the potential behavior problems here?

miscbits:
yeah so I have one question about this. 

The test you wrote here 'it_fails_when_id_is_not_mass_assignable'

How are you able to tell that it's returning the first row in the database? Is $failDemo->id returning as 1, or is it null? 

The point of the pull request I put in was to mutate attributes before searching for them in the database, so the only behavior change that should happen is instead of searching for exactly what you put in firstOrNew() the method will search for an item in the database after running the attributes through their setter mutators. I'm not sure that this would have caused firstOrNew to return the first row of the database unless I accidentally did something else unintentionally but basically the whole PR just added the getAttributes method before running the rest of the method. 

@taylorotwell I haven't looked at getAttributes() so I'm not sure how that behaves exactly, but do you know if it would give back an attribute that wasn't mass assignable? Also would it potentially give back a default value (like 1) if the attribute wasnt mass assignable? I can investigate further tomorrow, though I'm a little tied up at the moment.

tjmckenzie:
@miscbits The test does return $failDemo->id as 1. It seems as if the mutated attributes must be mass assignable, so in my case $mutatedAttributes is returning empty. Model::firstOrNew([]); returns the first record.

It seems to have worked well for me if I merge the arrays, so that any mutated ones will overwrite the original, but I'm not 100% sure this is the best approach.
`$mutatedAttributes = array_merge($attributes, $this->model->newInstance($attributes)->getAttributes());`

miscbits:
So wait firstOrNew([]) returns the first row? Seems like this is an issue in of itself no?

tjmckenzie:
It does. Confirmed in 5.3 and 5.2 as returning the first record.

However, 5.2 lets me use a non mass-assignable field with no problem with firstOrNew()

zdkroot:
@miscbits The problem is the `newInstance()` call. Any attributes in that array which are not mass assignable will not be present on the created model, so `getAttributes()` does not return them, which means they are not searchable.

zdkroot:
Also I think `firstOrNew([])` returning the first row - if one exists - is the correct behavior. It would be equivalent to `first()` then, which makes sense to me.

miscbits:
that call seemed like the only appropriate one for firstOrCreate which is the method with the error I was trying to fix. perhaps I could just use the forceFill method for first or new since I would not be creating an instance in the database when using firstOrNew. Does this seem like a reasonable fix?

tjmckenzie:
I don't feel there is a ned to use forceFill(), since a new'd instance already bypasses mass assignment. So far the array_merge suggestion seems to be working just fine with both non mass-assignables as well as returning attributes with proper mutators.

`/**
     * Get the first record matching the attributes or instantiate it.
     *
     * @param  array  $attributes
     * @return \Illuminate\Database\Eloquent\Model
     */
    public function firstOrNew(array $attributes)
    {
        $mutatedAttributes = array_merge($attributes, $this->model->newInstance($attributes)->getAttributes());

        if (! is_null($instance = $this->where($mutatedAttributes)->first())) {
            return $instance;
        }

        return $this->model->newInstance($attributes);
    }

    /**
     * Get the first record matching the attributes or create it.
     *
     * @param  array  $attributes
     * @param  array  $values
     * @return \Illuminate\Database\Eloquent\Model
     */
    public function firstOrCreate(array $attributes, array $values = [])
    {
        $mutatedAttributes = array_merge($attributes, $this->model->newInstance($attributes)->getAttributes());

        if (! is_null($instance = $this->where($mutatedAttributes)->first())) {
            return $instance;
        }

        $instance = $this->model->newInstance($attributes + $values);

        $instance->save();

        return $instance;
    }`

miscbits:
There is one issue with that though. If there is a mutator on an attribute that is not mass assignable, then the array merge fix will not mutate the attribute. This means that we come back into the original issue.

zdkroot:
@miscbits Ah  you're right, I was mistaken on how `array_merge()` works - I thought it overwrote keys that were present in the second array. `[array_replace()](http://php.net/manual/en/function.array-replace.php)` is the method we want, or something similar.

miscbits:
either that or we could introduce a new internal method to get all attributes regardless of whether or not they are fillable. I think forceFill seems like the better option though. 

Yes it is valid to have a mutator on an attribute that isn't mass assignable. you can still write a method to update a model without using mass assignment, therefore you can still put a mutator down on the method.

zdkroot:
@mcbits yes I think `forceFill()` is fine here - no reason to add methods if an existing one cleanly solves the problem :)

taylorotwell:
Yes, use `forceFill` please update when the PR is ready to merge.

miscbits:
didnt know if @tjmckenzie was gonna do it or not but I threw in a new PR to fix this issue then. #15003

Rainyeah:
ni men zai shuo shen me , lao zi kan bu dong!

miscbits:
@Rainyeah would you like someone to explain it to you?

overtrue:
@miscbits Sorry, Plz don't care about it. He's just a fool need of treatment.

JellyBool:
@miscbits Don't mind this guy , Sorry about that.

miscbits:
figured it was a troll but whatever. Thanks

Rainyeah:
@miscbits 
I'am sorry that i did not get the meaning of the code as i'am new.
Please ignore my previous comment.

Rainyeah:
@overtrue  i don't think calling other people fool is right thing. i'm still learning and everyone was new in the past.
/* ni yi wei ni hui shuo ying wen jiu hen 6 ma ,wo ye hui */

abrahamgreyson:
@Rainyeah Don't embarrass yourself.

taylorotwell:
This is on the entirely wrong repository.

grhaonan:
hej guys:
Does Lumen 5.4.6 still have this issue? I met the similar issue on 5.4.6 with firstornew function today and it was solved by added the search variables into the $fillable array in model (means these variables are mass-assignment supported)

Thanks

miscbits:
@grhaonan The PR that caused this issue was removed so it should be fine now.

PR #3903, opened 2016-08-23
Use class notation

jmarcher:


GrahamCampbell:
Thanks. We already fixed this on master. Develop will be fixed fin good time, though now develop represents Laravel 5.4, which is a while away. :)

PR #3905, opened 2016-08-24
[5.3] Removing strict mode for MYSQL

fernandobandeira:
IMO this was a bad change and one that will break a lot of installations, check the issue:

laravel/framework#14997

Maybe we should let this true but pass only some modes, the ```ONLY_FULL_GROUP_BY``` for example is breaking some things.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -62,7 +62,7 @@ return [
             'charset' => 'utf8',
             'collation' => 'utf8_unicode_ci',
             'prefix' => '',
-            'strict' => true,
+            'strict' => false,
             'engine' => null,
         ],
 


scoutman57:
I agree with this.. Not that my opinion matters.

taylorotwell:
I spoke with a MySQL developer who actually recommends we have this on. It can catch some gnarly bugs in your code. If you don't want it you can easily turn it off.

fernandobandeira:
@taylorotwell Is there a chance that we could magically add all selected columns to the groupBy when generating a query instead of forcing the user to add all of them manually?

DeeJaVu:
@taylorotwell Then please help me fix this code

```php
    public function getAllWithUsersCount()
    {
        $prefix = DB::getTablePrefix();

        return Role::leftJoin('role_user', 'roles.id', '=', 'role_user.role_id')
            ->select('roles.*', DB::raw("count({$prefix}role_user.user_id) as users_count"))
            ->groupBy('roles.id')
            ->get();
    }
```

i get this error

`SQLSTATE[42000]: Syntax error or access violation: 1055 'myappdev.roles.name' isn't in GROUP BY (SQL: select `roles`.*, count(role_user.user_id) as users_count from `roles` left join `role_user` on `roles`.`id` = `role_user`.`role_id` group by `roles`.`id`)`

PR #3907, opened 2016-08-24
Adding customization Global "From" Address and Name to .env file

rohmanhm:
I make a change in env file that 

> Global "From" **Address** and **Name**

 can be easily replaced.

Sometimes we want to quickly change the `Global From Address and Name`. And I thing it can be easier for a new start using Laravel


rohmanhm:
Sorry I forgot, if you have a lot of email might not work correctly

PR #3908, opened 2016-08-24
Complete menu example

borisdamevin:


taylorotwell:
It is fine how it is I think. It's common for websites to look like that.

PR #3915, opened 2016-08-29
Remove unused import

dimsav:
Importing `Illuminate\Support\Facades\Gate` is no longer needed.


GrahamCampbell:
See my other comment.

PR #3916, opened 2016-08-29
Remove unused import

dimsav:
The import of `Illuminate\Support\Facades\Event` is not needed.


GrahamCampbell:
Yeh, but we put it there so people don't have to import it for themselves. Really useful for beginners to avoid getting App\Event not found errors.

PR #3917, opened 2016-08-30
Remove Mandrill from Mail Driver options

juukie:
Mandrill has been [removed from the services.php config file](https://github.com/laravel/laravel/commit/ea52a963869f6f798ba7d3df995c4f0630775134) but is still mentioned in the mail drivers list.
[The mail documentation](https://laravel.com/docs/5.3/mail) doesn't speak about Mandrill so it would make sense to delete it from the list.

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -11,7 +11,7 @@ return [
     | sending of e-mail. You may specify which one you're using throughout
     | your application here. By default, Laravel is setup for SMTP mail.
     |
-    | Supported: "smtp", "mail", "sendmail", "mailgun", "mandrill",
+    | Supported: "smtp", "mail", "sendmail", "mailgun",
     |            "ses", "sparkpost", "log"
     |
     */


GrahamCampbell:
Mandrill is still supported.

GrahamCampbell:
https://github.com/laravel/framework/blob/5.3/src/Illuminate/Mail/Transport/MandrillTransport.php

GrahamCampbell:
Not marked as deprecated either.

juukie:
Than I suggest adding back the mandrill config to the services.php and mention it in the docs. Otherwise users will have to find out how to use the Mandrill driver..

GrahamCampbell:
I'd say adding it back to services.php would be fine, but not necessarily the docs. Plenty of things arn;t documented on purpose to discourage their use in case we later want to deprecate them. For example, the docs no longer feature facade aliasing from 5.3 upwards.

shahidnajam:
@juukie you care about users alot. You know how to use Mandrill driver. That's enough :)

juukie:
@GrahamCampbell makes sence. Should I create a PR reverting the removal of the Mandrill config?

@shahidnajam I do. I'm don't think that's enough. It makes no sence removing mandrill config and keeping mailgun, ses and sparkpost. We're talking about four lines of code.

PR #3920, opened 2016-08-31
Remove irrelevant comment line

legshooter:
I've been really busting my balls on this golden 3 char rule, only to realize wtf am I doing. Do your magic TayTay, 'cause that email line has become obsolete.

### DIFF
--- a/config/auth.php
+++ b/config/auth.php
@@ -81,9 +81,7 @@ return [
     | Resetting Passwords
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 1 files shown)


AbdelkaderBah:
Well this may be useful for some beginners don't be selfish ^_^

legshooter:
I'm not sure we understood each other, I am referring to #3823

taylorotwell:
I'll fix it.

PR #3922, opened 2016-09-02
Environment configuration for HTTPS only cookie

jasonmccreary:
As this likely changes between environments, figured it should be part of the environment configuration. Not entirely happy with the name though. While descriptive, `SESSION_HTTPS_ONLY_COOKIES` seemed long.


config/session.php
```diff
@@ -161,7 +161,7 @@
     |
     */
 
-    'secure' => false,
+    'secure' => env('SESSION_SECURE_COOKIE', false),
```
it-can: I think ```SESSION_COOKIE_SECURE``` is a better name

GrahamCampbell: Why not `SECURE_SESSIONS`?

jasonmccreary: Was trying to match the other names with the `SESSION_` prefix. If you're open to breaking trend, I like `SECURE_SESSION_COOKIES` best.

jeroennoten: I think `SESSION_SECURE` is the right choice, because we already have `SESSION_DOMAIN`.

PR #3925, opened 2016-09-04
[5.3] Lodash is unnecessary

hiendv:
I couldn't find any scripts in the boilerplate which depends on Lodash. Moreover, we should require Lodash methods separately instead of including the whole library, if necessary.

### DIFF
--- a/resources/assets/js/bootstrap.js
+++ b/resources/assets/js/bootstrap.js
@@ -1,6 +1,4 @@
 
-window._ = require('lodash');
-
 /**
  * We'll load jQuery and the Bootstrap jQuery plugin which provides support
  * for JavaScript based Bootstrap features such as modals and tabs. This
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
     "laravel-elixir": "^6.0.0-9",
     "laravel-elixir-vue": "^0.1.4",
     "laravel-elixir-webpack-official": "^1.0.2",
-    "lodash": "^4.14.0",
     "vue": "^1.0.26",
     "vue-resource": "^0.9.3"
   }


GrahamCampbell:
We're not forcing the dependency on anyone. I think this is a good default.

👎 For this PR.

resources/assets/js/bootstrap.js
```diff
@@ -1,6 +1,4 @@
 
-window._ = require('lodash');
```
crynobone: > We're not forcing the dependency on anyone. I think this is a good default.

Then I would think this line should be removed, to allow it not to be a "forced" default to anyone.

RryLee: Agree, laravel just give a alternative template, if you dont want then you can remove by yourself

GrahamCampbell: It's not forced on people because this is laravel/laravel, not laravel/framework. This line should remain here. It's not forced because people can delete it if they want.

hiendv: I know we are all talking about `laravel/laravel`. But we cannot just include bunches of **unnecessary** dependencies (_lodash is not even a lightweight one: 21kB gzipped_) and tell people to remove them if they don't like.

dwightwatson: I agree, it's not used so don't include it.

taylorotwell:
Going to keep it for now. Will evaluate in the future if necessary.

balping:
Re-evaluate for L5.5?

GrahamCampbell:
Laravel 5.5 already has a way to strip out the frontend assets. I'm not sure if it'll remove lodash though. It's a good shout. // cc @taylorotwell

balping:
@GrahamCampbell  Just tried it, `artisan preset none` doesn't remove lodash. In fact, it doesn't change `bootstrap.js` at all. This might be a bigger problem that deserves a separate issue I think, since even `require('bootstrap-sass');` is left in, which might be not desired.

devcircus:
Your entire frontend is customized to some extent on every install. Personally I don't even get the presets. Seems they were added to hopefully cut down on the complaints by devs that really don't understand just how simple it is to rip out the defaults and start fresh. 
it's always easier to delete things like this than to add it. Hell it's macroable. Make one called "clean" and drop it in the community frontend presets organization and call it a day.

hiendv:
Just want to remind people, tree-shaking is not compatible with lodash. You may want to consider importing the needed functions explicitly or using lodash-es with webpack2.

PR #3926, opened 2016-09-04
Ignore .DS_Store

BlueGood:


kamui545:
It should be ignored in your global .gitignore as it is OS specific.

See similar PR https://github.com/laravel/laravel/pull/3523

scoutman57:
Well if we are going to include .idea in the .gitignore we might as well include this because i would say this might be as common as .idea. We should A. include it or B. stop including IDE/OS or ENV specific stuff in .gitignore. I know @taylorotwell included it because he said he was tired of dealing with it. Maybe the better option is to remove it since global is the best place for this types of things. Then create a PR template that says DO NOT SUBMIT .gitignore file changes to PR. 

https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/

GrahamCampbell:
.idea shouldn't be there either. It just got merged to shut people up.

PR #3927, opened 2016-09-04
Updated config/mail.php, MAIL_FROM_* env options

ScraperDave:
Added MAIL_FROM_ADDRESS and MAIL_FROM_NAME instead of just nulls.

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -54,7 +54,7 @@ return [
     |
     */
 
-    'from' => ['address' => null, 'name' => null],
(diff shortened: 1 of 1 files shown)


tpraxl:
+1, because that is one of the first things, I do when setting up a new project. However, the changes should include the following lines in .env.example then:
```
MAIL_FROM_ADDRESS=null
MAIL_FROM_NAME=null
```

taylorotwell:
can add to your own app if needed.

PR #3928, opened 2016-09-05
Add missing space

olde86:
Add a missing space in the indentation. from 7 spaces to 8


GrahamCampbell:
I think this was missed out in order to lign it up with the dot above.

olde86:
Hah.. Developer zombie mode

PR #3930, opened 2016-09-05
Removing .idea from .gitignore

fernandobandeira:
IDE's and OS specific files shouldn't be in the project .gitignore, they should be added manually.

The reason i'm removing this even thought it was recently merged is because having this here will open the doors for many other PR's for specific IDE's and OS files, as you can check here: #3926 

Also at least when i start a new project PHPStorm asks to include .idea in the .gitignore automatically so theres no benefit on having this as default.

**Note**
At this point probably people are pretty tired of .gitignore PR's so maybe we should just add a [template](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/) for PR's and give a warning there to at least decrease the number of PR's including IDE's and OS files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 /node_modules
 /public/storage
 /vendor
-/.idea
 Homestead.json
 Homestead.yaml
 .env


jonagoldman:
is this ever going to end?

fernandobandeira:
@jonagoldman Having this file in the .gitignore will just increase the number of PR's, as it will encourage the creation of them. IK that it was merged because people were tired of it but it will only make things worse.

scoutman57:
This is why I suggested PR templates!!!!

GrahamCampbell:
sigh

GrahamCampbell:
This literally doesn't even matter. :/

fernandobandeira:
Well i'm closing this, honestly for me it doesn't matter if this file has every single IDE and OS files being ignored or not i just think that ignoring .idea and not .DS_STORE or any other common file is nonsense since there is no standard and it will encourage other people to create PR's adding the "missing" files.

FWSimon:
I think it's a good idea to have it

jonagoldman:
it does't matter. there will always be a PR to either add it or remove it. its kind of a quantum thing.

tpraxl:
You know what? Maybe it would help to replace `/.idea` in .gitignore with

```
# Do not put IDE or OS specific ignores in your local .gitignore file 
# Instead ignore it globally. See this link: https://help.github.com/articles/ignoring-files/
# /.idea
```

That way, chances are, that people visit the .gitignore file in order to add IDE specific ignores, but follow the better practice after they have read your comment. This might even reduce the amount of PRs on that.

fernandobandeira:
@tpraxl That sounds good, i'm not creating the PR though.

PR #3931, opened 2016-09-05
Added Home link to header when a user loggedin

azeemhassni:
After **make:auth** when you login on the **/** route 
you can still see the register & login routes. 

I've just added a check to add **Home** link if there is a login session.


taylorotwell:
It's fine.

azeemhassni:
OK

PR #3932, opened 2016-09-06
Remove the env() for pusher configuration values

christiansyl:
Ive seen that the pusher config values are set to null when trying to create a new pusher instance in the Pusher classe constructor. By removing the env setting the values were properly set and the pusher isntance was properly created.

### DIFF
--- a/config/broadcasting.php
+++ b/config/broadcasting.php
@@ -32,9 +32,9 @@ return [
 
         'pusher' => [
             'driver' => 'pusher',
-            'key' => env('PUSHER_KEY'),
-            'secret' => env('PUSHER_SECRET'),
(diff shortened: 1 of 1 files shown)


arubacao:
@christiansyl check out this [docs](https://laravel.com/docs/5.3/configuration#environment-configuration).

christiansyl:
Thanks for the reference to the docs. I guess I read it a while ago and I forgot about it (ok im pretty new to Laravel I must admit).... One thing that can be confusing is that only some conf elements are stored by default in the .env file in  afresh installation. Therefore you have some config files that includes entries with the env notation and some withouth...
But now that I re-read the doc it all makes sense :-|

c.

PR #3934, opened 2016-09-10
Update passwords.php

Shonetow:
Double qoutes into single quotes.


RonMelkhior:
@Shonetow You might want to escape the `'` in the middle of the string.

Shonetow:
My bad, don't ask.... :D

PR #3935, opened 2016-09-11
usused Illuminate\Support\Facades\Gate

dyktek:


GrahamCampbell:
Thanks, but this was included so people can use it.

RonMelkhior:
You should delete `composer.lock` from the request.

PR #3937, opened 2016-09-12
missing env() in config/app.php

sergeprokofev:
missing env() for name, timezone, locale, fallback_locale in app/config.php file

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -12,7 +12,7 @@ return [
     | any other location as required by the application or its packages.
     */
 
-    'name' => 'My Application',
+    'name' => env('APP_NAME', 'My Application'),
 
     /*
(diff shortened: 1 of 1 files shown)


config/app.php
```diff
@@ -12,7 +12,7 @@
     | any other location as required by the application or its packages.
     */
 
-    'name' => 'My Application',
+    'name' => env('APP_NAME', 'My Application'),
```
GrahamCampbell: This doesn't need to be an environment variable really.

ejdelmonico:
I agree. There is nothing there that needs to be hidden from anyone.

gmsantos:
Can this settings really vary in different environments?

PR #3938, opened 2016-09-13
Get cache key prefix from .env

BlueGood:
Sometimes, development environment and test environment use the same Memcached.


taylorotwell:
Feel free to change this in your own project.

PR #3939, opened 2016-09-13
`dependencies` instead of `devDependencies`?

itsazzad:
Should it not be `dependencies` instead of `devDependencies`?


GrahamCampbell:
No, I think the default setup is you build and commit production ready css/js and then the deployment server doesn't need to build anything.

taylorotwell:
Closing but of course you are free to modify it to whatever you want in your own project.

PR #3943, opened 2016-09-14
Added Http Kernel to createApplication

acacha:
As talked with @themsaid on issue:

https://github.com/laravel/framework/issues/15426

It seems a solution to "mimic" real app with tests when using programatically added Middlewares. But maybe there is a better solutions of something we miss.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -18,6 +18,8 @@ abstract class TestCase extends Illuminate\Foundation\Testing\TestCase
     {
         $app = require __DIR__.'/../bootstrap/app.php';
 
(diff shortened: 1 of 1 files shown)


themsaid:
Not sure if that's the best way to do it, but here's the problem:

While running tests we initialise the `Console/Kernel` and then call `bootstrap()` which will run the service providers, a call to `pushMiddlewareToGroup()` in a service provider expects the `Http/Kernel` to be initialised, however that won't be the case.

taylorotwell:
This isn't the right approach.

acacha:
@taylorotwel ok but then this is not a bug or I'm doing something wrong?

PR #3944, opened 2016-09-14
Default timezone to inherit php.ini date.timezone

heimer16:
date_default_timezone_get() defaults to UTC anyway if not set in php.ini

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -64,7 +64,7 @@ return [
     |
     */
 
-    'timezone' => 'UTC',
+    'timezone' => env('APP_TIMEZONE', date_default_timezone_get()),
 
     /*
     |--------------------------------------------------------------------------


GrahamCampbell:
👎 from me. This makes the app's behaviour unpredictable.

GrahamCampbell:
The whole point of this config was to not fallback to that php function surely?

scoutman57:
Yeah, I agree with Graham. This is a bad idea and can cause
unpredictability.

lancepioch:
> Also, in my case, when I do want it to inherit the timezone, there's no way to do it in .env, (unless there's some "inherit" option?). I have to edit app.php which is fine I guess, but it seems to defeat the purpose of .env.

The issue here is that .env changes represent being able to change to other environments. Putting this entry in the environment file basically tells people another variable that they have to go change it to their own value to work. But, unlike a database, you can't "migrate" the timezone or set up a blank redis instance or an api key... etc. Instead you have a list of hundreds of timezones with usually only 1 of them ever being invalid that you can't always guess.

For the majority of users, they either pick UTC or their local time zone and stick with it, forever (until they need internationalization and maybe switch to UTC). This really shouldn't be something that changes frequently or at all for that matter. It should be a one and done setup.

taylorotwell:
Will just leave as is. You can change it yourself if you want in your own application.

PR #3946, opened 2016-09-17
dependencies scope

crabbly:
I know this isn't a big deal, but it would be nice to separate dev dependencies from run time dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,12 +5,14 @@
     "dev": "gulp watch"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


rs-sliske:
> No, I think the default setup is you build and commit production ready css/js and then the deployment server doesn't need to build anything.

https://github.com/laravel/laravel/pull/3939

taylorotwell:
Will just leave as-is and let people adjust if they want / need to.

PR #3947, opened 2016-09-18
Fix pgsql schema and sslmode from .env

mikaelosterberg:


GrahamCampbell:
Thanks, but please send each change as a separate PR so they can be reviewed separately.

PR #3949, opened 2016-09-18
S3 Disk Filesystem config from .env

mikaelosterberg:


GrahamCampbell:
:-1: from me here.

mikaelosterberg:
@GrahamCampbell please explain to me way s3 secrets have to end up in the repository and way it is good to have this secrets hardcoded.

LKDevelopment:
You can do this in your own application, but it isnt something that needs to be out of the box.

dimkasta:
This would push ppl to follow good practice. +1 from me

And +1 to apply this in Socialite as well

mikaelosterberg:
@LKDevelopment let's do that then.

LKDevelopment:
@mikaelosterberg So and what is if the application doesn't need it? So change this on your own app... (not really my opinion, i want more env specifiy too but that is the arguments @taylorotwell or @GrahamCampbell saids in the history)

config/filesystems.php
```diff
@@ -56,10 +56,10 @@
 
         's3' => [
             'driver' => 's3',
-            'key' => 'your-key',
```
tpraxl: Tempts People to put their secrets directly into the source code that is version controlled. This would be a serious issue.

config/filesystems.php
```diff
@@ -56,10 +56,10 @@
...
-            'secret' => 'your-secret',
-            'region' => 'your-region',
-            'bucket' => 'your-bucket',
+            'key' => env('FS_S3_KEY', ''),
```
tpraxl: Guides developers (especially newbies and people under time pressure) to enter sensitive data at the right place.
If one of the env variables (e.g. `FS_S3_KEY`) is not present, then that's not a big deal, because the default value will be used.

mikaelosterberg: If only key is fetched from env you still have secret in source.

tpraxl: The line was just taken as an example. I intended to address all changes. Sorry for the irritation.

jakesyl: @tpraxl I don't think many newbies are using S3 file systems.

chenyongze:
The only supported ciphers are AES-128-CBC and AES-256-CBC with the correct key lengths.

config/filesystems.php
tpraxl: @jakesyl: Just a slightly offtopic note about #3932: What would have been the next step of christiansyl if env would not have been used there? I believe the requester put the sensitive data directly into config/broadcasting.php – that worked. But it would have been bad practice. So +1 for closing #3932 and +1 for merging this one.

tpraxl: @jakesyl: Consistency is an important point. I agree.

It makes no sense to put every configuration in the .env file. So you cannot be 100% consistent here. 

However, the consistent point should be: "sensitive data and configurations likely to change between local, testing, production should be configured in .env" (out of the box) from. Just my point of view.

JellyBool:
@chenyongze Don't make a comment like this. Your comment has nothing to do with this pr if I get it right .

taylorotwell:
Holding off on this since it seems to be fairly debated.

scoutman57:
I agree that any env specific info (dev, qa, prod) should by nature be pulled from the env or a env specific config file. I think what many people are missing is that all the files in your /config directory can be changed per env and do not even need to be checked into the repository. The way we have it is we create an _example file in /config  and then change them per env either manually or via CI/CD

PR #3950, opened 2016-09-19
Update AuthServiceProvider.php

lovePizza:
Removed import of unused class


jadjoubran:
This is kept here for convenience. You'd expect people to use the Gate facade here

PR #3952, opened 2016-09-20
Directive 'asp_tags' is no longer available in PHP

Edrard:
Good day
After upgrade from 5.6 to 7.0.10 on Laravel 5.2 I start receiving such error
http://i.imgur.com/wiXHUaP.png


scoutman57:
Yes, ASP style tags have been deprecated for a long time and finally
removed.  How is this related to Laravel and why the PR into master?

You should fix all your <% %> or <%= %> tags

PR #3963, opened 2016-09-24
added env SES_REGION

Jameron:
The services.php file should use env() for ses region.

### DIFF
--- a/config/services.php
+++ b/config/services.php
@@ -22,7 +22,7 @@ return [
     'ses' => [
         'key' => env('SES_KEY'),
         'secret' => env('SES_SECRET'),
-        'region' => 'us-east-1',
+        'region' => env('SES_REGION'),
     ],
 
     'sparkpost' => [


zaknes:
Can you add the default value back, please?

```php
env('SES_REGION', 'us-east-1')
```

taylorotwell:
It's fine as is. Can convert to environment in your own app if needed.

Jameron:
Seems inconsistent to me.

Jameron:
Also, this:

  'key' => env('SES_KEY'),
          'secret' => env('SES_SECRET'), 

would make more sense if it was 

 'key' => env('AWS_KEY'),
          'secret' => env('AWS_SECRET'),

Jameron:
This config file makes me uneasy bro.

erikwestlund:
There have been numerous recommendations to add extra ENV variables to the
config, and Taylor et al. have consistently given the answer Taylor just
did.

Not everyone uses AWS.  I don't think it's necessary to have AWS-specific
config in the env file when it is so easy to adapt.

Jameron:
Take a look at the config/services.php file ...

erikwestlund:
Missed that we're speaking of this specific file. My bad.

The region does not need to be kept out of version control as it's not
sensitive.  (See S3 links)

I think this is the guiding principle for the choice.

Erik

Jameron:
Not sensitive yes, but it can be specific to an environment.  E.g. using latency based routing policy with multi-region setups. West coast servers should use west coast SES endpoint, and East coast servers should use east coast SES endpoint. 

http://docs.aws.amazon.com/ses/latest/DeveloperGuide/regions.html

PR #3967, opened 2016-09-27
Moving User.php to Models/User.php & Updating config/auth.php

pfuri:
The User model is now in a dedicated Models folder rather than the root.  This keeps the root free of clutter and makes the default laravel file structure more developer friendly right out of the box.  The User model namespace and Auth config have been updated appropriately.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace App;
+namespace App\Models;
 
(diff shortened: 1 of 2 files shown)


rdrubis:
The older version of Laravel used to have a Models directory but they (Taylor and other core devs) agreed to remove it and store the models generated by **php artisan make:model** to the **app** directory by default. Of course you can recreate the Models directory as needed. It is a personal preference.

PR #3975, opened 2016-10-01
Update TestCase.php

anuuthomson1412:
easier to read


kamui545:
That does not follow the Laravel coding style, you can see it on StyleCI it is correct as it is.

http://www.php-fig.org/psr/psr-2/

PR #3978, opened 2016-10-02
Use app name in ExampleTest

danieliancu:
This will ensure tests continue to pass after changing the app name.


lagbox:
Why would it be failing? The `welcome.blade.php` view makes no reference to anything from the config and 'Laravel' is directly written into the view.

danieliancu:
@lagbox You're right, it was only happening in my app (using the app name on the home page).
Thanks for letting me know.

I am closing this now.

lagbox:
No problem.

PR #3976, opened 2016-10-01
update dependencies versions

mayoz:

PR #3979, opened 2016-10-03
Update to use Vue.js 2.0.x

etelford:
Minor updates to deal with deprecations in Vue.js 2. 

I've [submitted an accompanying PR for Passport](https://github.com/laravel/passport/pull/117).

### DIFF
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -16,5 +16,5 @@ require('./bootstrap');
 Vue.component('example', require('./components/Example.vue'));
 
 const app = new Vue({
-    el: 'body'
+    el: '#app'
 });
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,6 +1,6 @@
 const elixir = require('laravel-elixir');
 
-require('laravel-elixir-vue');
+require('laravel-elixir-vue-2');
 
 /*
  |--------------------------------------------------------------------------
--- a/package.json
+++ b/package.json
@@ -9,10 +9,10 @@
     "gulp": "^3.9.1",
     "jquery": "^3.1.0",
     "laravel-elixir": "^6.0.0-9",
-    "laravel-elixir-vue": "^0.1.4",
+    "laravel-elixir-vue-2": "^0.2.0",
     "laravel-elixir-webpack-official": "^1.0.2",
-    "lodash": "^4.14.0",
-    "vue": "^1.0.26",
-    "vue-resource": "^1.0.2"
+    "lodash": "^4.16.2",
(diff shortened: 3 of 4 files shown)


JellyBool:
@JeffreyWay said he will make a new pr for this.

https://github.com/laravel/laravel/pull/3976#issuecomment-250919074

etelford:
@taylorotwell @JeffreyWay I did the bare minimum to get the current functionality to work with Vue 2. Feel free to close if you'd rather take care of this.

taylorotwell:
@JeffreyWay what do you think about this PR? Look ok?

JeffreyWay:
@taylorotwell Looks good to me. Was going to say make a PR for Passport too, but looks like he already did, so I think we're all set here.

yoosuf:
Go ahead @taylorotwell

resources/assets/js/app.js
```diff
@@ -16,5 +16,5 @@ require('./bootstrap');
 Vue.component('example', require('./components/Example.vue'));
 
 const app = new Vue({
-    el: 'body'
+    el: '#app'
```
taylorotwell: Do we actually have an #app element in the default layout?

themsaid: No, it needs to be added.

taylorotwell:
OK let's add that first before merging this so it's not broken.

resources/assets/js/app.js
etelford: This got merged — should I still add an #app element? Incidentally, I don't think app.js is included by default either. I can add that too if you think it's appropriate.

unitedworx: @etelford regardless of manually including app.js and manually adding id="app" vue 2 wont work out of the box! 

There seems to be something wrong with the overall configuration. it keeps looking for the #app even if its added there manually!

etelford: @unitedworx Are you starting with a fresh Laravel install or attempting to update an existing install?

unitedworx: @eterfold fresh laravel install... my bad thought! turns out that including my app.js at the end of the body instead of the head saved the day!  That's what happens when you live enough time in the jquery world and everything lives inside a document ready function.

I guess welcome.blade.php needs to include a call call to js/app.js at the end of the body tag to make it clearer for others in the future. Maybe a div with an app ID as well as the example vue js component but commented out there in the code.

etelford: @unitedworx Good to hear.

Yes. Even before this PR I believe you had to manually add `app.js` to the welcome view.

PR #3983, opened 2016-10-05
Encrypt the user password on the model

shadoWalker89:
The logic for how the password is encrypted will be defined in one place.


yoosuf:
This is not necessarily needed

jeroennoten:
Note that user creation in the `RegisterController` also has to change then: https://github.com/laravel/laravel/blob/master/app/Http/Controllers/Auth/RegisterController.php#L68

shadoWalker89:
Nice catch @jeroennoten, thanks

GrahamCampbell:
Thanks but this has already been proposed and rejected.

PR #3984, opened 2016-10-05
The Application name should be on the .env file

rsm23:
Since the .env file holds most of the application’s configurations the
application name also should be there, to be reachable as easy as all
the .env configurations.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,7 +3,7 @@ APP_KEY=
 APP_DEBUG=true
 APP_LOG_LEVEL=debug
 APP_URL=http://localhost
-APP_NAME='Laravel APP'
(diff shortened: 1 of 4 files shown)


taylorotwell:
Your application name does not change from dev to production.

scoutman57:
I think people don't know what ENV really is suppose to be for. I think
they confuse it with a "application config" file and not a "environment
specific config" file

GrahamCampbell:
Yeh, I totally agree @scoutman57.

ejdelmonico:
Um. I would say that @rsm23 should read about the use of .env variables and the application thereof.

PR #3876, opened 2016-08-13
DocBlock addition for Model Factories

shehi:
For ```develop``` branch

PR #3987, opened 2016-10-07
Add typehint for Factory

barryvdh:
So that an IDE can autocomplete the calls.

### DIFF
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -10,7 +10,7 @@
 | database. Just tell the factory how a default model should look.
 |
 */
-
+/** @var $factory \Illuminate\Database\Eloquent\Factory */
 $factory->define(App\User::class, function (Faker\Generator $faker) {
     static $password;
 


JellyBool:
Nice job ! @barryvdh

spyric:
It was before #3876 and closed by Taylor

GrahamCampbell:
This is not a typehint, this is just a workaround for an IDE. We shouldn't need to change code at all for an IDE. That defeats.

GrahamCampbell:
Meh, leaving open incase Taylor wants to reconsider.

:-1: from me though.

barryvdh:
We're not changing code, it's just basically impossible for an IDE to detect where the $factories is coming from.
Even for us humans, we can't tell that by just looking at the code..

taylorotwell:
I'll merge this one but I really don't want to get carried away with these IDE work-arounds.

database/factories/ModelFactory.php
```diff
@@ -10,7 +10,7 @@
 | database. Just tell the factory how a default model should look.
 |
 */
-
+/** @var $factory \Illuminate\Database\Eloquent\Factory */
```
arcanedev-maroc: This typehint must be inverted, like this:

```
/** @var \Illuminate\Database\Eloquent\Factory $factory */
```

barryvdh: Ah yeah,  https://phpdoc.org/docs/latest/references/phpdoc/tags/var.html
Whoops

arcanedev-maroc:
@taylorotwell This *IDE work-arounds* may saves you a lot of time (autocomplete, navigate-to-method, documentation, etc).

taylorotwell:
So do we need to fix this? The type-hint is wrong format?

arcanedev-maroc:
@taylorotwell PR #3988

PR #3989, opened 2016-10-07
Add a default UsersTableSeeder file

howlowck:
Since there is a default factory for the User model, and a commented line in the `DatabaseSeeder` calling a `UsersTableSeeder`, it's the next logical step to add a default `UsersTableSeeder` file.  
For most projects, people can just uncomment the line in `DatabaseSeeder` and run the `db:seed` command.


taylorotwell:
Will just let people add what they need.

PR #3992, opened 2016-10-10
Update app.js

shogun88888888:
Change comment to point to #app instead of body


zaknes:
I would rephrase this.

Notice how every line in a configuration comment is three less characters than its parent?

taylorotwell:
I'll fix it.

PR #3995, opened 2016-10-11
Add Yarn lockfile

michaeldyrynda:


mantrax314:
What's for?

michaeldyrynda:
https://twitter.com/taylorotwell/status/785885272890183681, although Taylor subsequently said that he'll add this to the Laravel installer, so may opt to close this PR anyway.

lucasmichot:
I suppose homestead would come with yarn already installed then ?

michaeldyrynda:
Forge was updated, it makes sense homestead would be as well. Software parity 👌

GrahamCampbell:
I don't think we need to commit this in the main repo, just like we haven't commited the composer.lock file.

GrahamCampbell:
However, both these files are not "ignored", because users of this repo should commit those files to their apps.

arcanedev-maroc:
Just discovered yarn today and this is my reaction:

Feels gooood

But i think the `yarn.lock` file must be added in `.gitignore` instead.

michaeldyrynda:
Ignoring it kinda defeats the purpose of using the lockfile, no?

I'm in agreement with @GrahamCampbell I think.

arcanedev-maroc:
I'm also not sure about it but yarn is way better than npm (speed, no time out, BS errors ...), especially with Windows OS.

jehadassaf:
The Yarn lockfile is no different from the Composer one in terms of intended use.
This means that no, this should not be part of a fresh Laravel install and no, you should not add it to your .gitignore, as it would defeat its purpose.

taylorotwell:
I already got it added in the "laravel new" build. Thanks though!

jcloutz:
Any news on having yarn installed with homestead?

taylorotwell:
I’ll need to rebuild the box. I should be able to do it this week.

Regards,
Taylor

PR #3998, opened 2016-10-13
Just a security default of not letting out info about

alnutile:
if the user exists or not `We can't find a user with that e-mail address` is helpful for hacking a site in that you will eventually find emails that exist. A more neutral response is suggested from what I have read.

### DIFF
--- a/resources/lang/en/passwords.php
+++ b/resources/lang/en/passwords.php
@@ -17,6 +17,6 @@ return [
     'reset' => 'Your password has been reset!',
     'sent' => 'We have e-mailed your password reset link!',
     'token' => 'This password reset token is invalid.',
-    'user' => "We can't find a user with that e-mail address.",
+    'user' => "If you are a user in the system a password reset email has been sent.",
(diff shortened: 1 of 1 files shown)


themsaid:
Isn't that easy to hack by trying to register with an email and see if I get a "Email already exists" error?

taylorotwell:
Yeah, this doesn't actually solve anything as you can just check on registration.

Doodlepants:
Hey all, I'm new and sorry if this is not the ideal place to say this...
but to solve this issue wouldn't it be good to :
1- When any user tries to register ( or looks for registered emails ).
2- Email with a registration link to the applying user, even if already
registered.
3- If user exist, change email to just a notification that someone tried to
register in their name.

alnutile:
It is nice you can easily change this in the config so I get there is no need for a fix. One thing I want to bring up here is Laravel in the enterprise where many applications might not have registration. So that is why in this case I was only thinking here.

PR #3999, opened 2016-10-13
Reduce code in RedirectIfAuthenticated middleware

assertchris:


GrahamCampbell:
👎  from me.

assertchris:
Really? The method is already using a global function (`redirect`), and depends on the Auth facade not being removed from the app service provider, whereas the global function uses the container to resolve the auth manager. This change actually make the codebase less error prone. Could you explain why 👎 ?

yoosuf:
@taylorotwell  i think this is a good fix

GrahamCampbell:
👎 from me. The facade is easier to use and supports mocking. The function does not.

assertchris:
screen shot 2016-10-16 at 5 11 40 am

GrahamCampbell:
Bit of a hack that that works.

assertchris:
Not to put too fine a point on it, but that "bit of a hack" is how the documentation recommends developers mock facades: https://laravel.com/docs/5.3/mocking#mocking-facades...

GrahamCampbell:
> Not to put too fine a point on it, but that "bit of a hack" is how the documentation recommends developers mock facades

That's not what I was talking about. I was referring to you using the global function after mocking the facade. Mocking facades is fine - it's one of the reasons I suggested keeping the facade.

assertchris:
Oh, right. I see what you're saying.

taylorotwell:
Since we're using facades most other places I think I will keep it as a facade for now just to be consistent with other files.

PR #4006, opened 2016-10-17
Cleaning code: align translations in array

corentinheadoo:


taylorotwell:
They don't need to be aligned.

PR #4010, opened 2016-10-18
Add views to config/auth.php to customize the auth views

milose:
This will provide an easy way to customize the Auth views for packages. I also submitted the PR to change laravel/framework to use views from this configuration laravel/framework#15987


taylorotwell:
I really would rather just have a convention here rather than add more configuration options.

PR #4013, opened 2016-10-19
ship laravel base with clear:all as example

BartHuis:
give a base functionality as example how to simply make a new command


BartHuis:
see https://github.com/laravel/laravel/pull/4014 , didn't know how to do it in once...

PR #4014, opened 2016-10-19
ship laravel base with clear:all as example

BartHuis:
give a base functionality as example how to simply make a new command

### DIFF
--- a/app/Console/Commands/Clearall.php
+++ b/app/Console/Commands/Clearall.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+
+class Clearall extends Command
+{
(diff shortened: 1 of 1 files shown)


BartHuis:
see https://github.com/laravel/laravel/pull/4013 , didn't know how to do it in once...

taylorotwell:
I think those are all sort of very different things and your reasons for clearing them would be different for each case, so would rather keep them separate.

BartHuis:
there are cases that your evoirement "acts strange" and you don't know if it's your code, or something in one or another cache, so you want to delete everything and look if it still happens, then you know for sure it's your code ;)

BartHuis:
(i think symfony's cache:clear does the same i thought to remember from my symfony time?)

PR #4015, opened 2016-10-19
Make default mail from name and address more configurable

andylaw:
Moved configuration of default mail from address and name into the .env file.
Makes handling situations where development email is sent via one route (possibly 'fake') but production email is sent via a separate, real route. Having the configuration in the .env file keeps this separation clean.


taylorotwell:
You can make these changes in your own application as needed.

andylaw:
Your call - obviously - but having it in the 'core' file and relying on individuals to change it piecemeal leaves that file susceptible to accidental overwriting when upgrading a laravel instance "in place" and scatters email configuration in two separate places. In its current coded-in-the-file format, it makes having those separate addresses for development and production that bit harder to maintain.

I'm struggling to see the objection to the pull request if I'm honest. Is there a downside that I'm missing?

scoutman57:
This is one of the few times that I disagree with @taylorotwell on something. I do see where email could be ENV settings. A prime example is when you use one email for prod and a second email for Dev or QA.. That way when you really need to send emails for something you can sent them like this  from:dev_server1@test.com or for prod@mydomain.com.. When you receive several emails from integration test it's hard to keep them separate when you can't customize the mail message for various reasons. 

From emails can vary from env to env. That is the whole purpose of the env file.

PR #4016, opened 2016-10-20
Removed unused namespaces

willitscale:
Namespaces not used or referenced. I feel like unknowing users may leave them in without understanding their purpose.

If they are there for educational purposes then my mistake, but https://laravel.com/docs/5.1/testing#disabling-middleware highlights their use with corresponding implementations.


GrahamCampbell:
thanks, but these are here as examples

PR #4018, opened 2016-10-20
app time zone to env

ahmedash95:
Hi , sometimes working on a projects will launch in another area of the world like UK and you working in another like  Middle East   .. so It will be good if the timezone is in env by default .


ahmedash95:
sorry , i found many PRs for this and it's closed because of many reasons so i would close it :).

PR #4020, opened 2016-10-24

qteck:


taylorotwell:
So, let's fix the documentation instead?

PR #4021, opened 2016-10-26
Add "public" as a supported filesystem driver in docblock

michaeldyrynda:


taylorotwell:
Isn't public just an instance of the local driver usually?

taylorotwell:
What I mean is it's not an actual driver, it uses the local driver by default.

michaeldyrynda:
Yes, but isn't / shouldn't this be showing the supported (default?) disks?

Perhaps the docblock ought to be updated entirely at that point, seeing as it says `filesystem disk` but the supported examples are drivers.

taylorotwell:
Eh, yeah, it's a little confusing. You can name disks anything you want so it's not like you are tied to only those names. IMO there really doesn't need to be any "supported" list in that comment at all.

taylorotwell:
The supported list would actually be down lower where the disks are defined with a driver.

GrahamCampbell:
👎 This is not a driver name. All names of instances are arbitrary.

taylorotwell:
I fixed this comment in the file to make more sense.

michaeldyrynda:
Much better 🎉

PR #4029, opened 2016-11-01
Update gulpfile.js

aleixfabra:
Error: Entry module not found: Error: Can't resolve 'buble'

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,6 +14,5 @@ require('laravel-elixir-vue-2');
  */
 
 elixir((mix) => {
-    mix.sass('app.scss')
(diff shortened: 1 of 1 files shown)


taylorotwell:
The answer is not to remove this from the gulpfile.

aleixfabra:
I know, but I could find where to create an issue. So I decided to show it in that way.

JeffreyWay:
Submit your issue on the laravel/elixir repo instead.

aleixfabra:
Ok!

abachmann:
I had the same issue. You just need to install the latest node.js version.

PR #4030, opened 2016-11-02
dynamically setting the public directory

Ziadjb:


taylorotwell:
Would rather just stick with convention.

PR #4032, opened 2016-11-03
Alias axios to $http in the Vue object

scrubmx:
This change allows to keep using this.$http as before.


taylorotwell:
Eh, you can do this if you want. I don't think I will do this by default.

PR #4034, opened 2016-11-07
allow .env to override the timezone setting.

dierkerd:
allow a site/developer to provide a timezone in the .env.


scrubmx:
This has been proposed and turn down many times before: https://github.com/laravel/laravel/pull/4018, https://github.com/laravel/laravel/pull/3937, https://github.com/laravel/laravel/pull/3882, https://github.com/laravel/laravel/pull/3792, https://github.com/laravel/laravel/pull/3713

dierkerd:
I'm not really seeing a reason not to provide it. None of the ticket you point to actually provide a valid reason for not making this change, It's backward compatible, it's simple, it allows folks to be able to switch away from UTC on an app by app basis without having to hand edit the app.php.

scoutman57:
Setting for "app by app" basis should not be in .env.. The whole point of
the .env is for "environment" specific setting that could change for the
single app. Any setting that are constant for the  application no matter
the env should be set in the appropriate app config files.

dierkerd:
> Setting for "app by app" basis should not be in .env.. The whole point of
> the .env is for "environment" specific setting that could change for the
> single app. Any setting that are constant for the  application no matter
> the env should be set in the appropriate app config files.

A primary and valid reason for allow this change to config/app.php is that it directly effects log file rotation.
the config/app.php already has env values for changing out log file setup

> 'log' => env('APP_LOG', 'single'),
> 'log_level' => env('APP_LOG_LEVEL', 'debug'),

yet if you switch APP_LOG to daily in your .env you wind up having UTC timestamps on the log file which can prove to be naturally confusing.

If you're going to support .env values for these settings why not timezone?

scoutman57:
The reason the env exist for log rotation is because you don't normally rotate logs on the the same schedule from dev -> prod. 

I do see how someone could be a little confused on the UTC timestamps for the logs. This maybe should be cleared up with a little documentation.

dierkerd:
> The reason the env exist for log rotation is because you don't normally 
> rotate logs on the the same schedule from dev -> prod.

Same answer could/would apply to timezone... You might not be using the same timezone between dev and production. Heck you might not be using the same timezone on different instances of the same "app" installed to different environments/instances

I reiterate: It's a backward compatible, one line change, change.

PR #4035, opened 2016-11-08
allow PUSHER_CLUSTER option from env

wuwx:
if not set custer option, pusher debug console not show sent message


taylorotwell:
But won't it throw an error if the cluster is null?

taylorotwell:
I can't recreate this.

screen shot 2016-11-08 at 11 54 25 am

PR #4038, opened 2016-11-08
add arquivos compilados (css/js)

thompsonemerson:
Files compiled as `css` and `javascript` cause a lot of conflict when working as a team.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@
 Homestead.json
 Homestead.yaml
 .env
+/public/css/*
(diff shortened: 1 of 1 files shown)


taylorotwell:
You can add this to your .gitignore if you want. We don't have an opinion on this matter.

thompsonemerson:
Oh really @taylorotwell? But this is a real problem for us front end.
Would not it be cool to prevent this from happening? O.o

PR #4042, opened 2016-11-13
Add database Prefix in .env

AhsanAbrar:
i always do this manually so i think it should be come with core.

it's really helpful for me if you approve this little contribution.

i hope it'll be helpful other's too.

Regards


marktopper:
What about line [52](https://github.com/AhsanAbrar/laravel/blob/6fc79964740daa3a9ab42dd12da59c748c3d9970/config/database.php#L52) and [77](https://github.com/AhsanAbrar/laravel/blob/6fc79964740daa3a9ab42dd12da59c748c3d9970/config/database.php#L77).

AhsanAbrar:
ohh i'm really sorry, i often used mysql that's why forgot.

let me add these too.

GrahamCampbell:
👎 Using prefixes is very uncommon.

yoosuf:
👎🏿 using prefix is a personalised option, if you want do it within your project

overtrue:
👎 +1

pheliperocha:
.env is for settings that will differ according to the environment... It does not make sense to have different prefixes for different environments..

Mechtecs:
:-1: -1
It doesent make sense and is as stated above very uncommon.

scoutman57:
The only argument I have for putting the prefix in the .env is that we
currently have the "database name" in the env as well. The env can change
per environment just as the db name changes. Normally the prefix is still
considered part of the database name when you are referencing the db in
general.

Again that is really the only argument I can make. I can't say it's
uncommon as I have worked with many companies and have seen it go both ways
on if they use prefixes are not.

wangroute:
:-1:

thosewords:
:-1:

PR #4048, opened 2016-11-15
PHPDoc text

fredylg:
just some text


GrahamCampbell:
A function call has no return type or params.

PR #4050, opened 2016-11-15
Change the string column length to 191

nmfzone:
Because the database collation change from utf8_unicode_ci to utf8mb4_unicode_ci, it will cause [this](http://stackoverflow.com/questions/23786359/laravel-migration-unique-key-is-too-long-even-if-specified) issue.

"The reasoning for this length change is that regular utf8 charset and collation utilize only 3 bytes per character and utf8mb4 utilizes 4 bytes per character to support a larger range of characters." - [source](http://aoverton.com/web-development/using-mysql-utf8mb4-with-laravel-5/)

### DIFF
--- a/database/migrations/2014_10_12_000000_create_users_table.php
+++ b/database/migrations/2014_10_12_000000_create_users_table.php
@@ -16,7 +16,7 @@ class CreateUsersTable extends Migration
         Schema::create('users', function (Blueprint $table) {
             $table->increments('id');
             $table->string('name');
-            $table->string('email')->unique();
+            $table->string('email', 191)->unique();
             $table->string('password');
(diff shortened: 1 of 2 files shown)


Dominoli:
I'm not familiar with the base code of laravel, but you should find the standard length setting for possible index keys (eg string()) and change that to 191.  Otherwise you'll have to set the length in all migrations.

taylorotwell:
yeah if anything we would to change this at the global default level. however, i believe @fideloper told me recent versions of MySQL dont have this length limitation.

nmfzone:
Whoops, sorry. I haven't thought about it before.
Ok, I finally decided to upgrade mysql-server to the newest version.

PR #4051, opened 2016-11-16
Support predis v1.1.1

lokielse:
fix: `AUTH failed: ERR Client sent AUTH, but no password is set [tcp://127.0.0.1:6379]`

According to predix release log: https://github.com/nrk/predis/releases/tag/v1.1.1

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -111,9 +111,11 @@ return [
 
         'default' => [
             'host' => env('REDIS_HOST', '127.0.0.1'),
-            'password' => env('REDIS_PASSWORD', null),
             'port' => env('REDIS_PORT', 6379),
             'database' => 0,
+            'parameters' => [
+                'password' => env('REDIS_PASSWORD', '')
+            ]
         ],
 
     ],


GrahamCampbell:
Does this break support for other versions?

lokielse:
@GrahamCampbell The other versions of predis require a `password` field, set both `password` and `parameters.password`  works！ but it is a little bit ugly...

```php
'default' => [
    'host'       => env('REDIS_HOST', 'localhost'),
    'password'   => env('REDIS_PASSWORD', null),
    'port'       => env('REDIS_PORT', 6379),
    'database'   => 0,
    'parameters' => [
        'password' => env('REDIS_PASSWORD', null)
    ]
]
```

fernandobandeira:
Weird, this version was released month's ago we should've known about this earlier IMHO.

Also is there a relation with laravel/framework#15914 ? He's using predis 1.1.1 and he stated that 

> Method connection return null.

Perhaps it's related...

taylorotwell:
Meh, this PR is totally broken according to my testing on a fresh Forge server with Predis 1.1.1.

PR #4057, opened 2016-11-25
Extracted app name to .env file

Modelizer:
It will make easier for user to get started quickly and centralizing app name if they want to change. This will be handy if they want to use it in layouts file, email templates etc.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,8 +1,9 @@
+APP_NAME=Laravel
+APP_URL=http://localhost
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
 APP_LOG_LEVEL=debug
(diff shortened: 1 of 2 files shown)


taylorotwell:
This isn't something that changes per environment.

GrahamCampbell:
Maybe our documentation isn't clear enough that .env files are only for environment things such as secrets. It's not forbidden to edit the config files, in fact, encouraged, where appropriate.

Modelizer:
When the whole file is made up with vue It's difficult to keep sync with config data. And putting it into `.env` make it easy. I did face the problem.

Anyways thanks for the taunt @GrahamCampbell 👍  I do agree it was my bad.

PR #4059, opened 2016-11-28
[5.4] Update mounted instead of ready

insoutt:


GrahamCampbell:
I think ready is better. No need to match it to the name in the code.

GrahamCampbell:
:-1:

Mohamed-Elwan:
I think console logging is targeted to developers and reading the same name in the code would be some how good. It's not a big deal but it's not bad at all.

insoutt:
This is a small change. Although I think that mounted is more specific than ready

Mechtecs:
:+1: for mounted

PR #4060, opened 2016-12-01
added mail_from attributes to the default env file

RonMelkhior:
Since every app that uses email will eventually add this variables themselves, it should be added by default.


erikwestlund:
These aren't environment specific.

Mail from name and mail from address will be same in testing as in production, in almost all cases?

RonMelkhior:
These environment variables were added in pull request #4037

taylorotwell:
Please just add these if you need them.

PR #4061, opened 2016-12-02
Middleware implementation process

shijunti19:
laravel5.3
I will not choose a category to ask questions。

The normal process should be：
Routing -> middleware -> __ construct -> method！
Why your process is:
Routing->__ construct -> middleware ->  method！

I need to call the public method after middleware , before __ construct！
`$this->wx_api();
			$this->agent();`


fernandobandeira:
Check #15072 and close this...

PR #4062, opened 2016-12-02
Removed unused `use` statements

pjona:


taylorotwell:
They are there because you will probably need them.

pjona:
As you want, but I assume that if user need them he will add this lines. Default this lines are not used, so why keep them in repository? It makes only mess.

PR #4063, opened 2016-12-02
Removing @return void annotation on constructor

denjaland:
Scrutinizer (and probably other code quality verifiers) mark these as a minor issue.
As a constructor never returns a value, I thought it'd be nice to have these annotations removed indeed.

### DIFF
--- a/app/Http/Controllers/Auth/ForgotPasswordController.php
+++ b/app/Http/Controllers/Auth/ForgotPasswordController.php
@@ -22,8 +22,6 @@ class ForgotPasswordController extends Controller
 
     /**
      * Create a new controller instance.
-     *
-     * @return void
      */
     public function __construct()
     {
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -29,8 +29,6 @@ class LoginController extends Controller
 
     /**
      * Create a new controller instance.
-     *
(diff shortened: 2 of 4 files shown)


carkii:
I believe that, this return help the user how to use the function perfectly.

For me, it very useful and no matters what Scrutinizer believes:)

GrahamCampbell:
Yes, the void annotations are intended.

denjaland:
Ok fair enough - I would never put it on a constructor myself to be honest; that is just part of php language; ANY developer should know a constructor doesn't return anything...  

For your own methods, I agree it gives better hints for the developers using it.

denjaland:
I was just thinking about this again, and how I would never do it, but I was in fact wrong.
I believe you should update the annotations to return SELF actually.

Because the result of any constructor is that it creates an object of the class it belongs to, and returns it so it is placed inside the variable.

Looking at https://phpdoc.org/docs/latest/references/phpdoc/tags/return.html, it actually confirms that maybe you need to update it for your constructors to make it more correct - if you insist on keeping the annotation there :-)

GrahamCampbell:
phpdoc.org is wrong i'm afraid. They are just functions like any other, and they return whatever you actually return in them. (feel free to try this out)

Note however, that this did change in php 7.1, though, phpdoc.org defintely didn't use that definition because of that, it just got it wrong.

PR #4064, opened 2016-12-03
Correction values for "exists"

andrey-helldar:


taylorotwell:
That does not always make sense. You might use `exists` for checking for example if a country is a valid country based on a database, and in that case it doesn't make sense to say "already exists".

denjaland:
Hmm - maybe that use case would call for another validator... what about "allowed" rather that "exists".  I agree with @andrey-helldar that a validator named "exists" should validate whether something "exists" :-)

scoutman57:
Just and FYI "valid" and "exist" are not the same and should not be
considered the same . To use Taylor's example.. I might have a list of
countries in the DB and i would not use "exists to check valid and the
reason for that is this. If I have a product that can only ship to certain
countries but not others. You would do some thing like check the country is
"valid" to receive that product... There are many examples and this is just
one.

Considering "valid" and "exist" to be the same should be reconsidered..

Just my 2 sense.

scoutman57:
Not to mention that "is invalid" is a very bad error text then something
exist.. ."is invalid" why? In valid because it does not conform, invalided
because it exist? Now I have to go dig in the code.

https://www.nngroup.com/articles/error-message-guidelines/

if you want to keep it to "is invalid" please make the error message "The
selected :attribute is invalid because it already exists."

GrahamCampbell:
Well, you're free to change this in your app to whatever best fits your use case, that's why this is not hard coded in the framework.

teynon:
> The selected country is invalid because it already exists.
This would be an incorrect statement. I believe Taylor is correct here. This field can be used in a large variety of ways, but I see it mostly as a "this field must already exist". Therefore, in the majority of cases "because it already exists" would be incorrect. (Most would be something along the lines of "The selected :attribute does not exist.") While error messages should be precise, none of the proposed messages are correctly defining the problem except to say that it is "invalid". As Graham just stated, this is something each user should define if they use it in a different way.

gpressutto5:
What about this?

    'valid' /*or 'invalid'*/ => 'The selected :attribute is invalid.',
    'exists'                 => 'The selected :attribute already exists.',

fernandobandeira:
First off, this is not really that big of an issue,
Second you guys are kinda confusing ``` exists ``` with ``` unique ``` here.

> The field under validation must exist on a given database table.

So why would you throw this message when the field doesn't exist:
```
'The selected :attribute already exists.',
```
The message is displayed because the :atribute **doesn't exist** on the database so this message is just horrible and doesn't make sense at all.

PR #4065, opened 2016-12-04
delete useless comment rows

hanahiroAze:
I found needless comment out rows, and delete them.


scrwdnet:
I don't believe these are 'useless', I think they are in place to indicate that you can modify with your own code there.

GrahamCampbell:
Yeh. :)

PR #4070, opened 2016-12-12
Fixed csrf token broken js

brandy:
Every IDE I use kept saying this was broken, so I fixed it


WayneHarris:
@brandy Is your IDE set to ECMAScript 6. If not the IDE is likely to complain about the syntax.

brandy:
Ahh, I changed it to ECMAScript 6 and its fine now.

PR #4072, opened 2016-12-15
Add cluster and encrypted as option for Pusher

KevinHoughton:


taylorotwell:
I'm not sure we need to include these in the default config file as long as we document them.

PR #4073, opened 2016-12-16
Fix return types in PHPDoc

riesjart:
Both methods not always return instance of \Illuminate\Http\Response


diogoazevedos:
@taylorotwell Yeah, but when we strongly type the return, this will throw an error in API usage, because `Illuminate\Http\JsonResponse` extends `Symfony\Component\HttpFoundation\JsonResponse` and not `Illuminate\Http\Response`.

PR #4074, opened 2016-12-17
[5.3] Sort Composer packages by default

driesvints:
It's much easier to search for packages in the `composer.json` file when they're sorted by default. Having this on a fresh Laravel install from the beginning would be nice :)


GrahamCampbell:
Just saw this on the php-cs-fixer repo too. :)

PR #4077, opened 2016-12-21
Fixing problems with MariaDB due to mb4 charset

JuanDMeGon:
The utf8mb4 charset and utf8mb4_unicode_ci collation, are causing problems with MariaDB 10.x due to long key length.

**How to verify:**
With Laravel 5.4 (develop branch), try to execute the default users table migration, using MariaDB as database Engine.

It will produce a "Syntax error or access violation: 1071 Specified key was too long; max key length is 767 bytes (SQL: alter table `users` add unique `users_email_unique`(`email`))".

image

Removing the mb4 charset and collation, keeping the original utf8, solves the problem.


GrahamCampbell:
This keeps on going back and forward. :/

taylorotwell:
This isn't the solution. The solution is make the default string length easily configurable.

JuanDMeGon:
I don't get this solution. How dows make it easy configurable solve the problem?

If I understand, make it easy configurable means that for MariaDB, the default value for string must be 191m (767/4) while for MySQL can still being 255. At the end is it a MariaDB problem, and not a Laravel one? I'm afraid I did not get the idea.

I would like to help on it if you point me the right way to go.

👍

Dominoli:
If you can configure the key length, you can easily change it when using MariaDB, in your example. So it does solve the problem.
Considering MySQL, only new versions support the long key length.  Also on those systems it would be useful to change key length globally.

GrahamCampbell:
https://github.com/laravel/laravel/pull/3898

JuanDMeGon:
Thanks for the clarification and the context guys. I understand now why is it required in this way. 
I'll take a look on it. 
:D

PR #4081, opened 2016-12-27
Remove optimize command

diogoazevedos:


GrahamCampbell:
The optimize command is not deprecated, we actually just went straight ahead and removed the class compiled file generation from it. What remains of the otomize command is not deprecated. That said, I think we should actually deprecate the whole command.

GrahamCampbell:
I'm guessing the post install scripts already "clear compiled", so there's no point in running the optimize command, but not because it's "deprecated".

diogoazevedos:
@GrahamCampbell Thanks! I will recommit with an appropriated message.

taylorotwell:
Not sure this needs to be removed yet.

PR #4084, opened 2016-12-29
[5.4] Remove the "mail" driver as recommended, and add "array"

GrahamCampbell:


daftspunk:
Add some more detail to this. SwiftMailer no longer includes the `MailTransport` driver that was responsible for using php's `mail()` function.

https://github.com/swiftmailer/swiftmailer/issues/866

As a result, Laravel now treats the `mail` driver identical to the `sendmail` driver.

```
Swift_MailTransport -> Swift_SendmailTransport
```

PR #4085, opened 2016-12-29
Update server.php

Aamin1234:
change in comment


GrahamCampbell:
Thanks, but what we have is already ok. :)

PR #4086, opened 2016-12-30
The mb4 character set create a prefix limitation issue

rsm23:
UTF-16 can use up to 4 bytes per character which means that in an InnoDB table, you can’t have any keys longer than 191 characters.
With the ùtf8mb4`charset migrations like the Àuth`migrations will fail with an error like `Specified key was too long; max key length is 767 bytes`, because the `UNIQUE INDEX` on the **email** field requires at least 1020 bytes (255 * 4).

Unfortunately there is no real solution to this. the only options are to either reduce the size of the column, or use a different character set (like UTF-8).

## A preview of the error message

screen shot 2016-12-30 at 2 25 57 am


baskan:
see #4077

rsm23:
@baskan but still no solution for that

PR #4087, opened 2016-12-30
Create yarn.lock file

rsm23:
Adding the auto generated yarn.lock file


GrahamCampbell:
No need to include this, as with a composer.lock. You can include this in your own app, but not in the laravel/laravel base.

PR #4088, opened 2016-12-30
Fixed typo

tillkruss:
`DatabaseTokenRepository` actually uses this as seconds everywhere, not minutes.

### DIFF
--- a/config/auth.php
+++ b/config/auth.php
@@ -85,7 +85,7 @@ return [
     | than one user table or model in the application and you want to have
     | separate password reset settings based on the specific user types.
     |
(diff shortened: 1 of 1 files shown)


DojoGeekRA:
@tillkruss The constructor of the `DatabaseTokenRepository` is actually converting the value into [minutes](https://github.com/laravel/framework/blob/5.3/src/Illuminate/Auth/Passwords/DatabaseTokenRepository.php#L53).

tillkruss:
@DojoGeekRA: Thanks! I don't know why I missed that...

PR #4089, opened 2016-12-30
Adding the the Visual Studio Editor settings folder to .gitignore

rsm23:
It should be added because the **VS Code** editor is getting more and more popular


GrahamCampbell:
This should be in your global .gitignore file, as should `.idea` tbh.

PR #4090, opened 2016-12-31
Adding the Vue-Resouce Dependencies

rsm23:


zaknes:
Vue-Resource is being retired. We're using Axios now.

PR #4094, opened 2017-01-04
Fixed .gitignore

xchopin:
## Removed : 
`.idea` is a JetBrains file, it has to be included in the global .gitignore, not in the local one
(cf :  http://www.craftitonline.com/2013/10/ignore-symfony2-phpstorm-files-like-a-pro-with-global-ignoring)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,9 @@
+Homestead.json
+Homestead.yaml
+.env
+composer.lock
+composer.phar
 /node_modules
 /public/storage
 /storage/*.key
 /vendor
-/.idea
-Homestead.json
-Homestead.yaml
-.env
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 Homestead.json
 Homestead.yaml
 .env
-composer.lock
 /node_modules
 /public/storage
 /storage/*.key
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@ Homestead.json
 Homestead.yaml
 .env
 composer.lock
-composer.phar
 /node_modules
 /public/storage
 /storage/*.key


kevinvdburgt:
I agree with the `.idea` removal that it should be placed in the global `.gitignore`.

However, i disagree with the `composer.lock` file because it *is* a part of the project, when running `composer install` it tries to read the `composer.lock` file (it contain dependency version numbers etc..) so everyone has the same version of everything.

The same for `composer.phar`, it doesnt exists in the repository by default. And you should (recommend) to place the composer executable in your `$PATH`.

xchopin:
I understand your point of view thanks for sharing it with me. 
I updated the pull request

taylorotwell:
We're not changing this file anymore.

kevinvdburgt:
I'm still curious why the `.idea` is in the .gitignore file @taylorotwell  :) ?

JellyBool:
@kevinvdburgt check this : https://github.com/laravel/laravel/pull/3877

Maybe Taylor is tired of the pr about `.idea`.

kevinvdburgt:
Ah, I see, if that is the case I can understand that Taylor is tired about those PR's.

And it is just a single line I can remove from my project if I don't want it :)

RomainLanz:
> Maybe Taylor is tired of the pr about .idea.

Yes, but now we get other PR to remove it. 💀

JellyBool:
@kevinvdburgt I also think it's a personal thing about `.idea`,I use Phpstorm and sublime both.

@RomainLanz Saaad....

GrahamCampbell:
> I'm still curious why the .idea is in the .gitignore file @taylorotwell :) ?

Because we get spammed with PRs to add it.

> I also think it's a personal thing about .idea,I use Phpstorm and sublime both.

No IDE should have it's config put in our gitignore, since they belong in global ignore files, so it's nothing personal in that way.

PR #4096, opened 2017-01-06
Add properties docblocks for \App\User

lucasmichot:
Add properties docblocks for `\App\User`
This allows autocompletion with most IDE, out of the box


taylorotwell:
No plans on adding this. I don't like adding noise for IDEs.

PR #4097, opened 2017-01-09
update return type in docblock of render method

Gavrisimo:
That `parent` is `Illuminate/Foundation/Exceptions/Handler.php` and its `render` method is returning `\Symfony\Component\HttpFoundation\Response`, so just making it consistent. :)


taylorotwell:
I would generally rather advertise you can use the Illuminate Response here since it has more helper methods. Kind of a documentation / awareness thing.

PR #4100, opened 2017-01-17
Update s3 filesystem to use environment variables

ryzr:
This is a manual change I've had to make on a bunch of projects I work on.

Basically - I don't want AWS credentials committed to the repo, rather, I want to leave this config up to my deploy recipes which generate .env files for me. I believe this is the purpose of environment variables, but let me know if I'm misunderstanding.

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -56,10 +56,10 @@ return [
 
         's3' => [
             'driver' => 's3',
-            'key' => 'your-key',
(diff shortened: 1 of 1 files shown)


bbashy:
Not sure if it's best to set a default for those `env()` calls?

```
'key' => env('AWS_ACCESS_KEY_ID', ''),
```

taylorotwell:
I'll make this change on develop branch.

PR #4101, opened 2017-01-17
Fix voor IE & ME Auth

robindirksen1:
Do not use an underscore in the cookie name because Internet Explorer and Microsoft Edge fails to login.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -120,9 +120,12 @@ return [
     | instance by ID. The name specified here will get used every time a
     | new session cookie is created by the framework for every driver.
     |
+    | Do not use an underscore in the cookie name because Internet Explorer
+    | and Microsoft Edge fails to login.
+    |
     */
 
-    'cookie' => 'laravel_session',
+    'cookie' => 'laravelsession',
 
     /*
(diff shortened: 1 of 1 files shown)


RonMelkhior:
IIRC the issue is with domains/subdomains with underscore in them. Not cookies with underscores. Otherwise it'd be a much bigger issue.

robindirksen1:
@RonMelkhior I used this solution in a domain without a underscore (mysite.online is the domain), maybe it's because the .online TLD?

RonMelkhior:
@robindirksen1 I kind of doubt it.

It seems some people experience the same issue as you from a quick search on Google, but also some of them have underscores in their domain/sub-domain name.

bbashy:
If this was true, Google Analytics wouldn't work. They use `_ga` for their cookie name.

robindirksen1:
@bbashy yes, that's true but the only think that I changed in my config was the `_` in the config file. Maybe if you put the `_` in front of the cookie name it works?

bbashy:
@robindirksen1 Well GitHub's is `user_session`. Does this site not work for you on IE/Edge?

PR #4104, opened 2017-01-18
Improve .gitignore

anchiceratops:
Ignored
- .DS_Store
- Thumbs.db
- .log

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@
 Homestead.json
 Homestead.yaml
 .env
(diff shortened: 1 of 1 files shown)


kevinvdburgt:
No, this is not an improvement. You have to place those ignore rules in your global git config.

anchiceratops:
@kevinvdburgt Whoaa, interesting. Not necessarily agree with what you have to say. But maybe you're right about `this is not an improvement`.

PR #4105, opened 2017-01-19
Setting the maximum count of log files

andrey-helldar:
When using the daily log mode, Laravel will only retain five days of log files by default. If you want to adjust the number of retained files, you may add a log_max_files configuration value to your  app configuration file.

https://laravel.com/docs/5.3/errors#configuration

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -121,6 +121,8 @@ return [
     */
 
     'log' => env('APP_LOG', 'single'),
+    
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't know that we need to bloat the configuration file with this by default. It can be added when needed by the developer.

andrey-helldar:
Logically)

PR #4108, opened 2017-01-21
Fixed a typo in a block comment queue.php

dbudwin:
Normally, words that start with vowels (like "unified") are preceded by "an" instead of "a."  However, "unified" is an exception to the rule.

Source: https://www.a-or-an.com/a_an/unified

### DIFF
--- a/config/queue.php
+++ b/config/queue.php
@@ -7,7 +7,7 @@ return [
     | Default Queue Driver
     |--------------------------------------------------------------------------
     |
-    | The Laravel queue API supports a variety of back-ends via an unified
+    | The Laravel queue API supports a variety of back-ends via a unified
     | API, giving you convenient access to each back-end using the same
     | syntax for each one. Here you may set the default queue driver.
     |


RomainLanz:
You break the minus 3 characters rules :(

dbudwin:
Can you explain?  I didn't see a reference to what you're talking about in the [guidelines](https://laravel.com/docs/5.3/contributions).

Mozartted:
In the Laravel Framework, the lines of the comments Taylor creates are always 3 or more characters lesser than the comment lines before them for example 

```php
/*
| Run The Application
|
| Once we have the application, we can handle the incoming request
| through the kernel, and send the associated response back to
| the client's browser allowing them to enjoy the creative
| and wonderful application we have created for them.
|
*/
````

But from the fourth line you could see they are not exactly necessary... But following that pattern would be cool....

RomainLanz:
@dbudwin No I'm just joking!

@taylorotwell Like to have minus 3 characters at every line in a comment. That's only that! 😛

dbudwin:
Haha, I do see that pattern a fair bit in the comments.  Although if you include punctuation, sometimes it's only -2 characters.  Is this a hard-and-fast rule?  Do I need to reword it more so that this PR can get merged?

taylorotwell:
Haha, contributors don't have to worry about that.

PR #4109, opened 2017-01-21
[5.4] Add font path to fix glyphicons font-face.

yajra:
Without this variable, the compiled path output is wrong:

`/css//fonts/glyphicons-halflings-regular.woff2?448c34a56d699c29117adc64c43affeb`

And the correct path should be:

`/fonts/glyphicons-halflings-regular.woff2?448c34a56d699c29117adc64c43affeb`

### DIFF
--- a/resources/assets/sass/_variables.scss
+++ b/resources/assets/sass/_variables.scss
@@ -35,3 +35,6 @@ $input-color-placeholder: lighten($text-color, 30%);
 
 // Panels
 $panel-default-heading-bg: #fff;
+
+// Fonts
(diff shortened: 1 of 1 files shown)


taylorotwell:
@JeffreyWay is this correct?

JeffreyWay:
I think the fonts do compile to `/fonts`. So we can either set it as a variable like this PR, or update Laravel Mix to export fonts to the `/css` directory instead. Preference?

taylorotwell:
Hmm, maybe just have them export to css. Thanks.

PR #4111, opened 2017-01-23
Use const and destructuring assignment

diogoazevedos:
Is a good practice use `const` with `require` and more elegant use destructuring assignment.


RomainLanz:
We should also remove semicolon, but it's another story

taylorotwell:
It doesn't matter.

taylorotwell:
Jeffrey says its fine so I will merge I guess.

JeffreyWay:
const sucks.

PR #4112, opened 2017-01-24
chore(mix): extract by default vendor js

RomainLanz:
Hey!

By default the `app.js` file is using `jQuery` and `VueJs`. Because we provide those two dependencies and people will probably use them it may be better to extract them by default into Laravel Mix.

This will avoid bad assets management for newcomers.

### DIFF
--- a/webpack.mix.js
+++ b/webpack.mix.js
@@ -12,4 +12,5 @@ const { mix } = require('laravel-mix');
  */
 
 mix.js('resources/assets/js/app.js', 'public/js')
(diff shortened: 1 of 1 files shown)


taylorotwell:
People can add this if they want.

RomainLanz:
Of course they can add it if they want, same for the VueJs boilerplate or everything. I would prefer encourage good practice by default instead of letting the developer change every new project this file by adding this line.

JeffreyWay:
I'd prefer that we didn't add this, personally.

PR #4116, opened 2017-01-24
Add dev dependencies and rollback scripts

diogoazevedos:


taylorotwell:
Needs full description.

taylorotwell:
Please re-open with a full description of what this fixes, how, etc. Thanks.

JeffreyWay:
Laravel Mix already includes all of these script additions except Vue. They're unnecessary.

https://github.com/JeffreyWay/laravel-mix/blob/master/package.json

diogoazevedos:
@JeffreyWay Yes, but when using Yarn they don't create the .bin folder like NPM does.

RomainLanz:
@diogoazevedos So this is not the problem of Laravel if `yarn` doesn't install correctly the dependencie.

diogoazevedos:
@RomainLanz Yarn install all dependencies correctly, it just don't create the `.bin` folder like NPM does, and this make the things blow up if the command you are using isn't in dependencies (or global).

JeffreyWay:
We have a different fix incoming that doesn't require laravel/laravel to re-declare those dependencies that are requested by Mix.

PR #4118, opened 2017-01-25
Add mix-manifest.json to gitignore

ahjorth:
This file will be immediately overwritten upon a build if versioning is active. It seems like a pointless thing to version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@
 /.idea
 Homestead.json
 Homestead.yaml
+mix-manifest.json
 .env


GrahamCampbell:
Not if you commit your production assets rather than building them at deploy.

ahjorth:
Hmm I guess you're right. For some reason I thought `public/js` and `public/css` were ignored as well. I guess adding it doesn't make sense if these folders are excluded from gitignore by default too.

adiachenko:
@GrahamCampbell That's something I don't quite understand for two following reasons.

- How would you manage git merges with a bunch of compiled gibberish (i.e. `public/js/app.js`) in your version control?
- Won't your `.git/` folder be growing unreasonably large with time?

I am asking because I wanted to write a pull request for ignoring compiled public assets in Laravel default installation and then saw this pull request. 

Not saying that I'm in the right, I may be missing some popular workflow. It's just kinda disheartening with Laravel default installation being so slick to remember edit `.gitignore` each time I initialize new project. I would appreciate if you explain the reasoning behind this decision.

PR #4119, opened 2017-01-25
Fixes FQN

legshooter:


brunogaspar:
What it is now is the correct FQCN.

DojoGeekRA:
@legshooter i think that you're missing the point.

It is intentional to use `\App\Http\Middleware\TrimStrings::class` as a wrapper. It extends `Illuminate\Foundation\Http\Middleware\TrimStrings` to provide more functionality, in this case, not to trim the `password` and `password_confirmation ` fields.

legshooter:
Yeah, my bad.

PR #4120, opened 2017-01-26
removed the default database character set issue on 5.4

ayimdomnic:


RonMelkhior:
IIRC you only need this if you run MySQL <5.7.7

PR #4124, opened 2017-01-28
Remove redundant check from bootstrap/autoload.php

vantoozz:
Redundant check removed since compiled class file generation was removed
https://github.com/laravel/framework/commit/09964cc8c04674ec710af02794f774308a5c92ca#diff-7b18a52eceff5eb716c1de268e98d55dL870


taylorotwell:
This entire file could probably be removed basically.

PR #4126, opened 2017-01-30
Append '@' to let composer resolve PHP executable

timgws:
To execute PHP scripts, you can use `@php` which will automatically resolve to whatever
php process is currently being used.

See https://getcomposer.org/doc/articles/scripts.md#executing-php-scripts
for more information.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,18 +29,18 @@
     },
     "scripts": {
         "post-root-package-install": [
-            "php -r \"file_exists('.env') || copy('.env.example', '.env');\""
+            "@php -r \"file_exists('.env') || copy('.env.example', '.env');\""
         ],
         "post-create-project-cmd": [
-            "php artisan key:generate"
+            "@php artisan key:generate"
         ],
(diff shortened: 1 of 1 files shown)


lioannou:
@taylorotwell - this function `@php` was only added to composer 1.3.* which was only released 4 weeks ago: https://github.com/composer/composer/blob/master/CHANGELOG.md

Given the script wasnt broken in the first place(?) - is it worth the possible hassle from people running into issues?

Just a thought...

timgws:
@lioannou it was (and always was) broken for me on Windows.

Also, if I used a different PHP that was not on my path, the version of PHP that was used to run composer would not be the same version of PHP used to execute the PHP command.

See also: https://github.com/Microsoft/BashOnWindows/issues/1643

lioannou:
Ok fair enough... might be worth doing a PR to the docs and listing composer >=1.3 as a requirement then

timgws:
:+1:

I can get to it when I am back at a computer a little later :)

PR #4128, opened 2017-01-30
Make Axios automatically send `X-CSRF-TOKEN` in the header

monomichael:
Axios doesn't automatically send the `X-CSRF-TOKEN` that [the passport token guard](https://github.com/laravel/passport/blob/f2a8af3a48adac02288779f59de95f46fef69b7a/src/Guards/TokenGuard.php#L204) is looking for.

See https://github.com/laravel/passport/issues/256 and https://github.com/laravel/docs/pull/3045

### DIFF
--- a/resources/assets/js/bootstrap.js
+++ b/resources/assets/js/bootstrap.js
@@ -28,6 +28,7 @@ window.Vue = require('vue');
 window.axios = require('axios');
 
 window.axios.defaults.headers.common = {
+    'X-CSRF-TOKEN': window.Laravel.csrfToken,
     'X-Requested-With': 'XMLHttpRequest'
 };
 


ianrodriguesbr:
@monomichael  #4127

taylorotwell:
Passport does automatically send a header it just sends X-XSRF-TOKEN. We need to update passport to check for both headers I think?

cheykeodina:
I end up with this kink of error why? 
TypeError: window.Laravel is undefined
	'X-CSRF-TOKEN': window.Laravel.csrfToken,

rickbolton:
With axios you don't require to set this so I would remove this. 

If you want to fix the window.Laravel is undefined then add it before your JS file within your Laravel blade template e.g.

```
<script>
        window.Laravel = <?php echo json_encode([
            'csrfToken' => csrf_token(),
        ]); ?>;
</script>
```

conterno:
OK so you have to publish the Laravel Auth views in order for the error to go away.

cheykeodina:
@rickbolton  I follow the document of Laravel https://laravel.com/docs/5.4/passport#consuming-your-api-with-javascript and watch video lesson in Laracast as well. However, it still show error:Unauthorized. Is there any problem with the document. I use Apache2, PHP version 5.6, MySQL5.5, is there any problem with this environment.

Paltond:
I'm experiencing the same problems as @cheykeodina

MarGul:
@cheykeodina Not sure if this is your issue but after you have authenticated you need to make a "dummy" GET request to get the laravel_token. This is if you have an SPA that makes a POST request to authenticate.

Better yet is to extend CreateFreshApiToken middleware and make it send back a fresh token on both POST and GET. See here: https://github.com/MarGul/Bidder/blob/master/app/Http/Middleware/CreateFreshApiToken.php

PR #4134, opened 2017-02-02
Timezone should be part of .env

jolamar:
Laravel is big enough now that nearly all apps will have to setup their timezone to be something other than UTC, if they are going to make use of date functions. Changing this after models have been created with UTC dates would make dates on older models UTC and newer ones a different timezone.


jolamar:
May be worth adding a comment of support timezones http://php.net/manual/en/timezones.php

RonMelkhior:
Uh, most apps still use UTC, due to it not having DST, and then just convert the timezone via Carbon or their own solution.

jolamar:
Ah makes sense. Thanks Ron!

PR #4141, opened 2017-02-16
Update package.json

coolpraz:
update laravel mix to new version

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "axios": "^0.15.3",
     "bootstrap-sass": "^3.3.7",
     "jquery": "^3.1.1",
-    "laravel-mix": "^0.7.2",
+    "laravel-mix": "^0.8.1",
     "lodash": "^4.17.4",
     "vue": "^2.1.10"
   }


RomainLanz:
Why not using [greekeeper](https://greenkeeper.io/) instead of having every week PR like that?

GrahamCampbell:
If we wanted to just blindly update the version, we could just use a wildcard. I don't see why greenkeeper is useful tbh.

RomainLanz:
It's not blindly update....

Greenkeeper will clone the repository, applies the update, run the test-suite (that we don't have) of the JavaScript code, then create a PR to update the package with all result from the test-suite, changelog of the package, commits from the latest update, etc.

GrahamCampbell:
> Greenkeeper will clone the repository, applies the update, run the test-suite (that we don't have) of the JavaScript code, then create a PR to update the package with all results from the test-suite, changelog of the package, commits from the latest update, etc.

Basically a blind update then. Doesn't do anything to solve the so called PR problem described.

RomainLanz:
The "PR problem" is that someone need to create the PR.

Same goes for all other JS dependencies.
```
"devDependencies": {
  "axios": "^0.15.3",
  "bootstrap-sass": "^3.3.7",
  "jquery": "^3.1.1",
  "laravel-mix": "^0.8.1",
  "lodash": "^4.17.4",
  "vue": "^2.1.10"
}
```

PR #4142, opened 2017-02-16
Use single quotes for passwords.user translation

caiquecastro:


caiquecastro:
Why?

marky291:
What is the actual reason for this ? Pointless

DojoGeekRA:
@caiquecastro i think that because of the double quotes, there is no need to escape the single quote and the message reads better.

PR #4143, opened 2017-02-17
Change hard-coded '/login' param to named route

kneipp:
Change hard-coded '/login' param to named route 'login'.

Adding a bit of flexibility to work with translations w/o need change Handler class:
`
Route::get('entrar', 'CustomController@showLoginForm')->name('login'); `

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -60,6 +60,6 @@ class Handler extends ExceptionHandler
             return response()->json(['error' => 'Unauthenticated.'], 401);
         }
 
-        return redirect()->guest('login');
+        return redirect()->guest(route('login'));
     }
 }


lioannou:
@taylorotwell @kneipp - there is one small problem with this PR: on a brand new 5.4 install, you cause a "InvalidArgumentException - Route [login] not defined" if you try and hit the default `my.app/api/user` which is registered.

In saying that - the old behavior before this PR looks like it was a "404 NotFoundHttpException" (since there is no `\login` on a base install.

Not sure which is the lesser of two evils? Maybe the default should be to `\` so the "out of the box" install works, and let people customise it themselves as they see fit?

Leave it with you...

kneipp:
Are you trying to hit /api/user without run art make:auth?

Guess not related with this PR. May be /api/user need be on make:auth scaffold related.

lioannou:
I'm not personally trying. The point is the default Exception Handler now has a default route named `login` that does not exist on a default Laravel installation.

It's not a massive drama either way - but I'm thinking of "new" Laravel people trying to use it, and getting an error out of the box, might cause some confusion.

PR #4144, opened 2017-02-17
Use 5.4's new fluent route syntax in API routes file

svenluijten:


taylorotwell:
Specifying middleware first feels better to me.

PR #4146, opened 2017-02-18
APP_NAME variable for Application Name

abixalmon:
Added APP_NAME environment variable for Application Name in config/app.php

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,6 +3,7 @@ APP_KEY=
 APP_DEBUG=true
 APP_LOG_LEVEL=debug
 APP_URL=http://localhost
+APP_NAME=Laravel
 
 DB_CONNECTION=mysql
 DB_HOST=127.0.0.1
--- a/config/app.php
+++ b/config/app.php
@@ -12,7 +12,7 @@ return [
     | any other location as required by the application or its packages.
     */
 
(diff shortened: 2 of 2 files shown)


mateusjatenee:
I particularly don't find this very interesting as the app's name is not
going to change very often.
Em sáb, 18 de fev de 2017 às 02:48, Abi Narayan <notifications@github.com>
escreveu:

overtrue:
-1

mrkatz:
It might not change often, however it would be useful to have a central location to declare the app name and be refrenced from there

astierler:
I agree that while not very useful, it does kind of make it more consistent to the rest of the app config settings. (Or if deploying the same Laravel app for multiple different sites - not saying I've done this, but can't hurt)

GrahamCampbell:
This has already been proposed. Just edit the config files for your app. They're there to be edited after all, otherwise we'd not include them in the app structure by default.

The env variables are for environment specific variables, such as your database password, and not for things specific to your application.

GrahamCampbell:
> however it would be useful to have a central location to declare the app name and be refrenced from there

We already have this, from the config file. You should never try to read env variables outside of config files in laravel, because of config caching, and php threading bugs.

PR #4147, opened 2017-02-18
Added --watch-poll

ilyahoilik:
There is problem with `npm run watch` (webpack doesn't detects a changes) without `--watch-poll`. I use Homestead.


taylorotwell:
@JeffreyWay thoughts?

JeffreyWay:
File watching is always a little strange with VMs. Some people have no issues, others need to opt for polling. Polling, as I understand it, puts a bit more strain on things, so it probably shouldn't be the default. 

Maybe we can make it more clear in the docs that `--watch-poll` is an alternative. I'll include that with my Mix PR.

mlantz:
Could you not add another script called watch-poll which would leave it built in but as an additive

taylorotwell:
That might be a better option.

PR #4148, opened 2017-02-20
Remove .env files from versioning

m1guelpf:
This change will protect enviroment files like .env.local .env.prooduction .env.remote from versioning

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,5 @@
 /.idea
 Homestead.json
 Homestead.yaml
-.env
+.env*
+!.env.example


marky291:
You are assuming such files will exist.

m1guelpf:
@marky291 It's common to have a .env.production and if some of that files doesn't exist, noting happens.

frezno:
imho no need for this.
It's up to the user to add or delete some files which may or may not have an affect on gitignore. But that's up to the user in this case.
Otherwise you might have to add eg a manual.md file and a changelog file etc. as well since those might be added by the user as well.

thecodemill:
FYI, whilst not always practical, it's recommended NOT to use .env.production files, or even the .env file itself on production servers. Such config options should be passed to the environment from the server configuration to reduce overhead - https://github.com/vlucas/phpdotenv#usage-notes

marky291:
Laravel is built to use .env even for production.

PR #4150, opened 2017-02-20
ignore log files

AustinPaquette:
Prevent the accidental commit of any log files - typically generated by npm in the event of an error `npm-debug.log`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 Homestead.json
 Homestead.yaml
 .env
+*.log


FakeHeal:
Why would a log file be important to *all*?

taylorotwell:
No plans to do this out of the box.

AustinPaquette:
@taylorotwell totally fair, figured I'd submit a PR and see what the response was like.

AustinPaquette:
> Maybe better to just specify the npm error log name then assume all logs.

@marky291 Is there any kind of `*.log` file that would be worth keeping? I can't think of one off-hand myself, perhaps I'm missing something?

PR #4151, opened 2017-02-21
Refactor npm scripts

JeffreyWay:
Per #4147, this adds a new `npm run watch-poll` command, and dries up some of those long repetitive npm scripts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,14 @@
 {
   "private": true,
   "scripts": {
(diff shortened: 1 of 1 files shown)


SeoFood:
@JeffreyWay  you sure that the production script will work? `mix.config.inProduction` is always false for me.

JeffreyWay:
@SeoFood The latest version of Laravel Mix will check for the `-p` flag. If you're on an older version, you can either keep your existing NPM scripts, or update to v0.8.2.

SeoFood:
thx @JeffreyWay that old version :)

JeffreyWay:
Gosh - why on Earth would they make it so that `$npm_package_config_webpack` works for Mac and Linux, but you have to use `%npm_package_config_webpack%` for Windows specifically? How are folks supposed to know this?? Fixing this now.

PR #4152, opened 2017-02-21
csrfToken is auto added to window.Laravel

conterno:
On line 31 of [`/resources/assets/js/boostrap.js`](https://github.com/laravel/laravel/blob/master/resources/assets/js/bootstrap.js) there is this header `'X-CSRF-TOKEN': window.Laravel.csrfToken,` added to the default axios headers. You get an error out of a fresh install of Laravel because the window.Laravel.csrfToken is not being set anywhere by default. You have to know about the csrf token and that Laravel expects you to put it in your html manually if you want to use the new Laravel Mix functionality and do things like `npm run dev`.

Yes, `php artisan make:auth` will put a new default layout view but if they don't do this... ?

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -6,6 +6,9 @@
         <meta name="viewport" content="width=device-width, initial-scale=1">
 
         <title>Laravel</title>
+        
+        <!-- CSRF Token -->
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't think it makes sense to put this on the welcome view. There is no form nor is it a layout.

PR #4153, opened 2017-02-23
Add .vagrant folder to .gitignore.

JimMerioles:


RomainLanz:
This is project specific and shouldn't be in every default installation...

svenluijten:
@RomainLanz I disagree, since it's mentioned you can do this in [the documentation](https://laravel.com/docs/master/homestead#per-project-installation), it might help some users who forget to add the folder to their `.gitignore` manually 🙂

taylorotwell:
Yes your global gitignore file. That is what it is there for.

PR #4156, opened 2017-02-28
Checking version php.

anhduca4:
Can you add checking version php in index.php on public.


app/Exceptions/Handler.php
```diff
@@ -60,6 +60,6 @@ protected function unauthenticated($request, AuthenticationException $exception)
             return response()->json(['error' => 'Unauthenticated.'], 401);
         }
 
-        return redirect()->guest('login');
+        return redirect()->guest(route('login'));
```
GrahamCampbell: What if we crashed before we booted up rooting?

GrahamCampbell:
Looks like there are lots of other changes in here by mistake.

PR #4159, opened 2017-03-03
Migration bug fixed

Advaith3600:
Migration error when doing `php artisan migrate` in a fresh project give an error like - Specified key was too long error, Fixed

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -2,8 +2,8 @@
 
 namespace App\Providers;
 
-use Illuminate\Support\ServiceProvider;
 use Illuminate\Support\Facades\Schema;
+use Illuminate\Support\ServiceProvider;
 
 class AppServiceProvider extends ServiceProvider
 {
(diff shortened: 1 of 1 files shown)


joshmanders:
This is already talked about in the docs: https://laravel.com/docs/5.4/migrations

app/Providers/AppServiceProvider.php
```diff
@@ -13,7 +14,7 @@ class AppServiceProvider extends ServiceProvider
      */
     public function boot()
     {
-        //
+        Schema::defaultStringLength(191);
```
Advaith3600: But If I am not wrong, everyone faces this problem. There are many who don't know how to fix this.
So I think it is better to fix it by default

jayaregalinada: Actually __NOT EVERYONE__ are facing the same problem. However, I don't know why others do not have this kind of issue. I tried installing the same provisions, programs, software, mysql and mariadb versions on a different kind of machines. Other machines are failed and faces this issue, others are just as expected and have a successful migration.

PR #4161, opened 2017-03-04
Faker locale can be set in config, add that to the default config files

vpratfr:
At some point in time, Laravel allowed to change the locale used by Faker. However that configuration variable has remained undocumented and people are probably still doing it with a custom provider binding.

Hence I just added the variable to the default project's `app.php` file.

See also: http://stackoverflow.com/questions/34496720/change-faker-locale-in-laravel-5-2


taylorotwell:
We can document the option instead.

PR #4162, opened 2017-03-04
[5.4] Fix cross-env bin path

eklundkristoffer:
https://github.com/kentcdodds/cross-env/commit/aae44dff49494912a76423a7e45d3d1afd630e4a

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,11 +1,11 @@
 {
   "private": true,
   "scripts": {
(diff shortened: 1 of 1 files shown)


willvincent:
@pagchen Yarn does not create that node_modules/.bin directory. So unless/until it does (Refer to [this yarn PR](https://github.com/yarnpkg/yarn/pull/1210)) this should _not_ be changed to node_modules/.bin/... paths.

PR #4166, opened 2017-03-06
Fix cross-env path

kentcdodds:
Due to npm's [path](https://docs.npmjs.com/misc/scripts#path) behavior with npm scripts, you do not need (and should not use) the full path of scripts in your npm scripts. They will be available to you automatically when you run your scripts with `npm run`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,11 +1,11 @@
 {
   "private": true,
   "scripts": {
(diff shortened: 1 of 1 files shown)


jankal:
Related to #4162

kentcdodds:
Huh... I use yarn exclusively (as do many others) and I've never had the issues that are talked about in that PR. But feel free to do whatever you like. Like I said, it's very unlikely that this will change in the future. That said, I make no gaurantees.

PR #4167, opened 2017-03-06
Fix cross-env binary

driesvints:
Resending in #4166 because it *is* actually the correct behavior. The only thing what was missing is that since you're directly using the binary it *needs* to be added to the devDependencies which will make it available in the `.bin` directory (even if you use Yarn). You can drop the full path like @kentcdodds reasoned. 

This way #4162 won't be an issue anymore in the future.

/cc @kentcdodds @jankal @willvincent @eklundkristoffer @pagchen

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,15 +1,16 @@
 {
   "private": true,
   "scripts": {
-    "dev": "node node_modules/cross-env/dist/bin/cross-env.js NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
-    "watch": "node node_modules/cross-env/dist/bin/cross-env.js NODE_ENV=development node_modules/webpack/bin/webpack.js --watch --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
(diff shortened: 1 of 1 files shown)


driesvints:
@JeffreyWay you might be interested in this as well.

taylorotwell:
@JeffreyWay can you confirm this?

JeffreyWay:
Only thing would be that this forces Laravel to be "responsible" for the `cross-env` dependency, instead of Mix. But that basically amounts to a single line in the default package.json file, so not a biggie. 

@taylorotwell I've tested this and it does work.

driesvints:
@JeffreyWay you actually "want" your app to be responsible for that because it directly uses the dependency. If the dependency was only used internally inside Mix then that wouldn't be an issue.

The current issues are an example of that. Because `cross-env` is required as a child-dependency of Mix, it doesn't gets added to the `.bin` directory and then you can't use it in your app (you don't directly want to reference it like @kentcdodds reasoned).

JeffreyWay:
I'm not disagreeing. I'm saying it's a trade-off. Same reason why Mix auto-installs Webpack for you, too.

driesvints:
@JeffreyWay true that.

PR #4168, opened 2017-03-06
Adding Same-site to config

fernandobandeira:
This PR add's the same-site option for the config/sessions file, I've created it so we don't forget to add it 😄 

Comment needs to be refactored =/


taylorotwell:
Merged.

PR #4169, opened 2017-03-07
artisan.bat

saidbakr:
instead of writing php artisan every time you want to use the generator this batch file do it without php. just write artisan followed by your command

### DIFF
--- a/artisan.bat
+++ b/artisan.bat
@@ -1,0 +1,3 @@
+REM instead of writing php artisan every time you want to use the generator this batch file do it without php. just write artisan followed by your command
+@ECHO OFF
+php artisan %*


joshmanders:
I think `php artisan` makes it more universal. Best to keep this as something you use in your own environment available in your path. 

I do this in macOS. It just checks to make sure that `artisan` is a file in the current working directory otherwise throws an error.

saidbakr:
@joshmanders I think that there is something like batch file in MacOS, I think it is called bash script or something like that, so creating one with the same name `artisan` for MacOS  it should work fine and artisan.bat will be neglected there.

joshmanders:
@saidbakr yes, I'm saying if putting `artisan.bat` then for consistency there should be bash too, but that just adds a bunch of clutter, and this is best to be left out, IMHO.

zaknes:
Josh is right in that if we have a file for Windows command lines, we'd need one for bash, as well.

If `php artisan <command>` bothers you, you can always make an alias.

This is what I use:
```bash
alias art="php artisan "
```

([Cmder](http://cmder.net/) is a good console emulator for Windows).

ismailzafar:
@saidbakr This feature is not worth to be added. You can add alias in git bash by using @zaknes method.

ludo237:
It's seems an overhead, you can create a simple alias for that even on windows

ichmbch:
Well, actually I think "laravel/installer" needs to be bundled with artisan-cli (something similar to gulp-cli or grunt-cli). While the ability to call artisan without php prefix could be a good thing, I think the main benefit would be the ability to call artisan from project's subdirectories. 

Someone already tried to do that (though I haven't tested it) but I think laravel needs official support for this: https://github.com/johnnyfreeman/artisan-cli

PR #4170, opened 2017-03-07
Globally initialized JQuery

renjithvk:
Resolve "Uncaught ReferenceError: jQuery is not defined" while try to install plugins


taylorotwell:
You make this change as needed in your own application.

PR #4172, opened 2017-03-09
Pass request object to register create method

TheGIBSON:
framework: https://github.com/laravel/framework/pull/18278


taylorotwell:
I don't really want to introduce a breaking change here. You can access the current request via the `Request` facade or `request` helper.

TheGIBSON:
It was intended for the new version. But ok.

PR #4173, opened 2017-03-09
replaced the default favicon with laravel logo

SamAsEnd:
---

- Removed the `0` byte favicon placeholder
- Used laravel.com favicon for placeholder


SamAsEnd:
What do you think about this @themsaid ?

taylorotwell:
Going to hold off on this because if we do this soon we'll be getting PRs for apple touch logos, and all kinds of other logos. 😆

martindilling:
It could be a artisan command to make all those needed fav/apple/android/windows icons as templates, always a pain to figure out what icons are needed xD

lioannou:
There are services that generate these all from one image automatically. All favicons, apple icons, browseconfig.xml. manifest etc in one command. I use http://realfavicongenerator.net/ personally - but there are a few others.

PR #4175, opened 2017-03-10
sqlite.database needs to be absolute to be able to work for artisan and serve

hannesvdvreken:
When `DB_DATABASE=storage/database.sqlite` then it doesn't work
for both `php artisan tinker` and `php artisan serve`.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -35,7 +35,7 @@ return [
 
         'sqlite' => [
             'driver' => 'sqlite',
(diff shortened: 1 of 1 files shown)


taylorotwell:
Usually I just put the full path the .env file so it's not an issue. I might leave it as is for now at least for the 5.4 release and then re-evaluate it for 5.5. Thanks.

hannesvdvreken:
Ok 👍

PR #4183, opened 2017-03-22
Specified key was too long

afiqiqmal:
fix on Syntax error or access violation: 1071 Specified key was too long; max key length;


mingyoung:
#4159

PR #4185, opened 2017-03-22
Add comment for lodash in bootstrap.js

scrubmx:
This will make the file more consistent and at the same time will serve as a hint for new developers as to why we are including lodash in the first place.

### DIFF
--- a/resources/assets/js/bootstrap.js
+++ b/resources/assets/js/bootstrap.js
@@ -1,3 +1,8 @@
+/**
+ * Lodash is a general purpose library which provides utility functions to
+ * solve common programming tasks like manipulating strings, arrays and
+ * objects with a clean and consistent syntax that's very performant.
+ */
 
 window._ = require('lodash');
 


resources/assets/js/bootstrap.js
```diff
@@ -1,3 +1,8 @@
+/**
+ * Lodash is a general purpose library which provides utility functions to
+ * solve common programming tasks like manipulating strings, arrays and
+ * objects with a clean and consistent syntax that's very performant.
```
GrahamCampbell: Makes it sound like the syntax is performant, which it may well be, for the programmer, but I don't think that was the intended meaning of this phrase, rather, it was meant to say the algorithms are "fast"?

scrubmx: @GrahamCampbell You are right, my english is not the best. How about this:

```js
/**
 * Lodash is a general purpose library which provides utility functions to
 * solve common programming tasks like manipulating strings, arrays and
 * objects using an elegant syntax while still being very performant.
 */
```

taylorotwell:
Don't think we need this right now.

PR #4188, opened 2017-03-26
Require laravel-mix v0.*

themsaid:


deframe:
If you're allowing 0.9 of Mix to be installed then you'll need to update webpack.mix.js to autoload jQuery or the Bootstrap JS will break (perhaps along with other things).

`
mix.autoload({
    jquery: ['$', 'jQuery', 'window.jQuery']
})
`

jQuery was autoloaded implicitly before 0.9.

PR #4189, opened 2017-03-27
Composer -o

tabuna:
Composer allows us to optimize the autoloader using the --optimize flag or shortly -o.


lucasmichot:
@taylorotwell is `php artisan optimize` still useful then ?

PR #4194, opened 2017-03-30
why request redirect root:url ?

thandonguocmo2020:
Laravel redirect to root on request validation error .

I have a required field in a Laravel Request Class and when that field is not present in the request, the request is redirected to root '/'.

I am sending the request via Postman.

File request of me. 

```<?php

namespace App\Api\V1\Admins\Requests;

use Illuminate\Foundation\Http\FormRequest;

class LoginRequest extends FormRequest
{

    public function authorize()
    {
        return true;
    }

    public function rules()
    {
        return [
            'email' => 'required|email',
            'password' => 'required'
        ];
    }

    public function messages()
    {
        return parent::messages(); // TODO: Change the autogenerated stub
    }
}
```
Uri : 
```php
   Route::post('/login', function(\App\Api\V1\Admins\Requests\LoginRequest $request){
      dd($request->all());
 
    });
```
// post ajax send 👍 

if error => redirect path "/"

else success => redirect next();

how to fix bug help.

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -18,6 +18,7 @@ class Kernel extends HttpKernel
         \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
         \App\Http\Middleware\TrimStrings::class,
         \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
+        \App\Http\Middleware\TrustProxies::class,
     ];
 
     /**
--- a/app/Http/Middleware/EncryptCookies.php
+++ b/app/Http/Middleware/EncryptCookies.php
@@ -2,9 +2,9 @@
 
 namespace App\Http\Middleware;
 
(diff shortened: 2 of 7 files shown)


thandonguocmo2020:
Laravel redirect validator root url "/" by it check ajax == false. if it return true, it run

If ($ this-> expectsJson ()) {
             Return new JsonResponse ($ errors, 422);
         } 

I solved the problem by adding  setup POST MAN in Headers: 


I hope it will be useful to the next person when the validator does not work as you expect, you need to set it up like sending ajax to the routes api:


"Key" => "value"

X-Requested-With => XMLHttpRequest

Post Man need config setting some ajax

Otherwise laravel will not be able to return error code 422.


If ($ this-> expectsJson ()) {
             Return new JsonResponse ($ errors, 422);
         }

==> return false if not have "X-Requested-With" : "XMLHttpRequest" in headers POST MAN.

ajax or not ajax active other #. 

sorry i do not know english.

PR #4198, opened 2017-04-01
Update bootstrap.js

Artiik-fr:
Require VueJS only if it is not defined.


taylorotwell:
You can change these files however you want. They are in your app.

PR #4200, opened 2017-04-02
Prevent email and password being identical

tillkruss:
Don't allow users to signup with a password that equals their email address. It's a fast check and raises awareness.

See https://github.com/laravel/framework/pull/18622.


taylorotwell:
Feels like a really strange check to add. Should we add checks to prevent other common things like "can't set password to "123456", etc.? Going to hold off for now.

PR #4202, opened 2017-04-03
Use ~ import in app.scss

mihai-scurtu:
I'm new to webpack and I thought that importing from node_modules is the correct way to go, which didn't work.

And that led me to the answer: https://github.com/JeffreyWay/laravel-mix/issues/412

Shouldn't the example file be that way too?


taylorotwell:
The file is currently working.

PR #4205, opened 2017-04-03
Remove blank lines in assets

mdavis1982:
This PR removes the blank lines at the start of the JS and SCSS files.

It's entirely possible that these are there for a reason that I don't know about - but I remove them each time I start a new project and I've not had any issues 😄


taylorotwell:
I just like it better with them. Sorry :/

PR #4206, opened 2017-04-03
Make cache key prefix configurable in environment

mdavis1982:
This PR makes the cache key prefix configurable in the environment file.

## Why?
There are some situations where two instances of a project (e.g. staging / UAT) that could share the sae underlying cache implementation. By allowing the cache key prefix to be configurable in the environment, each project can effectively namespace their cache preventing problems.

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -86,6 +86,6 @@ return [
     |
     */
 
-    'prefix' => 'laravel',
+    'prefix' => env('CACHE_KEY_PREFIX', 'laravel'),
 
 ];


taylorotwell:
You can just do this in your app as needed.

mdavis1982:
How do you do this in the app? I could be wrong, but I thought that the prefix is appended to each cache entry, which makes it hard to change without changing the configuration file?

laurencei:
@mdavis1982 - Taylor means just change it in your own Laravel app once it is downloaded - i.e. make that same change yourself in your configuration file - just on your local copy once installed.

mdavis1982:
But that was my point - if two copies of the app share the same server, and potentially the same caching storage (for example, in the case of a staging environment and a UAT environment) each could overwrite the other's caching storage causing unusual side effects.

If this was configurable in the environment, then you could prefix one with `myapp_staging` or `my app_uat` without any problems, and without needing the change it in the committed codebase.

laurencei:
@mdavis1982  - I'll ping you on slack - lets not talk here

uyab:
For anyone else who have same issue, simply edit your config file `cache.php`:
```
...
// before
//'prefix' => 'laravel',

// after
'prefix' => env('CACHE_PREFIX'),
...
```

PR #4208, opened 2017-04-06
Add to axios defaults instead of overwriting them

misenhower:
This change ensures the default "Accept" header specified by axios is retained.

Previously, checking `$request->wantsJson()` on a request made from axios would return false since the "Accept: application/json... (etc)" header was omitted.

This change uses the format specified in the axios docs here https://github.com/mzabriskie/axios#global-axios-defaults


a-komarev:
@misenhower thank you for this catch!

PR #4209, opened 2017-04-07
Add *.swp to .gitignore

connectkushal:
Git tracks the swap files which often becomes a headache for vi users and other programs that creates them.


GrahamCampbell:
These should be added to your global .gignore file.

PR #4210, opened 2017-04-07
Fix possible errors on DocBlocks

arikaiy:


taylorotwell:
I keep it this way on purpose. To show that that class has more functionality available.

PR #4215, opened 2017-04-10
Middleware returns ControllerMiddlewareOptions

dersvenhesse:
The `return void` documentation is incorrect since the called `middleware` function in `Routing\Controller` does return `ControllerMiddlewareOptions`.


taylorotwell:
Can you send this to 5.4 branch? Also, constructor signatures should always be `@return void`

GrahamCampbell:
> The return void documentation is incorrect since the called middleware function in Routing\Controller does return ControllerMiddlewareOptions.


👎Constructorss cannot return values.

dersvenhesse:
I think theoretically they can, since its valid syntax. It might not make sense because there is no way to use the return value.

PR #4216, opened 2017-04-10
OCD order change

laurencei:
Clean up the order - more PSR-2 stylish


a-komarev:
:( What is a pros of the order by string length? As for me - it's much easier to see all classes grouped by vendor\package and etc.

laurencei:
I didnt notice the class names were alphabetical :)

I thought it was just a random order.

All good - we can just leave it.

PR #4217, opened 2017-04-10
Remove underscore from default session name

laurencei:
Solves https://github.com/laravel/framework/issues/18357

There is no denying it is an edge case that occurs rarely and only in certain situations - but there is no harm in removing the underscore - so seems worth it?

_Random question; while we are at it - should we remove "laravel" from the default session name? Perhaps call it 'websession' or something? Seems like an unnecessary leaking of knowledge that the application is built on Laravel?_


taylorotwell:
Hmm, this feels somewhat edge case and be changed at the app level. Lots of cookies have underscores.

PR #4218, opened 2017-04-11
Update to latest cross-env

dwightwatson:
Updates the dependency constraint to the latest version, so Laravel isn't shipping with any outdated dependencies. [From looking at the breaking changes](https://github.com/kentcdodds/cross-env/releases) we should be fine for the setting of `NODE_ENV`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "devDependencies": {
     "axios": "^0.15.3",
     "bootstrap-sass": "^3.3.7",
-    "cross-env": "^3.2.3",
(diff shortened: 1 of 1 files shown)


askaoru:
There are still some issues with the v4 cross-env for windows platform, so in my opinion I think we should hold this off for a while until it's resolved.

https://github.com/kentcdodds/cross-env/issues/103
https://github.com/kentcdodds/cross-env/issues/99

dwightwatson:
Ah fair enough, too quick off the bat!

PR #4221, opened 2017-04-12
Correct namespaces for default listeners

watscho:
Added backslashes to namespaces, otherwise `php artisan event:generate` is creating files in `app/Providers/App/Events/...` instead of `app/Events/...`


GrahamCampbell:
Looks like a bug in the command. Backlashes here are actually wrong.

taylorotwell:
Works perfectly for me on a fresh Laravel installation using the `event:generate` command.

watscho:
Apparently I launched the command wrong. sorry

PR #4222, opened 2017-04-12
Using app name from .env file

soufyanekaddouri:
Using app name from .env file instead of configuration file

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="{{ config('app.locale') }}">
+<html lang="{{ env('APP_NAME') }}">
     <head>
         <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


DojoGeekRA:
@soufyanekaddouri you can't fetch values from `env` function after you do a `php artisan config:cache` (which is the case of every deployed app).

[Here](https://laravel.com/docs/5.4/configuration#configuration-caching) is the reference in the documentation.

al0mie:
I guess app name and locale are different things.

taylorotwell:
Correct, we are not supposed to fetch using env outside of configuration files since they are cached.

GrahamCampbell:
> Correct, we are not supposed to fetch using env outside of configuration files since they are cached.

Yes. This is VERY important. Laravel doesn't actually load the environment variables when config is cached.

PR #4223, opened 2017-04-12
use name from config file

al0mie:


m1guelpf:
@al0mie Since it's an example view, and also contains laravel-related links, I would leave it as it is.

taylorotwell:
Will leave as-is for now.

PR #4226, opened 2017-04-17
Remove extra autoload file

diogoazevedos:
The autoload file is not used anymore, just for define a constant.


joshmanders:
This breaks `artisan`, as you need to update it to point to the `vendor/autoload.php` too.

PR #4227, opened 2017-04-17
Fix break from #4226

joshmanders:
https://github.com/laravel/laravel/pull/4226 removes `bootstrap/autoload.php` and breaks `artisan`. This PR fixes `artisan` to work.

### DIFF
--- a/artisan
+++ b/artisan
@@ -1,6 +1,8 @@
 #!/usr/bin/env php
 <?php
 
+define('LARAVEL_START', microtime(true));
+
 /*
 |--------------------------------------------------------------------------
 | Register The Auto Loader
--- a/artisan
+++ b/artisan
@@ -13,7 +13,7 @@
 |
 */
 
-require __DIR__.'/bootstrap/autoload.php';
+require __DIR__.'/vendor/autoload.php';
 
 $app = require_once __DIR__.'/bootstrap/app.php';
 
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit backupGlobals="false"
          backupStaticAttributes="false"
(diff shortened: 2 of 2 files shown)


laurencei:
You need to also add `define('LARAVEL_START', microtime(true));` in there as well

GrahamCampbell:
👎 Other PR should be reverted. It breaks phpunit tests too.

joshmanders:
Good catch there @GrahamCampbell, I'll update this PR to fix that too.

GrahamCampbell:
Already send a revert PR. It's the best fix for the issue.

GrahamCampbell:
No point in removing the existing file. It did a perfect job.

GrahamCampbell:
PHPUnit is broken in the sense that the constant doesn't get defined there.

joshmanders:
No use keeping a single file that requires `vendor/autoload.php` only..

joshmanders:
I just did a search of the codebase (quickly using find in files) and `LARAVEL_START` is only ever defined, never used...

laurencei:
@joshmanders - yes, but current packages/apps will be expecting it to be there.

joshmanders:
Good point @laurencei

laurencei:
@joshmanders - one option though is to define it in the CreatesApplication trait? That would solve the problem.

joshmanders:
I thought about that, but seems hacky.

joshmanders:
I guess I'll leave PR open and see what @taylorotwell says.

DojoGeekRA:
@joshmanders i'm very curious about what is the purpose of `LARAVEL_START`, what is it used for?

laurencei:
@DojoGeekRA - it is a constant that marks the start of the Laravel Application.

Although not directly used by Laravel itself - it is available for developers/packages etc to determine performance/run times etc.

DojoGeekRA:
Thanks @laurencei!

PR #4228, opened 2017-04-17
Revert "Remove extra autoload file"

GrahamCampbell:
Reverts laravel/laravel#4226

PR #4230, opened 2017-04-18
Fix Laravel Start for PHPUnit

laurencei:
@GrahamCampbell - will this fix the issue for https://github.com/laravel/laravel/pull/4228?

### DIFF
--- a/tests/CreatesApplication.php
+++ b/tests/CreatesApplication.php
@@ -13,6 +13,8 @@ trait CreatesApplication
      */
     public function createApplication()
     {
+        define('LARAVEL_START', microtime(true));
+        
         $app = require __DIR__.'/../bootstrap/app.php';
 
         $app->make(Kernel::class)->bootstrap();
--- a/tests/CreatesApplication.php
+++ b/tests/CreatesApplication.php
@@ -14,7 +14,7 @@ trait CreatesApplication
     public function createApplication()
     {
         define('LARAVEL_START', microtime(true));
-        
+
         $app = require __DIR__.'/../bootstrap/app.php';
 
         $app->make(Kernel::class)->bootstrap();


GrahamCampbell:
This won't work, because we can't define a constant more than once.

laurencei:
yeah - was just thinking that...

even using 'is_defined' wont work - because we need to re-define it for each test?

taylorotwell:
I don't think we need to be sweating over this. There isn't even a clear reason to use the constant and if someone has one they can make the file themselves.

laurencei:
@taylorotwell - the constant is used by lots of packages. 

https://github.com/barryvdh/laravel-debugbar is the most popular.

So @GrahamCampbell is right - there are going to be side effects...

arcanedev-maroc:
I think it's wise to keep the old `bootstrap/autoload.php` file because:

 * It's a breaking change for many packages that uses the `LARAVEL_START` constant.
 * Useful for the `old` shared hosting that doesn't gives you the php.ini to edit and if you want to solve issues like this:

```php

<?php 
// bootstrap/autoload.php
ini_set('xdebug.max_nesting_level', 500); // To fix the `Maximum function nesting level of 'xxx' reached, aborting!`


define('LARAVEL_START', microtime(true));
```

GrahamCampbell:
> ini_set('xdebug.max_nesting_level', 500); // 

No point in that. The new default in xdebug is large enough anyway.

arcanedev-maroc:
I talking about the `old` shared hosting.

GrahamCampbell:
Meh, I'm massively against adding that for old shared hosting. Use better hosting, or fix it yourself would be my approach. No reason to have xdebug in production (in most cases anyway).

arcanedev-maroc:
Yeah, tell that to the "Client" :)

PR #4232, opened 2017-04-19
Ignore the npm debug log

JuanDMeGon:
When an NPM run fails, is created a npm-debug.log file. This file can be ignored and should not be published on a Git based repository.


m1guelpf:
@JuanDMeGon I think this should be added to your global gitignore

PR #4233, opened 2017-04-19
Use dirname() instead of realpath()

williamgb:
See: https://github.com/kalessil/phpinspectionsea/blob/master/docs/probable-bugs.md#phar-incompatible-realpath-usage

### DIFF
--- a/bootstrap/app.php
+++ b/bootstrap/app.php
@@ -12,7 +12,7 @@
 */
 
 $app = new Illuminate\Foundation\Application(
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
👎 This will never be used in a phar.

joshmanders:
Why would you put Laravel in a phar?

taylorotwell:
This exists in laravel/laravel so if you need to change it change it in your own application.

PR #4236, opened 2017-04-23
Allow filesystem to be changed on config

caiquecastro:
Allow to change the filesystem storage on the fly.
For example, you can swap the storage disk with a fake one with
`Storage::fake` for tests.


GrahamCampbell:
Seems fair. 👍

laurencei:
Do you need the `FILESYSTEM_DEFAULT` and `FILESYSTEM_CLOUD` in the default `.env.example` though? It's starting to get fairly large in there...

So to clarify; I'm saying just have the changes to `config/filesystem.php` as part of this PR, and not the `.env.example` as well?

caiquecastro:
Remove the env variables from .env.example

troccoli:
I think we do. If I see something in .env.example I know it's something I
can customise.

caiquecastro:
Actually the default and cloud keys are mapped to disks array, so, after thinking more about it, it doesn't make sense to add them to .env, only if there are more disks declared on disks array.

taylorotwell:
Should be FILESYSTEM_DRIVER like the rest of the "default" values (queue, etc.)

caiquecastro:
So, should I replace FILESYSTEM_DEFAULT by FILESYSTEM_DRIVER? What about the cloud driver?

taylorotwell:
Yes.

Cloud variable name is fine.

caiquecastro:
Thanks. I'd updated the PR.

ConnorVG:
Seeing `_DRIVER` and `_CLOUD` is semi-frustrating; especially considering `_DRIVER` and `_CLOUD_DRIVER` looks equally silly.

Bummer!

PR #4237, opened 2017-04-24
Update default homepage

zachweix:
Change URL for logging in and registering on default homepage. Right now Laravel checks if there is a login route and then it sends the person to /login even if their login route is somewhere else, such as at /admin/login (e.g. they are using Backpack for Laravel). Therefore, instead this will check the exact URL for the login route.


taylorotwell:
I don't want to make this change right now.

PR #4238, opened 2017-04-25
Added ajax get and post request helpers

stea4lth:
I use these to test ajax protected routes.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -7,4 +7,20 @@ use Illuminate\Foundation\Testing\TestCase as BaseTestCase;
 abstract class TestCase extends BaseTestCase
 {
     use CreatesApplication;
+    
+    /**
+    * Make ajax POST request
+    */
+    protected function ajaxPost($uri, array $data = [])
+    {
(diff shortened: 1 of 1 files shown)


taylorotwell:
We already have postJson, etc.

stea4lth:
Yes, but they do not actually make an ajax request (they don't set the
headers). So, I ran into this issue...  I had a route protected with
if($request->ajax()) ... etc... when I tried to use postJson() the
conditional didn't execute because the request was not an ajax request, it
just expected that the response was json. I have noticed this problem with
others on other forums and laravel is to beautiful and elegant to not have
a very simple way of making a true ajax request, without having to do
$this->post($uri, $data, array('HTTP_X-Requested-With' =>
'XMLHttpRequest')); , etc...

I'm not really sure where they should go, I just stuck them in my
TestCase.php, but I think this should be added to the language in some
form, for this reason.

By the way, I would like to Thank You, you are my favorite teacher and have
taught me more about programming then I ever could have imagined.

THANK YOU

Chris

PR #4240, opened 2017-04-26
Update readme.md

B-07:


tejomayonline:
I am using ubuntu 14.04 LTS.It does not come up with mysql 5.7 .then I hav to tweak the configuration.right?

PR #4241, opened 2017-04-26
doctype uppercase to lowercase

B-07:
no harm, only benefits

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="{{ config('app.locale') }}">
     <head>
         <meta charset="utf-8">


GrahamCampbell:
> no harm, only benefits

How do you mean. Last time I checked, it should be upper case?

B-07:
[here's html5bp's reasoning from uppercase to lowercase](https://github.com/h5bp/html5-boilerplate/issues/1522). Also noticed that Taylor himself using lowercase in laravel.com

GrahamCampbell:
> Also noticed that Taylor himself using lowercase in laravel.com

We actually don't know this for sure. It's likely CloudFlare have changed this.

a-komarev:
Supporting it keep all html document structure completely lowercase.

PR #4243, opened 2017-04-26
fix indentation

B-07:
a tab was missing


taylorotwell:
I think it's fine as-is.

PR #4247, opened 2017-04-28
Revert Queue Prefix

laurencei:
Although it wont affect anything - it might lead to confusion if people think it is still an option.

It can be added back if the feature is re-added at a later stage.

### DIFF
--- a/config/queue.php
+++ b/config/queue.php
@@ -19,19 +19,6 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Queue Prefix
-    |--------------------------------------------------------------------------
-    |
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
👍

GrahamCampbell:
I don't like the idea of a queue prefix anyway. 👎 on having that feature, even if the implementation did work.

mateusjatenee:
I'm a little out of the loop — why was the feature removed on `laravel/framework`?

GrahamCampbell:
https://github.com/laravel/framework/pull/18860#issuecomment-298024056

taylorotwell:
@mateusjatenee because it didn't work.

PR #4250, opened 2017-04-30
Update index.php

muba00:


imabug:
[PSR 2](http://www.php-fig.org/psr/psr-2/).2 - The closing ?> tag MUST be omitted from files containing only PHP.

overtrue:
👎

muba00:
@imabug Thank you for information.

GrahamCampbell:
> tag MUST be omitted from files containing only PHP.

The reason for this is because if you accidently put a newline after the closing tag in a php file, then php will start sending the request to the client.

fengchenghao:
@GrahamCampbell LOL. It makes sense. Thank you for your explanation!

PR #4251, opened 2017-05-01
Better .htaccess dealing with ending slashes

adamyi:
This solves the wrong redirection to remove ending slashed if laravel is installed in a sub-directory of the server.

For instance, if my instance is installed at http://xx.xx.xx.xx/laravel, and I want to access the test route. If I access http://xx.xx.xx.xx/laravel/test/, the original would redirect to http://xx.xx.xx.xx/test, which yields a 404. This patch makes it redirect to http://xx.xx.xx.xx/laravel/test.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -7,7 +7,8 @@
 
     # Redirect Trailing Slashes If Not A Folder...
     RewriteCond %{REQUEST_FILENAME} !-d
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I think this has already been proposed in the past?

taylorotwell:
You can make any change to this file you need to. It lives in your application.

PR #4253, opened 2017-05-03
Add sqlsrv as group connection

insign:
Since the doc says that the Laravel supports SQL Server out of the box, makes sense add it, out of the box.


uxweb:
Great!!

PR #4257, opened 2017-05-09
Make sessions more accessible from .env

fwartner:
This will make the session configuration more flexible using the .env file.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -70,7 +70,7 @@ return [
     |
     */
 
-    'connection' => null,
+    'connection' => env('SESSION_CONNECTION', null),
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
No need to add this by default I think.

fwartner:
I often came into a situation were i need to edit these variables directly in the config files.. Becomes sometimes a pain..

GrahamCampbell:
> I often came into a situation were i need to edit these variables directly in the config files.. Becomes sometimes a pain..

Editing the config file is the default option, and that's not a bad thing. The environment variables should only be used for secrets, or things that change per environment. Things that are the same in your application should be set by editing the config files.

fwartner:
You are talking about "secrets"..

```
BROADCAST_DRIVER=log
CACHE_DRIVER=file
SESSION_DRIVER=file
QUEUE_DRIVER=database
```

Are these lines (all of them are by *default* in the .env file) some sort of "secrets"? ^^

GrahamCampbell:
> You are talking about "secrets"..

I didn't say just secrets. I said:

> used for secrets, or things that change per environment.

taylorotwell:
You can add this to your own app as needed.

PR #4258, opened 2017-05-09
Disable indexes

m1guelpf:
I believe this should be in the framework by default, as I've found lots of Laravel installations without this, and it's a basic security rule.


taylorotwell:
You can change this as needed in your own application already. If anything I would rather just get closer to matching other framework .htaccess file's like Symfonys.

askaoru:
Things like this are supposed to be handled on the (web) server level instead of the application level.

forj75:
Wow enough

Da iPad

PR #4260, opened 2017-05-10
Make window.Laravel.csrfToken unneccesarry

DrowningElysium:
I don't like the fact we have to add 
```
<script>
        window.Laravel = {!! json_encode([
            'csrfToken' => csrf_token(),
        ]) !!};
</script>
```
To have it working, when the docs suggest adding only the meta tag. This will get the token from the meta tag.


taylorotwell:
This was merged. Not sure why it shows it wasn't.

PR #4262, opened 2017-05-11
Add composer.lock to .gitignore

m1guelpf:
From StackOverflow:

> `composer.lock` is build metadata which is not part of the project. The state of dependencies should be controlled through how you're versioning them (either manually or as part of your automated build process) and not arbitrarily by the last developer to update them and commit the lock file.

> If you are concerned about your dependencies changing between composer updates then you have a lack of confidence in your versioning scheme. Versions (1.0, 1.1, 1.2, etc) should be immutable and you should avoid "dev-" and "X.*" wildcards outside of initial feature development.

> Committing the lock file is a regression for your dependency management system as the dependency version has now gone back to being implicity defined.

> Also, your project should never have to be rebuilt or have it's dependencies reaquired in each environment, especially prod. Your deliverable (tar, zip, phar, a directory, etc) should be immutable and promoted through environments without changing state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ Homestead.json
 Homestead.yaml
 npm-debug.log
 .env
+composer.lock


GrahamCampbell:
Composer.lock should be committed to applications.

Qh0stM4N:
I have a scenario I'm using ignore composer.lock files.

The application is distributed as a package and "git" based development is done.
New installations are done with ssh `$ git clone https://mysite/myuser/myrepo.git`. And the `$ composer update` command. This setup model sometimes changes  the composer lock file.
As a result, when we want to update the system, we have to do something like `$ git fetch origin master && git reset --hard origin/master` before the `$ git pull -ff origin master` command. The primary priority will be to keep the main repository equal to the updates there.
Where is application layer separated from dependency, Is it completely wrong to ignore the composer lock file ?

crynobone:
https://blog.engineyard.com/2014/composer-its-all-about-the-lock-file

If it doesn't affect you, add `composer.lock` to your local application `.gitignore`. But most of the time `laravel/laravel` is used for an application where you develop locally and deploy to production, which definitely needs `composer.lock` stored in via git.

fajarwahyu:
As for Qh0stM4N, you can do "$ composer install" instead of "$composer update". That way, your composer.lock file will not changed, thus leaving your git working directory intact.

forj75:
Stop 

Da iPad

RomainLanz:
@Qh0stM4N You should definitively use `composer install`.

It will install all of your dependencies inside your `composer.lock` file with the same version. So you are sure that the version of dependencies you have in local or in other installation are the same.

If you use `composer update` you will have different versions everywhere and that will become hard to know where the bug is.

Qh0stM4N:
thanks @RomainLanz @crynobone @fajarwahyu  this way is wront. From now on deleted /.composer.lock in gitignore files. composer update only main repository, testing application. If i am sure there are no problems pull last change clone projects and run composer install commands. :sweat_smile:

PR #4264, opened 2017-05-11
Wrong redirect to www-root instead of laravel root when removing trailing slashes

m1guelpf:
Fix https://github.com/laravel/framework/issues/19153

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -7,7 +7,8 @@
 
     # Redirect Trailing Slashes If Not A Folder...
     RewriteCond %{REQUEST_FILENAME} !-d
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I think this has already been proposed a few times before?

taylorotwell:
You can change the .htaccess as needed in your own application. Every change I have made to the .htaccess in the past bsed on a PR has later been reverted because it was wrong for someone else.

nolimitdev:
This change should not be wrong for someone else. It just fixes broken redirecting which can not be usefull for nobody.

PR #4265, opened 2017-05-12
Added database charset and collection environment variables.

abhishekdeshkar:
I have had hard time configuring database collection & charset by going to config/database.php file. It will be better if we add two environment variables for database charset as DB_CHARSET and database collection as DB_COLLECTION.


taylorotwell:
You can add the environment variables to your own application if you want to. You control that code.

joshmanders:
> I have had hard time configuring database collection & charset by going to config/database.php file. 

How is that hard?

PR #4267, opened 2017-05-14
Fix erroneous string size for email field

xdega:
This commit addresses an issue with the email field size being too big for an index in MySQL, thus causing the current default migrations to fail every time.

See this reference: https://dev.mysql.com/doc/refman/5.6/en/innodb-restrictions.html
The limit is 767 bytes. With MySQL assuming 3 bytes per character, this comes to 256 characters rounded up.

By simply specifying the length of the field, this issue is avoided.

### DIFF
--- a/database/migrations/2014_10_12_000000_create_users_table.php
+++ b/database/migrations/2014_10_12_000000_create_users_table.php
@@ -16,7 +16,7 @@ class CreateUsersTable extends Migration
         Schema::create('users', function (Blueprint $table) {
             $table->increments('id');
             $table->string('name');
-            $table->string('email')->unique();
+            $table->string('email', 256)->unique();
             $table->string('password');
             $table->rememberToken();
             $table->timestamps();
--- a/database/migrations/2014_10_12_100000_create_password_resets_table.php
+++ b/database/migrations/2014_10_12_100000_create_password_resets_table.php
@@ -14,7 +14,7 @@ class CreatePasswordResetsTable extends Migration
     public function up()
     {
         Schema::create('password_resets', function (Blueprint $table) {
-            $table->string('email')->index();
+            $table->string('email', 256)->index();
             $table->string('token');
             $table->timestamp('created_at')->nullable();
         });


lagbox:
>"If you are running a version of MySQL older than the 5.7.7 release or MariaDB older than the 10.2.2 release, you may need to manually configure the default string length generated by migrations in order for MySQL to create indexes for them."

[Laravel 5.4 - Database - Migrations](https://laravel.com/docs/5.4/migrations#indexes)

This?

xdega:
@lagbox Ok. But for the sake of a few minor insertions, why wouldn't you just fix the issue instead of requiring a workaround for a, no doubt, significantly large portion of users?

Besides, isn't it a generally better practice to specify your field lengths?

xdega:
For the sake of clarity. This fix is targeted at repairing the default migrations that come in a fresh install. Users should be able to run these migrations out of the box, as long as they satisfy the requirements listed here: https://laravel.com/docs/5.4/installation#server-requirements

This PR addresses that, makes it possible where it wasn't before.

lagbox:
Its very simple. Laravel supports more than 1 database out of the box. This 'issue' (which is a MySQL issue solely) affects only people using an 'older' version of MySQL. I don't even use MySQL. Why should there be a change that is intended to fix 1 specific issue with 1 specific database that will affect everyone regardless?

Its already documented for people who choose not to use newer version of their database how to get around the problem with their database.

xdega:
@lagbox what field length do you, a non-MySQL user, need for an email field? If it is 256 characters or less, how does this negatively affect you?

xdega:
Also. This limit still applies to the 'current' version of MySQL: https://dev.mysql.com/doc/refman/5.7/en/innodb-restrictions.html

lagbox:
>"If you are running a version of MySQL older than the 5.7.7"

If you were talking about 5.7.0 yes, but that isn't the current version of 5.7. Version 5.7.7 changes this, that is why the Laravel docs point out that version number.

>"Alternatively, you may enable the innodb_large_prefix option for your database. Refer to your database's documentation for instructions on how to properly enable this option." - Laravel docs.

This is what 5.7.7 does. 

https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-7.html#mysqld-5-7-7-feature

lagbox:
No change needs to be made to any migrations. This is something to do with you and your database, again. There are numerous ways to correct it as its a problem with your database.

xdega:
@lagbox Fair enough. Point taken. This isn't my issue personally, as I was addressing it for someone else. 😛

lagbox:
No worries, its good to bring it up :-). I'm sure it catches people off guard. Not everyone makes it through all the docs and will stumble into it as there are lots of people still on MySQL 5.6.

Perhaps it is something that could be discussed again, but it seems the default 255 has worked out well, besides this MySQL issue with older versions due to the default charset change in Laravel 5.4.

xdega:
If I may. I am going to reopen this. As you suggested, it may be useful to prevent duplicate PRs. That is unless it is @taylorotwell 's goal to keep the PR section "clean". I'll let the "boss man" make the final call on closure.

taylorotwell:
I agree with prior commenters on this.

PR #4269, opened 2017-05-15
Return json if sucessfull login expects it

Quezler:


Quezler:
Sibling of laravel/framework#19212

taylorotwell:
No plans to make this change.

Quezler:
@taylorotwell might i ask why? all login responses except success have json responses.

PR #4274, opened 2017-05-17
Added redirectTo property to RedirectIfAuthenticated middleware.

CJEdgerton:
I needed to change the redirect location in `RedirectIfAuthenticated` so I thought I would add a `$redirectTo` property like in `LoginController`, `RegisterController`, and `ResetPasswordController`. If this update is appropriate I can also update [the docs](https://laravel.com/docs/5.3/authentication#included-authenticating) before the pr is merged.

### DIFF
--- a/app/Http/Middleware/RedirectIfAuthenticated.php
+++ b/app/Http/Middleware/RedirectIfAuthenticated.php
@@ -7,6 +7,14 @@ use Illuminate\Support\Facades\Auth;
 
 class RedirectIfAuthenticated
 {
+
+    /**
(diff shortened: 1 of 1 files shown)


lagbox:
It doesn't seem like there is a need to define a variable as you have the ability to adjust that path yourself in the code directly.

taylorotwell:
As noted, you can change this code already. You own it.

CJEdgerton:
Ok, sounds good. Maybe adding something to the docs would be useful since that functionality in the auth controllers is documented though.

PR #4276, opened 2017-05-25
Call currently used php binary in composer instead of php located in $PATH.

dragoniscz:
Current composer.json needs path to php based in system $PATH variable. Better way is use `@php` ([Scripts - Composer](https://getcomposer.org/doc/articles/scripts.md#executing-php-scripts)) that will use the same php binary as for composer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,18 +30,18 @@
     },
     "scripts": {
         "post-root-package-install": [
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks, but please send to master. 5.3 is only receiving security fixes. :)

PR #4277, opened 2017-05-25
Call currently used php binary instead of php located in $PATH.

dragoniscz:
Current composer.json needs path to php based in system $PATH variable. Better way is use `@php` ([Scripts - Composer](https://getcomposer.org/doc/articles/scripts.md#executing-php-scripts)) that will use the same php binary as for composer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,18 +29,18 @@
     },
     "scripts": {
         "post-root-package-install": [
-            "php -r \"file_exists('.env') || copy('.env.example', '.env');\""
+            "@php -r \"file_exists('.env') || copy('.env.example', '.env');\""
         ],
         "post-create-project-cmd": [
-            "php artisan key:generate"
+            "@php artisan key:generate"
         ],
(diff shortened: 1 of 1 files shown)


laurencei:
This was a previously reverted PR: https://github.com/laravel/laravel/pull/4126

The problem is it changes the minimum composer requirement to 1.3 (when that feature was introduced).

At the time that version had only been out for 4 weeks. It's been almost 6 months now - so might be ok - but worth noting it is a change in requirements.

GrahamCampbell:
Maybe worth only changing on development branch then?

taylorotwell:
You can change it in your own file if you want. You own the file.

taylorotwell:
Sure it's fine if we change it on development branch.

dragoniscz:
There is problem, that if you use "composer create-project", the command fails because no php is found. And "composer create-project" is one of preferred way to install laravel.

PR #4278, opened 2017-05-25
Use current PHP version instead of using PATH

m1guelpf:
See #4277

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,18 +30,18 @@
     },
     "scripts": {
         "post-root-package-install": [
-            "php -r \"file_exists('.env') || copy('.env.example', '.env');\""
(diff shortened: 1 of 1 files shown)


krwu:
This leads to an error:

```bash
$ composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Nothing to install or update
> Illuminate\Foundation\ComposerScripts::postAutoloadDump
You made a reference to a non-existent script @php artisan package:discover
```
The last line:
**You made a reference to a non-existent script @php artisan package:discover**

GrahamCampbell:
You need to use the latest version of composer.

krwu:
@GrahamCampbell 
It works now. Thank you.

PR #4283, opened 2017-06-01
Simplify mix require

JeffreyWay:
Note that I left your dang `const` as it is, even though it's stupid. `let` for life. 💃💃💃💃

### DIFF
--- a/webpack.mix.js
+++ b/webpack.mix.js
@@ -1,4 +1,4 @@
-const mix = require('laravel-mix');
+let mix = require('laravel-mix');
 
 /*
  |--------------------------------------------------------------------------
--- a/webpack.mix.js
+++ b/webpack.mix.js
@@ -1,4 +1,4 @@
-const { mix } = require('laravel-mix');
+const mix = require('laravel-mix');
 
 /*
  |--------------------------------------------------------------------------


joshmanders:
Yo I'mma `let` you PR this, but `const` is the best keyword of all time. Of. All. Time.

taylorotwell:
Thanks <3

xargr:
@taylorotwell i think const is better solution than let becuse require is immutable.. perhaps...

JeffreyWay:
const doesn't mean immutable. None of this matters even remotely.

joshmanders:
`const` is visual debt, @xargr. Get out of our faces with that shizzle!

xargr:
constant cannot change through re-assignment, and it can't be redeclared... i have saw over 10 tutorial for js framework and all developer uses const in require statement

JeffreyWay:
I don't like `const` because it makes you assume that the value is immutable, when it's not. 

https://www.dropbox.com/s/gnrz1r4nu0nkg5b/Screenshot%202017-07-19%2011.52.55.png?dl=0

It makes zero difference in this file. `var` would be equally fine.

matthewlilley:
However, it does not allow the rebinding of that variable. Where as let does. 

The question here is "Does mix ever need to be rebound?"

JeffreyWay:
I think "slight immutability" is an oxymoron. :)

I use `let` for everything. Anyways, like I said it doesn't matter here. Zero people will reassign the `mix` variable.

matthewlilley:
I think const should be default if no rebinding if needed. And aside from that, It's good to get into a habit of using the most appropriate variable keyword, more importantly, distinguishing the two.

joshmanders:
I think bike-shedding is not really ideal, and if you don't want to use `let` just update it in your own `webpack.mix.js` file and be done with it.

PR #4284, opened 2017-06-04
[5.5] Always respond with json on api routes

mateusjatenee:
It just automatically adds the `accept application/json` header to any request targeting an api endpoint. That's one thing I usually do on my projects and I thought it might be an interest addition


denjaland:
Personally I don't like request headers to be modified by the server at all; it will override the only reality.  Can't you write middleware that always returns json rather than forcing it in this - for me - unnatural way?

mateusjatenee:
@denjaland yes, I was thinking that exactly. This PR is just an idea, of course.

taylorotwell:
No plans to add this out of the box.

PR #4285, opened 2017-06-06
Update welcome.blade.php

tamjidmurtoza:
please update this old type of doctype


bluegr:
The doctype is not case sensitive:

https://stackoverflow.com/questions/7020961/uppercase-or-lowercase-doctype

Thus there's no "old" or "new" way to write it

PR #4286, opened 2017-06-06
Update welcome.blade.php

tamjidmurtoza:
Please update font comment for easy to understand.

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -7,7 +7,7 @@
 
         <title>Laravel</title>
 
-        <!-- Fonts -->
+        <!-- Google Fonts -->
         <link href="https://fonts.googleapis.com/css?family=Raleway:100,600" rel="stylesheet" type="text/css">
 
         <!-- Styles -->


MatthewSH:
This isn't really all that necessary.

Mechtecs:
The fonts means fonts in general and is not limited to "Google Fonts". It should be kept as it is now.

ConnorVG:
A change like this will just make other people PR stuff like:
```diff
diff --git a/resources/views/welcome.blade.php b/resources/views/welcome.blade.php
index 606e59e7e3..51161200b4 100644
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -8,8 +4,8 @@
 
         <title>Laravel</title>
 
-        <!-- Styles -->
+        <!-- CSS Styles -->
         <style>
```

#nothankyou

MatthewSH:
@ConnorVG is absolutely correct. There is no point to specify or expand on the comments that are there. It's simple and straightforward enough. There is no need for further explanation. 

Which begs the question, why am I making this secondary comment if what's needing to be explained has already been explained?

PR #4290, opened 2017-06-13
Add Models namespace, move User.php to Models namespace, add base Model

thehiredgun:


overtrue:
👎 You can make this change in your own application if you want to. You control that code.

thehiredgun:
Yes - Meant to merge into my own fork. Sorry!

decadence:
By the way it isn't bad change at all because it's needed almost everytime.
Though it wouldn't be merged I'm sure.

thehiredgun:
@decadence Agreed! Models deserve their own namespace - putting them within App\ makes no sense.

troccoli:
+1

On 13 Jun 2017 2:09 pm, "Nick Wakeman" <notifications@github.com> wrote:

@decadence <https://github.com/decadence> Agreed! Models deserve their own
namespace - putting them within App\ makes no sense.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/laravel/laravel/pull/4290#issuecomment-308109705>, or mute
the thread

a-komarev:
The only thing I'm against - making abstract model. But moving models to `App\Models` namespace will be great for me. Doing it in 100% of my apps.

m1guelpf:
Opened https://github.com/laravel/internals/issues/637

decadence:
@m1guelpf anyway it can be changed only in form of PR.

connectkushal:
https://github.com/laravel/laravel/pull/3821#issuecomment-228748576

This seems like a nice workaround.

jordanmiguel:
I agree with it. Models deserve their own space.

zaknes:
I think they should be in there by default and a simple command can be used to move it and update the code everywhere else.

And of course, the `artisan make:model` should have an option for it to be placed into the `App\` directory.

okanlawonema:
Am very sure this will be declined, cause model means different thing in
different context, you can always do that yourself, but if we can use
artisan command to set custom name of Models namespace, like we do for the
App namespace

jordanmiguel:
@okanlawonema exactly. Maybe something like "php artisan app:models App/Models"

PR #4291, opened 2017-06-15
faker documentaion

tejomayonline:
Please add faker factory based documentation with all possible data types ,combinations,api so that anybody can create datas they want.Laravel  db seeding with faker is an awesome feature but lags in doc.


scoutman57:
This information is already well documented at the faker repository. https://github.com/fzaninotto/Faker

There is absolutely no need for it to be documented in Laravel. That is like asking for Laravel to document every function of the PHP language.

PR #4292, opened 2017-06-20
Update .htaccess

mxvin:
An error 403 Forbidden may possible, because FollowSymLinks Options is not declare/enabled.


taylorotwell:
The .htaccess lives in your application. You can customize it however you want.

PR #4293, opened 2017-06-21
[5.5] Let unauthenticated handle custom exception message and code.

lucasmichot:
We might throw a custom exception inheriting `\Illuminate\Auth\AuthenticationException`

---

This PR lets unauthenticated method handle custom exception message and code.


taylorotwell:
People can customize this as needed since its in the app repo.

PR #4295, opened 2017-06-21
PHPUnit testing named routes not working

alex-petrea:
So I'm running the following test which does a post request to the login controller.
public function test_login_post()
    {
        $response = $this->post('/login', [
            'email' => $this->user->email,
            'password' => $this->user->password,
            '_token' => csrf_token()
        ]);
        dd($response);
    }

Once the login is successful then the user should get redirected to the /app/invoices

If I set the /app/invoices route like this and run the test
Route::get('/app/invoices', 'InvoicesController@index')->name('account.invoices');
I get an error in the response that route('account.invoices') was not found.

If I name the route this way:
Route::get('/', [
        'as' => 'account.invoices',
        'uses' => 'InvoicesController@index'
    ]);
I get no errors.

I don't know if this is a bug or I'm doing something wrong


edenreich:
@alex-petrea hey, I've tried to replicate this, but I need some more info. I get 200 for each request with phpunit.

I have noticed that in your question you specify different paths. the first one you are testing goes to /app/invoices and the second one is only / is that exacly what you were testing or did you test:

Route::get('/app/invoices', [
'as' => 'account.invoices',
'uses' => 'InvoicesController@index'
]);

and you get error no found?

alex-petrea:
@edenreich I mistyped it. I was testing

Route::get('/app/invoices', [
'as' => 'account.invoices',
'uses' => 'InvoicesController@index'
]);

but never mind as I thought the name() function could be used instead of 'as' so I could keep the code shorter but I just realised it's used for the controller action ... silly me

PR #4296, opened 2017-06-23
Use null coalescing operator in UserFactory

chabib:
Since the next version of Laravel will require PHP 7.0+, I suggest to start using this operator


chabib:
Sorry, seems like I merge to wrong branch. It's my first time contributing

PR #4297, opened 2017-06-23
Use null coalescing operator

chabib:
Since the next version of Laravel will require PHP 7.0+, I suggest to start using this operator

### DIFF
--- a/database/factories/UserFactory.php
+++ b/database/factories/UserFactory.php
@@ -19,7 +19,7 @@ $factory->define(App\User::class, function (Faker $faker) {
     return [
         'name' => $faker->name,
         'email' => $faker->unique()->safeEmail,
-        'password' => $password ?: $password = bcrypt('secret'),
+        'password' => $password ?? bcrypt('secret'),
         'remember_token' => str_random(10),
     ];
 });


database/factories/UserFactory.php
```diff
@@ -19,7 +19,7 @@
     return [
         'name' => $faker->name,
         'email' => $faker->unique()->safeEmail,
-        'password' => $password ?: $password = bcrypt('secret'),
+        'password' => $password ?? bcrypt('secret'),
```
mateusjatenee: `$password ?? $password = bcrypt('secret');`, right?

chabib: I just realized that this line is related to performance regarding bcrypt usage after reading [this](https://stackoverflow.com/questions/44037895/how-does-the-static-password-work-in-the-default-laravel-user-factory). It is pointed out in the comment that this is obscure. What if the password assignment moved out from returned array with some explanation comment?
Something like this:

```
$factory->define(App\User::class, function (Faker $faker) {
    /*
    / By using a static variable for password, we only have to call
    / bcrypt(which is expensive) once, and then use that same hash value
    / on every subsequent call to the factory.
    */
    static $password;

    $password = $password ?? bcrypt('secret');

    return [
        'name' => $faker->name,
        'email' => $faker->unique()->safeEmail,
        'password' => $password,
        'remember_token' => str_random(10),
    ];
});
```

chabib: And why i suggest null coalescing operator over elvis operator is that null coalescing operator only check if variabel is null, and doesnt check its value according to [this](https://stackoverflow.com/questions/34571330/php-ternary-operator-vs-null-coalescing-operator/34571460#34571460)

PR #4298, opened 2017-06-23
Use null coalescing operator in UserFactory

chabib:
Since the next version of Laravel will require PHP 7.0+, I suggest to use this operator that specified for only checking if variable is null. Compared to elvis operator that also check variable's value that casted to boolean, which is not necessary in this case.
[Reference](https://stackoverflow.com/a/34571460/4306661)


GrahamCampbell:
Why do we need this? The current code is already fine?

taylorotwell:
Current code works.

PR #4304, opened 2017-06-29
:wrench: Reduce discoverability of session cookie name.

coderabbi:
Derives session.cookie from `SESSION_COOKIE`, falling back to `APP_NAME . '_session'`, falling back to `'laravel_session'` (current) in order to make it less discoverable, thereby (slightly) reducing threat vector.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -161,7 +161,7 @@ return [
     |
     */
 
-    'secure' => env('SESSION_SECURE_COOKIE', false),
+    'secure' => env('SESSION_SECURE_COOKIE', ! env('APP_ENV') == 'local'),
 
     /*
     |--------------------------------------------------------------------------
--- a/config/session.php
+++ b/config/session.php
@@ -122,7 +122,7 @@ return [
     |
     */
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks for the pr. Should definitely go to develop branch though. :)

coderabbi:
Done.

brunogaspar:
What if the `APP_NAME` contains spaces?

coderabbi:
Good call @brunogaspar!

config/session.php
```diff
@@ -161,7 +161,7 @@
     |
     */
 
-    'secure' => env('SESSION_SECURE_COOKIE', false),
+    'secure' => env('SESSION_SECURE_COOKIE', ! env('APP_ENV') == 'local'),
```
GrahamCampbell: 👎

coderabbi: Yeah, pollution from a different pull request. This one has been resubmitted against <develop>, that one is forthcoming, you can thumbs down it there (though analysis would be nice...). 😄

PR #4305, opened 2017-06-29
:wrench: Reduce discoverability of session cookie name.

coderabbi:
Derives session.cookie from SESSION_COOKIE, falling back to (snake_cased) APP_NAME . '_session', falling back to 'laravel_session' (current) in order to make it less discoverable, thereby (slightly) reducing threat vector.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -122,7 +122,7 @@ return [
     |
     */
 
-    'cookie' => env('SESSION_COOKIE', snake_case(env('APP_NAME', 'laravel')).'_session'),
+    'cookie' => env('SESSION_COOKIE', str_slug(env('APP_NAME', 'laravel'), '_').'_session'),
 
     /*
     |--------------------------------------------------------------------------
--- a/config/session.php
+++ b/config/session.php
@@ -122,7 +122,7 @@ return [
     |
     */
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Could any characters be added to the app name which would in turn make this cookie name invalid (spaces, special characters, etc.?)

coderabbi:
In practice, no. 

https://stackoverflow.com/questions/1969232/allowed-characters-in-cookies 

Though perhaps it would be "good citizenship" to limit to those in RFC 6265 (http://www.ietf.org/rfc/rfc6265.txt), i.e. all alphanumeric plus !#$%&'*+-.^_`|~ are fair game.
 
I can strip all but those out, if you'd like, or not, as either approach is technically valid.

GrahamCampbell:
Don't we have a slug function that does that?

coderabbi:
`Str::slug()` is a bit more stringent that even RFC 6265 requires, but that would certainly work, yes (albeit with the interesting side effect that @ gets converted to 'at').

coderabbi:
Updated it to use str_slug, so is compliant with RFC 6265.

taylorotwell:
Seems OK, Thanks.

PR #4306, opened 2017-06-29
:wrench: Defaul session.secure to true when APP_ENV != 'local'.

coderabbi:
The world is HTTPS. Forge makes HTTPS a snap. Why aren't we encouraging HTTPS?

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -161,7 +161,7 @@ return [
     |
     */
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
👎  Would opt to just default this to `true` on develop, and don't do anything else.

taylorotwell:
People can change this to whatever value they wish.

PR #4307, opened 2017-06-29
Use config function instead of env function

DojoGeekRA:
`app.name` already uses `Laravel` as a default value in case of the `APP_NAME` environment variable is not set. This allows to remove some duplication.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -124,7 +124,7 @@ return [
 
     'cookie' => env(
         'SESSION_COOKIE',
-        str_slug(env('APP_NAME', 'laravel'), '_').'_session'
+        str_slug(config('app.name'), '_').'_session'
     ),
 
     /*


GrahamCampbell:
Unfortunately, the config function is not available while config itself is being loaded. :)

DojoGeekRA:
@GrahamCampbell thanks for the feedback!

I actually tried this before submitting the PR, i have the following config:

captura de pantalla 2017-06-29 a la s 14 44 32

and then i hitted tinker to fetch the value:

captura de pantalla 2017-06-29 a la s 14 44 04

Actually i even tried with this with the config cached to see if some problem can emerge:

captura de pantalla 2017-06-29 a la s 14 47 26

Am i missing something?

GrahamCampbell:
What about just clearing the cache?

DojoGeekRA:
Yup, still works:

captura de pantalla 2017-06-29 a la s 14 56 31

GrahamCampbell:
I'm sure there are issues with calling config within config files. Ping @taylorotwell.

rappasoft:
I think the reason you shouldn't use config calls in other config files is you can not be sure that the config file you're calling to is going to be loaded before the config file you're calling from.

DojoGeekRA:
@GrahamCampbell you're right, now i have the same feeling. I think that more importantly that this being merged, is to understand how this works and maybe this can serve as precedent to PRs like this in the future 😃  (to accept them or reject them right away).

@rappasoft that's a good theory, let's dig a bit further and confirm that 👍

thecrypticace:
The config files would be loaded in the order defined returned by the filesystem which may not be alphabetical. `app.php` could be available at the time but it also might not. You can't guarantee it unless you sort them before loading. And even if you do that you still can't reference something in a later config file that hasn't yet been processed.

DojoGeekRA:
@thecrypticace great to know, thanks!

PR #4310, opened 2017-07-02
Use the new route view helper

m1guelpf:
See https://github.com/laravel/framework/pull/19835.


taylorotwell:
Would prefer to leave this as a regular route because I often use it to test various things.

PR #4312, opened 2017-07-03
[5.5] Enable double rate limit.

lucasmichot:
Enable double rate limit, introduced in https://github.com/laravel/framework/pull/19807


taylorotwell:
I'll let users adjust this as needed.

PR #4323, opened 2017-07-05
Sort packages by name

mayoz:


zlodes:
Are u serious? :D

a-komarev:
@zlodes why not? Look in the composer.json config section:
```
"config": {
    "sort-packages": true,
}
```

a-komarev:
But... PHP isn't sorted... Maybe because it's not a package ^_^

PR #4324, opened 2017-07-05
Add Global "From" Address

ahmed-aliraqi:
Add global email from address to .env.example file


overtrue:
👎

yljphp:
👎

GrahamCampbell:
No need for all the 👎. All that needs explaining is that you are free to modify the application config file. The env file is just there for things that always change between environments, or secrets. :)

lucasmichot:
two-thumbs-down

manogi:
While I don't think this warrants changing the .env.example file: I almost always have some mail addresses in the env files because I want to have different recipients / senders locally for testing than in production. So I think yes, the .env file is a good place for this (when you don't have these addresses somewhere else, like your DB), but no, we don't have to have this in .env.example (too specific, the needs and names will be different across applications).

Also: just a thumb down with no comment is harsh, no time for writing one sentence explaining why you are thumb-downing?

LasseRafn:
I don't think I'd need this, however I see the reason for your suggestion 👍 It does make sense to have it in your env, as its unlikely you'll be sending from same mail locally and production. 

However, I guess most people use mailtrap or similar, so locally/dev the from does not matter.

I'd be fine with this being merged (won't make anything worse)

LasseRafn:
Locale especially is a thing I actually feel **SHOULD** be environment-specific, for multi-language applications (its far easier to use different installations with a .env file each) than changing language in the application (Try doing queues with multiple languages and ONE application; it can be done but its rarely nice to do)

medozeus:
15941350_1922760991288008_7527559961410518668_n

Sheyma83:
قصف بالجبهة 😂😂😂😂

RryLee:
👎

PR #4325, opened 2017-07-05
Use APP_TIMEZONE environment variablee in application config

vladimir-yuldashev:
It will be handy if default `config/app.php` will have `APP_TIMEZONE` environment variable.


joshmanders:
You can add this to your own project.

decadence:
Every PR's change can be added to your own project.

joshmanders:
We don't need every config variable available as an env variable because of one or two people's usecase of wanting completely different database collation or using different timezones.

The config is meant to be altered by you, it's not hard to add the env in stuff you want to change. I personally only change what actually needs changed in different environments and it's a hassle having to remove everyone's edge use cases and change back to what actually matters.

PR #4326, opened 2017-07-07
Using env variable for Redis database number

v-Stein:
Update .env and database config to use an environment variable for Redis database number.

Avoids conflicts when using multiple Laravel apps on a single Redis server


overtrue:
👎

joshmanders:
This doesn't need PR'd in, you can alter it in your own app.

v-Stein:
I did it, but I assume this should be set by default for a new Laravel app.

joshmanders:
It is set by default, the value is `0`.

What I am saying is not everything needs to be configurable in `.env` that's what `config/` directory is for.

v-Stein:
Every present option for a Redis server is set in .env by default, except the database index number.
Any explanation?

taylorotwell:
It's less common for that to be environment specific.

PR #4327, opened 2017-07-10
The app.url is used also on e-mail templates.

geodeveloper:
The config `app.url` is also used here:

`Illuminate/Auth/Notifications/ResetPassword.php`
`lluminate/Mail/resources/views/html/message.blade.php`
`lluminate/Mail/resources/views/markdown/message.blade.php`

I think a bit of clarification would be good.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -49,6 +49,8 @@ return [
     | the Artisan command line tool. You should set this to the root of
     | your application so that it is used when running Artisan tasks.
     |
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think it could be use a lot of places in the future. I don't want to name each one for now.

geodeveloper:
Good to know. I hope it'll be updated this section in near future then. It would be useful for new "laravels".

PR #4328, opened 2017-07-11
5.3 Is it impossible add access to change created_at?

kindslayer:
hello, I need to update created_at time to today ! because I send sms for posts created on this month and need to change update created_at of special post. how can i do it


hrodrigues1984:
Have you tried?
https://laravel.com/docs/5.3/eloquent-mutators#defining-a-mutator

PR #4330, opened 2017-07-12
delete not using classes

shindakioku:
AuthServiceProvider and EventServiceProvider include classes and they don't using their. I deleted it.


brayniverse:
They're there for convenience. There have been rejected pull requests to remove this before.

PR #4331, opened 2017-07-14
DB_SCHEMA environment variable for PostgreSQL

srmklive:
Currently there is no option to override schema value when using postgres. We can override the username, password etc through .env, but can't override the schema value. This PR adds the `DB_SCHEMA` environment variable through which we can override the default schema attribute value of postgres configuration.


taylorotwell:
You can make this an environment variable in your own app if you wish.

PR #4332, opened 2017-07-14
Warn about storage and bootstrap/cache file permissioning

MichaelHoughton:
At the moment, when the folders:

`storage` and `bootstrap/cache` permissions are not set, Laravel simply shows a blank white screen.

This is confusing for new comers and can cause issues when deploying an application.

The following PR is a potential fix for this - but I can appreciate that there might be a better place to put this than in the public/index.php file.

This simply warns users about the error and prompts them to fix the permissions - rather than seeing the blank screen which happens currently.

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -37,6 +37,26 @@ $app = require_once __DIR__.'/../bootstrap/app.php';
 
 /*
 |--------------------------------------------------------------------------
+| Lets make sure the writable files have their permissions set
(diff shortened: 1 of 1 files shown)


taylorotwell:
No plans to add this. We recently updated the Laravel installer to set the permissions for you and warn you if they aren't correct.

decadence:
But installer doesn't handle this when we clone existing project.

PR #4333, opened 2017-07-15
Fixed CI

DCzajkowski:
Corrected indentation and prepended a class name with a forward slash.

P.S. What is the reason for indenting three times instead of once?


GrahamCampbell:
I'd prefer to just use if statements than multi-multiline ternary expresions. :P

PR #4339, opened 2017-07-21
Temporary Filesystem

luizguilhermefr:
Often, we must read files temporarily and then discard them.

This PR proposes to include a new default disk, the **temporary** disk.

The default would be the PHP's [built-in](http://php.net/manual/en/function.sys-get-temp-dir.php) `sys_get_temp_dir()` function, but the user would have liberty to use another location using env `TMP_DIR` key.


taylorotwell:
I think anyone can add this to their config if needed.

PR #4342, opened 2017-07-23
Fixed trailing slash redirection for subdirectory installs.

djh101:
Previously redirection to remove trailing slashes would fail if Laravel was not installed in the root directory.


djh101:
Accidently pull requested to master.

PR #4345, opened 2017-07-23
- making the application name dynamic in the default view layout

lukeparkinson:
- Making the application name dynamic in the default view layout, as it is in the auth stub files. This is currently hard coded.


taylorotwell:
This view isn't intended to be kept so will leave as-is for now. Thanks!

PR #4352, opened 2017-07-27
Check if jQuery already exists

marianvlad:


taylorotwell:
I think that's the point of the try / catch.

PR #4354, opened 2017-07-29
Ignore nbprojects.

lloricode:
Signed-off-by: Lloric Mayuga Garcia <emorickfighter@gmail.com>


a-komarev:
Developers should ignore IDE related temporary files and folders globally.

PR #4357, opened 2017-08-01
Forth mandatory parameter (cluster) for Pusher service added.

Tpojka:


GrahamCampbell:
Thanks. Where is the documentation for this now being required please? Surely this is a breaking change?

Tpojka:
It is said [here](https://github.com/pusher/pusher-http-php#pusher-constructor). Rule is that four parameters are being passed to constructor, except `us-east-1` is [default if nothing passed](https://pusher.com/docs/clusters#what-is-a-cluster). I left double fall back to that one: null in `.env.example` and second argument in `config/broadcasting.php` file.

arcdigital:
It doesn't look like this is required, the constructor says the options array is optional. If not passed, it defaults to an empty array, which is what Laravel also passes through.

https://github.com/pusher/pusher-http-php/blob/master/src/Pusher.php#L28

GrahamCampbell:
Laravel will need pusher 2.0 I assume?

taylorotwell:
I'm fine with this being a var but the default should just be us-east-1 in the env example file.

Tpojka:
Actually, I made mistake: it should be `mt1` which is shortcut key used for `us-east-1`. Explanation can be seen in [docs](https://pusher.com/docs/clusters#on-the-server-side): 

> The `cluster` option is required if the Pusher app was created in a different cluster to `us-east-1`.

For example, when user registers Pusher application, it's needed to select cluster region. For me in Bosnia Pusher selected `eu` (assuming by IP). Anyone from Europe when see automatically selected `eu` would leave as that. And if you leave empty cluster option in laravel application, Pusher would calculate `mt1` region. While doing refactoring old code that used old Pusher library and trying to switch it to new pusher 3.0, I had problem until I set cluster in laravel app to be `eu` which is setting of registered Pusher app. 
All in all for using Pusher 3.0 version and leaving potential users to register their region cluster as Pusher settings, this is needed. 
But only need to be changed to `mt1` in `.env` as @taylorotwell asked.

taylorotwell:
I don't know what mt1 means.

Tpojka:
It's shortcode explained [here](https://pusher.com/docs/clusters#how-do-you-configure-the-cluster-option).

kn100:
This should be merged - Pusher has multiple 'clusters' or servers that users can select when creating their Pusher app. If the user doesn't specify the cluster here in the client library, Pusher will a assume that the cluster is 'mt1' which is one of Pushers clusters. This PR specifies the cluster explicitly, making it more obvious that it needs to be changed. If the user does not change the cluster and their app exists on a different cluster, for instance 'eu' - then their application will not work.

Could you look at this again?

PR #4359, opened 2017-08-02
fix user table migration problem, change utf8mb4 to utf8

afcn007:


overtrue:
WTF？

micos7:
utf8, unlike utf8mb4  doesn`t even support non latin based languages.

afcn007:
I have get error when I migrate user table on 5.4.30 version, mysql says the email column is over the limit 767bytes, after change the utf8mb4 to utf8, migrate goes well, what other solution?

ready4god2513:
```
SET GLOBAL innodb_file_format="Barracuda";
SET GLOBAL innodb_large_prefix="on";
SET GLOBAL innodb_file_per_table=true;
```

That's the solution @afcn007

PR #4362, opened 2017-08-03
Using space in APP_NAME surrounded by quotes fails Str:startsWith method

KevinHoughton:
I've come across this problem with one of my sites, using quotes in the .env files as described.

What it does is that the function Str::startsWith() returns false when it checks if APP_NAME starts with '"'. 

I have not found a solution yet for this problem.

.env
`APP_NAME="Laravel Test"`

config/app.php
```
/*
    | Application Name
    |
    | This value is the name of your application. This value is used when the
    | framework needs to place the application's name in a notification or
    | any other location as required by the application or its packages.
    */

    'name' => env('APP_NAME', 'Laravel'),
```

My own project's app.blade.php
```
<title>{{ config('app.name', 'Laravel') }}</title>
```

So now when you use this in the default app.blade.php for example as the title of the site. It returns the default: "Laravel" instead of the expected "Laravel Test"

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,4 @@
-APP_NAME=Laravel
+APP_NAME="Laravel Test"
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true


KevinHoughton:
Live example https://laravelcomponents.com/

.env on forge: 
schermafbeelding 2017-08-03 om 12 59 50

title being set in app.blade.php:
schermafbeelding 2017-08-03 om 13 00 49

title on live environment:
schermafbeelding 2017-08-03 om 13 00 13

baskan:
did you tried to clear config cache with `artisan config:clear` command?

KevinHoughton:
@baskan I did, nothing changed.

KevinHoughton:
Got it working, though I still find it weird that Str::startsWith($value, '"') returns false. Even if I put APP_NAME=""Laravel Test"". Why is this if there?

taylorotwell:
I don't think we want the default name to be "Laravel Test".

PR #4364, opened 2017-08-05
[5.5] add whoops to dev dependencies

tomirons:


ankurk91:
It is already here
https://github.com/laravel/framework/blob/master/composer.json#L77

GrahamCampbell:
> It is already here
https://github.com/laravel/framework/blob/master/composer.json#L77

Dev dependencies are not inherited.

ankurk91:
Oh, I completely forgot about that.
Why don't `laravel/framework` move it to `dependencies` ?

m1guelpf:
Because it's a dev dependency

PR #4365, opened 2017-08-07
Removed some unnecessary redundancy

guilhermesiani:


app/Console/Kernel.php
```diff
@@ -12,14 +12,12 @@ class Kernel extends ConsoleKernel
...
      */
-    protected $commands = [
-        //
-    ];
+    protected $commands = [];
 
     /**
      * Define the application's command schedule.
      *
-     * @param  \Illuminate\Console\Scheduling\Schedule  $schedule
```
GrahamCampbell: Our PHPDoc style is to use the full name, so that humans can immediate read it.

app/Console/Kernel.php
```diff
@@ -12,14 +12,12 @@ class Kernel extends ConsoleKernel
      *
      * @var array
      */
-    protected $commands = [
-        //
```
GrahamCampbell: This is there so people can easily add commands to it.

app/Http/Kernel.php
```diff
@@ -30,7 +30,6 @@ class Kernel extends HttpKernel
             \App\Http\Middleware\EncryptCookies::class,
             \Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse::class,
             \Illuminate\Session\Middleware\StartSession::class,
-            // \Illuminate\Session\Middleware\AuthenticateSession::class,
```
GrahamCampbell: This is there on purpose so it can be uncommented easily.

tests/Unit/ExampleTest.php
```diff
@@ -3,8 +3,6 @@
 namespace Tests\Unit;
 
 use Tests\TestCase;
-use Illuminate\Foundation\Testing\DatabaseMigrations;
```
GrahamCampbell: These are purposefully here, and should not be removed.

PR #4366, opened 2017-08-09
Show mail driver name for "mail" function usage

aik099:


taylorotwell:
I believe this was deprecated by SwiftMailer.

aik099:
Then we should update text above `Laravel supports both SMTP and PHP's "mail" function as drivers for the ...`, because PHP's "mail" function is supported now, but support will be eventually dropped by Swift mailer.

PR #4377, opened 2017-08-16
Create signIn method in TestCase class

alexandrebordiere:
This method is a "better version" of the actingAs method.

If no user is given, it will create a new one using the factory.

It's very handy because on some tests, you just want to be signed in, whatever the user is.

```php
/** @test */
public function an_authenticated_user_cannot_reach_the_login_page()
{
    // Without using the signIn method.

    $user = factory('App\User')->create();

    $response = $this->actingAs($user)
        ->get('login');

    $response->assertStatus(302);
}

/** @test */
public function an_authenticated_user_cannot_reach_the_login_page()
{
    // By using the signIn method

    $response = $this->signIn()
        ->get('login');

    $response->assertStatus(302);
}
```

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -12,14 +12,15 @@ abstract class TestCase extends BaseTestCase
      * Set the currently logged in user for the application.
      *
      * @param  mixed  $user
+     * @param string|null  $driver
      * @return $this
      */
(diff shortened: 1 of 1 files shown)


taylorotwell:
You can add this to your own application as needed.

PR #4378, opened 2017-08-16
Change route->get with route->view

zgabievi:
There is no need to use a function to return the simple view. We can you route view instead.


brunogaspar:
Duplicate of #4310

zgabievi:
Oops. Closing.

PR #4380, opened 2017-08-17
password field should not be fillable

SamAsEnd:
The `password` field is *hidden by default*, why should it be fillable too?

Developers should loosen up their mass assignment depending on the app requirement.

Laravel it self is giving a bad example!

Plus Why not!

P.S: sorry about the last pull request :disappointed:


taylorotwell:
Again, you can change this in your app if you want. The only danger I see here is someone messing up their own password and just hurting themselves... (not that this operation uses mass assignment anyways).

PR #4381, opened 2017-08-17
Using the new Route::view() helper

ConsoleTVs:
Since laravel 5.5 brings the new Route::view() helper, now the basic route that comes with laravel can be simplified to a more modern and clean approach.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -11,6 +11,4 @@
 |
 */
 
-Route::get('/', function () {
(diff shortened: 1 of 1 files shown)


zaknes:
Duplicate of #4310.

ConsoleTVs:
Oh whops! closing it then, thanks for pointing that out @zaknes

PR #4383, opened 2017-08-18
Fix in comment in config/app.php

scrubmx:
I know it is a small thing but it's driving me crazy!


zaknes:
Haha thanks for catching that.

PR #4384, opened 2017-08-18
Remove unnecesary if

m1guelpf:
Removes an unnecesary if and makes the syntax clearer.


hrodrigues1984:
Shouldn't this function return some kind of value?

decadence:
You missed return statement

GrahamCampbell:
It's much more readable before.

PR #4387, opened 2017-08-20
[5.5] Return glyphicons

balping:
I'm not sure if this is desired. I asked a question about it in [internals](https://github.com/laravel/internals/issues/745), but received no answer.

Quoting from there:

> Is there a reason why glyphicon fonts are not included in `public/fonts` by default?
> 
> They [used to be there](https://github.com/laravel/laravel/tree/5.0/public/fonts) in 5.0, but in later versions they have been excluded. ([This](https://github.com/laravel/laravel/commit/5e11f87de2c8f9303e6e296d74ec5f7f2093fd9b) was the particular commit of removal.)
> 
> Since JS and CSS mix builds are shipping with laravel, it would be logical to include these font files too. As they are produced by the default mix build.
> 
> But these might be intentionally excluded from `public/fonts/vendor/bootstrap-sass/bootstrap`.  If it is intentional indeed, then what's the reason behind it?


kamui545:
It's a 3rd party and should not be versioned.

Try doing this:

```
npm install
npm run dev
```

screen shot 2017-08-20 at 15 51 49

PR #4389, opened 2017-08-26
[5.5] Add sticky to config

laurencei:
I'm not actually 100% if we should add this or not....

Anyone who wants to add a read/write connection needs to modify the config for the `read` and `write` hosts. I have a PR for the docs for that section now - so people will see the new `sticky` option there.

So should `sticky` be in the default config? It would raise awareness of the option, but its not needed unless they modify the config anyway... I'm in two minds about it...

Happy if you dont accept this - I see benefits either way.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -52,6 +52,7 @@ return [
             'prefix' => '',
             'strict' => true,
             'engine' => null,
+            'sticky' => true,
         ],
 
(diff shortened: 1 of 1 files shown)


laurencei:
Actually - I'll change the docs to make it better and we dont need this.

PR #4390, opened 2017-08-27
Remove extra comma

ankurk91:


GrahamCampbell:
Personally, I'd rather actually see the Oxford comma added in all cases, rather than removed here.

(please don't make that change in this PR, tho, will need discussion first)

alicompu:
Clean code, Beautiful.

subotkevic:
Looks good. I have tested the changes. Seems to work.

ankurk91:
Some resources :

https://en.m.wikipedia.org/wiki/Serial_comma

https://english.stackexchange.com/questions/412/should-i-put-a-comma-before-the-last-item-in-a-list

michaeldyrynda:
The Oxford comma is used correctly in this instance.

PR #4392, opened 2017-08-29
Updates Unauthenticated JSON Exception Response

sixlive:
The default framework method for rendering this exception uses the `message` keyword, this should be consistent.

https://github.com/laravel/framework/blob/master/src/Illuminate/Foundation/Exceptions/Handler.php#L218

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,5 @@
 Homestead.json
 Homestead.yaml
 npm-debug.log
+yarn-error.log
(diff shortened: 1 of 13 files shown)


taylorotwell:
Can you resend this to develop? 👼

sixlive:
My bad, doesn't need it in develop, looks like it was removed in favor of foundation rendering. 💖

PR #4391, opened 2017-08-28
Update web.php

yappkahowe:
Make use of `Route::view` helper.

PR #4367, opened 2017-08-10
Add route view

Te7a-Houdini:

PR #4395, opened 2017-08-30
Update welcome route to use the new Route::view() method

mdavis1982:
Now that we have the new `Route::view()` method, should we update the Welcome route to use this since it's only function is to return a view?

Happy for this to be closed if it's not preferable to update it, but thought it might make it clear that this is now possible 😄

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -11,6 +11,4 @@
 |
 */
 
-Route::get('/', function () {
-    return view('welcome');
-});
+Route::view('/', 'welcome');


decadence:
Here we go again
https://github.com/laravel/laravel/pull/4391

MehediDracula:
Duplicate of

#4310
#4367 
#4378 
#4381 
#4386

MehediDracula:
Lots of PR for this :laughing:

mdavis1982:
Hadn't seen the other ones! I'll close this 😄

joshmanders:
Can we get a `CONTRIBUTING.md` file added so that a notice that this is not needed can be added... We're at 50+ PR's for this.

PR #4399, opened 2017-08-30
Fixed composer's invalid argument exception

mmkjony:
Fixing a conflict with the latest composer while installing a new laravel 5.5.0 project.
Solving composer's exeption: 
"[InvalidArgumentException]
  Script "post-install-cmd" is not defined in this package"

And, added a script to show the installed laravel version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,6 +46,9 @@
         "post-autoload-dump": [
             "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
             "@php artisan package:discover"
+        ],
+        "post-install-cmd": [
+            "@php artisan --version"
         ]
     },
     "config": {


a-komarev:
Great idea about version!

laravelish:
i dont have that error

```
Running composer as root/super user is highly discouraged as packages, plugins and scripts cannot always be trusted
You made a reference to a non-existent script @php -r "file_exists('.env') || copy('.env.example', '.env');"
Running composer as root/super user is highly discouraged as packages, plugins and scripts cannot always be trusted
You made a reference to a non-existent script @php artisan key:generate
Running composer as root/super user is highly discouraged as packages, plugins and scripts cannot always be trusted
> Illuminate\Foundation\ComposerScripts::postAutoloadDump
You made a reference to a non-existent script @php artisan package:discover
Application ready! Build something amazing.
```

only the env file is not generated

taylorotwell:
You just need to update your Laravel installer, as noted in release notes.

gayansanjeewa:
I just got the same Issue. Thanks @taylorotwell for your comment. I forgot to do a `composer global update` before the `laravel new <project-name>`. Once I did that, It's all good. Long live L5.5 ;)

a-komarev:
@laravelish https://getcomposer.org/doc/articles/scripts.md#executing-php-scripts

laravelish:
@a-komarev thanks for the info, i had installed composer through apt-get which is using the version 1.1.3
i did not know it was using a old version.. time to install it from https://getcomposer.org/installer

jo2k:
I was getting that exception error until I updated Composer with `composer global update`. Now I can do `laravel new blog` and it works without errors, all the way, including setting the application key.

It would be nice if the Laravel installer's message was a bit more specific,  especially for us newbies to Laravel and Composer.

imkane:
This is happening to me with Laravel 6.2. I did "composer global update" too, but that didn't help :(

PR #4398, opened 2017-08-30
Use Route::view

vyuldashev:

PR #4400, opened 2017-08-31
Update web.php

Artiik-fr:
Use the new Route::view method from Laravel 5.5

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -11,6 +11,4 @@
 |
 */
 
-Route::get('/', function () {
-    return view('welcome');
-});
+Route::view('/', 'welcome');


fletch3555:
There have literally been 50 PRs for this already, and they have all been declined so far.  I assure you, yours will be no different.

a-komarev:
I will just leave it here: #4398 #4396 #4391 #4386 #4381

Artiik-fr:
Okey

a-komarev:
It will be easier to add view helper in routes and keep one more route with closure for tests to stop this types of PRs in future :}

ghazanfarmir:
@a-komarev and then expect even more PRs to remove one of the two :)

a-komarev:
@ghazanfarmir this will happen with possiblity of 0.0000001% and now we are getting PRs with `Route::view` each day.

PR #4402, opened 2017-09-01
Redirect using intended url after successful login

ericdowell:
When a user is found to be a guest the full url is stored in the session key `url.intended` before they're sent to the login page. Once successfully authenticated we should respect redirecting the user to `url.intended`, otherwise use default `/home`.


taylorotwell:
This could create redirect loops.

PR #4405, opened 2017-09-04
Adding laravel icon

yamenarahman:
I think it's okay to initiate laravel with its beautiful logo rather than nothing.


bbashy:
This PR is broken. Your files weren't up-to-date.

PR #4406, opened 2017-09-05
Add S3 endpoint for compatible services

jbrooksuk:
This makes it easy to use something like Digitalocean Spaces.

Potentially, this should be on the 5.5 (maybe 5.4 too?) branch.


taylorotwell:
Send to 5.5 if it's not breaking.

jbrooksuk:
Will do!

PR #4412, opened 2017-09-09
Update User.php

keangkai:


scoutman57:
I don't know why you took away the * in the doc block.

PR #4418, opened 2017-09-13
Set redis database in .env file

leo108:


taylorotwell:
You can add this to your application if you need it.

PR #4419, opened 2017-09-13
[5.5] Move tinker to dev dependencies

frovere:
Probably not used in production env and it's an extra package
It works in dev env with autodiscovery

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,12 +7,12 @@
     "require": {
         "php": ">=7.0.0",
         "fideloper/proxy": "~3.3",
-        "laravel/framework": "5.5.*",
-        "laravel/tinker": "~1.0"
+        "laravel/framework": "5.5.*"
     },
     "require-dev": {
         "filp/whoops": "~2.0",
         "fzaninotto/faker": "~1.4",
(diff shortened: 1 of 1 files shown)


marky291:
And please explain how I am suppose to cache / clear caches in a simplistic manner or use git artisan deployment commands.

arubacao:
@marky291 all of the `php artisan x` commands would still work, including cache clearing. Only `php artisan tinker` would be "gone".
However, unless this doesn't lead to a significant performance boost, I don't think this will be merged.

GrahamCampbell:
Tinker is often used in production systems to query the state of things. Very useful.

muShubbar:
I don`t think Tinker is an extra package :)

michaeldyrynda:
It has been a standalone package since 5.4

Mechtecs:
+1 i'm missing tinker, too
It is a good interactive shell and should be included by default as a dev dependency

michaeldyrynda:
It's a [default dependency](https://github.com/laravel/laravel/blob/master/composer.json#L11) in Laravel, it's just a separate package rather than part of core framework as it was previously.

PR #4422, opened 2017-09-18
Added "-Indexes" option

andrey-helldar:
If you open the `/css`, `/js`, or any other folder on the default server, you can see the list of files in the directory.
The `-Indexes` option forbids viewing directory files via the web interface.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -1,6 +1,6 @@
 <IfModule mod_rewrite.c>
     <IfModule mod_negotiation.c>
-        Options -MultiViews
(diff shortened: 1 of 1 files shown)


m1guelpf:
@taylorotwell I know you don't like to modify the `.htaccess` file, but I think this could be an useful change. Also, [here is a more complete explanation of what does the `-Indexes` option do](https://stackoverflow.com/a/15838100/6386024).

gabriel-caruso:
Ow, that's a serious security fix :clap:

PR #4424, opened 2017-09-20
use the new helper method Route::view

bardouni:


scoutman57:
Seriously! Please close this as it's been attempted 100+ times. You would have known this if you even attempted to check out the history.

Damiloju:
Lol is there no way to stop people from doing this 
The same thing every time and the same reply

scoutman57:
yes, set some PR templates would help a lot.

scoutman57:
Or added a doc block to that piece of code.

scoutman57:
@taylorotwell what do you think of something like this? It even carries the 3 character shortness. Which by the way is much harder than it looks. 

https://github.com/scoutman57/laravel/commit/37978f9e9779ade8733a5925d768e4fd1c87addb

brayniverse:
This has been suggested a lot since 5.5 came out and understandably some people are likely to find it somewhat annoying. Personally I still think it's funny but am glad to see it's happening less frequently. This is a trivial change and I wouldn't have expected someone to check closed pull requests to see if it has already been attempted - so it's an easy thing to miss if you're not watching this repository. 

Thank you for your contribution @bardouni and I'm sorry it wasn't wanted at this time. Nevertheless I hope you keep contributing ([this](https://github.com/laravel/framework/pull/21113) was a particularly nice one).

👍

joshmanders:
>This is a trivial change and I wouldn't have expected someone to check closed pull requests to see if it has already been attempted

Being trivial means that you should suspect someone has already made such a request, and to check closed and open pull requests. Regardless of trivialness, that's best etiquette to check to see if what you're doing was attempted or submitted already.

PR #4425, opened 2017-09-21
fix problem when creating new project with laravel installer

g0110280:
This exception showed up when running `laravel new laravel`.

Solution: Add `post-install-cmd` to `composer.json` and run `php artisan key:generate` in it.


m1guelpf:
@g0110280 You have to update the Laravel Installer.

g0110280:
My bad, sorry.

PR #4427, opened 2017-09-22
ValidationException message

gguney:
Message in Illuminate\Validation\ValidationException constructor was static.


taylorotwell:
This isn't something you would display to your users.

PR #4436, opened 2017-09-28
Update .env.example

jirawat001:


rsm23:
I think **Taylor Otwel** will keep it as it is so that people won't forget **Homestead**  in favor of **Valet**

PR #4439, opened 2017-10-01
Route view helper for welcome page.

SidharthRaveendran:
Used the Laravel 5.5 route helper for views.

PR #4443, opened 2017-10-04
Use `Route::view()` in default routes file

svenluijten:
Not sure we'd want to push this to the front and include it in the default routes file, or keep it as kind of a "hidden feature", but I'll leave that decision up to you 🙂

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -11,6 +11,4 @@
 |
 */
 
-Route::get('/', function () {
-    return view('welcome');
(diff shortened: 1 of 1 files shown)


svenluijten:
#4439, nevermind. 🙊

m1guelpf:
@svenluijten You aren't the first one who proposes this change, but you're the first that has searched the PR section for duplicates, so 👏 👏

PR #4446, opened 2017-10-07
Update server.php

keangkai:
change a little comment


joshmanders:
I know this is for hacktoberfest to get a shirt, but why not do meaningful PR's? You think changing this one file's comment style from what is widely used and set by Taylor to something else, is going to get merged?

PR #4448, opened 2017-10-10
Update artisan wording in comment

jmarcher:
In other parts of the system, the application is mentioned as "your application" I taught it makes sense to change it here to the same convention.


taylorotwell:
Throws off the spacing.

PR #4450, opened 2017-10-12
Add exception throw for import failures

pori:
It would save a lot of debugging time to add an exception around the import's try-catch block. I realized this when seeing that upgrading to Bootstrap 4 produced silent failures.


pori:
Any particular reason why this was rejected? Just trying to make sure I was being compliant with the contributing guidelines.

rs-sliske:
they are in a try catch as a way of making them optional dependencies

aka if someone doesnt want to use jquery, they can just not install it. they dont have to edit that file

PR #4455, opened 2017-10-14
added base model file so it can be edited

Naxon:
I added a base Model class to the app folder so it can be editable (like the base controller can be).

[PR #21670 on laravel/framework](https://github.com/laravel/framework/pull/21670)


m1guelpf:
If you're going to do this, you have to make the User model extend this.

crynobone:
If we going this route I prefer to have the class name as `App\Eloquent` and make it `abstract class`.

devcircus:
This takes all of 10 seconds to do manually and everyone uses a different structure.

PR #4462, opened 2017-10-22
Publish register function

Laraveldeep:
Many beginners stumble on how to modify default registration behavior. For example, instead of automatically logging in, send verification mail and do other chores when user registers.

This question has been asked many times in Laracast before.

This PR publishes the `register` method from `Illuminate\Foundation\Auth\RegistersUsers` on `App\Http\Controllers\Auth\RegisterController` so that users can modify (override) registration process as per their need immediately.

### DIFF
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -68,4 +68,22 @@ class RegisterController extends Controller
             'password' => bcrypt($data['password']),
         ]);
     }
+
+    /**
+     * Handle a registration request for the application.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function register(Request $request)
+    {
+        $this->validator($request->all())->validate();
+
+        event(new Registered($user = $this->create($request->all())));
+
+        $this->guard()->login($user);
+
+        return $this->registered($request, $user)
+                        ?: redirect($this->redirectPath());
+    }
 }


renoirfaria:
Hey @Laraveldeep why did you closed this pull request? It was a great idea!

Laraveldeep:
@renoirfaria I tried another PR for same but @taylorotwell closed it. Probably its obvious that we are using `RegistersUsers` trait.

renoirfaria:
@taylorotwell desculpa incomodar mas voce poderia nos elucidar sobre este topico? Qual a melhor maneira de resolver isto?

jeystaats:
@renoirfaria  You can overwrite the function if you want.
By default the RegistersUsers trait is used in here . But if you write the function in the LoginController it will grab that function. 

So in that way you can rewrite the logic. No need to rewrite and 'publish' the code.

renoirfaria:
@jeystaats right, thats exactly what i actually do. But when it comes to add some verifications after user makes login it became a bit overwhelming, maybe we can work together on a better aproach to help people that needs to customize the logic after login/register.

rappasoft:
I think you're missing this then:

```
/**
     * The user has been authenticated.
     *
     * @param Request $request
     * @param         $user
     *
     * @return \Illuminate\Http\RedirectResponse
     * @throws GeneralException
     */
    protected function authenticated(Request $request, $user)
    {

    }
```

renoirfaria:
@rappasoft right, but this method runs only after the user is already authtenticated... if i want to add a rule to check if the user is activated (for example) you will have to check it and logout the same user after check it.

or you have to rewrite the entire login method just to add one or two lines of code.

My point here is that laravel could have some methods that customize the logic in the middle of the login/register proccess to avoid rewriting methods just to add few lines of custom code.

rappasoft:
I mean I don't see anything wrong with doing this:

https://github.com/rappasoft/laravel-5-boilerplate/blob/feature/5.5/app/Http/Controllers/Frontend/Auth/LoginController.php#L60

Laraveldeep:
@jeystaats 

> You can overwrite the function if you want.

That's what I am trying to do.
Its not very obvious to new users how to override these methods. 

Actually users might want to modify registration form with data from your DB, `showRegistrationForm` method also needs to be visible here.

Actually I am just bringing that forward in app. By default code executes from `Illuminate\Foundation\Auth\RegistersUsers` when we don't do anything.

Now after bringing that in front, same code runs from `App\Http\Controllers\Auth\RegisterController`

It helps users one step easier to modify/override the logic right away if they need after this. Nothing fancy.

Laraveldeep:
@rappasoft 

By default its like this
```
    /**
     * Show the application registration form.
     *
     * @return \Illuminate\Http\Response
     */
    public function showRegistrationForm()
    {
        return view('auth.register');
    }

```
Supposing I need a registration form where I need to collect user's data with their location, city, country etc. (of course with user table mofified), I need to do something like this
```
    /**
     * Show the application registration form.
     *
     * @return \Illuminate\Http\Response
     */
    public function showRegistrationForm()
    {
        return view('auth.register')->with(some data pulled from DB);
    }

```
It will be easier if we bring this to front (same code), where user can modify right away without digging deep.

That's the logic

Eventually we have to override this method but stumbling block is removed for new comers I guess.

rappasoft:
Yeah but if you bring it to the front and something changes in the framework repository, then another change needs to be made in the laravel repository.

This is the most basic of OOP, people need to learn to dig for the information to solve their problems, that is the best way to learn.

jeystaats:
I agree with @rappasoft . People need to learn that this a possibility with OOP. No need to hint every  possibility with a framework. This is just a learning progress throughout programming and if people want to learn this then they can read tutorials or people can write tutorials for them. 

A framework shouldn't be a tutorial on itself on the possibilities of OOP.

PR #4469, opened 2017-10-25
Showing app name on welcome page.

sidis405:
Upon installation of a fresh copy of Laravel, the first thing i do every single time is update the welcome.blade.php to reflect the project title i/m working on.
Especially for projects that will take a bit to have a landing page, but have internal pages that are defined quicker/

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -5,7 +5,7 @@
         <meta http-equiv="X-UA-Compatible" content="IE=edge">
         <meta name="viewport" content="width=device-width, initial-scale=1">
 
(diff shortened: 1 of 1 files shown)


m1guelpf:
@sidis405 This has been discussed before, and Taylor decided to keep it.

sidis405:
@m1guelpf I see. On the other hand this is already implemented on the app stub that the framework provides upon make:auth. So the it's fine when auth is scaffolded but not when the app is first set up? 
As far as branding goes, it still defaults to Laravel so i don't see an issue with that either.

Anyhow, thanks for the heads up man :)

PR #4470, opened 2017-10-26
Update AWS environment variable defaults

michaeldyrynda:
Aids those users that are using the AWS CLI tools, by matching the `filesystems.disks.s3`
configuration values with those that would be set in a user's native environment already.

AWS Documentation [reference](http://docs.aws.amazon.com/cli/latest/userguide/cli-environment.html).

Resolves laravel/internals#778

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -57,9 +57,9 @@ return [
 
         's3' => [
             'driver' => 's3',
-            'key' => env('AWS_KEY'),
-            'secret' => env('AWS_SECRET'),
-            'region' => env('AWS_REGION'),
+            'key' => env('AWS_ACCESS_KEY_ID'),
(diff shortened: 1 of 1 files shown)


m1guelpf:
@michaeldyrynda Maybe fallback to the old env key?

michaeldyrynda:
Can do, though it shouldn't matter - this will only affect new installs and the old key doesn't exist in the `.env` file

brunogaspar:
> this will only affect new installs and the old key doesn't exist in the .env file

This statement is not true.

If someone, like me, syncs from time to time the laravel/laravel repository by pulling in the changes from master (which is the latest **stable** release), the application itself might break due to the fact that your `.env` file is not out of sync and using incorrect variables, because changes like this are easy to miss and it's usually something you'll not have merge conflicts, at least not that often.

Nonetheless, i appreciate the change.

PR #4471, opened 2017-10-27
Factory code completing

lkmadushan:
IDE code completing support

### DIFF
--- a/database/factories/UserFactory.php
+++ b/database/factories/UserFactory.php
@@ -13,6 +13,8 @@ use Faker\Generator as Faker;
 |
 */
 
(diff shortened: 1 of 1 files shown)


database/factories/UserFactory.php
```diff
@@ -13,6 +13,8 @@
 |
 */
 
+/* @var Illuminate\Database\Eloquent\Factory $factory */
```
GrahamCampbell: This is not actually valid phpdoc. You need two stars at the start, otherwise, it's just a plain old comment.

lkmadushan: oh didn't see it. thanks

PR #4478, opened 2017-11-02
Updated fzaninotto/faker dependency in composer.json

gabriel-caruso:
Updated `fzaninotto/faker` to `~1.7`.


taylorotwell:
This isn't needed because ~1.4 will pull in 1.7.

PR #4474, opened 2017-10-30
Added new hashing configuration file.

morloderex:
In reference to [#21885](https://github.com/laravel/framework/pull/21885).

PR #4480, opened 2017-11-06
Added hashing configuration.

morloderex:
Resubmitted #4474 aganist develop.

### DIFF
--- a/config/hashing.php
+++ b/config/hashing.php
@@ -1,0 +1,19 @@
+<?php
+
+return [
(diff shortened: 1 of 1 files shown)


m1guelpf:
@morloderex Wouldn't it be better to name the file `hash.php` instead of `hashing.php`?

taylorotwell:
For now I will let people make this manually until we get a few versions deeper into PHP.

PR #4482, opened 2017-11-09
For Patreon (Alan Storm/Pulse Storm)

astorm:


SAutushka:
I think, incorrect URL is added (copy&past issue).

astorm:
@SAutushka You're correct.  Working in the evening isn't as glamorous as it used to be :)  Fix in this PR: https://github.com/laravel/laravel/pull/4486

PR #4484, opened 2017-11-09
Fix a small grammar

pranayaryal:
Fix a small grammatical error


taylorotwell:
This grammar also isn't right. I'll fix it.

PR #4488, opened 2017-11-15
Add \Schema::defaultStringLength(191); to AppSer..

enimiste:
Add \Schema::defaultStringLength(191);  commented to the AppServiceProvider to make it easy to enable and avoid searching for the syntaxe on each project.


KristofMorva:
For new projects why don't you use a new database version instead?

enimiste:
I agree with you.
But it is not the case all the time.

ruhul0:
https://github.com/laravel/laravel/pull/4972

PR #4489, opened 2017-11-15
add use RefreshDatabase in test files

laurencei:
We include the file, but dont actually `use` it in the example. This turns it on by default, to help highlight what is the minimum is to use it.


taylorotwell:
This could cause problems if people haven't configured a database yet.

PR #4490, opened 2017-11-15
Option 2 - put RefreshDatabase into the abstract TestCase

laurencei:
This seems like a better approach - that way all test classes extend the one file, rather than constantly redeclaring it?

### DIFF
--- a/tests/Feature/ExampleTest.php
+++ b/tests/Feature/ExampleTest.php
@@ -3,7 +3,6 @@
 namespace Tests\Feature;
 
 use Tests\TestCase;
-use Illuminate\Foundation\Testing\RefreshDatabase;
 
 class ExampleTest extends TestCase
 {
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -2,8 +2,8 @@
 
 namespace Tests;
 
-use Illuminate\Foundation\Testing\TestCase as BaseTestCase;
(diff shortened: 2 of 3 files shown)


KristofMorva:
I guess it's like that because in some cases you don't want to reset the database - maybe if you have many no-database-touching unit tests to gain performance?

fireynis:
@KristofMorva I think that is exactly why. This is forcing someone to reset their DB, maybe they don't want to do that at all and run tests against a persisting DB.

Right or wrong, forcing them to do it is not better.

fletch3555:
Agreed.  If you do anything about this, it should be Option 1 (#4489), not this.  But I don't really see the benefit to that one either.

laurencei:
Makes sense. I'll let Taylor close it when he decides...

rosDNsf:
@laurencei running the database refreshing in EVERY test is not even close to be a sensible default.

For example, why custom validation rules, hashing classes or third party services which should have independent unit test that don't rely on the database be obligated to pay for the performance hit of the "RefreshDatabase" trait?

> You can still change it to suit your application as needed.

I would say the same to you, if you require to do this just because you don't want to use the trait when it is needed, then do it in your own applications.

laurencei:
ok - fair enough - it was just an idea 😄

PR #4492, opened 2017-11-19
Git ignore screenshots from Laravel Dusk

jonhassall:
Git ignore screenshots from Laravel Dusk


KristofMorva:
I think placing a .gitignore in the folder is a better solution, so permissions for the folder can be set even after cloning, instead of the whole Browser folder.

jonhassall:
I agree. I will close this request.

PR #4493, opened 2017-11-20
Upgrade Vue

m1guelpf:
This PR upgrades the Vue dependency to its latest version.

:octocat:


DojoGeekRA:
@m1guelpf the current constraint already allows `2.5.7`

`^2.1.10` is the same as `>=2.1.10` and `<3.0.0`

https://getcomposer.org/doc/articles/versions.md#caret-version-range-

PR #4498, opened 2017-11-26
Update welcome route syntax

nikolaynizruhin:
Since laravel has `Route::view()` method I think it will be more clear way to determine welcome route.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -11,6 +11,4 @@
 |
 */
 
-Route::get('/', function () {
-    return view('welcome');
-});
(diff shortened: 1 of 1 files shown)


michaeldyrynda:
Taylor has mentioned in the past he relies on the definition as-is for testing purposes. Been a few PRs [opened about this](https://github.com/laravel/laravel/pulls?utf8=✓&q=is%3Apr+in%3Abody+route%3A%3Aview) previously, too.

nikolaynizruhin:
Aha got it.

PR #4500, opened 2017-11-28
Fix cookie name for Edge 14/15

agentmarine:
Edge has issues with session names containing underscores in certain versions. so replacing the underscores with a hyphen should resolve this issue.


taylorotwell:
That sounds more like a Edge problem? Underscores seem valid in cookie names across the Internet. However, you can change this in your own app as needed since it's a configuration file owned by the user.

PR #4502, opened 2017-12-02
Add redis database to .env

a-komarev:
Proposal to make `REDIS_DATABASE` configurable via .env file.

Making it because this is one of must have changes on each project creation stage. It's required to separate production & staging databases, otherwise if queue stack is stored in same redis database - there will be issues with it (staging will handle production queues and production will handle staging queues occasionally).


fletch3555:
I'm going to assume the answer is going to be along the lines of "You own those files. You're welcome to do this in your project."

a-komarev:
@fletch3555 no problem. Then let's remove all other env definitions in configuration files: `REDIS_PORT`, `REDIS_HOST`, `REDIS_PASSWORD` and so on :+1:

PR #4505, opened 2017-12-06
update AppServiceProvider

mahdi4187:
to remove error for migrate database


deleugpn:
This is an issue for only a subset of people and should not be assumed that everybody starting a new Laravel project wants that behavior.

PR #4506, opened 2017-12-06
Prevent register controller to be accessible for guest users

sebastiennoirie:
In the case of the auth controller are used for backend,
it's a better solution to allow register form only for authentificated users.

Otherwise, this allow an anonymous user to create an account and access to the backend.

I think that it's a better to be more restrictive :)


Te7a-Houdini:
by this way no one can register to your application

sebastiennoirie:
Hum, yes of course !
I will update it to allow register only on local env.

PR #4507, opened 2017-12-06
Allow register controller only in local environment

sebastiennoirie:
In the case of the auth controller are used for backend,
it's a better solution to allow register form only for authentificated users.

Otherwise, this allow an anonymous user to create an account and access to the backend.

I think that it's a better to be more restrictive :)

### DIFF
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -36,7 +36,7 @@ class RegisterController extends Controller
      */
     public function __construct()
     {
-        $this->middleware('guest');
+        $this->middleware('auth');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Te7a-Houdini:
it's the same idea #4506  your application won't allow to register anyone from outside unless it's authenticated and that's not the case in most of the applications

deleugpn:
You can override the `construct` method to do whatever you want in your application. The Laravel default template assumes a public application where anybody can/should be able to register themselves (like GitHub, Facebook, Forums, Laracast, etc).

madeITBelgium:
You can use some third-party package to prevent or allow registration per domain name. https://github.com/madeITBelgium/Laravel-Email-Domain-Validation

PR #4510, opened 2017-12-15
Add config for app whoops editor.

yajra:
In relation to https://github.com/laravel/framework/pull/22350.


m1guelpf:
@yajra Why does it default to sublime?

yajra:
Not really sure what to put on default but since Taylor uses sublime afaik then I guess it's a good default? I use `vscode` on my personal config. What do you suggest?

m1guelpf:
@yajra Setting it to `false` or `null` should disable it

yajra:
@m1guelpf updated default to `null`. Thanks!

ElfSundae:
If we add this, it should use `env` and with empty default value:
```php
 'editor' => env('APP_EDITOR'),
```
Then anyone can configure it in his `.env` file, and there is no issues  when the default editor is not installed.

taylorotwell:
Would rather just let people add this manually.

a-komarev:
@yajra maybe it would better to make this extension via package? I've started small one yesterday, and we could extend it's functionality to fit all requirements: https://github.com/cybercog/laravel-whoops-editor

a-komarev:
@yajra thanks for pointing out this requirement.
This feature was added to https://github.com/cybercog/laravel-whoops-editor package.
It's looking for `app.editor` config value, if not found or return `null` then `APP_EDITOR` value from `.env` file will be used.
Additionally it allows to overwrite Homestead file path on the fly.

PR #4512, opened 2017-12-16
whoops package updated

pavinthan:
whoops package updated


ankurk91:
The **~** sign in `"filp/whoops": "~2.0",` denotes that composer will automatically pick 2.1 when available
Read more
https://getcomposer.org/doc/articles/versions.md#tilde-version-range-

pavinthan:
Thanks @ankurk91 and @GrahamCampbell

PR #4513, opened 2017-12-17
Ignore the MacOS .DS_Store in .gitignore

vuchastyi:
It's annoying when i create laravel project, then push to git and it contains .DS_Store files.


asoorm:
This should be in your global gitignore

GrahamCampbell:
Thanks for the PR @vuchastyi. Buy, as @asoorm points out, the generally accepted way to deal with platform specific files is to put them in system's global ignore file.

PR #4514, opened 2017-12-17
[5.6] Adds nunomaduro/collision on composer require-dev

nunomaduro:
This PR proposes the addition of the package [Collision](https://github.com/nunomaduro/collision) as **dev** dependency of every Laravel app. Collision is an detailed & intuitive error handler framework for PHP **console** applications, and includes an adapter/service-provider for **Laravel**.

The package project is [https://github.com/nunomaduro/collision](https://github.com/nunomaduro/collision) and bellow is an screenshot example.

example-2
Those are the Collision **dependencies**:

```json
    "require": {
        "php": "^7.1",
        "filp/whoops": "^2.1.4",
        "symfony/console": "~2.8|~3.3|~4.0"
    }
```

Concerning the integration Laravel, If the exception implements the  `Symfony\Component\Console\Exception\ExceptionInterface`, [Collision](https://github.com/nunomaduro/collision) still uses the default `\Illuminate\Contracts\Debug\ExceptionHandler::renderForConsole` method:

screenshot 2017-12-17 22 31 30
Otherwise Collision will render the exception details itself giving an detailed explanation of the problem:

screenshot 2017-12-17 22 57 56
If someone needs details about package or about how it works behind the scenes please don't hesitate to ask.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,7 @@
     },
     "require-dev": {
         "filp/whoops": "~2.0",
+        "nunomaduro/collision": "~1.1",
         "fzaninotto/faker": "~1.4",
         "mockery/mockery": "~1.0",
         "phpunit/phpunit": "~6.0"


GrahamCampbell:
Thanks, but I don't think we should include a third party library like this tbh.

timacdonald:
As this package is "Whoops for the console" I don't see a huge problem (from my cosy little bubble of working with Laravel on small projects over here) with including "Whoops for Http" and "Whoops for console". If anything it makes things more consistent as now we see exceptions in the console like we do in the browser. Otherwise you could just argue that "Whoops" should not be included - but we all love that.

I have just started using this and its awesome. Definitely going to be installing in any future project.

ankurk91:
Laravel 5.5 supports php v7.0
https://github.com/laravel/laravel/blob/3f0e742421d8a922ba8a3b075189aa1bfd103ab6/composer.json#L7-L8

While collision requires php v7.1

https://github.com/nunomaduro/collision/blob/d4b71295d47d269f385f6f4355e16eeb07a66eff/composer.json#L16-L20

ElfSundae:
@ankurk91 It is for 5.6, https://github.com/laravel/laravel/blob/develop/composer.json

timacdonald:
Looks like `5.6` which this is targeting is going to be min PHP `7.1` so should be sweet: https://github.com/laravel/framework/blob/master/composer.json#L18

ankurk91:
But GitHub indicates that target branch is `master` not `develop`.

screenshot from 2017-12-18 10 35 21

ElfSundae:
@ankurk91 Oh, sorry, you're right. 
@nunomaduro L5.6 is on develop branch, you may change the base branch to `develop`

taylorotwell:
Seems pretty nice to me. Thanks

taylorotwell:
@nunomaduro one thing that I think would be nice is to disable collision when PHPUnit is running. Do you have any thoughts on that?

nunomaduro:
@taylorotwell You are right! I just tagged `v1.1.17` where Collision is disabled on testing environment.

https://github.com/nunomaduro/collision/commit/312f56cf740185ed831301096890a54ea5a02eb9

taylorotwell:
Thanks

PR #4516, opened 2017-12-19
Update SQS config

pavinthan:
Use env for SQS config


pavinthan:
Thank you @taylorotwell

PR #4518, opened 2017-12-19
Trusted proxy to version 4.0 for Laravel 5.6

fideloper:
Trusted proxy now either trusts the `Forwarded` header, or all `X-Forwarded-*` headers to determine the correct client info when a proxy is used.

Symfony 4 no longer supports using custom headers. This takes away an unfortunate feature, but simplifies the package a bit.

### DIFF
--- a/app/Http/Middleware/TrustProxies.php
+++ b/app/Http/Middleware/TrustProxies.php
@@ -15,15 +15,9 @@ class TrustProxies extends Middleware
     protected $proxies;
 
     /**
-     * The current proxy header mappings.
+     * The headers used to detect proxies.
      *
      * @var array
      */
-    protected $headers = [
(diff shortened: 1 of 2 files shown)


svenluijten:
Should the variable maybe be renamed to `$header` instead of `$headers`?

KristofMorva:
@svenluijten I don't see how would that change be relevant to this PR. It is called the same way, and it still seems to be an array.

fideloper:
It's no longer an array.

In one case is one header (`forward`), in another case, it's allowing multiple (all `x-forwarded-*` headers), so either way.

KristofMorva:
@fideloper ah okay, I've thought it based on the PhpDoc, which is `array`.

fideloper:
Ah, yep, thanks for pointing that out, I'll get it fixed up.

PR #4520, opened 2017-12-21
[5.5] Timezone and Locale config set as env values

s2kagency:
Timezone and Locale should be set as variable depending on environmental values, avoiding to change config/app.php file (the same project could be deployed on different servers with different settings).


Jam0r85:
Pretty sure this has been suggested before and the changed wasn't implemented.

fletch3555:
https://github.com/laravel/laravel/search?q=env%28%27APP_TIMEZONE%27&type=Issues&utf8=%E2%9C%93

Already been suggested and denied twice

s2kagency:
ok, sorry but is it possible to know why it was denied? assuming to keep the UTC as default timestamp why not giving the possibility to edit in .env without modifying the php file?
Especially for the Locale it is dependent on the environment.

rs-sliske:
the config file is part of your project, you are free to make these changes yourself if you need them

personally i dont really think these do make much sense to be changed across environments. if they were going to be changed shouldnt it be based on the current user?

fletch3555:
You're free to do it in your own app (https://github.com/laravel/laravel/pull/3792#issuecomment-222945960)

s2kagency:
@rs-sliske if the same project is released in different versions for different URL (project.it, project.eu, project.com) with different settings why can't the locale and/or timezone be related to that specific version?

rs-sliske:
@s2kagency if that is needed for your project, then you can make the change in your project

ElfSundae:
Using `env` or not both OK for me. I submitted `APP_TIMEZONE` [before](https://github.com/laravel/laravel/pull/3792), so I'd like to share my two cents here.

If many (maybe MOST I guess) of us change the timezone config for every project, why not make it environmental for convenient?  Besides avoiding to change the app config file, we will get more, e.g. different locale staging/production app using the same code (maybe different branch or tag). Or using the config value in another config like `session.cookie` using `env('APP_NAME')`.
Another case is setting timezone for testing, you can see Laravel framework's [bootstrap.php](https://github.com/laravel/framework/blob/5.5/tests/bootstrap.php#L30) and Orchestral [Testbench's way](https://github.com/orchestral/testbench#overriding-application-timezone), I prefer to configure it in the `phpunit.xml` like:
```xml
<php>
    <env name="APP_ENV" value="testing"/>
    <env name="APP_TIMEZONE" value="Asia/Shanghai"/>
    <env name="CACHE_DRIVER" value="array"/>
</php>
```

Are they specific to the environment? I think it depends. Think about the `app.name` configuration, it is not environmental before 5.4. The `mail.from` config as well. 

I just want to say different people have different kind of view. Actually now I stand on the current behavior: no env, until we have a strong reason.

PR #4522, opened 2017-12-22
NPM replace with yarn in scripts

pavinthan:
NPM replace with yarn in scripts


m1guelpf:
@pavinthan Not everyone has Yarn! You can make the change in your own application if you want.

pavinthan:
Thanks @m1guelpf and @taylorotwell

PR #4523, opened 2017-12-23
[5.6] Phpunit pretty result printer

carusogabriel:
With this PR, I want to propose a better output for our tests with [`mikeerickson/phpunit-pretty-result-printer`](https://github.com/mikeerickson/phpunit-pretty-result-printer). In the same way we do with `Whoops`, for better output in the Browser, this package adds a better output for tests in the terminal.
Here are screenshots comparing:
- Before:
captura de tela de 2017-12-23 10-52-28
- After:
captura de tela de 2017-12-23 10-52-55

I just need to add:
```sh
codedungeon/phpunit-result-printer
```
in our dev-dependencies in `composer.json` and
 ```xml
printerClass="Codedungeon\PHPUnitPrettyResultPrinter\Printer"
```
in the `phpunit` tag inside `phpunit.xml`.

Hope you find it interesting :smile:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
         "nunomaduro/collision": "~1.1",
         "fzaninotto/faker": "~1.4",
         "mockery/mockery": "~1.0",
-        "phpunit/phpunit": "~6.0"
+        "phpunit/phpunit": "~6.0",
+        "codedungeon/phpunit-result-printer": "~0.4"
     },
     "autoload": {
         "classmap": [
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -6,6 +6,7 @@
          convertErrorsToExceptions="true"
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
+         printerClass="Codedungeon\PHPUnitPrettyResultPrinter\Printer"
          processIsolation="false"
          stopOnFailure="false">
     <testsuites>


GrahamCampbell:
I don't think this needs to be in the core. If it were core material, it'd already be shipped with phpunit.

carusogabriel:
@GrahamCampbell [Someone already suggested](https://twitter.com/s_bergmann/status/944119498914914304) to @sebastianbergamann, and it probably will become a third party extension. And as we support `Whoops`, I thought would be interesting.

deleugpn:
Isn't it already a third-party extension?

brunogaspar:
@carusogabriel Why would you think this is good for everyone to use by default?

a-komarev:
I like it. It's much more informative and easier to understand what tests are failed, without diving deep into error messages.

carusogabriel:
@deleugpn From what I know, no.

@brunogaspar It drastically reduces the amount of dots (`.`) and lines in large suites. Se [the example](https://github.com/mikeerickson/phpunit-pretty-result-printer#screenshot) in the repo.

brunogaspar:
> It drastically reduces the amount of dots (.) and lines in large suites. Se the example in the repo.

@carusogabriel That's a very weird excuse to be fully honest, reduce the amount of "dots". Tests output should give information, not hide it, that package does nothing that helps either situation, it just gives a "weirder" output, in my opinion.

@a-komarev You like it, cool, i respect it, however i personally don't like the output of this package.

Here's a better one https://github.com/LimeDeck/phpunit-detailed-printer which tells you the failing tests without the need to wait for the entire suite or grouped tests to end, which is something this package you guys want to add by default don't do nor does PHPUnit do by default.
I know it creates a bigger output, but gives you immediate access to the failing tests, which is a bonus.

But either way, it's a 👎  from me, as i mentioned, this type of stuff should be added by the user to their own applications and not enforced by Laravel.

a-komarev:
@brunogaspar reduces dots amount, but incremented lines amount drastically. I'm not sure it's a benefit :D

brunogaspar:
> reduces dots amount, but incremented lines amount drastically.

@a-komarev That's why i said

> I know it creates a bigger output, but gives you immediate access to the failing tests, which is a bonus.

> I'm not sure it's a benefit :D

Right. The package on this PR doesn't give any benefit either, just removes dots from the output lol

a-komarev:
I've found one issue with it. I had pretty long namespace in one of the projects, and it's cutted (only last part is displayed).

a-komarev:
Solution which @brunogaspar is really awesome. It converts function names to readable text, write called class complete namespace and displays time of each test execution.

carusogabriel:
@a-komarev @brunogaspar `Detailed PHPUnit Printer` is a really good package. If this PR do not get accepted, you can PR it :smile:

Damiloju:
I think this should be a thing of choice  Nice package but surely not everyone will like it

MichaelDeBoey:
@taylorotwell Is [thinking about this idea](https://twitter.com/taylorotwell/status/944613326734192643) for sure 🙂

a-komarev:
Not sure if it will be added to the core, but I've created PR #4524 to have an alternative suggestion which seems more solid to me.

taylorotwell:
I will revisit this before 5.6 release.

PR #4524, opened 2017-12-23
[5.6] Detailed PHPUnit Printer

a-komarev:
Alternative to: #4523

Initially @carusogabriel suggested to integrate PHPUnit Pretty Result Printer. But as revealed after some integrations in different projects it has issues:
- long namespaces cutted down (only ending part is visible) what makes understanding whats broken more complicated;
- it doesn't show what test is failed, it only shows how much success\fail tests were in Test Case.

Then @brunogaspar shared link on another great tool: https://github.com/LimeDeck/phpunit-detailed-printer which I'm proposing to add instead of solution described above.

Benefits of PHPUnit Detailed Printer:
- Displays fully qualified namespace of Test Case.
- Displays each test from Test Case in human readable format by converting `function it_can_fill_body_attribute` to `It can fill body attribute`.
- Displays execution time for each test with highlighting slow tests.

I don't know if it will be good to have such detailed information in projects with thousands of tests, but it works great in most of the cases by giving you much more overview about your tests.

Thanks to @carusogabriel and @brunogaspar for the suggestions. Even if this wouldn't be added to the core - this will be helpful starting point to add this manually to the project.

And here is a screenshot of how it will look like:

detailed-printer

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,7 @@
     "require-dev": {
         "filp/whoops": "~2.0",
         "fzaninotto/faker": "~1.4",
+        "limedeck/phpunit-detailed-printer": "^3.1",
         "mockery/mockery": "~1.0",
         "phpunit/phpunit": "~6.0"
     },
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -6,6 +6,7 @@
          convertErrorsToExceptions="true"
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
+         printerClass="LimeDeck\Testing\Printer"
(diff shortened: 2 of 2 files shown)


a-komarev:
Another one solution which seems to be even more better readability inspired from Mocha: https://github.com/Sempro/phpunit-pretty-print

GrahamCampbell:
There also exists other ways to get good output, such as, by telling phpunit to output junit xml, then any junit consumer can give output, such as from an IDE.

taylorotwell:
I have it on my list to revisit this right before 5.6 release.

PR #4526, opened 2017-12-24
Include cluster in pusher config

themsaid:


Tpojka:
Exact the same PR I proposed several (3–4) months ago. But, maybe now is good moment for merging. :)

PR #4530, opened 2017-12-26
Update from original laravel

keroslee:
Update from original laravel


keroslee:
Update from original laravel

keroslee:
Update from original laravel

PR #4531, opened 2017-12-29
Change the locale and fallback_locale and fallback_locale from env vars

marcospaegle:
Change the config/app.php to get timezone, locale and fallback_locale from env vars. This away we dont need to change the config/app.php


taylorotwell:
You own the app, you can change these if needed.

PR #4532, opened 2017-12-30
Update AppServiceProvider.php

brunolsantos:
Fixed string Lenght error when migrating.

Traversy tip:
https://youtu.be/4pc6cgisbKE?t=11m56s

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -2,8 +2,8 @@
 
 namespace App\Providers;
 
-use Illuminate\Support\ServiceProvider;
 use Illuminate\Support\Facades\Schema;
+use Illuminate\Support\ServiceProvider;
 
 class AppServiceProvider extends ServiceProvider
 {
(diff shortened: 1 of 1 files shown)


lagbox:
not everyone needs to do this .. this is a work around for using the wrong version of a database

pavinthan:
@brunolsantos  It doesn't need because, new MySQL servers support it, use the latest version of MySQL server is good

brunolsantos:
Yes @lagbox and @pavinthan, but like @enimiste said there

> [ ... ]
But it is not the case all the time.

I think this update will improve the laravel usage for people that use XAMPP like programs.

m1guelpf:
@brunolsantos You own the files. You can change them locally

lagbox:
how about people actually use a real setup instead of these silly stacks that obviously come with dated software ... if only laravel was nice enough to figure out a way to provide a fully functional environment with all the latest software and didn't need to do any work arounds ... oh wait, homestead

PR #4533, opened 2017-12-30
Make welcome.blade.php use config for name

DrowningElysium:
We had some users at the IRC channel that were wondering why config('app.name') wasn't working in `app.blade.php` generated by `php artisan make:auth`.

This will make it so that `welcome.blade.php` also uses the config file for the name.


m1guelpf:
@DrowningElysium This has been discussed before, and Taylor decided to keep Laravel instead of your app name.

DrowningElysium:
@m1guelpf Thanks, I didn't look it up to be honest. I'll leave it open if Taylor has changed his mind, he can merge it and else close it.

PR #4288, opened 2017-06-06
Fix return type to the correct one

nmfzone:
Changed return type of render method on the Exception handler from `\Illuminate\Http\Response` to `\Symfony\Component\HttpFoundation\Response`.

PR #4535, opened 2018-01-05
Fix return typo

carusogabriel:
Recently, I ran [`PHPStan`](https://github.com/phpstan/phpstan) in my application, and it detects this miss typo. Makes sense, as we are calling `parent`,and in `Illuminate/Foundation/Exceptions/Handler::render` [we return](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Foundation/Exceptions/Handler.php#L165) `\Symfony\Component\HttpFoundation\Response`


mingyoung:
#4210 #4434 #4288

PR #4538, opened 2018-01-05
added an extension to the .htaccess file, commented out

kinablelieselot:
It is not complete but it's a start and I find it quite interesting!


GrahamCampbell:
Thanks, but I don't think we need all this in the default app.

Badlapje:
more security and some default performance in the default app doesn't hurt, especially as it's all commented out.  
+1 for this one

PR #4541, opened 2018-01-07
Report or log an exception need return

baixinxing:
Object Oriented need return


m1guelpf:
@baixinxing Parent class doesn't return anything

PR #4543, opened 2018-01-08
Remove autoload for database factories

jerguslejko:
Files in this directory (`database/factories`) are loaded by the framework itself. There is no need to generate a classmap for them.  (https://github.com/laravel/framework/blob/5.5/src/Illuminate/Database/Eloquent/Factory.php#L202)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,8 +19,7 @@
     },
     "autoload": {
         "classmap": [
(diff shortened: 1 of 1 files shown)


taylorotwell:
Going to leave this in place for now in case devs are taking unique class based approaches here.

PR #4544, opened 2018-01-11
[5.5] Added User factory type hint annotation.

ed-fruty:


ed-fruty:
@taylorotwell , Can you provide any description, why do you close it without merge? As I see, master and develop branches doesn't have this changes.

PR #4545, opened 2018-01-12
Server Requirements

tabuna:
- PHP >= 7.0.0
- OpenSSL PHP Extension
- PDO PHP Extension
- Mbstring PHP Extension
- Tokenizer PHP Extension
- XML PHP Extension


GrahamCampbell:
:-1: For adding these. We have composer so that it recursively resolves dependencies.

PR #4546, opened 2018-01-14
Remove alias "Eloquent"

MCMatters:
The class `Illuminate\Database\Eloquent\Model` is abstract, so when you are trying call something static, it calls `__callStatic` method which makes `(new static)` and cause an exception 
```Cannot instantiate abstract class Illuminate\Database\Eloquent\Model```
I suggest removing this alias

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -203,7 +203,6 @@ return [
         'Cookie' => Illuminate\Support\Facades\Cookie::class,
         'Crypt' => Illuminate\Support\Facades\Crypt::class,
         'DB' => Illuminate\Support\Facades\DB::class,
-        'Eloquent' => Illuminate\Database\Eloquent\Model::class,
         'Event' => Illuminate\Support\Facades\Event::class,
         'File' => Illuminate\Support\Facades\File::class,
         'Gate' => Illuminate\Support\Facades\Gate::class,


taylorotwell:
Eh, I think something else is causing your issue.

MCMatters:
@taylorotwell I've just tried on fresh Laravel installation and got the error. If you don't believe me, try it yourself.

Here the screenshot with proof
http://take.ms/pHrWt

deleugpn:
Why would you call make on Eloquent itself?

michaeldyrynda:
You wouldn't ever do that in practice.

The Eloquent alias likely only exists as backwards compatibility for when `Model` used to be referenced as `Eloquent` in [previous versions of Laravel](https://github.com/laravel/laravel/blob/v4.2.11/app/models/User.php).

In v5, it was changed to reference `Illuminate\Database\Eloquent\Model` [directly](https://github.com/laravel/laravel/blob/v5.0.0/app/User.php).

MCMatters:
@deleugpn, @michaeldyrynda  it is just an example. Since `Model` is macroable, people may add some functions and someone wants call it from common facade, because it is presented in `aliases` section. What's the problem to delete it, because it confuses some people

m1guelpf:
@MCMatters Maybe remove it in `5.6`?

PR #4547, opened 2018-01-16
fix some php7 type hints

AlaaAttya:


pavinthan:
@AlaaAttya void support in PHP 7.1, what about php 7.0?

TomasVotruba:
:+1: for adding PHP 7.0 return typehints

marky291:
Can’t we just use the documentation @return, seems pointless unless I’m missing something here.

deleugpn:
Not that I'm in favor of this, but I think Laravel 5.6 will be 7.1+ anyway?

mateusjatenee:
Does Laravel use return types anywhere elae? Unless it was added to the
core recently I don’t remember seeing it anywhere.

app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -45,7 +45,7 @@ public function __construct()
      * @param  array  $data
      * @return \Illuminate\Contracts\Validation\Validator
      */
-    protected function validator(array $data)
+    protected function validator(array $data): \Illuminate\Contracts\Validation\Validator
```
a-komarev: Import it?

app/Exceptions/Handler.php
```diff
@@ -44,9 +44,9 @@ public function report(Exception $exception)
      *
      * @param  \Illuminate\Http\Request  $request
      * @param  \Exception  $exception
-     * @return \Illuminate\Http\Response
+     * @return \Symfony\Component\HttpFoundation\Response
      */
-    public function render($request, Exception $exception)
+    public function render($request, Exception $exception): \Symfony\Component\HttpFoundation\Response
```
a-komarev: Import it?

taylorotwell:
Not a fan.

PR #4548, opened 2018-01-16
[5.5] Add typehint comment for the $factory variable

carusogabriel:
Following https://github.com/laravel/framework/pull/22553

### DIFF
--- a/database/factories/UserFactory.php
+++ b/database/factories/UserFactory.php
@@ -13,6 +13,7 @@ use Faker\Generator as Faker;
 |
 */
 
+/* @var \Illuminate\Database\Eloquent\Factory $factory */
 $factory->define(App\User::class, function (Faker $faker) {
     return [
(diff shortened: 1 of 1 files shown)


carusogabriel:
@taylorotwell The StyleCI change should be applied and replicate to framework?

a-komarev:
@carusogabriel search in issues. There are tons of PRs with this change. Some of them were accepted, then reverted, and again and again.

carusogabriel:
@a-komarev Thanks, I'll replicate this to Framework as well :+1:

GrahamCampbell:
The StyleCI change is happening because the position is not valid phpdoc, so it gets converted to a comment.

GuidoHendriks:
@GrahamCampbell By that you mean that it's invalid _according to StyleCI_? Because it actually is valid PHPDoc.

PR #4549, opened 2018-01-17
Develop

udanpe:
Its better then fillable.
```
$model = new Model;

$rules =  [
    'name' => 'required',
    'surname' => 'required',
];

$request->validate($rules);

$model->saveOnlyValidatedFields($model, $rules, $request);

function saveOnlyValidatedFields($model, $rules, $request) {

    foreach($rules as $key => $v) $model->$key = $request->$key;

    $model->save(); // return $model
}

```

UPD.
i want new method in new version laravel
`$model->saveOnlyValidatedFields($request)`


GuidoHendriks:
I'm confused. 🤷🏼‍♂️

MichaelDeBoey:
@udanpe I don't think this was supposed to be in this repo?

udanpe:
@MichaelDeBoey Why? I think this is a repo for new  feature. Or not? Ok. Sorry.

jvanbaarsen:
@udanpe Can you add some more information about what you're trying to archieve with this PR? What does the code you are trying to add do? What problem does it solve?

udanpe:
@jvanbaarsen more clean code and more secure
`$user->fill($request->all())->save();  // with protected $fillable not secure`
 
```
$request->validate([
    'value1 ' => 'required',
    'value2 ' => 'required',
    'value3 ' => 'required',
    'value4' => 'required',
    'value5 ' => 'required',
//.....
'value45 ' => 'required',
]);


$user->value1 = $request->value1;
$user->value2 = $request->value1;
$user->value3 = $request->value1;
$user->value4 = $request->value1;
// ......
$user->value45= $request->value45;


// duplicate code
```

i want new function in new version laravel
`$model->saveOnlyValidatedFields()`

fletch3555:
What if there are fields you don't want to validate?  I'm at a loss for an example at the moment, but I'm sure one exists.

udanpe:
@fletch3555 don't use this futures.
Or
```
$request->validate([
    'NotSecureField' => '',
]);
```

fireynis:
I can actually see how this would be useful. It would be nice to grab validated fields from a request to use as filliable, I do this all the time. 

Does this exist anywhere in the framework currently?

udanpe:
about protected $fillable and secure. Part of fields cant change only administrator if you add all fields to fillable every users can be administrator or part of code will be duplicated. Protected fillable its bet idea. Better fillable only validated fields

PR #4550, opened 2018-01-17
Updated PHP DocBlock for TrustProxies::$proxies to note that "**" can be used to trust all IP addresses.

qmcree:


KristofMorva:
First of all, I think you are not limited to use IP addresses, domain names can also be used. In addition, "IP addresses are non deterministic" may need some clarification, personally I don't understand what you mean, as IP addresses are certainly deterministic, maybe the processing order of them is not (but why would that matter anyways if it's only a whitelist?).

deleugpn:
Nondeterministic mean dynamic ip address.

KristofMorva:
@deleugpn that was my guess, but I'd say "dynamic IP address" is a much clearer expression than "non-deterministic IP address".

PR #4552, opened 2018-01-18
Update AppServiceProvider.php

Mickey248:
update laravel php artisan make:auth
and php artisan migrate for error laravel5.4-5.5


Jam0r85:
Same thing has been submitted many previous times and it only applies to certain databases.

lagbox:
#4532

PR #4553, opened 2018-01-18
Add force flag to ensure phpunit will overwrite environment variables

agostlg:
This is the solution for a problem I faced today when setting up my testing environment for a newly created Laravel project.

Main issue and solution described here: [https://laravel.io/forum/laravel-possibly-overwriting-phpunit-env-variables](https://laravel.io/forum/laravel-possibly-overwriting-phpunit-env-variables)

Also I did some basic proof by creating a fresh install:

First I set APP_ENV=local

```
export APP_ENV=local
```

With the non-modified phpunit.xml file

```
➜  blog git:(master) ✗ cat phpunit.xml | grep APP_ENV
        <env name="APP_ENV" value="testing"/>
```

I ran phpunit with the current values and die at the first test printing APP_ENV

```
➜  blog git:(master) ✗ vendor/bin/phpunit
PHPUnit 6.5.5 by Sebastian Bergmann and contributors.

local%
```
Now I do the same but with the force key

```
➜  blog git:(master) ✗ cat phpunit.xml | grep APP_ENV
        <env name="APP_ENV" value="testing" force="true"/>
➜  blog git:(master) ✗ vendor/bin/phpunit
PHPUnit 6.5.5 by Sebastian Bergmann and contributors.

testing%

```

I had the same behaviour running phpunit on my local (with the ENV set beforehand) and on my docker container.

The thread that adds the force flag is: [https://github.com/sebastianbergmann/phpunit/issues/2353](https://github.com/sebastianbergmann/phpunit/issues/2353)

Thanks.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -23,9 +23,9 @@
         </whitelist>
     </filter>
     <php>
-        <env name="APP_ENV" value="testing"/>
-        <env name="CACHE_DRIVER" value="array"/>
-        <env name="SESSION_DRIVER" value="array"/>
-        <env name="QUEUE_DRIVER" value="sync"/>
+        <env name="APP_ENV" value="testing" force="true"/>
+        <env name="CACHE_DRIVER" value="array" force="true"/>
+        <env name="SESSION_DRIVER" value="array" force="true"/>
+        <env name="QUEUE_DRIVER" value="sync" force="true"/>
     </php>
 </phpunit>


taylorotwell:
Can't recreate with this simple test:

image

agostlg:
@GuidoHendriks Yeah you're right, my intention here was only to fix for others who have environment keys being set on docker-compose.yml for local and Jenkins configuration for the testing server, which is my case.

Still makes sense to me that phpunit.xml should overwrite the env variables regardless of where it stays, and the only way to assure that at the moment is to use the flag.

CImrie:
@taylorotwell I just recloned an existing project that was previously working and this issue suddenly started happening for me too.
My install was refusing to use my test DB during phpunit tests and kept using the one provided in .env file despite `app()->environment()` and `app()->environmentFile()` both dd'ing as `testing` and `.env.testing` respectively.

I've done some research and can reproduce/fix with following setup:
PHP 7.2.8
PHPUnit 6.5.9

PHPUnit appears to be setting the env variables and then vlucas/phpdotenv package refuses to override any because they are already set. I think phpdotenv moved to 'immutability' as the default when going to v2.x.

I can resolve this by changing `Line 27` in `Illuminate\Foundation\Bootstrap\LoadEnvironmentVariables`
from: 
```php
        try {
            (new Dotenv($app->environmentPath(), $app->environmentFile()))->load();
        } catch (InvalidPathException $e) {
            //
        }
```

to:
```php
        try {
            (new Dotenv($app->environmentPath(), $app->environmentFile()))->overload();
        } catch (InvalidPathException $e) {
            //
        }
```
(i.e. changing `load()` to `overload()` as provided in the phpdotenv link below).

See these issues/answers as reference for what lead me to the above change:
https://github.com/sebastianbergmann/phpunit/issues/2353
https://stackoverflow.com/a/41827901
https://github.com/vlucas/phpdotenv#immutability

CImrie:
Ok, in addition to this (as vlucas/phpdotenv does prevent overriding) and after finally fixing my own system without changing the above lines I found my exact issue:
I had 'dotenv' plugin enabled in ~/.zshrc (oh-my-zsh) which was setting system env variables to the local laravel environment - because phpdotenv package doesn't override in the current setup, this was causing it to always use my local environment.

I do still think I'd expect .env (Laravel) to override system-defined ENV variables, but nonetheless I hope this helps anyone who comes across a similar issue.

PR #4557, opened 2018-01-23
Update laravel-mix

nikolaynizruhin:
Update laravel-mix to 2.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
         "bootstrap-sass": "^3.3.7",
         "cross-env": "^5.1",
         "jquery": "^3.2",
-        "laravel-mix": "^1.0",
+        "laravel-mix": "^2.0",
         "lodash": "^4.17.4",
         "vue": "^2.5.7"
     }


taylorotwell:
Will have to wait on @JeffreyWay to indicate if this should be pulled in yet or not.

taylorotwell:
I also don't know what breaking changes it may involve.

xandor:
(Potentially) Breaking
[PR #1367](https://github.com/JeffreyWay/laravel-mix/pull/1367) Distinguish better between font and image SVGs. (Depending upon your project structure, this may change the output path for certain SVG files.)

if there are any SVGs in found in your CSS/SASS/LESS(/...) files, they have bin collected in the folder public/fonts.
From v2.0 onwards these svg files are copied to public/images unless they where in assets/fonts.

JeffreyWay:
Fine to merge. 

As @xandor, the only change worthy of bumping to 2.0 is a tweak to how we set the output path for SVG files. We now do a better of a job deciding if it's a font SVG or not.

AGhasemzadeh:
Hello,
I think we can have it in version 5.6 and also jQuery 3.3.

PR #4559, opened 2018-01-24
Auto detect resource root

MordiSacks:
Since we know the absolute app URL (as it is defined in the .env file)  
Why not use it in laravel-mix.  
I personally, find this very useful, As I develop in windows using [WAMP](http://www.wampserver.com/en/), and rarely use virtual hosts, This results in project urls that look like `127.0.0.1/Projects/foo/public`.  
I think this PR can be very beneficial.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
         "axios": "^0.17",
         "bootstrap-sass": "^3.3.7",
         "cross-env": "^5.1",
+        "dotenv": "^4.0.0",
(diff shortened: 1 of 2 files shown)


ankurk91:
The package `dotenv` will be installed by laravel-mix automatically
https://github.com/JeffreyWay/laravel-mix/blob/049485f8d0f251355bd859b63cd9418b89f45755/package.json#L43

BTW - I am not in favor of this hack

PR #4560, opened 2018-01-26
Update to jQuery 3.3

AGhasemzadeh:
Hello,
I think we can use jQuery 3.3.
Also we need to rebuild assets.


GrahamCampbell:
The originally version constraint already resolves 3.3, why change it?

taylorotwell:
This constraint already supports 3.3.

PR #4562, opened 2018-01-26
Semicolon end of line

gizburdt:


a-komarev:
Why is it downvoted if all the rest lines ends with semicolon in this file?

gizburdt:
I have no idea!

resources/assets/js/bootstrap.js
```diff
@@ -43,7 +43,7 @@ if (token) {
  * allows your team to easily build robust real-time web applications.
  */
 
-// import Echo from 'laravel-echo'
+// import Echo from 'laravel-echo';
```
yoosuf: does it really matter?

a-komarev: JavaScript works with semicolons at the end of lines as well as without them.

It's all about code style. If you are using semicolon every line end - it does matter. Thats exactly what we have in current file.

gizburdt: What's cool about Laravel (among many other things) is the code styling. In code styling small things do matter I think :)

JediPoker: I am admittedly very new to Laravel but isn't this a comment line?  Do we need to worry about styling comments?  And while this my be the preference for some (to use the optional semi-colon) it is likely not the preference for all.  

When we think of a code/feature/style change shouldn't the change be a) necessary and b) a benefit for the majority of the community?  

I don't see how this meets either point

a-komarev: But look below, there are 2 commented out definitions with semicolons at the end:

```js
// window.Pusher = require('pusher-js');


// window.Echo = new Echo({
//     broadcaster: 'pusher',
//     key: 'your-pusher-key',
//     cluster: 'mt1',
//     encrypted: true
// });
```

PR #4564, opened 2018-01-30
AWS SES region to env

pavinthan:
AWS SES available for 3 regions so, It's good to move to env


taylorotwell:
You can make this change in your own application.

PR #4527, opened 2017-12-24
Redis Database Env upgraded

BTaskaya:
Make REDIS_DATABASE configurable.

PR #4571, opened 2018-02-11
Load `database.redis.default.database` from `.env` file.

ModernArtisan:
In some cases, conflicts may arise when several applications works on the same server and uses the same redis DB.
Related to this issue - https://github.com/laravel/horizon/issues/279


sisve:
What happens if there's a conflict? How would someone know if they are affected by this?

a-komarev:
Related: #4502 #4527

a-komarev:
I don't really understand why this variable so unfavorable. There are a lot of other things which already configurable and not useful for all people. And in case of Redis database I've described main reason of adding this variable. If several applications are hosted on same machine - there will be conflicts with Queues and other third party packages. Especially this applicable for small startups with one server for staging and production applications.

fletch3555:
Your app owns the file, you're free to make the change there if you need it.  @taylorotwell has noted in the past that he doesn't want to change the configs anymore (except for adding new options I imagine)

PR #4576, opened 2018-02-15
Update EventServiceProvider.php

vuchastyi:
Not needed `use` statement.


taylorotwell:
It's there for convenience because it is often used.

PR #4582, opened 2018-02-19
Update faker version

micos7:


deleugpn:
Won't 1.7 already be pulled by the current constraint?

fletch3555:
Yes it will @deleugpn

pavinthan:
@micos7 `~1.4` will pull `~1.7`, no need the changes, please see [Tilde Version Range](https://getcomposer.org/doc/articles/versions.md#tilde-version-range-)

PR #4583, opened 2018-02-20
Update axios

nikolaynizruhin:
Update axios to 0.18


fletch3555:
`^0.17` includes `0.18`.  No need to change it

nikolaynizruhin:
Hmm...when I run yarn upgrade I get 0.17.1

fletch3555:
perhaps something else is preventing 0.18 from being found?  I'll admit I'm not very familiar with node packaging.

Version format reference: https://stackoverflow.com/a/25861938

PR #4586, opened 2018-02-21
Excluded from saving the repository *orig files

andrey-helldar:
Files with the extension `*.orig` are created when trying to combine problem commits.


michaeldyrynda:
This kind of exclusion is better placed in your own [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

PR #4587, opened 2018-02-21
[5.6] Add nunamaduro\collision Listener to default phpunit.xml

DenzoNL:
Hey all,

Since `nunamaduro\collision` is now included by default in Laravel 5.6 (and recently updated to `~2.0`, which supports `phpunit ~7.0`), I thought it would make sense to add it to the phpunit configuration by default, so everyone gets those sweet exceptions and stack traces when running tests.

The reason I've sent this PR to the `master` branch instead of `5.6` is because it stops `phpunit` from running when `nunamaduro\collision` is still at version `~1.1`, and I assume many early adopters still have that version set in their `composer.json` file.

I've tested this change against the latest `master` commit after running `composer install` and the configuration change seems to work fine and causes no issues.

Example screenshot:
phpunit-collision

I understand if there's hesitation to add yet another default setting based on a dependency not everyone might use, but I think it's a sensible decision given that the default Laravel install already installs `phpunit` and `collision`.

Thanks in advance!

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -28,4 +28,7 @@
         <env name="SESSION_DRIVER" value="array"/>
         <env name="QUEUE_DRIVER" value="sync"/>
     </php>
+    <listeners>
+        <listener class="NunoMaduro\Collision\Adapters\Phpunit\Listener" />
+    </listeners>
 </phpunit>


laurencei:
Anyone installing a new copy of this package would get collision v2.0 - so
there is no breaking change issue with this PR. No reason it can’t go to
5.6 (if it is to be accepted).

nunomaduro:
I am using the collision Phpunit listener for all my projects. But I never had the opportunity of testing it in some edge cases, such as different CI tools or different test suites configurations.

taylorotwell:
I would prefer this aspect of collision to be opt-in.

PR #4591, opened 2018-02-26
Update services.php

nicolas-lutic:
To able to add the region conf in the .env


taylorotwell:
You can do this in your own application if necessary.

PR #4592, opened 2018-02-26
Update hashing.php

ericdowell:
Add the ability to set the hash driver from .env file.


taylorotwell:
You can do this in your own application as needed.

PR #4597, opened 2018-03-01
APP_NAME from .env file

ersanserkan:


taylorotwell:
I think it's fine as is.

PR #4598, opened 2018-03-03
Added view route for welcome page.

themeyellow:


decadence:
This was declined thousands of times

warlight:
Tell me, why?) I don;'t like example that exists

brayniverse:
Because Taylor uses this closure route for testing.

PR #4601, opened 2018-03-08
No closure based route by default to allow Route caching

gauravmak:
Many projects do not touch API routes at all and this prevents Route caching on the server without removing it.

### DIFF
--- a/routes/api.php
+++ b/routes/api.php
@@ -13,6 +13,4 @@ use Illuminate\Http\Request;
 |
 */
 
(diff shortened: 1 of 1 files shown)


deleugpn:
Just remove it from your project?

gauravmak:
I do, everytime. Just thought it would better if it's not closure based by default.

robindirksen1:
@gauravmak if this get accepted you should also add the APIController to the repo, it doesn’t exists.

PR #4603, opened 2018-03-10
[5.6] Improving page load

votemike:
The fonts requested via the Google stylesheet take time to load and a fair amount of time to lookup the DNS. So prefetch the DNS for the font files.
If you load the Google CSS file via the `@import` in app.css, then you have to wait for the CSS file to be downloaded and parsed before the file starts downloading. The Google CSS file should be downloaded in the HTML so that it can be downloaded in parallel with app.css. (Another option would just be to preload the Google Stylesheet from the HTML and leave the CSS reference in app.css)

See corresponding PR in laravel/framework: https://github.com/laravel/framework/pull/23479

### DIFF
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -1,7 +1,4 @@
 
-// Fonts
-@import url("https://fonts.googleapis.com/css?family=Raleway:300,400,600");
-
 // Variables
 @import "variables";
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Doesn't have same font weights available?

votemike:
My thinking behind this PR was that if a user would like to use the 300,400,600 fonts then they should add a link tag for "https://fonts.googleapis.com/css?family=Raleway:300,400,600" in their template file in a similar way to this: https://github.com/laravel/framework/pull/23479/files#diff-93ede835a0f5c2ef4bab2b523a85ac36R18

PR #4605, opened 2018-03-11
[5.6] Autoload migrations as well

krmroland:
Before this PR, running `php artisan make:migration create_users_table` more than once would create duplicates of the users table migration  

This is apparently because the migrations are not autoloaded and the `ensureMigrationDoesntAlreadyExist` method of `Illuminate\Database\Migrations\ MigrationCreator::class` never throws am exception even when the class exists.

This Pr will autoload the database/migrations to fix that

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
     "autoload": {
         "classmap": [
             "database/seeds",
-            "database/factories"
+            "database/factories",
+            "database/migrations"
         ],
         "psr-4": {
             "App\\": "app/"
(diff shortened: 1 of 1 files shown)


brunogaspar:
This was removed here https://github.com/laravel/laravel/pull/4340 so we should further discuss what should happen, either autoload or not or to implement another solution to detect the issue you're having.

krmroland:
@brunogaspar ,Probably I thought it was a bug since `Illuminate\Database\Migrations\MigrationCreator` still performs some check to ensure that the migration doesnt exist.
```
 protected function ensureMigrationDoesntAlreadyExist($name)
    {
        if (class_exists($className = $this->getClassName($name))) {
            throw new InvalidArgumentException("A {$className} class already exists.");
        }
    }
```

krmroland:
How bad can requiring once all migrations files be? https://github.com/laravel/framework/pull/23498

PR #4607, opened 2018-03-13
[5.6] Set MAIL_DRIVER to array in phpunit.xml

shadoWalker89:
Set the array driver as a default for testing to avoid sending emails with smtp the default driver which slows down tests

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -27,5 +27,6 @@
         <env name="CACHE_DRIVER" value="array"/>
         <env name="SESSION_DRIVER" value="array"/>
         <env name="QUEUE_DRIVER" value="sync"/>
+        <env name="MAIL_DRIVER" value="array"/>
     </php>
 </phpunit>


ludo237:
What about `<env name="MAIL_DRIVER" value="log"/>`? is it faster or slower?

GuidoHendriks:
@ludo237 Why would you need the emails of your tests being logged?

ludo237:
@GuidoHendriks it can be useful when you need to verify the content of the email

GuidoHendriks:
@ludo237 That's not something you'd want to use logging generally, I'd say. And definitely won't make a good default setting in my opinion.

shadoWalker89:
My guess is during tests you don't need to open the log to check the email because your have your tests checking them for you.
In theory array driver is faster then log because of the i/o operations of the log driver.
I guess when you want to work on the ui of your email you should render the mailable in the browser or use a service like mailtrap

PR #4610, opened 2018-03-14
Update Bootstrap.js

darkwebside:
Hi Taylor, I have found an issue. if you run document.head.querySelector('meta[name="csrf-token"]') on the console, you get null.
It wouldn't be document.querySelector('meta[name="csrf-token"]')

### DIFF
--- a/resources/assets/js/bootstrap.js
+++ b/resources/assets/js/bootstrap.js
@@ -30,7 +30,7 @@ window.axios.defaults.headers.common['X-Requested-With'] = 'XMLHttpRequest';
  * a simple convenience so we don't have to attach every token manually.
  */
 
-let token = document.head.querySelector('meta[name="csrf-token"]');
+let token = document.querySelector('meta[name="csrf-token"]');
 
 if (token) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
When did that head get added? Has it always been there?

taylorotwell:
Both approaches work for me?

darkwebside:
It is a little bit strange, because in a clean Laravel installation in my laptop, both approaches
 works fine. But in my current project which is using DataTables with Bootstrap 4 in my mac desktop. I´m getting this odd error. I attach you a capture.
error

ludo237:
@darkwebside based on the screenshot I am assuming you are using Firefox. This is mine using debian-based docker container on Windows 10 

image


Maybe something on your side?

darkwebside:
Hi @ludo237  Probably the problem is only in this project setup, because I'm experimenting that odd behavior in all browsers installed in this mac machine with. As I said in a clean current Laravel installation on other machine there was no prob. I´m using this panel  [https://coreui.io/](https://coreui.io/) (html version) which is using gulp and I'm using Laravel mix to bring all js libraries into my project.

PR #4622, opened 2018-03-29
[5.6] Update app.php to respect PHP date.timezone

dancryer:
This change updates `app.php` to use the following line instead of setting the timezone to UTC by default.

`'timezone' => env('APP_TIMEZONE', date_default_timezone_get())`

Fix for: https://github.com/laravel/framework/issues/23743

The end result here is the same. Laravel would respect the server's timezone setting, but if it was not set, the default is UTC anyway.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -65,7 +65,7 @@ return [
     |
     */
 
-    'timezone' => 'UTC',
+    'timezone' => env('APP_TIMEZONE', date_default_timezone_get()),
 
     /*
     |--------------------------------------------------------------------------


GrahamCampbell:
I don't think we should do this. The app should control it's own timezone.

dancryer:
@GrahamCampbell I agree the app should be able to control it, hence adding the `APP_TIMEZONE` .env variable.

This is just about giving the framework a sensible default value. Most servers are configured for the timezones they are designed to serve, and `date_default_timezone_get()` will default to UTC if they are not.

The current setting is particularly problematic in the UK. During the winter, UK time is UTC, so you could happily build your app under the assumption that Laravel is properly using the PHP `date.timezone` setting, only for the clocks to change to BST in March and your dates to all be an hour out. 

With the above change, that problem goes away, without really changing the status-quo for most developers at all.

GrahamCampbell:
I think a sensible default is one that is app controlled, and not derived from the system like that.

dancryer:
@GrahamCampbell I couldn't disagree more strongly. From a purely academic standpoint I understand your point, but being pragmatic about it, this just causes unnecessary problems.

The current behaviour puts all of the burden on the developer to notice an obscure setting in the `app.php` file, and leads to confusion when the clocks change and things are suddenly broken, but all of your other non-Laravel apps are fine (because they respect the server's timezone settings.)

PHP has the timezone functionality for a reason, and even goes as far as to warn you if you don't have a timezone set in php.ini.

As far as I can tell, both Symfony and Zend Framework rely on PHP's `date.timezone` setting by default, so it doesn't seem like an unreasonable argument to be making.

GrahamCampbell:
I think if people want this in their apps, they'll just edit the config. I think the existing default is best.

ludo237:
It's no an obscure setting ... it's like one of the first file you should check on a new laravel installation 😕

EdanBrooke:
Those who come from a hardware/networking background and maintain LEMP/LAMP stacks would expect the app to use the timezone already specified in the PHP configuration for the server. Laravel should default to that timezone - it's a global setting.
If you want to change it by app, you can do that by editing `app.php`. Further, as a lot of settings in `app.php` are extracted as environment variables to `.env`, it makes good sense to also extract this timezone setting to `.env` as OP suggested. All **ENVIRONMENT** configuration should be done between `.env` and the web server itself.

himalnischal:
how to fix such type of error?
suresh@suresh-Inspiron-3442:~/Desktop/laravel/blog$ php artisan serve
PHP Warning: require(/home/suresh/Desktop/laravel/blog/vendor/autoload.php): failed to open stream: No such file or directory in /home/suresh/Desktop/laravel/blog/artisan on line 18
PHP Fatal error: require(): Failed opening required '/home/suresh/Desktop/laravel/blog/vendor/autoload.php' (include_path='.:/usr/share/php') in /home/suresh/Desktop/laravel/blog/artisan on line 18

EdanBrooke:
A pull request isn't a support forum. I suggest you read the getting started tutorials.

`composer install`

PR #4625, opened 2018-04-04
Edited .gitignore

hivokas:
As for me, these folders & file shouldn't be committed.
What do you think about this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,10 @@
 /node_modules
 /public/hot
 /public/storage
+/public/css
+/public/js
+/public/vendor
+/public/mix-manifest.json
 /storage/*.key
 /vendor
 /.idea


a-komarev:
I'm sure you will get an answer that this file is under your control in your own application and you are free to change it for your requirements.

dpde:
This might work with your project but probably not for everyone.
I worked on many projects where all frontend assets are compiled locally to not rely on any build process during deployment (which might fail and brake the deployment). In that case, the assets in /public are required.

a-komarev:
In my current project there are no assets at all. It's only API application.

rappasoft:
Yeah but that doesn't mean it has to change for everyone..

taylorotwell:
I honestly like committing them. Hate compiling and stuff on the production server.

nunomaduro:
@hivokas Not everyone compiles this stuff on the production server.

PR #4626, opened 2018-04-04
add lang pt-BR

fabiojaniolima:


zaknes:
From what I understand, translations are kept out of this repository because it was tough to maintain in the past. I believe that [this repository](https://github.com/caouecs/Laravel-lang) is used for translated versions of the base Laravel application.

fabiojaniolima:
I understand that the translations are kept in an external repository. Thanks for the explanation.

PR #4628, opened 2018-04-08
[5.6] Remove "stopOnFailure" from phpunit.xml

carusogabriel:
"stopOnFailure" is `false` [by default](https://phpunit.de/manual/current/en/appendixes.configuration.html).

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -6,8 +6,7 @@
          convertErrorsToExceptions="true"
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Hmm, I don't think this should be removed. Its presence makes it super easy to toggle.

carusogabriel:
@GrahamCampbell Maybe the other way, bring common configs to be easily toggled?

GrahamCampbell:
I think what's currently there is fine tbh. I don't think we need to change anything.

PR #4631, opened 2018-04-09
Update Bootstrap to 4.1

socieboy:


GrahamCampbell:
Does the `^` operator not already resolve according to semver, so there's nothing to do here?

socieboy:
@GrahamCampbell  got it!

PR #4635, opened 2018-04-12
[5.7] Maintain ANSI output

JayBizzle:
Maintain ANSI output when `artisan package:discover` is run through composer


GrahamCampbell:
What if composer was run in no-ansi mode?

jhoff:
Confirmed that if composer is run with --no-ansi, the output of `package:discover` is still in ansi.

taylorotwell:
See above.

PR #4637, opened 2018-04-14
Allow set configuration timezome, locale from environment

thanhphuongat:


cyrildewit:
This was rejected already for a number of times: #4531, #4520, #4325, #4134, #4034, #4018, #3882, #3792, #3764, #3713, #3508.

GrahamCampbell:
Thanks for the PR, but this can be added to your own app if you want. The config files are designed for editing. :)

PR #4640, opened 2018-04-17
Add new Environment variables

guumo:
Add new Environment variables for timezone, language, and fallback_locale.


overtrue:
:-1:

cyrildewit:
This was rejected already for a number of times: #4531, #4520, #4325, #4134, #4034, #4018, #3882, #3792, #3764, #3713, #3508 and #4637.

Quouting some answers:

> You own the app, you can change these if needed.

>Feel free to add this to your own app.

> You can add this to your own project.

And so on...

No offense 😄 !

guumo:
Hi! Thanks for reply! :)

I don't understand the reasons for reject the proposal. 

> You own the app, you can change these if needed.
> Feel free to add this to your own app.
> You can add this to your own project.

In the config/app, multiple variables using the environment variables to get value. 
What is the difference between the "apo name" and "app timezone"?

According to that criterion, I also own the application to change the name of the application

Thanks!

nicknick-io:
If we added every single variable to the configs, the code would get
bloated really fast.

Just because you need it, doesn't mean that everybody else needs it.

That is why your proposals get rejected.

DanielGuerrero94:
A recent pull request #4629 was merged and i think it makes sense because in config/services all other keys were already looking in .env. But if that's the criterion i don't get it, if i make a pull request for every key in config it gets bloated as @nicknick-io said and if i add only one will be rejected.

GuidoHendriks:
You probably have spent more time in this PR than you’d ever spend on adding it to all your future projects projects manually. 💁🏻‍♂️

guumo:
It is probable. It is also likely that with this way of thinking, laravel would never have existed, since my PR is not for my projects, it is a criterion for the benefit of laravel.

GuidoHendriks:
Spending time on changes like this that don’t solve any problems isn’t very productive. If this would be merged, we could as well add all of the config variables to `.env`. That doesn’t benefit Laravel at all.

GrahamCampbell:
Thanks, but this really was intentionally not provided out of the box.

PR #4645, opened 2018-04-20
Unused import

ForzaSFerrari:
As long as various things will not be used, before the developer imports it, it should not be included in the files.


GrahamCampbell:
This was included on purpose, so it can be later used.

PR #4657, opened 2018-05-09
Feature dynamic home route

dionysiosarvanitis:
If we define a default home route we can avoid the necessity of the hardcoded route in `redirect('/home')` and also avoid having a nonexistent route in the code.


bbashy:
I always forget about this.

dionysiosarvanitis:
Even if you still have to modify some code for a route definition, `web.php` is the best place to do it, especially compared to a middleware! You should also consider the fact that it solves the problem of a route that should have been defined somewhere, but it is not.

best regards

PR #4660, opened 2018-05-15
Add setter for password and remove hashing from register user fucntion

shahvirag07:
Instead of user hashing at time store user record, use setter for password hash.


jmarcher:
Why the extra steps? I do not think this is necessary...

shahvirag07:
What if laravel app give setter be default in User model than no need of extra step.
It will helpful when password need to set from anywhere else (by super-user role), at that time m=no need to encrypt.

PR #4662, opened 2018-05-19
Update .editorconfig to use two spaces for styles

scrubmx:


bloggervista:
Yes I also hate 2 space. Due to callback hell js people are using 2 space but in PHP no callback hell. And Due to async await I guess its time for people to move from 2 space to 4 space. 2 space looks so bad to my eyes :(

scrubmx:
You guys are aware the current [app.scss](https://github.com/laravel/laravel/blob/0869d963aa96e2a9d4ab58cb451e35e9a31b07d9/resources/assets/sass/app.scss) uses two spaces right?

PR #4664, opened 2018-05-21
Configure axios headers for Authorization via API

JeyKeu:
This will give a pre-configured axios headers with `api_token`.
Since the `users` table does not have an `api_token` field we have left this configuration commented.
All the we have to to do is add the `api_token` field to the `users` table and we are all set to use axios with authorized API requests.


JeyKeu:
Related PR https://github.com/laravel/framework/pull/24264

taylorotwell:

Thanks!

PR #4669, opened 2018-05-28
group namespaces classes

samuelsonbrito:
Grouping included classes with "use"


GrahamCampbell:
Thanks for the PR, but our CI will actually revert this if it were merged, to restore the other code style: https://github.styleci.io/analyses/XVL7pw.

PR #4670, opened 2018-06-01
Add iMi digital new Patreon Sponsor to readme

amenk:


amenk:
Ref: http://patreon.com/imi_digital

PR #4672, opened 2018-06-03
[5.6] Add global mail recipient config

caiquecastro:
There is this option on the docs, but it's not present on laravel base project. If you are OK with this PR, I can work on some adjusts for the docblock.


taylorotwell:
I don't really find this option of much use by default? In development shouldn't you be using some mail trap or log driver or something like that?

caiquecastro:
I was thinking about using it to send contact emails from contact us page without explicitly declaring it.

PR #4673, opened 2018-06-05
[5.6] Added french localization

Max13:
Unlike package developments, I think that translations are critical enough and likely to be done again and again by any non English project, to either be included in the main repo or have an official one.

If this PR is rejected, I suggest that the team creates an official translation repository, to avoid having to choose between X and Y repos when they should or must be the same.

### DIFF
--- a/resources/lang/fr/auth.php
+++ b/resources/lang/fr/auth.php
@@ -1,0 +1,19 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Lignes de traduction de "Authentification"
+    |--------------------------------------------------------------------------
+    |
+    | Les lignes suivantes sont utilisées durant l'authentification
+    | pour afficher divers messages à l'utilisateur. Vous êtes libre de modifier
+    | ces lignes de langue suivant les besoins de votre application.
+    |
+    */
+
+    'failed' => 'Ces identifiants sont incorrects.',
+    'throttle' => 'Trop de tentatives de connexion. Veuillez réessayer dans :seconds secondes.',
+
+];
--- a/resources/lang/fr/pagination.php
+++ b/resources/lang/fr/pagination.php
@@ -1,0 +1,20 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Lignes de traduction de "Pagination"
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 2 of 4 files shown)


markwalet:
I was just creating an issue on the `laravel/issues` repo to suggest the idea of a official translation repo.

I personally don't think that you should install every language by default. This only results in more files I have to delete from a fresh install.

My issue:
https://github.com/laravel/ideas/issues/1201

GrahamCampbell:
btw: https://github.com/caouecs/Laravel-lang

Max13:
I’ve done this PR because I added fr localization to the `spark-aurelius` repo, and it was merged pretty quickly. The comments I’ve written here was the same as I wrote for `spark`.

@markwalet: If you’re worrying about extra unnecessary files, with an official repository the issue would be the same, you will likely have to install all lang files even if you only need one. But in either way, an artisan command could easily be used to clean the unnecessary files.

I still think it would be better to include all Lang files in the default repository, because Lang files can be customized (they’re in the user land, not a « dependency »).

markwalet:
If you take a look at my issue you see that I want to make it possible to install a single language. That should be possible and I'm planning on implementing that myself :)

Max13:
I did, and saw that. Personally, I still prefer them in the main repo.

If my PR is rejected, My second opinion is to have an official repo, so I’ll follow you and help you if you need

markwalet:
Thanks! I personally prefer the `make:auth` style. But you already knew that by now haha :)

taylorotwell:
It creates too many PRs to the core to fix things I have no way of verifying.

SimonJ:
@taylorotwell But you merged it on Spark https://github.com/laravel/spark-aurelius/pull/154

Max13:
@SimonJ I understand @taylorotwell's decision even if he had merged it on `spark-aurelius`.

`Spark` and `Laravel` are different products, `Spark` aims at easing users starting up a new billable project. As a paid project, I imagine there is a LOT less PRs and users as `laravel` framework. Translations can be contained and monitored on `spark` but it will be "polluting" here.

At least, that's what I understood from @taylorotwell. Now I hope the team will really consider laravel/ideas#1201

ludo237:
One way to verify this is to create a dedicated team with one or two people for each language. What do you think?

Max13:
@ludo237 I personally agree. Now, we have to wait their decisions on laravel/ideas#1201

PR #4674, opened 2018-06-05
Configuration for sending logs to specific slack channel

pankaj-garg:
Added a configuration parameter in logging.php, so that it can be set from .env file to send logs to different channels.

### DIFF
--- a/config/logging.php
+++ b/config/logging.php
@@ -57,6 +57,7 @@ return [
             'username' => 'Laravel Log',
             'emoji' => ':boom:',
             'level' => 'critical',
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think you can configure this in your own application if you need it. Typically incoming webhook hooks are already configured for a given channel.

pankaj-garg:
@taylorotwell Actually this is already implemented to read from config https://github.com/laravel/framework/blob/5.6/src/Illuminate/Log/LogManager.php#L281

I thought to make it configurable for different environments.

PR #4675, opened 2018-06-06
Add currency settings to config/app.php

Max13:
I think currency settings should be in the config. If not in `laravel/laravel`, at least in `laravel/cashier`.

It will be easier to automatically set `Cashier::useCurrency()` in `CashierServiceProvider::boot()`, also easier to add it as a custom `VueJS` property `VueJS`  in `Spark`.

If rejected, please indicate if it could be merged if I create a PR to `laravel/cashier` adding a config file which will have to be published to allow user to set currency.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -95,6 +95,27 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Application Currency Configuration
+    |--------------------------------------------------------------------------
+    |
+    | The application currency determines the default currency that will be used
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I don't think this is necessary to include in the default distro for everyone tbh.

Max13:
@GrahamCampbell Thanks for the input. That's why I also suggest creating a vendor config file in `Cashier`. Do you think it may be merged? Or should I open a `[Proposal]` first on `Cashier`?

PR #4676, opened 2018-06-07
Remove unused constant LARAVEL_START

rellect:
It's not used.
Also the start time is available globally via $_SERVER['REQUEST_TIME_FLOAT'] if needed.

### DIFF
--- a/artisan
+++ b/artisan
@@ -1,8 +1,6 @@
 #!/usr/bin/env php
 <?php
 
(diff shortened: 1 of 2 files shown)


ludo237:
I forget what is the purpose of `LARAVEL_START` to be honest.

nunomaduro:
@ludo237 Yeah. Is useful for profiling proposes. Anyway, that would be a breaking change.

PR #4677, opened 2018-06-07
Specify required PHP extenstions in composer.json

gauravmak:
A bunch of PHP extensions are required for Laravel as specified in server requirements (https://laravel.com/docs/installation#server-requirements).

We can specify them in the require config as a Platform packages (https://getcomposer.org/doc/01-basic-usage.md#platform-packages) to provide instant feedback while installing the framework.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,6 +6,13 @@
     "type": "project",
     "require": {
         "php": "^7.1.3",
+        "ext-ctype":"*",
+        "ext-json":"*",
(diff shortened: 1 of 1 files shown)


ankurk91:
If I remember correctly composer can not detect extension installed via pecl.
Is your solution cross platform?

GrahamCampbell:
Thanks, but dependencies are always resolved transitively, so there's no need to re-specify them.

PR #4678, opened 2018-06-07
[5.7] Add a custom generator.php config file.

lucasmichot:
Let the user override the namespaces, for all generator commands (`artisa make:*`).

For instance if `'Model'` is set to `'App\Models'`, all models generated will be created in the `\App\Models` namespaces, which means in most setup in `app/Models`

If not value is set, the path will use the fallback default set in the custom generator command.

---

This PR requires https://github.com/laravel/laravel/pull/4679

### DIFF
--- a/config/generator.php
+++ b/config/generator.php
@@ -1,0 +1,23 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Generator paths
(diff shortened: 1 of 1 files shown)


marky291:
This removes typehinting?

lucasmichot:
See https://github.com/laravel/laravel/pull/4680

PR #4680, opened 2018-06-07
[5.7] Allow to set namespaces for each generated class

lucasmichot:
Even if Laravel comes with sensible defaults, the developer sometimes want to have a different structure for a project.

For instance, putting models either in `\App` either in `\App\Models` has been a long discussed topic.

This PR enables the user to override paths for each generated files in a custom `generator.php` configuration file.
If nothing is set, the many generator commands will use their own defaults.

This works with everything but the migrations files.

---

This PR requires https://github.com/laravel/framework/pull/24501

### DIFF
--- a/config/generator.php
+++ b/config/generator.php
@@ -1,0 +1,23 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Generator paths
+    |--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


taylorotwell:
I can already do php artisan make:model Models/Foo though? I'm not sure I want to take this approach right now.

PR #4683, opened 2018-06-09
[5.6] Add env variable for redis client

hkp22:
I wanted to use `phpredis` instead of `predis` in one of my laravel project. It took me a couple of hours to find out where to tell the Laravel to use `phpredis` instead of `predis`. 

I have seen similar questions on stackoverflow too. 

So, I believe env variables for `REDIS_CLIENT` would be helpful at least for new laravel developers.


m1guelpf:
You own this files. You can make the change in your own project.

hkp22:
@m1guelpf  @GrahamCampbell Yes, I know I can make this change in the project. But I thought if there was env variable for redis client, It would not have taken a couple of hours to find out where to specify phpredis.  So, I believe env variable should be there.

overtrue:
@hkp22 I don't need this change. Thanks.

rajneet-rw:
I agree with @hkp22

gmsantos:
Hey @hkp22 this information is already on docs https://laravel.com/docs/5.5/redis#introduction

Do you think is needed more information on config class? I think that the supported clients doc line could help with that, without putting the environment var. 

Any thoughts @GrahamCampbell ?

PR #4684, opened 2018-06-12
[5.7] rename parameter of Daily Driver logger from days to max_files

allok:
It will be more clear to rename parameter 'days' of Daily Driver logger to 'max_files' because it show how many files will keep on disk.
Also it supports '0' value for unlimited files.

Related to https://github.com/laravel/framework/pull/24562

### DIFF
--- a/config/logging.php
+++ b/config/logging.php
@@ -48,7 +48,7 @@ return [
             'driver' => 'daily',
             'path' => storage_path('logs/laravel.log'),
             'level' => 'debug',
-            'days' => 7,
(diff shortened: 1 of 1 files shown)


marky291:
Each day is its own file ?

Believe this would require refactoring in the core framework also for the new name change.

allok:
@marky291 No, parameter 'max_files' affect only on deleting old files. And this pass only once by method parameter.

taylorotwell:
I don't see a big point in this breaking change.

PR #4685, opened 2018-06-12
Use auth() and validator() helper when possible.

lucasmichot:
Use `auth()` and `validator()` helpers when possible; and save some imports.


taylorotwell:
I'm really fine with the facades here.

PR #4688, opened 2018-06-20
[5.6] Using environment variable set app timezone

medz:


Te7a-Houdini:
i think you can do this in your application

michaeldyrynda:
This’ll get closed, but timezone isn’t typically something that changes per-environment anyway, so I don’t see much value in having it in the `.env` file anyway.

medz:
I am constructing an open source program image based on Laravel's implementation. There is no guarantee that every user will use UTC. Just like the database configuration information of `.env`, no one will change it from time to time! But if the configuration in the out of construction is nothing wrong!

medz:
@Te7a-Houdini @michaeldyrynda I do not agree with your point of view. If you take this point of view, most of Laravel's `.env` is meaningless. The purpose of environment variables is to perform some configuration processing during the development of the program or at the beginning of the build. After that we will use `config:cache` for cache configuration, so it does not affect!

medz:
@Te7a-Houdini I think you are right, I should implement this in my program! I've added it to my program and I'm happy to work. Building a generic Docker image is also great!

Te7a-Houdini:
@medz  Good Luck :+1:

GrahamCampbell:
I agree that this isn't something that is usually environment specific, so this PR should not be merged. Instead, your should edit and commit the config in your own app, for everything that is not environment specific.

PR #4689, opened 2018-06-21
Email Verification (Skeleton)

taylorotwell:
This PR contains the application skeleton changes needed for optional email verification.

To use:

`php artisan make:auth`

`Route::auth(['verify' => true]);`

`class User implements MustVerifyEmail`

### DIFF
--- a/app/Http/Controllers/Auth/VerificationController.php
+++ b/app/Http/Controllers/Auth/VerificationController.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace App\Http\Controllers\Auth;
+
+use Illuminate\Http\Request;
+use Illuminate\Routing\Controller;
+use Illuminate\Foundation\Auth\VerifiesEmails;
+
+class VerificationController extends Controller
+{
+    /*
+    |--------------------------------------------------------------------------
+    | Email Verification Controller
+    |--------------------------------------------------------------------------
+    |
+    | This controller is responsible for handling email verification for any
+    | user that recently registered with the application. Emails may also
+    | be resent if the user did not receive the original email message.
+    |
+    */
+
+    use VerifiesEmails;
+
+    /**
+     * Where to redirect users after verification.
+     *
+     * @var string
+     */
+    protected $redirectTo = '/home';
+
+    /**
+     * Create a new controller instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        $this->middleware('auth');
+        $this->middleware('signed')->only('verify');
+        $this->middleware('throttle:6,1')->only('verify', 'resend');
+    }
+}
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/Auth/VerificationController.php
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Create a new controller instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        $this->middleware('auth');
+        $this->middleware('signed')->only('verify');
+        $this->middleware('throttle:6,1')->only('verify', 'resend');
```
jpinedapetplan: what is this 6, 1?

bloggervista: may be max 6 request per minutes?

taylorotwell: You can call those routes 6 times per minute.

database/migrations/2014_10_12_000000_create_users_table.php
```diff
@@ -17,6 +17,7 @@ public function up()
             $table->increments('id');
             $table->string('name');
             $table->string('email')->unique();
+            $table->boolean('email_verified')->default(false);
```
m1guelpf: What about a timestamp here (confirmed_at)? You can check using a getter, and you get to know the consent date

arcanedev-maroc: Or `verified_at` ?

taylorotwell: What would such a date be used for?

sweet-io: * confirming when a user activated their account
* reporting on the lag between signup and confirmation of account
* finding all users who confirmed their account (and agreed to terms) before a certain important date (terms changed, onboarding changed, etc in an application

jcmurrayii: * confirming when a user activated their account
* reporting on the lag between signup and confirmation of account
* finding all users who confirmed their account (and agreed to terms) before a certain important date (terms changed, onboarding changed, etc in an application

arcanedev-maroc: As @jcmurrayii said, a `date` record has mutliple usages than just `boolean` value.

taylorotwell:
OK I'm fine with changing this to a date.

taylorotwell:
Updated to use a date.

ankitjain28may:
@taylorotwell Are we saving the token sent along with verification mail ?

zaknesler:
@ankitjain28may Nope, since we are using a [signed route](https://laravel.com/docs/5.6/urls#signed-urls).

hivokas:
My first thought after announcement of signed routes was definitely about email verification with usage of them. Like it!

PR #4695, opened 2018-07-05
[5.6] Add config for EnsureRequestIsSecure middleware

dwightwatson:
This is the other side to laravel/framework#24747 - adding the boilerplate configuration for the new middleware. Not sure if you wanted to include the middleware in the HTTP kernel by default or not. I think it would be better to include it because otherwise there will be impact when you change `app.https`.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -56,6 +56,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Force HTTPS
+    |--------------------------------------------------------------------------
+    |
+    | If an unsecured request is received by your app it can redirect
+    | back to the same resource over HTTPS. This prevents you from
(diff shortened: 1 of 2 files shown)


config/app.php
```diff
@@ -54,6 +54,19 @@
...
+    | Force HTTPS
+    |--------------------------------------------------------------------------
+    |
+    | If an unsecured request is received by your app it can redirect
+    | back to the same resource over HTTPS. This prevents you from
+    | serving duplicate content over HTTP and HTTPS schemes.
+    |
+    */
+
+    'https' => env('APP_HTTPS', false),
```
MichaelDeBoey: Maybe also add this to the `env.example` file?

PR #4696, opened 2018-07-06
[5.6] Update composer.json

voyula:
composer create archive without unnecessary files.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,5 +55,8 @@
         "optimize-autoloader": true
     },
     "minimum-stability": "dev",
-    "prefer-stable": true
+    "prefer-stable": true,
+    "archive": {
+        "exclude": [".idea/", "vendor/"]
+    }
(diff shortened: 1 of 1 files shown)


michaeldyrynda:
These are changes you can make in your own app, I think. Not everybody uses PhpStorm, they may need to exclude other directories from their archives. 

Best to keep the project `composer.json` fairly generic.

voyula:
Hmm.

An another way without use composer.json ;
We can ignore this files via .gitattributes "export-ignore" prototype. Composer reading and using .gitattributes file "export-ignore"'s.

This PR isn't unnecessary, i'm using composer archive command and this is problem for me.

michaeldyrynda:
That may be so, but `.idea` is in the `.gitignore` file, so it shouldn't ever end up _in_ your repository, in the first place.

PR #4697, opened 2018-07-07
5.4

HuantanaCorp:
Method [validateIn(0,1)] does not exist.

  $data['state_tax']='0';
        $validator = \Validator::make($data, 
            [ "state_tax" => 'in(0,1)',
]);
dd($validator->passes());

how fix that?? I need to set 'state_tax' with default value - 0,1 -  but validateIn no exit , its weird


michaeldyrynda:
Your usage is incorrect. 

You need to use `in:0,1` in your validation, or use the `Rule` class for fluent construction per the [documentation](https://laravel.com/docs/5.4/validation#rule-in)

PR #4698, opened 2018-07-09
Add beyondcode/laravel-dump-server as dev dependency

mpociot:
This PR adds the https://github.com/beyondcode/laravel-dump-server package as a dev-dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,7 @@
         "laravel/tinker": "^1.0"
     },
     "require-dev": {
+        "beyondcode/laravel-dump-server": "^1.0",
         "filp/whoops": "^2.0",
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Why not just contribute this feature to laravel/framework?

mpociot:
I discussed this with Taylor and we decided it would be best to use the package as a dev dependency. 

@nicknick-io instead of complaining here, why don’t you open a reproducible issue on the actual package repository and explain what’s “damn buggy”?
A lot of people are happily using this already and complaining here is not helping at all.

michaeldyrynda:
@nicknick-io its a dev dependency, so not like it’s making it to production and if you don’t want to use it in dev, you can just...  not.

PR #4705, opened 2018-07-13
[5.7] Allow to use default view for rendering exception

mnabialek:
This PR changes 2 things:

1. It allows to use `default` view for all exceptions except those defined in custom views. Now it's impossible so in case you want to define one view for 404 and one for all the others cases it's impossible - you need to create views for multiple statuses and you are not even sure if you cover all of them or not.

2. Move some logic from Framework to App so it's easier to change the logic.  After the update it's easier to customize this logic because it's enough just to implement custom logic here and all uncovered cases will be handled by default framework logic.

If approved additional changes in Framework can be done to simplify code a bit after moving some logic to app.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -3,7 +3,9 @@
 namespace App\Exceptions;
 
 use Exception;
+use Illuminate\Support\ViewErrorBag;
 use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
+use Symfony\Component\HttpKernel\Exception\HttpException;
 
 class Handler extends ExceptionHandler
 {
@@ -49,5 +51,23 @@ class Handler extends ExceptionHandler
     public function render($request, Exception $exception)
     {
         return parent::render($request, $exception);
+    }
+
+    /**
+     * Render the given HttpException.
+     *
+     * @param  \Symfony\Component\HttpKernel\Exception\HttpException  $e
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    protected function renderHttpException(HttpException $e)
+    {
+        if (view()->exists($view = 'errors.' . $e->getStatusCode()) || view()->exists($view = 'errors.default')) {
+            return response()->view($view, [
+                'errors' => new ViewErrorBag(),
+                'exception' => $e,
+            ], $e->getStatusCode(), $e->getHeaders());
+        }
+
+        return parent::renderHttpException($e);
     }
 }
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -4,8 +4,8 @@ namespace App\Exceptions;
 
 use Exception;
 use Illuminate\Support\ViewErrorBag;
-use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
 use Symfony\Component\HttpKernel\Exception\HttpException;
+use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
 
 class Handler extends ExceptionHandler
 {
@@ -61,7 +61,7 @@ class Handler extends ExceptionHandler
      */
     protected function renderHttpException(HttpException $e)
     {
-        if (view()->exists($view = 'errors.' . $e->getStatusCode()) || view()->exists($view = 'errors.default')) {
+        if (view()->exists($view = 'errors.'.$e->getStatusCode()) || view()->exists($view = 'errors.default')) {
             return response()->view($view, [
                 'errors' => new ViewErrorBag(),
                 'exception' => $e,


laurencei:
But people can just add this functionality themselves if they want?

> If approved additional changes in Framework can be done to simplify code a bit after moving some logic to app.

If that occured - wouldnt that would be a breaking change to *existing* apps - because they would have to add this code, otherwise it'll disppear from the framework and not exist in the old app code?

ankurk91:
@mnabialek
I can understand your pain. Right now i have these error pages. I have to repeat this in each and every project. Because i want same UI for all of the error pages.
There should be a generic way to do this.
screenshot from 2018-07-14 10 48 53


I would like to see a `error.default` template as fallback template in v5.7.

mnabialek:
@laurencei I agree if we make changes in framework it will be breaking change. But this targets 5.7 so breaking changes are allowed I believe and of course we should add this in upgrade guide. 

We can also avoid this to be breaking change for now and we could leave framework code unchanged - it would be a bit of code duplication but for now for upgrading existing applications no change would be required (but then default wouldn't work for them).

Also that's true - they can add it themselves but before they would need to know which method from Framework they had to override. Now this method would be extracted to app (similar as `render` and `report`)  so it would be much easier to adjust it to your needs.

mnabialek:
@ankurk91 The worst thing is that you never know if some other status code won't be used and then none of your custom views will be used at the moment.

taylorotwell:
Why put this in the application skeleton? Why not just put it in the framework code?

mnabialek:
@taylorotwell Probably we can do it also there but probably it would mean a bit more changes because with current version we cannot just use `default` as we are using `errors` namespace and it would still read file from framework, not user defined `default`.  

But probably any of those would be fine, so if you think it's better to implement this in framework probably we can do it also in there.

taylorotwell:
I can't just do this?

image

That will still load from the users resources/views/errors directory.

mnabialek:
From my test it seems it won't work (assuming you haven't touched any other code). If I have only view `errors/default.blade.php` and I'm using invalid url (404), it will still use Framework 404 page and not custom `errors/default.blade.php` for above code.

But of course we cannot just swap conditions, because we would like to still use custom pages for some types of errors. For example in one of app I've defined `errors/404.blade.php` to display custom 404 page and `errors/default.blade.php` to display custom page for any other types of errors

rodruiz:
@mnabialek,
following @taylorotwell code, to fix the issue you have to replace "**::**" with "**.**"
```php
"errors::{$e->getStatusCode()}"
...
"errors.{$e->getStatusCode()}"
```

taylorotwell:
@rodruiz yeah, that solution should work.

taylorotwell:
@mnabialek what do you think about @rodruiz's change?

michaeldyrynda:
Swapping it to `errors.default` would mean that it never uses the framework default, though? Can the defaults be made publishable?

fletch3555:
Can't you put your overrides in the `errors` namespace with something like `vendor/errors/404.blade.php`?  Then `errors::404` would work normally

mnabialek:
@taylorotwell Yes, it would work, but finally it wouldn't use `errors.default` when it exists. The logic order should be: 
- use custom view with status (for example errors/404) when exists
- otherwise verify if errors/default exists - if yes, use it
- otherwise use framework view if it exists
- finally use default exception rendering.

So to sum up, I think we could completely remove `registerErrorViewPaths` path method (it would be too simple), and we could rewrite `renderHttpException` method like this:


```
protected function renderHttpException(HttpException $e)
{
    View::replaceNamespace('errors', [__DIR__ .'/views']);

    $view = Arr::first([
        "errors.{$e->getStatusCode()}",
        'errors.default',
        "errors::{$e->getStatusCode()}",
    ], function ($view) {
        return view()->exists($view);
    });

    if ($view) {
        return response()->view($view, [
            'errors' => new ViewErrorBag,
            'exception' => $e,
        ], $e->getStatusCode(), $e->getHeaders());
    }

    return $this->convertExceptionToResponse($e);
}
```

If you agree, I can make PR to framework then instead of changing code as initially proposed in this PR.

mnabialek:
@taylorotwell  Any decision about this?

laurencei:
The proposed solution feels a bit messy to me - it feels hacky creating a special view that `errors.default` somehow overrides other things. What happens if I want a "catch all except 404" style page? That wont work in the proposed solution.

A simplier "cleaner" solution would be to move all default views out of the framework, into the application. Then you can simply have the system go:

- Look for specific http error file in application
- Look for default error file in application
- Display exception rendering

It immediately solves all of this - and removes the need for any hacks and it is very easy to "reason with". You can then either have a single 'default' error view for everything, or you can use it as a 'catch all except 404' function. The developer is in complete control without changing a line of code.

To make upgrading to from older version to 5.7 - you could have a simple "artisan publish:errors" or something, so all previous installs get the default error files moved to application to prevent a breaking change.

hwmaier:
Just stumbled across this discussion after I read Tayler's tweet about only one PR left open.

I think one should avoid polluting the app with boilerplate code and templates. So I would not want to see having to publish the error pages into the app folder unless they really need to be customized.

Most people are probably OK with just customizing the `404.blade.php` and the `error/layout.blade.php` template which gives all error pages a proper site-specific look and feel.

I never saw a point to customize a 429 error page, I am happy that just a message is output and and that message is yielded into the custom error template so it will look in line with the rest of the site. The issue may be different for other languages where the message text needs to be translated. But this can be achieved by adding the messages to the translation files..

I can see that for other error codes which are not handled by the framework a default error page is sensible, but that again should be part of the framework in the first place and look and feel could be customized via the layout in most cases.

So my vote is for:

- Look for specific http error file in application
- Look for specific http error file in framework
- Look for default` error.blade.php` file in application
- Look for default ` error.blade.php` file in framework
- Extend (as currently done) all framework pages from `errors::layout` which allows customization of presentation while re-using the framework pages
-  Add error messages to lang translation files to allow i8n

based on the principle that specific pages always override generic pages (and not the other way).

taylorotwell:
@laurencei's solution may be the best solution.

stevelacey:
Anyone else who got here because their custom error pages stopped working:

Rename your `errors/layout.blade.php` to `errors/illustrated-layout.blade.php` 👌

PR #4707, opened 2018-07-16
Fixed minimal security problem

voyula:
Options -Indexes moved to global for never directory listing


taylorotwell:
You can modify your .htaccess as needed.

PR #4709, opened 2018-07-18
Use dirname() instead realpath()

rellect:
It's cleaner and should be faster

### DIFF
--- a/bootstrap/app.php
+++ b/bootstrap/app.php
@@ -12,7 +12,7 @@
 */
 
 $app = new Illuminate\Foundation\Application(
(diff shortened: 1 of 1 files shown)


GuidoHendriks:
Very good arguments.

antonkomarev:
@rellect then make benchmarks to prove that its faster.

rellect:
realpath() needs to resolve the path and make sure it exists on the filesystem.
dirname() is not aware of the filesystem and simply operates on the given string.

If you insist on benchmarks I can do it later..

PR #4713, opened 2018-07-23
Update docblocks to match parent class

ejunker:
While testing [Larastan](https://github.com/nunomaduro/larastan) warned that `render()` had the wrong return type based on the docblock. I also updated the docblock for `report()`


taylorotwell:
I don't really understand why `report` would throw an exception.

PR #4715, opened 2018-07-24
[5.7] Add editor option to config

mnabialek:
As mentioned here https://twitter.com/marcelpociot/status/1016784890128470017 there is a hidden editor config that many people are not aware because it's not set directly in `app.php`

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -43,6 +43,20 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | IDE/Editor used for development
+    |--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


rweber7912:
already discussed in  #4704 and #4510

mnabialek:
@rweber7912 Thanks, I haven't noticed it. So I'm closing

antonkomarev:
@mnabialek you could install development package https://github.com/cybercog/laravel-whoops-editor and it will add this key to configuration automagically and control it via `.env` file.

PR #4717, opened 2018-07-27
[5.7] Make timezone config convenient for remote friendly teams 

introwit:
- Makes timezone config convenient for remote friendly teams by allowing each developer on the team to use his/her preferred timezone via the env variable 🙂 
- It is very useful to monitor records in the database during local development for remote devs.

UPDATE: If the creators & maintainers feel that this "non-breaking convenience" isn't worthy, feel free to scrap & close it, no worries ❤️


deleugpn:
https://github.com/laravel/laravel/pull/4688

scoutman57:
My time works remotely and I don't see how this would be beneficial. The "team" should all be working off the same time zone settings. If you have a team across different timezones then it's best to use the unified timezone of "UTC"

introwit:
@deleugpn thank you :) To answer @GrahamCampbell 's answer on that PR:

> I agree that this isn't something that is usually environment specific, so this PR should not be merged. 

The timezone of records created & updated is a part of the development environment. CRON jobs, etc. are dependent to such values. Let me explain with an example: 
- What if the client is from Australia & the development agency is based in the US? Will the developers working in US on that client project do development according to the Australian timezone? 
- What if a product is built on Laravel and the dev team of the product is spread across the globe? I feel timezone is an integral part of the development environment because lot of things in your business logic depends on it 🙂

introwit:
@scoutman57 if that works for your team, thats great :) I guess we aren't losing or breaking anything by making it configurable?

deleugpn:
You can always make that change in your Laravel installation whenever you need it.

introwit:
@deleugpn may I know whats the issue if the base install provides that convenience with the default value of UTC as the current one? :)

scoutman57:
> What if the client is from Australia & the development agency is based in the US? Will the developers working in US on that client project do development according to the Australian timezone?
What if a product is built on Laravel and the dev team of the product is spread across the globe? I feel timezone is an integral part of the development environment because lot of things in your business logic depends on it 

It's funny that you mention this specific example as I work from Texas US for a company in Australia . We use "UTC" for all times in the database as we should. As far as development goes, that has nothing to do with how times are stored. Your individual msyql clients generally show the time in the developers local times. 

I am trying to understand your bullet points: Will the developers working in US on that client project do development according to the Australian timezone? Is this related to working hours or storing datetime records? If storing datetime records, you would store as UTC and display as users local time if required.

"I feel timezone is an integral part of the development environment because lot of things in your business logic depends on it" This comment validates why "UTC" should be the only timezone, if users are across the globe then the system needs 1 timezone. If you need something in business logic like show the time this happened in the "current user" local timezone, then you would just convert that in business logic, not on the system side.

introwit:
@scoutman57 

- Thank you for marking my understanding as "funny" 🙂 

> Your individual msyql clients generally show the time in the developers local times.

By clients you mean "Sequel Pro", etc? As far as I am aware it's not possible, but would love to be proved wrong and take home a new learning :) 

- Also, my whole point as noted in the PR description is to make local development a bit quicker where if I need to interact with times stored in the DB during the local dev, I don't have to calculate stuff. 

- We aren't breaking anything to provide this convenience as the default value that will be shipped is UTC like the existing one 🙂

rrpadilla:
@introwit,
best practice is to save as UTC always.
Then your views has to render the dates using user timezone.
You have 2 options:
1. Get user timezone from database.
2. Get timezone from users browser using javasctipt. 

For example, [Laravel Nova](https://nova.laravel.com/) is getting the timezone from users browser. 
Then if you travel from US to Australia you do not have to update the timezone setting in the database because your app is using the timezone from users browser. .

robclancy:
Having different timezones between development and production is asking for problems let alone different developers. What is used by servers should have no bearing on what a user sees, if you aren't localizing then you're doing it wrong.

taylorotwell:
Feel free to add to your own app if you want. :heart:

introwit:
@taylorotwell is it some sarcasm that means I should stop the 'not so useful PRs'? The tone of the reply and the sweet ❤️ at the end are contradictory confusing 🙈 Its just that I am shooting PRs for everything I am improving while working on my projects, not affected at all by closed PRs 😄Just feel like I should present my idea practically so I create a PR directly 😅

PR #4723, opened 2018-08-02
Use `node_modules/.bin` as more robust storage for exec. node scripts

lptn:
there is a small probability that a path to npm executable scripts will changed. 
`node_modules/.bin/` directory has flat list of symlinks to all  executable scripts and will not be changed

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,12 +2,12 @@
     "private": true,
     "scripts": {
         "dev": "npm run development",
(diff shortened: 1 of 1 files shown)


taylorotwell:
If something actually becomes broken we will change it but since it isn't I see no reason to potentially introduce a breaking change to people's systems.

PR #4724, opened 2018-08-03
Add AWS s3 endpoint to support s3 compatible object storage

stkari:
Before this PR, the s3 filesystems configuration works great for AWS, but using these parameters with S3-compatible object storage from different providers doesn't work. Because the aws SDK automatically of course uses AWS endpoints. Adding the 'endpoint' parameter makes it possible.


michaeldyrynda:
Does this continue to work in the current manner if `AWS_ENDPOINT` is not set and subsequently `null`?

It may be better to document the config value than change the default behaviour.

taylorotwell:
Feel free to add to your own application.

PR #4727, opened 2018-08-14
Changed font to new one

Advaith3600:
Updated the font to Nuntio from Raleway


GuidoHendriks:
Why should this be changed in the framework? 🤔

Advaith3600:
So that there is a uniformity all over the areas.

PR #4729, opened 2018-08-15
Develop

gargighorai:
why vendor folder is missing


gargighorai:
artisan serve is seeking for vendor DIR

ankurk91:
> why vendor folder is missing

Run this command - `composer install`
This will create missing `vendor` folder.

PR #4731, opened 2018-08-16
Localize responses

dmcbrn:
Related to: https://github.com/laravel/framework/compare/5.6...dmcbrn:patch-1

---details---

When creating an API I'm getting untranslated responses. This PR (in combination with the one mentioned above) will extract the 'Unauthenticated' text to language files making it easy to translate it.

### DIFF
--- a/resources/lang/en/auth.php
+++ b/resources/lang/en/auth.php
@@ -15,5 +15,9 @@ return [
 
     'failed' => 'These credentials do not match our records.',
     'throttle' => 'Too many login attempts. Please try again in :seconds seconds.',
+    
+    'responses' => [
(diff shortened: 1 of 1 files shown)


taylorotwell:
Just use the __ helper in the creation of that message.

dmcbrn:
Let me elaborate a bit.

I'm using DataTables js library to present data. If my session expire and I try to do something I get a 401 response. Along with the 401 status response I get the text message saying 'Unauthorized'. Since this text message is included in the response I thought it would be reasonable to offer the localization of that text since localization is included in Laravel. It seems like a consistent way to do it.

PR #4732, opened 2018-08-19
[5.7] test bootstrap

halaei:
This is something I do in almost all of my laravel projects:

1. Create a custom bootstrap for my tests.
2. In that bootstrap create a laravel application and clean up the application state, e.g. refresh the database and clean the cache.
3. Make sure my tests are running in a separate environment, so the state of my manual-testing/development database does not affect the tests, and the tests does not drop my development data.

I am not sure if laravel and phpunit have alternative approach for this, so probably others would like to do the same. This PR adds the bootstrap file, while doing nothing by default, it has some commented codes that might be handy.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit backupGlobals="false"
          backupStaticAttributes="false"
-         bootstrap="vendor/autoload.php"
+         bootstrap="tests/bootstrap.php"
          colors="true"
          convertErrorsToExceptions="true"
          convertNoticesToExceptions="true"
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -1,0 +1,13 @@
+<?php
+
+require __DIR__.'/../vendor/autoload.php';
+
+//$app = require __DIR__.'/../bootstrap/app.php';
+//
+//$app->make('Illuminate\Contracts\Console\Kernel')->bootstrap();
+//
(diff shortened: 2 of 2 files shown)


michaeldyrynda:
Why not just do this in the [CreatesApplication](https://github.com/laravel/laravel/blob/master/tests/CreatesApplication.php) trait?

halaei:
As far as I know, createApplication() is called before every test. It would be slow to migrate:refresh before every test.

michaeldyrynda:
You can add it to the `setUpBeforeClass` method instead of `setUp`, then it does it once per class rather than once per test.

halaei:
Once per class is still more than once 😉

GrahamCampbell:
Thanks for the PR. My feeling is that the current default is best, and this can of course be added to your app if you want. laravel/laravel is designed to be a good starting point, but everyone has different requirements, and so will make changes as they see fit.

halaei:
I think of it as a way to speedup the process  of testing, from developer time perspective. Still the decision is all yours. Thanks for consideration.

halaei:
Find out, there is a relatively new trait `RefreshDatabase` which is introduced for the purpose of running migrations only once. Generally, that is fine, but it doesn't work in one case for me, in which I have to rollback migrations instead of deleting all the tables. I wonder if it is OK to send a PR to add some configurations to `RefreshDatabase` with that regard?

fobin:
Exactly almost same iasue as @halaei I also need to seed so currently I've done it on tearDown I run Artisan command to migrate and seed. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #4733, opened 2018-08-20
Exposed Redis database selection to configuration through .env

unrealization:
This should be useful for environments where multiple applications share a single Redis server.


fletch3555:
This has been proposed a few times before (#4502 #4527) and declined both times.  You're free to add this to your own project.  laravel/laravel is intended as a generic starting point for everyone, and everyone has different needs.

I personally don't see any issue with this change, but merging it in opens the flood gates for a host of other similar PRs that may provide little to no actual benefit to the project.  In either case, it's not really my decision to make

unrealization:
Thanks for the info. I just figured this would be a convenient place to put this as it would less likely to be affected or cause problems than doing it in config/database.php

PR #4735, opened 2018-08-25
Extract 2 core configurations for sessions to environment

danijelk:
In the spirit of JMac & LaravelShift I've extracted 2 session variables to the environment file. Not 'just because' but that one is able to set SESSION_DRIVER to redis, but unable to set a connection without touching the core session config file.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -70,7 +70,7 @@ return [
     |
     */
 
-    'connection' => null,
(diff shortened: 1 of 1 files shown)


danijelk:
Glad you agree @taylorotwell Btw is it ok to make another PR to master to get it into 5.6? As I guessed now that develop is the 5.7 branch whilst master is 5.6.

PR #4738, opened 2018-08-29
add root .htaccess

yuchonghua:


decadence:
Just point your DOCROOT to `public` folder

devilsafe:
nginx php7.2 is ok for this wonderful job?

yuchonghua:
In fact, many virtual host operators do not support modifying the root directory, they need their own vps, and the purchase price of vps is 10 times that of the virtual host.

michaeldyrynda:
There are multiple VPS providers (Vultr, Linode, Digital Ocean) with hosting for as little as $2.50 per month. What virtual host are you paying 25c a month for?

PR #4741, opened 2018-08-31
[5.7] Define mix as const

mattmcdonald-uk:
`mix` isn't re-assigned (and shouldn't be) so `const` is more appropriate.

### DIFF
--- a/webpack.mix.js
+++ b/webpack.mix.js
@@ -1,4 +1,4 @@
-let mix = require('laravel-mix');
+const mix = require('laravel-mix');
 
 /*
  |--------------------------------------------------------------------------


brunogaspar:
@mattmcdonald-uk Not to discourage you or anything, but this was kinda discussed before https://github.com/laravel/laravel/pull/4283

mattmcdonald-uk:
So it has!

Well close if needed, but I'd disagree with the relevance of an immutability argument (which seems to be the focus of the previous discussion).

`const` is an indicator that the variable won't be and shouldn't be re-assigned and makes the code easier for others to follow.

Excellent piece on that here: https://mathiasbynens.be/notes/es6-const

brunogaspar:
Yeap. I also agree with your point, but, we never know the outcome of these pull requests, it's like winning the lottery :)

taylorotwell:
🤷‍♂️

PR #4745, opened 2018-09-04
Use trait and not contract

m1guelpf:


taylorotwell:
You need both for the feature to work actually.

PR #4747, opened 2018-09-05
[5.7] Remove unnecessary use on verification controller

eliurkis:
Remove unnecessary use of the Request on the auth verification controller.


bkintanar:
I think this is placed there so that new devs can use it when needed.

DCzajkowski:
🤷‍♂️

edantonio505:
xdr dd dr. d CDC z xdr x dsz ZERO FSA

PR #4750, opened 2018-09-06
Removing unnecessary use from User Model

kp-syginteractive:
The base class Illuminate\Foundation\Auth\User already uses the MustVerifyEmail trait.

### DIFF
--- a/app/User.php
+++ b/app/User.php
@@ -2,7 +2,6 @@
 
 namespace App;
 
-use Illuminate\Auth\MustVerifyEmail;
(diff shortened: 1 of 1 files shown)


kp-syginteractive:
In the release notes the docs don't mention using the trait - https://laravel.com/docs/5.7/releases. 

And the functionality all works fine even if you do not use the trait since it is already used in the base class.

bbashy:
I'd send a PR to the docs combined with this change then.

PR #4759, opened 2018-09-09
use npx in npm scripts

tianyong90:
NPX is a new and powerful tool, using it can simplify the scripts.


ankurk91:
Not everyone has npm installed, I use yarn.

PR #4761, opened 2018-09-15
Seeded users should be verified by default

gauravmak:
the email_verified_at column value is used to determine whether a user has verified her account.


scoutman57:
I disagree but it should maybe be random.

PR #4762, opened 2018-09-16
Fix 'resent' ambiguity

olipayne:


troccoli:
What was the ambiguity exactly?

fletch3555:
resent - verb - to feel bitterness toward

resent - verb - past tense of resend, to send again

troccoli:
There is no ambiguity, as the passive form is used and the two verbs would be

- To be resented
- To be resent

PR #4763, opened 2018-09-18
Preserve colors

tillkruss:
I like seeing the colors, makes it easier to scan.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,11 +43,11 @@
             "@php -r \"file_exists('.env') || copy('.env.example', '.env');\""
         ],
         "post-create-project-cmd": [
-            "@php artisan key:generate"
+            "@php artisan key:generate --ansi"
         ],
         "post-autoload-dump": [
             "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
-            "@php artisan package:discover"
+            "@php artisan package:discover --ansi"
         ]
     },
     "config": {


nunomaduro:
Nice one.

alish00:
nice
[ساندویچ پانل](https://mammut-group.com/)

phanan:
A similar PR was submitted here https://github.com/laravel/laravel/pull/4635 and closed by @taylorotwell and @GrahamCampbell for (IHMO) valid reasons. By forcing `--ansi` we're basically removing the ability to force `--no-ansi`, which can be useful in certain cases (OTOMH: debug, poor console support, color accessibility). In short, this decision should be users' to make, not Laravel.

tillkruss:
‪Isn't it enough to just remove it for users who need to use --no-ansi? My guess would be that's less than 1%.

phanan:
I'm not sure what you mean by "removing it for users who need to use `--no-ansi`," nor do I have the number to back either of the arguments up. I guess the key point here is (according to linked PR) `--no-ansi` will always be overriden by `--ansi` thus makes it impossible for users to disable colors.

antonkomarev:
@phanan this changes located in `laravel\laravel` repository - what means in your application, and not hardcoded in framework. So you are free to change `composer.json` file to fit your requirements.

phanan:
👍 I guess it shouldn't be a big deal. Thanks for listening to my rant 😄

PR #4764, opened 2018-09-19
Fix API authentication failure response

AegirLeet:
Fixes https://github.com/laravel/framework/issues/25636.

Detailed explanation [here](https://github.com/laravel/framework/issues/25636#issuecomment-422898097).

### DIFF
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -2,6 +2,7 @@
 
 namespace App\Http\Middleware;
 
+use Illuminate\Support\Facades\Route;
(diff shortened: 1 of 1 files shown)


taylorotwell:
You are free to change this in your own application if necessary.

AegirLeet:
@taylorotwell Sure, but I don't think having broken routes in the default install is appropriate. Setting up a fresh Laravel project, hitting one of the supplied example routes and getting an Exception is really not what you'd want to see as a developer imo.

anastasioscho:
I agree with @AegirLeet. And the worst part in my opinion is that you don't even have a chance to catch an AuthenticationException in App\Exceptions\Handler.

PR #4766, opened 2018-09-23
make:view package

vvmspace:
Add php artisan make:view


GrahamCampbell:
Thanks, but I don't think this needs including out of the box.

PR #4767, opened 2018-09-24
[5.7] Set logs to daily by default

bertheyman:
*Why consider this?*

Daily logging makes an application more "production ready" out of the box. I know developers are considered to set this config, but for many new or inexperienced devs this isn't common knowledge.

Sticking to daily by default makes more sense to me:
- Way easier of getting an overview of what happened in a chronological order.
- No chance of consuming lots of disk space by huge logs.

I understand this is also partly a personal preference, so I'd welcome any feedback or constructive criticism why this should remain as is or changed to something different.

### DIFF
--- a/config/logging.php
+++ b/config/logging.php
@@ -36,7 +36,7 @@ return [
     'channels' => [
         'stack' => [
             'driver' => 'stack',
-            'channels' => ['single'],
+            'channels' => ['daily'],
         ],
 
         'single' => [


mnabialek:
Good idea, this is the first thing I always changed in new app :) However I would consider changing default 7 days to something more, for example 30 days if you need to investigate some problems more in past

weseldridge:
yes! this is great.

michaeldyrynda:
Daily logs in development suck, though!

laurencei:
> Daily logs in development suck, though!

Yeah. @michaeldyrynda - maybe it should be

`'channels' => [env('LOG_STACK, 'daily')],`

Then you can change easily for dev vs production?

gmsantos:
Just a reminder that `RotatingFileHandler` (the monolog handler used by daily log) is considered a workaround by monolog itself.

https://github.com/Seldaek/monolog/blob/1.x/src/Monolog/Handler/RotatingFileHandler.php#L16-L24

PR #4768, opened 2018-09-25
[5.7] Use Mailhog Settings

browner12:
MailTrap is a paid **external** service (with a optional free tier).  Mailhog is a free **local** service, and comes preinstalled on Homestead machines.

There may be additional things that MailTrap offers that Mailhog does not, but I think switching from an external to local service (for the default) is better.

Also, I'm not sure we should give Mailtrap preference, since there's lots of external mail catching services, and people are still free to use what they'd like.


gpressutto5:
I believe mailtrap is more "development ready", since it requires nothing but an account. Also, its "optional free tier" provides pretty much anything you could need when starting developing a new app.

mnabialek:
Just as side note - remember that not everyone uses Homestead for development. I use Docker, many people on Windows use Xampp.

laurencei:
@mnabialek - yep. Obviously everyone is free to change the settings.

I was just saying that most other "default" Laravel features/config (such as database) start with an assumption of using Homestead, so it makes sense to continue that all the way through.

ludo237:
I don't have MailHog on my docker container, what about it? 

How changes like this can improve the Laravel codebase at all?

Spoiler Alert: they don't

browner12:
I'm always a huge fan of consistency, and since we use Homestead settings for the database, to me it makes sense to use these settings for the default mail.

For people that use the tools provided by the Laravel ecosystem, we are looking to provide a friction free process to get them up and running their Laravel sites.

taylorotwell:
Eh, will hold off on this for now.

PR #4770, opened 2018-09-26
[5.7] Update log config: 30 daily + easy overwrite channel for dev

bertheyman:
As per request [here](https://github.com/laravel/laravel/pull/4767):

- Default max daily logs to 30
- Easily overwrite the channel for dev environment

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,6 +5,7 @@ APP_DEBUG=true
 APP_URL=http://localhost
 
 LOG_CHANNEL=stack
+LOG_STACK=daily
 
 DB_CONNECTION=mysql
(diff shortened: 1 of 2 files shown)


laurencei:
If this was accepted - you would want to update `.env.example` with the name there, so people could see it more obviously it could be updated.

bertheyman:
Good idea! I've updated my pull request.

taylorotwell:
Eh, I don't really agree with this change. The LOG_STACK in practice could be multiple drivers (an array).

bertheyman:
Ok, thanks for the feedback!

PR #4771, opened 2018-09-26
[5.7] Change default days to 30 for daily channel

mnabialek:
As we changed default channel to daily in https://github.com/laravel/laravel/pull/4767 I think it makes sense to keep logs for more than 7 days. In case any problem appears, you often need to take a look a bit further than just 7 days.

### DIFF
--- a/config/logging.php
+++ b/config/logging.php
@@ -49,7 +49,7 @@ return [
             'driver' => 'daily',
             'path' => storage_path('logs/laravel.log'),
             'level' => 'debug',
-            'days' => 7,
+            'days' => 30,
         ],
 
         'slack' => [


fletch3555:
I think 7 is a perfectly reasonable default.  You're free to update the config as you see fit for your projects

fletch3555:
In all honesty, should the framework REALLY be concerned with whether a dev would "forget" to change a setting? To me, that sounds like fault falls 110% on the dev(and/or their team) who didn't change the value.

mnabialek:
If you think a bit more about it you will see that:

- before all logs were stored in single file (so devs didn't have to change it all and had access to all logs)
- changing default from 7 to 30 doesn't make any harm - just let you to keep your logs for more days. You don't need to use them if you don't want to
- if you want to change to 7  days, you can do it, but it's much more reasonable to have logs from more days by default when you need to investigate any repeating problem

taylorotwell:
I'll meet in the middle at 14 days. :)

PR #4239, opened 2017-04-26
Develop

tejomayonline:
in laravel 5.4+ after a fresh installation and auth scaffoding when i wisht to migrate small errors shows related to user table artibutes character limitation.
I have found it multiple times. I have fixed it using schema  facade's native method  in AppServiceProvider's boot mettod ,i.e. Schema::defaultStringLength(200);

the error message is
[Illuminate\Database\QueryException]                                                                                                           
  SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 767 bytes (SQL: alter table `users` add  
   unique `users_email_unique`(`email`))                                                                                                         
                                                   [PDOException]                                                                                                   
  SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 767 bytes  

please resolve it.
thank you 
tejomay......

PR #4774, opened 2018-09-27
Fixed the "Specified key was too long error" Bug

MuhaddiMu:
Fixed Laravel 5.4: Specified key was too long error Bug in Laravel Which was not fixed yet.
https://laravel-news.com/laravel-5-4-key-too-long-error

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -3,6 +3,7 @@
 namespace App\Providers;
 
 use Illuminate\Support\ServiceProvider;
+use Illuminate\Support\Facades\Schema;
 
 class AppServiceProvider extends ServiceProvider
 {
@@ -14,6 +15,7 @@ class AppServiceProvider extends ServiceProvider
     public function boot()
     {
         //
+        Schema::defaultStringLength(191);
     }
 
     /**


bertheyman:
This bug / limitation depends on the used database system. While it can be needed for some to limit the default string length, I'm not sure it needs to be like that for all database systems.

antonkomarev:
Related to #4239 #4488

MuhaddiMu:
@GrahamCampbell I'm agree with that and according to @bertheyman That's not the case with all of the databases out there. 
So if we let it be that would be also Okay. But to figure out if the majority is facing the same.

martindilling:
I agree it shouldn't be there by default, but the first few times I ran into this problem it was really really annoying and had no clue what was wrong. I can really imagine new people just throwing up Laravel on Wamp or something for the first time, would be a bit demotivated hitting this roadblock as some of their first Laravel experience :(
I don't have a really good solution though, maybe if it's possible to catch the error and add to the error message or something xD

laurencei:
What about if it was included, but commented out:

`// Schema::defaultStringLength(191);`

?

MuhaddiMu:
@martindilling I faced the same issue in starting and after Googling I come up with the solution but then I don't know where the actual **AppServiceProvider.php** file is located.

@laurencei It'll be good as well, Or maybe just better explain it in Installation Docs specifically for **Phpmyadmin** or XAMPP or WAPP or whatever the database is.

whendytakashy:
I found problem to, and then I found it to solve this problem with other method.

[ThisLink](https://whendy.net/laravel/laravel-syntax-error-or-access-violation-1071/)
Or
[ThisLink](https://stackoverflow.com/questions/23786359/laravel-migration-unique-key-is-too-long-even-if-specified)

ruhul0:
https://github.com/laravel/laravel/pull/4972

PR #4776, opened 2018-09-28
[5.7] Update config/cache.php slug generation to conform to config/session.php

SammyK:
The `config/session.php` file generates a slug with the `str_slug()` helper. This PR makes `config/cache.php` generate a slug in the same manner. :)

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -1,7 +1,5 @@
 <?php
 
-use Illuminate\Support\Str;
-
 return [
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I actually would prefer to change `session` because I may eventually do away with these global helper aliases. So, I'm trying to prefer using Str::* and Arr::* where possible. I'll fix the other config file. Thanks.

PR #4777, opened 2018-09-28
[5.7] Make app path stream safe

SammyK:
This PR ensures that the app base path is stream-safe instead of the [potential unexpected behavior of using `realpath()`](https://github.com/kalessil/phpinspectionsea/blob/master/docs/probable-bugs.md#phar-incompatible-realpath-usage).


taylorotwell:
Thanks!

PR #4778, opened 2018-09-29
:wheelchair: Extract timezone and locale configurations to environment

frowhy:
I'm not American, and I don't use coordinated universal time. So, I need to modify the configuration file directly every time, which I don't think is good.

Signed-off-by: frowhy <frowhy@gmail.com>


GrahamCampbell:
Thanks for the PR. The convention here would be to change the config files in your app. For app specific config, you are meant to edit the config files, only for environment specific things should the environment variables be used, like for passwords.

PR #4780, opened 2018-09-30
Use correct facade

zaknesler:
We should be checking if the route is defined, not if it exists in the request.

Related PR: https://github.com/laravel/laravel/pull/4775


GrahamCampbell:
Thanks. 👍

PR #4784, opened 2018-10-02
[5.7] Set hash rounds to 4 to speed up tests

shadoWalker89:
This improvement was originally tweeted by taylor 

https://twitter.com/taylorotwell/status/943135617466105856

And it does really make a difference.
I was really surprised that it's not included by default in 5.7
So there it is, enjoy the speed.

### DIFF
--- a/tests/CreatesApplication.php
+++ b/tests/CreatesApplication.php
@@ -2,6 +2,7 @@
 
 namespace Tests;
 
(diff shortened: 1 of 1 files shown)


mnabialek:
It seems it was already done here: https://github.com/laravel/laravel/pull/4643/files

PR #4786, opened 2018-10-02
[5.7] Remove duplication of middleware

pedrofurtado:
@taylorotwell


driesvints:
Can you give some more info regarding this change?

taylorotwell:

Thanks!

taylorotwell:
I don't work PRs without explanations.

pedrofurtado:
Sorry guys. I forgot to explain why: This file defines a middleware to be available for routes, named "bindings". It's used by default in "api".
But it is also used in "web". But, different of "api", they use the full-qualified class name of middleware ("Illuminate\Routing\Middleware\SubstituteBindings::class") instead of name "bindings". So , I suggest to remove this duplicate in the file 👍 

let me know if it is ok or if any change is required 🤝

ludo237:
Why not the inverse? I mean why don't use the full-qualified name inside the `api` key? 

I don't think that Kernel need any update on this

PR #4788, opened 2018-10-04
Security Precaution

v1shky:
Enable by default important security config in debug blacklist to prevent security issues.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -224,4 +224,28 @@ return [
 
     ],
 
+    'debug_blacklist' => [
+        '_ENV' => [
+            'APP_KEY',
+            'DB_PASSWORD',
+            'REDIS_PASSWORD',
+            'MAIL_USERNAME',
+            'MAIL_PASSWORD',
+            'PUSHER_APP_KEY',
+            'PUSHER_APP_SECRET'
+        ],
+        '_SERVER' => [
+            'APP_KEY',
+            'DB_PASSWORD',
+            'REDIS_PASSWORD',
+            'MAIL_USERNAME',
+            'MAIL_PASSWORD',
+            'PUSHER_APP_KEY',
+            'PUSHER_APP_SECRET'
+        ],
+        '_POST' => [
+            'password',
+        ],
+    ],
+
 ];


v1shky:
I have some rookie developers and during alpha to debug they enable but for black box testing they provide some others the link and each debug reveals their security so thought this would be useful.

nicknick-io:
Do it in your own project.

v1shky:
Ok Thanks.

barryvdh:
I do think it's probably a good idea to set this by default, even if it's just to show the option to users. Could be just the example from the docs:

```php
    'debug_blacklist' => [
        '_ENV' => [
            'APP_KEY',
            'DB_PASSWORD',
        ],

        '_SERVER' => [
            'APP_KEY',
            'DB_PASSWORD',
        ],

        '_POST' => [
            'password',
        ],
    ],
```

pmingram:
I'm in agreement with @v1shky and @barryvdh. At the very least it would be a good idea to set _something_ by default, even if it's just the example from the docs.

> You shouldn't have APP_DEBUG enabled in production at all

This is correct, you shouldn't, but there are already way too many examples on the Internet of people doing things you shouldn't do. A quick Google search will turn up a bunch of websites where you can freely view the .env file, or who have Laravel DebugBar enabled (I've seen this on production sites!) and _way_ too many sites that have `APP_DEBUG` enabled in production, showing full information on errors.

The default `User` model has the `password` and `remember_token` columns hidden - I fail to see how this is any different?

savankaneriya:
I think it does't bother if it is set by default. One will realise when one have developed the whole app in windows and live server have linux , many thing needed to have quick look in production mode. so at that time this will be really handy to turn on the DEBUG and resolve the issue without reviling the secrets. good work @v1shky .

barryvdh:
FYI, I've created this PR: https://github.com/filp/whoops/pull/612
With wildcards it should make more sense to provide better defaults.

GuidoHendriks:
> The default `User` model has the `password` and `remember_token` columns hidden - I fail to see how this is any different? – @pmingram 

There's a clear difference. The `$hidden` property [is used for JSON representation](https://laravel.com/docs/5.7/eloquent-serialization#hiding-attributes-from-json), which actually impacts production code.

pmingram:
@GuidoHendriks Right, but we're talking about the security implications of disclosing sensitive information here. Not how data is represented.

In that regard, the issue of hiding sensitive information in a model, and hiding sensitive information in an error message (that shouldn't be visible anyway - but that's another problem) are one and the same.

pmingram:
@GuidoHendriks Yes you could. So what's the problem with discussing and implementing this one, which seems like a no-brainer to me and clearly a number of others.

PR #4789, opened 2018-10-04
add verify option to hashing configuration as a hint

Chrissi2812:
Adding this as a little hint to this https://github.com/laravel/framework/commit/5fd4b899cc42d266fab34ee2d5f92fb47ca34fd0.


taylorotwell:
I think this will just raise more questions honestly without further explanation. "Verify" sounds like something I want? Why wouldn't I want verification? It sounds like a positive thing to want regarding password hashing so I can see people just enabling it without understanding it.

PR #4790, opened 2018-10-04
Additional Config Options Documentation

rexfordkelly:
It was not clear to me that .env files had higher precedence over application config files. While it makes total sense, now that I know it lead to unexpected behavior. So I added simple comment blocks to each config file to make this behavior clear in the source.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -1,5 +1,16 @@
 <?php
 
+/*
+|--------------------------------------------------------------------------
+| Config Option Precedence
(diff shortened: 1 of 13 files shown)


fireynis:
They don't take precedence really though, only if you use env(KEY, 'something') will it override 'something' with whatever is in KEY. If you just put in 'configval' => 'something' that will never be overridden.

rexfordkelly:
Right, I can see that, thanks!

PR #4795, opened 2018-10-06
Update UserFactory password

connectkushal:
The password set by the default UserFactory is `secret` which is 6 characters. This is less than the required minimum 8 character length as per new NIST standards on memorized secrets (i.e. user passwords) as stated here  laravel/framework#25957

Although this is **enforced in the next version** of laravel by #4794, I think the UserFactory default should be changed in the current version itself.

By this pr, the new default is set to `password`

---

### DIFF
--- a/database/factories/UserFactory.php
+++ b/database/factories/UserFactory.php
@@ -18,7 +18,7 @@ $factory->define(App\User::class, function (Faker $faker) {
         'name' => $faker->name,
         'email' => $faker->unique()->safeEmail,
         'email_verified_at' => now(),
-        'password' => '$2y$10$wOY5GEMqZ2onAWGeNXlTd.DEgxnT.dIovWTvxfgMqvRCk1LRbjPIm', // secretpass
+        'password' => '$2y$10$92IXUNpkjO0rOQ5byMi.Ye4oKoEa3Ro9llC/.og/at2.uheWG/igi', // password
         'remember_token' => str_random(10),
     ];
 });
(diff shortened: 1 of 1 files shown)


DojoGeekRA:
@connectkushal have you considered `password` instead?, seems more intuitive and has 8 characters too.

connectkushal:
@DojoGeekRA yes i did, and here is why i thought its not good, password it is not a clear password

As you can see it is confusing even to say it, and also vague for any new user reading the code comment.

Lastly, for the sake of comparison `secretpass` is 10 characters long, so a little bit more secure compared to `password` and a bit more neat and convenient than something like `secretpassword` or `secret123`

antonkomarev:
Since its used only for the tests `password` will be the best choice!

connectkushal:
@antonkomarev  if many think so i am willing to change

ankurk91:
Who will update auth related controllers?
https://github.com/laravel/laravel/blob/a273cea667438858b8b3d8459021ade9e3eb18a1/app/Http/Controllers/Auth/RegisterController.php#L54

connectkushal:
@ankurk91 Please read the description again, properly.

ankurk91:
You send it to `master` branch, it means you want it to be merged in 5.7.

I don't see any description about Auth related controllers. They are still enforcing min 6 length passwords

connectkushal:
@ankurk91
 
> Although this is enforced in the coming version of laravel by #4794, I think the UserFactory default should be changed in the current version itself.

antonkomarev:
But master is 5.8. Isn't it?

connectkushal:
Changing required length for 5.7 may affect users of websites using current version. It is a breaking change, and is best to **enforce** the change in minimum password length in 5.8. 

Since UserFactory is used for testing and development, any changes can be made safely to it, as done by this pr. 

If maintainers raise the same issue please create a new pr as you have pointed this out.

taylorotwell:
Develop branch?

PR #4796, opened 2018-10-06
Add Keys for mail sender details in .env template

gauravmak:


taylorotwell:
You can add them in your own file if you need them.

PR #4797, opened 2018-10-06
[5.8] Update UserFactory password in line with #4794

connectkushal:
The new requirement for user password is of 8 characters, as set by #4794 and laravel/framework#25957.

This pr updates the UserFactory password to `password`


tohidhabiby:
THANK YOU SO MUCH, Excellent.

PR #4799, opened 2018-10-10
adding trustedproxy config file. Supports AWS ELB.

rrpadilla:
**PROBLEM**: Hard coded settings in the TrustProxies middleware class.
- Production: Probably you will use ELB.
- Development: Probably you are not using ELB.

**SOLUTION**: Moving settings to config file. Config file get settings from env file. Really easy to define your settings based on environments.

To configure Trusted Proxies when needed.
Change your .env if:
- you're using AWS ELB:
    - __TRUSTEDPROXY_PROXIES__="*"
    - __TRUSTEDPROXY_HEADERS__="__HEADER_X_FORWARDED_AWS_ELB__"
- IP address (or range) of your proxy
    - __TRUSTEDPROXY_PROXIES__="192.168.1.1,192.168.1.2"
    - __TRUSTEDPROXY_PROXIES__="192.168.1.0/8"
    - __TRUSTEDPROXY_HEADERS__="__HEADER_X_FORWARDED_ALL__"
    - __TRUSTEDPROXY_HEADERS__="__HEADER_FORWARDED__"

If you are using AWS ELB you can configure your Nginx like this:
```nginx
server {
    # all traffic, secure or otherwise, comes in on 80 from the ELB
    listen 80;
    server_name YOUR-DOMAIN.COM;
    root /home/USERNAME/sites/YOUR-DOMAIN/public;

    # ELB stores the protocol used between the client
    # and the load balancer in the X-Forwarded-Proto request header.
    # Check for 'https' and redirect if not
    if ($http_x_forwarded_proto != 'https') {
        return 301 https://$server_name$request_uri;
    }
    .......
}
```

### DIFF
--- a/app/Http/Middleware/TrustProxies.php
+++ b/app/Http/Middleware/TrustProxies.php
@@ -3,21 +3,49 @@
 namespace App\Http\Middleware;
 
 use Illuminate\Http\Request;
+use Illuminate\Contracts\Config\Repository;
 use Fideloper\Proxy\TrustProxies as Middleware;
 
 class TrustProxies extends Middleware
 {
     /**
-     * The trusted proxies for this application.
+     * Create a new trusted proxies middleware instance.
      *
-     * @var array
+     * @param \Illuminate\Contracts\Config\Repository $config
      */
-    protected $proxies;
+    public function __construct(Repository $config)
+    {
+        parent::__construct($config);
+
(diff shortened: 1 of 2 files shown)


taylorotwell:
IMO this is too much code in an app middleware that we ship. You can already make these changes for your own application if necessary.

PR #4804, opened 2018-10-12
[5.7] Add "pretend" key to the mail config

hooch:
The docs say to "set the pretend option in the config/mail.php configuration file to true".  This PR adds a placeholder pretend option that's set to false.

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -120,4 +120,17 @@ return [
         ],
     ],
 
+    /*
+    |--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


driesvints:
I don't think this is available anymore in the latest Laravel release. Can you please link to the docs where you see this? I don't see this in the docs of the current release.

hooch:
Thanks @driesvints you're right, I had landed in the 5.0 docs.

PR #4807, opened 2018-10-13
[5.7] Fixes a small grammar error in public/index.php

robbyrice:
:grinning:

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -16,7 +16,7 @@ define('LARAVEL_START', microtime(true));
 |
 | Composer provides a convenient, automatically generated class loader for
 | our application. We just need to utilize it! We'll simply require it
-| into the script here so that we don't have to worry about manual
+| into the script here so that we don't have to worry about manually
 | loading any of our classes later on. It feels great to relax.
 |
 */


taylorotwell:
It breaks the spacing 😱

robbyrice:
So, this is a no go then?

shirshak55:
@robbyrice taylor don't want to break space so why don't you move 'info' to upper and add it :P

robbyrice:
@shirshak55 not quite sure I understand what you mean.

shirshak55:
@robbyrice  something like 

|
| Composer provides a convenient, automatically generated class loader for
| our application. We just need to utilize it! We'll simply require it into
| script here so that we don't have to worry about manually
| loading any of our classes later on. It feels great to relax.
|
*/

robbyrice:
Oh ok, I see. However I'm not going to move forward with this, so feel free to create a new Pull Request if you would like.

shirshak55:
@robbyrice yeah i am also not going because taylor may say it don't add any value to framework which is correct IMHO :)

samundra:
Besides, Taylor is very strict about 3 character thing, IYKWIM.

PR #4817, opened 2018-10-15
[5.7] RFC 5785 .well-known and security.txt

melbahja:
add: .well-known folder : https://tools.ietf.org/html/rfc5785
add: security.txt file: https://securitytxt.org/
thanks :)

### DIFF
--- a/public/.well-known/security.txt
+++ b/public/.well-known/security.txt
@@ -1,0 +1,1 @@
+# generate your security text here: https://securitytxt.org/


m1guelpf:
out-of-scope imo

connectkushal:
I think as a security practice its both educating and essential. Plus the folder will be hidden and its discovery will certainly push new users to learn something new.

@melbahja thanx for the links :)

GrahamCampbell:
Thanks for the PR, but I worry this will only result in lots of Laravel sites with that path serving the following text:

```
# generate your security text here: https://securitytxt.org/
```

PR #4819, opened 2018-10-16
[5.7] Remove the home directory from all controllers, and middlewares.

GabMic:
To be honest, every time i make a fresh install of laravel, the first thing i do is to remove the `home` directory from all controllers. people do not use `/home`, they use just `/` like 99.999% of sites worldwide.
This will save some time, and make it more reasonable.

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -25,7 +25,7 @@ class LoginController extends Controller
      *
      * @var string
      */
-    protected $redirectTo = '/home';
+    protected $redirectTo = '/';
 
     /**
      * Create a new controller instance.
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -28,7 +28,7 @@ class RegisterController extends Controller
      *
      * @var string
      */
-    protected $redirectTo = '/home';
+    protected $redirectTo = '/';
 
     /**
      * Create a new controller instance.
--- a/app/Http/Controllers/Auth/ResetPasswordController.php
+++ b/app/Http/Controllers/Auth/ResetPasswordController.php
@@ -25,7 +25,7 @@ class ResetPasswordController extends Controller
      *
      * @var string
      */
-    protected $redirectTo = '/home';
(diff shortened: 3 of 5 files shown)


driesvints:
`/` is already reserved for your landing page. `/home` is the page where you end up when you login, register, etc...

m1guelpf:
While I'm against removing /home for /, I'd say /dashboard is a much common url. Should I do some research on the usage of both and submit another PR for consideration?

driesvints:
@m1guelpf yeah I agree but `/home` is now the default and what people are used to. It's also the default used in Spark, etc. So changing it would mean people having to figure out why it changed and it's not worth the hassle imo.

GabMic:
Well, the fact it is (along with the controller, routes) for auth users, and no user is existing after the auth command, it is really unnecessary. 
The landing page should be removed after running the auth command, as it is generating a new home view. that view should be visible from the main path `/`, and of course not be for only authenticated users. you can see this in probably 99.999% of projects, sites web apps etc'.
All this will simplify workflow and save some time for everyone on a fresh install.

narwy:
> / is already reserved for your landing page. /home is the page where you end up when you login, register, etc...

No way.

I agree with removing `/home` it's better to have it all as `/`

driesvints:
> All this will simplify workflow and save some time for everyone on a fresh install.

Well, yeah but this will save time for yourself and others who agree with you but not for others who prefer the current approach. They'll have to make the changes every time they use a fresh Laravel install. It's just an opinion on how the auth behavior should be.

narwy:
`/home` is nothing related do a dashboard, to be honest why even restructure a route to display something so simple.

I agree with all the changes to make it `/` or even `/dashboard`

hulkur:
Make it configurable, with default `/home` to preserve current state

GabMic:
That's what i am saying. I think the majority of people do not use this imo. As evidence, i really don't see any site practically using this path.

ludo237:
What about `/private` ... 

I mean the choice are almost infinite how do you decide? Maybe @hulkur is right but again, it's literally a 4 letter replace DIY

taylorotwell:
No changes planned here.

PR #4820, opened 2018-10-16
changed syntax for validation

brainmaniac:
Changed the syntax for the validation to be more aligned to the proposed way of implementing custom validations in the docs: https://laravel.com/docs/5.7/validation#custom-validation-rules


sadegh-shahbazi:
it's too hard for me to read another programmer's code
tylor you put many too many code and syntax for one job!
who care i can write 1 line code instead of 3 line
i can't focuse to understanding the project
just do something like golang 
thank you

PR #4823, opened 2018-10-17
Revert validation syntax change

SjorsO:
https://github.com/laravel/laravel/pull/4820 was merged earlier today without any discussion. I don't think the change is a good idea, so i've opened up this PR to start a discussion (and hopefully revert the change)

The old syntax is shorter, cleaner, and is used in all the [validation docs](https://laravel.com/docs/5.7/validation#custom-validation-rules) examples that don't use a closure or custom rule. The validation in the `RegisterController` doesn't use any custom rules, so there is no reason to write the rules as an array.

### DIFF
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -49,9 +49,9 @@ class RegisterController extends Controller
     protected function validator(array $data)
     {
         return Validator::make($data, [
-            'name' => ['required', 'string', 'max:255'],
-            'email' => ['required', 'string', 'email', 'max:255', 'unique:users'],
-            'password' => ['required', 'string', 'min:6', 'confirmed'],
+            'name' => 'required|string|max:255',
+            'email' => 'required|string|email|max:255|unique:users',
+            'password' => 'required|string|min:6|confirmed',
         ]);
     }
 


shirshak55:
One thing i don't like about simple syntax is i cannot use rule object :D

ludo237:
@shirshak55 is right, the new default syntax with array allows an easy drop in of `Rule` and custom rules

SjorsO:
The majority of validators don't use custom rules. You can change a rule to array syntax when you actually need to add a custom rule.

ludo237:
Same goes in reverse, you can go for a string-based syntax when you actually need it

taylorotwell:
Eh, I dunno. If you want to drop-in a custom rule for your user validation, you would have to rewrite the whole thing into an array?

nicknick-io:
If a person is using make:auth, then the person is most likely new to
Laravel, and is trying to learn the basics.

Now comes my question:
"How likely is it that a person who just started using Laravel would create
a custom rule?"


Not relevant for the first part ::
Personally i would never trust the security in the "make:auth" part anyway,
since it has shown some pretty gaping holes over the last couple of years.
So i would never see this code.

martindilling:
@nicknick-io
So you would rather build an authentication system yourself? And you would of course singlehandedly know about and secure against all the obscure and complex security holes, that have been discovered and patched in a framework with a lot of users who battle test it? 
Makes perfect sense xD

michaeldyrynda:
I think the array syntax is generally much more familiar to people than the pipe-separated string. Validation is a list of rules, and lists of things are best represented as an array in PHP, imo.

nicknick-io:
@martindilling Did you know that you are **_NOT_** forced to use the build in scaffolding, you can make your own controller and your own views, and yet still use the authentication system. I never said I would build my own authentication system, did I? No, that would be silly, using a framework and all.

PR #4827, opened 2018-10-21
[5.8] Update welcome.blade.php

voyula:


driesvints:
@GavG actually, this would work fine. The named route `home` comes with the Auth scaffolding and the routes are only displayed when the auth scaffolding has been added so it's perfectly fine to use the `route` helper here. Fresh Laravel apps won't have issues with this.

GavG:
@driesvints You're quite right, didn't notice the surrounding @if (Route::has('login')) block in the template.

voyula:
This is right... Because has in auth control.

shirshak55:
but it don't improve framework which might be wanted by taylor :)

PR #4829, opened 2018-10-22
Update channels.php

kloentje2:
Improve PSR2 compatibility


driesvints:
This isn't a PSR2 convention as far as I can see and it doesn't follows Laravel's code styling.

PR #4842, opened 2018-10-31
[5.7] Add framework version to welcome page

Farris27:
add the version of laravel.

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -1,10 +1,11 @@
+ @php  $laravel = app()->version(); @endphp
 <!doctype html>
 <html lang="{{ str_replace('_', '-', app()->getLocale()) }}">
     <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1">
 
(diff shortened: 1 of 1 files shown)


ludo237:
Weird implementation honestly

resources/views/welcome.blade.php
```diff
@@ -1,10 +1,11 @@
+ @php  $laravel = app()->version(); @endphp
 <!doctype html>
 <html lang="{{ str_replace('_', '-', app()->getLocale()) }}">
     <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1">
 
-        <title>Laravel</title>
+        <title>Laravel - {{$laravel}}</title>
```
troccoli: Why don't you do `{{ app()->version() }}` directly?

resources/views/welcome.blade.php
```diff
@@ -1,10 +1,11 @@
+ @php  $laravel = app()->version(); @endphp
```
GrahamCampbell: Better to pass this from the controller, than to do this inline.

PR #4844, opened 2018-11-02
Changes default welcome view to use route::view instead of get

oheck:
Laravel introduced Route::view. Changing the Route::get closure to Route::view will allow for route caching.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -11,6 +11,4 @@
 |
 */
 
-Route::get('/', function () {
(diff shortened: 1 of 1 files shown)


deleugpn:
There are probably over 30 PRs that tried this before. It's likely to be the most tried thing to change on the skeleton project. 
Taylor already said that he prefers this way as it's easy to prototype / test anything on a fresh Laravel installation by just running code on that closure.

PR #4845, opened 2018-11-04
[5.7] Bind LARAVEL_START into the container

deleugpn:
When writing test code that covers features involving LARAVEL_START, it is impossible to not have it leak to other tests when using PHPUnit without using `isolateProcess` (https://github.com/laravel/telescope/pull/328/files#diff-1d02b08fbcaa478f6b508a29d704aad8R20). Binding the start time into the container allow the developer to rely on the container value instead of the constant and, therefore, allow value modification during PHPUnit tests.

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -49,6 +49,8 @@ $app = require_once __DIR__.'/../bootstrap/app.php';
 |
 */
 
+$app->instance('LARAVEL_START', LARAVEL_START);
+
 $kernel = $app->make(Illuminate\Contracts\Http\Kernel::class);
 
 $response = $kernel->handle(


narwy:
I agree, it would make the PHPUnit tests much easier.

laurencei:
This doesnt make sense. I do unit and package testing with `LARAVEL_START` all the time without issue.

You just need to define it with your test application, and it works perfectly.

What problem does this actually solve?

deleugpn:
@laurencei you cannot define a constant twice unless you use isolate process. That makes phpunit a lot slower. And then every consecutive test uses a value defined in the 1st test that defined it.

narwy:
@taylorotwell re-open?

PR #4850, opened 2018-11-09
[5.7] Update app.js

arcanedev-maroc:
This change allows us to use a custom `name` specified in Vue component and to avoid if there is any duplication in components.

Example:

```vue

// ./components/ui/SidebarTogglerComponent.vue

<template>
    <a class="nav-link" @click="toggle">
        <i class="fa fa-fw fa-bars"></i>
    </a>
</template>

<script>
    export default {
        name: "ui-sidebar-toggler-component",

        methods: {
            toggle() {
                // Toggle the sidebar
            }
        }
    }
</script>
```

Then you can use it like this: `<ui-sidebar-toggler-component></ui-sidebar-toggler-component>`

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -22,7 +22,9 @@ Vue.component('example-component', require('./components/ExampleComponent.vue'))
 // const files = require.context('./', true, /\.vue$/i)
 
 // files.keys().map(key => {
-//     return Vue.component(_.last(key.split('/')).split('.')[0], files(key))
+//     let file = files(key)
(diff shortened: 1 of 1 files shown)


taylorotwell:
Feel free to add this to your application if you want.

arcanedev-maroc:
**cc**: @reinink

reinink:
Cool idea! I agree though, seems like something you could easily setup on a per project basis, should you need this level of customization.

PR #4851, opened 2018-11-12
Changed public path to public_html

fraigo:
Changed default `public` to `public_html`


fraigo:
Not in this repository

PR #4852, opened 2018-11-13
[^5.6] Printing blade directive from database in laravel

robertnicjoo:
It would be great to return blade directives inside another directive (data stored in database)

# Example + Code

`{{$category->title}}` shows category title but it also stored in another table in database like `category {{$category->title}} has 45 posts and etc.` (some additional texts) then I get this in my blade:

`{!!$myCustom!!}` so this `myCustom` has to print something like:

```
category LARAVEL has 45 posts and etc.
```

but what I get currently is:

```
category {{$category->title}} has 45 posts and etc.
```

The title part won't print.

### reason I suggest/ ask for this ability

If you check [this plugin](https://www.mageworx.com/magento-seo-meta-templates-extension.html) of Magento you'll see the reason. This type of functionality can save lots of time for taking care of websites SEO also can be used for lot's of other purposes.

### DIFF
--- a/resources/lang/en/pagination.php
+++ b/resources/lang/en/pagination.php
@@ -13,7 +13,7 @@ return [
     |
     */
 
-    'previous' => '« Previous',
-    'next' => 'Next »',
+    'previous' => '&laquo; Previous',
+    'next' => 'Next &raquo;',
 
 ];


robertnicjoo:
@ladea this is the only place provided for all of those you've mentioned, maybe they should open issues so we publish such things there.

Anyway to answer your question this is combined of all of those `a need as request` `an issue of blade` and `a request for making it happen`.

laurencei:
Hi there,

Welcome to Laravel and we are glad to have you as part of the community.

Unfortunately this GitHub area is not for ideas, suggestions etc. This is only for issues/bugs with the framework code itself.

I will be closing your ticket here. You are able to open a ticket at https://github.com/laravel/ideas

Alternatively you are able to open a PR using the Contributions guidelines: https://laravel.com/docs/5.7/contributions#which-branch

If you feel I've closed this issue in error, please provide more information about how this is a framework issue, and I'll reopen the ticket.

Thanks in advance.

robertnicjoo:
@laurencei hi, thanks man (I think you're right) let this be closed I'll open an issue in ideas.

Thank you.

PR #4853, opened 2018-11-13
[5.7] Remove lodash dependency when auto registering Vue components

aryehraber:
Love this new addition, but don't often use the full Lodash library. By using `pop()` instead of `_.last()`, everything will continue to work without Lodash being pulled in from `bootstrap.js`.


reinink:
This is tight. Could even make this a one-liner if you wanted:

```js
const files = require.context('./', true, /\.vue$/i)
files.keys().map(key => Vue.component(key.split('/').pop().split('.')[0], files(key)))
```

aryehraber:
@reinink I like it, I'll submit an additional PR

PR #4855, opened 2018-11-15
[5.7] Add `log_channel` configuration option

sebdesign:


sebdesign:
Aligning those comments took more time than implementing the actual feature!

config/mail.php
```diff
@@ -120,4 +120,17 @@
...
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Log Channel
+    |--------------------------------------------------------------------------
+    |
+    | If you are using the "log" driver, you may specify a different channel
+    | name if you prefer to keep e-mail messages separated from other log
+    | messages. Otherwise, Laravel will select the default log channel.
```
fletch3555: Might want to remove the period.  It makes this line one character too long :)

driesvints: It doesn't. Indentation is 3-2.

sebdesign: I noticed some other comment lines that end with a period would be 2 characters shorter than the line above them, so I guess that's acceptable.

config/mail.php
```diff
@@ -120,4 +120,17 @@
...
+    |
+    */
+
+    'log_channel' => null,
```
samundra: what about picking it from environment variable. 

```
'log_channel' => env('LOG_CHANNEL', null)
```

sebdesign: Would you use the `log` driver across different environments? I suppose this driver is mostly used for local development, or so I don't feel it's essential to have an environment variable for this option.

samundra: Yes, In production we output it to stdout while in local development we use file driver. This was intentional as we didn't wanted to create log file on production and staging rather output to stdout and let filebeat pick it.

driesvints: Yeah I think `MAIL_LOG_CHANNEL` is twtg.

PR #4857, opened 2018-11-19
[5.7] Add missing phpdoc comment

Erulezz:


app/Exceptions/Handler.php
```diff
@@ -31,6 +31,7 @@ class Handler extends ExceptionHandler
      *
      * @param  \Exception  $exception
      * @return void
+     * @throws Exception
```
antonkomarev: Use Fully Qualified Class Names please

taylorotwell:
This method doesn't typically throw an exception.

PR #4860, opened 2018-11-20
[5.5] Update AuthServiceProvider.php

romaindutartre:
Removed unused Gate Facade that stayed from an older version


taylorotwell:
It's there as a convenience because it's common use the Gate facade in this service provider.

PR #4861, opened 2018-11-20
[5.7] Update Kernel.php

unstoppablecarl:
fixes https://github.com/laravel/framework/issues/26571

I assume this is ok to go into master?

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -65,7 +65,7 @@ class Kernel extends HttpKernel
     /**
      * The priority-sorted list of middleware.
      *
-     * This forces the listed middleware to always be in the given order.
+     * This forces non-global middleware to always be in the given order.
      *
      * @var array
      */


GrahamCampbell:
I think this makes things less clear. Instead of removing "the listed" and saying "non-global", which is actually not quite accurate. At the end of the original text, it should say something about how it only affects the priority of middlewares within the router layer.

unstoppablecarl:
@GrahamCampbell 
Can you explain what about this is not accurate? Looking at the code global middleware is run in the order defined in the `App\Http\Kernel.php` `$middleware` property. Then all other middleware is merged and sorted. What would be more clear to communicate this than calling it "non-global" ?

It was merged here: https://github.com/laravel/framework/pull/26575
Whatever it ends up being the language should match in both places.

ludo237:
Too much of a bargain for a word that users can change by themselves

PR #4872, opened 2018-12-01
Prepare app.js for Mix 4.0 (and future)

Francismori7:
Since Mix 4.0 is still beta but is based on Babel7, it will now need to behave like `import`, which requires us to mention `default` to know what is `require`d.

Not sure if this is required right now, but it will be required once the switch is made. AFAIK, this is a non-breaking change for people who don't upgrade.


taylorotwell:
Please submit when you KNOW it is not breaking, not when "AFAIK".

PR #4881, opened 2018-12-14
[5.7] Add default schedule timezone setting

shrft:
this is related to this one:
https://github.com/laravel/framework/pull/26794

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -71,6 +71,18 @@ return [
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


shrft:
@ludo237 
I am not sure about that. 
I set it to null since I think most people want default schedule timezone and application timezone to be the same value.

taylorotwell:
Doesn't follow conventional comment formatting. Also, people are able to override the method in the Kernel.

PR #4882, opened 2018-12-14
Bump to Mix v4

JeffreyWay:
This PR includes all necessary changes for the Mix v4 update. A few notes for those interested:

1. You'll see three new Sass-specific dependencies in `package.json`. In the past, we included all Sass dependencies with each install of Mix, regardless of whether the user desired Sass compilation for their app. With v4, Mix now install those dependencies on the fly - only if the user requests it with `mix.sass()`. However, because laravel/laravel includes Sass boilerplate, I added them here to skip the extra download step. 

2. As part of the vue-loader 13 update, if you use the CommonJS-style `require` for importing ES modules, you must now add `.default`. See the [vue-loader release notes](https://github.com/vuejs/vue-loader/releases/tag/v13.0.0) for more details, if interested. (Personally, I think this repo should be updated to exclusively use ES modules, but that's a different PR.)

3. I did a fresh production recompile for the `app.js` and `app.css` files that are included with Laravel out of the box.

Okay, think that's it.


taylorotwell:
Thanks!

ankurk91:
What about updating presets

https://github.com/laravel/framework/tree/f546835092206037ce1d9b4103be204e2e9272cc/src/Illuminate/Foundation/Console/Presets

JeffreyWay:
Yes, that would be good, too. If someone wants to submit a PR to update them, that’d be great. Otherwise, I’ll makd a note to do it on Monday.

browner12:
I think we need to get into the habit of only allowing Laravel employees to submit changes for compiled files. No one is actually reviewing all of the changes in those files due to their enormity, and we've seen in past weeks how easy it is for people to slip malicious code into packages, just as it would be with these files.

Jeffrey is obviously a unique case where he's not an employee, but a huge part of the ecosystem, and I'm definitely not accusing him of anything here.  I'm more so advocating for a procedure in place that only employees can PR compiled files, so we can be defensive in case an attack like this was ever leveled against us.

I would be more than happy to submit a PR to the contribution guide if people agree with this.

ludo237:
You have a good point @browner12. In general I do recompile those assets as soon as I start a new project just for the sake of avoid any problem at all. Not that I don't trust Jeffrey ofc.

taylorotwell:
@browner12 I agree with that and generally try to stick with that. Usually I will pull down PRs to my machine and compile them myself.

PR #4888, opened 2018-12-20
Laravel 5.7 error variables/ or session not showing in blade view

adisazhar123:
Laravel 5.7 error variables/ or session not showing in blade view. It works on localhost server, but once uploaded to production it doesn't show. Fixed it by moving        ` \Illuminate\Session\Middleware\StartSession::class` in `protected $middlewareGroups` array to `protected $middleware` in Kernel.php


adisazhar123:
Sorry, wrong thread

PR #4889, opened 2018-12-26
Update title of Welcome view to reflect application name

Jhnbrn90:
Use the application name as defined in the .env file. If none is set, the default title will be 'Laravel'.


taylorotwell:
Welcome view isn't used by your actual application.

PR #4890, opened 2019-01-02
Add APP_TIMEZONE in config

ahmadmayahi:
As a team, sometimes we have different time zones for different developers.
In the current implementation of Laravel, there's no setting in `.env` file for the timezone.


config/app.php
```diff
@@ -67,7 +67,7 @@
     |
     */
 
-    'timezone' => 'UTC',
```
samundra: Timezone is app specific and rarely different per environment `local`, `production` or `staging`. I don't find this helpful.

GuidoHendriks:
This has been tried before but never made it in. Just do it in your own project if you need to.

ahmadmayahi:
@GuidoHendriks Not if you are working in a big team where a bunch of them use different time zone.
Modifying the `config/app.php` is a big deal, because somebody may forget, and commit his changes.

samundra:
@ahmadmayahi If app timezone varies per developer then it's a requirement which is very specific to your project. If individual developer are making changes in common timezone config then something is wrong with application config. It still doesn't really justify this PR. For most people, the timezone is configured for whole application and that happens once. 

If I may ask, could you please let us know why individual developers have their own set of config for `Timezone`? Once, we know more about why you need to make this change then perhaps we can discuss more.

As far as I know, UTC is the most frequently used one and I have rarely seen anyone changing it to something else. Rather the date/time conversions are handled at application level which makes more sense.

However, I am still not convinced on having this PR on the code base.

GuidoHendriks:
@ahmadmayahi I'm not saying they should modify their `config/app.php`, but you can just modify it once to use an environment variable. This just won't make it in, [as it never did before](https://github.com/laravel/laravel/pulls?utf8=✓&q=app_timezone).

samundra:
@ahmadmayahi https://github.com/laravel/laravel/pull/3792#issuecomment-222945960  looks like it's been discussed multiple time.

ahmadmayahi:
Since it's a duplicate #3792 (comment)  so I'm closing it.

PR #4894, opened 2019-01-07
[5.8] Update SQS queue configuration to use a url instead of prefix

alexkb:
As per this [PR over in laravel/framework](https://github.com/laravel/framework/pull/27086) the default Amazon SQS configuration needs a full url, rather than a prefix.


taylorotwell:
Just make the full URL your queue name this in this case and remove the prefix option entirely from your configuration.

PR #4895, opened 2019-01-07
[5.7] Add missing env MAIL_FROM_NAME

richardkeep:
Since it is specified here https://github.com/laravel/laravel/blob/master/config/mail.php#L60 it makes sense to have it on the env file anyway.


richardkeep:
I know there are lots of ENV vars not specified, but this is kinda crucial. The default value 'Example' cannot be used in any way in production, meaning that when you specify mail configurations you have to edit 2 files: Both .env file and config/mail.php file instead of 1 file.

koenhoeijmakers:
`config/mail.php` already uses these env variables? All you have to change is your environment variables which you'll have to do anyways.

richardkeep:
> `config/mail.php` already uses these env variables? All you have to change is your environment variables which you'll have to do anyways.

You missed my point. I mean, when you are configuring the mail, you will edit the env file for most details, and reach out to config/mail.php to ONLY edit the MAIL_FROM_NAME. You can't go with 'Example' as default name.

For the rest of the default ENV variables in the system which are not specified, the default values are mostly null, or boolean which can be used as it is, but this is exceptional

samundra:
I also agree with @driesvints 

I don't think it's required to be listed on `.env.example`. We can always change it from configuration page when required. Besides, if we keep adding such env to `.env.example` the list will keep on growing and growing.

richardkeep:
Ok. I'll take I back

PR #4898, opened 2019-01-10
add htaccess file for when we use shared host

ritonx:
when we use shared host use a httaccess for shouting request to index.php in public dir


driesvints:
Please see https://github.com/laravel/laravel/pulls?q=is%3Apr+htaccess+is%3Aclosed

PR #4904, opened 2019-01-14
[5.8] Modify RedirectIfAuthenticated middleware to accept multiple guards

Te7a-Houdini:
Same as #3685 , this will allows to

```php
Route::get('foo', 'Foo@foo')->middleware('guest:web,api');
```

### DIFF
--- a/app/Http/Middleware/RedirectIfAuthenticated.php
+++ b/app/Http/Middleware/RedirectIfAuthenticated.php
@@ -19,7 +19,7 @@ class RedirectIfAuthenticated
     {
         foreach ($guards as $guard) {
             if (Auth::guard($guard)->check()) {
-                return redirect('/');
+                return redirect('/home');
             }
         }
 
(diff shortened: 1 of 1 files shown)


gpressutto5:
Why redirecting an api request?

Te7a-Houdini:
@gpressutto5  it's just an example , but in real case scenarios we maybe doing this

```php

Route::get('admins-only-login-form', 'AdminController@loginForm')->middleware('guest:admin,moderators,users');
```

taylorotwell:
Why did you also change the redirect URL?

Te7a-Houdini:
@taylorotwell sorry changed by mistake

tomhorvat:
@taylorotwell @Te7a-Houdini 

Does this break 'guest' middleware? I'm able to access /login on clean 5.8-dev install without redirect...

Tested on valet, twice.

PR #4906, opened 2019-01-15
[5.7] Make var more descriptive.

markotitel:
When creating environments for Laravel, this setting confuses me even after dozens of fresh environment deployments.


driesvints:
At the very least because this is targeting the current release it will be breaking for anyone using the current env var.

Te7a-Houdini:
@markotitel  develop branch

PR #4897, opened 2019-01-10
upgrade trustedproxy package to 4.1 and adding config

rrpadilla:
@fideloper will release [4.1](https://github.com/fideloper/TrustedProxy/releases) after this PR has been submitted. See [fideloper/TrustedProxy/pull/117 ](https://github.com/fideloper/TrustedProxy/pull/117).

**PROBLEM**: Hard coded settings in the TrustProxies middleware class.
- Production: Probably you will use ELB.
- Development: Probably you are not using ELB.

**SOLUTION**: Moving settings to config file. Config file get settings from env file. Really easy to define your settings based on environments.

Features:
1. Accept IPs separated by comma.
    - You can add this config to the .env file:
```config
TRUSTEDPROXY_PROXIES='192.168.1.1, 192.168.1.2'
# More options
TRUSTEDPROXY_PROXIES='*'
```

2. Accept headers as a string.
    - You can add this config to the .env file:
```config
TRUSTEDPROXY_HEADERS="HEADER_X_FORWARDED_ALL"
# More options
TRUSTEDPROXY_HEADERS="HEADER_FORWARDED"
TRUSTEDPROXY_HEADERS="HEADER_X_FORWARDED_AWS_ELB"
```

3. Added support for AWS ELB.

**Maybe you want to add this to the laravel documentation.**
If you are using AWS ELB with HTTPS you can configure your Nginx like this:
```nginx
server {
    # all traffic, secure or otherwise, comes in on 80 from the ELB
    listen 80;
    .......

    # ELB stores the protocol used between the client
    # and the load balancer in the X-Forwarded-Proto request header.
    # Check for 'https' and redirect if not
    if ($http_x_forwarded_proto != 'https') {
        return 301 https://$server_name$request_uri;
    }
    .......
}
```

Thanks.

PR #4907, opened 2019-01-15
[5.7] Update Trusted Proxies

rrpadilla:
See PR #4897 for more details.
- Delete default TrustProxies middleware because empty property declaration.
- Using original middleware provided by the TrustedProxy package instead.
- Trusted Proxy middleware disabled by default.
- Added config file to read configuration from environment.

Thanks.


taylorotwell:
Configuration comment formatting doesn't match rest of framework.

PR #4909, opened 2019-01-15
[5.8] Make var more descriptive.

markotitel:
#4906 
When creating environments for Laravel, this setting confuses me even after dozens of fresh environment deployments.


markotitel:
I set the value to path to the S3 bucket.
`http://s3-aws-region.amazonaws.com/bucket`

taylorotwell:
I'm not sure you really need to set it at all though. The point of this variable I think is not to point to the bucket but to be any URL you want and it will override the default base URL that would normally be used.

PR #4910, opened 2019-01-16
[5.7] Update Trusted Proxies. Format Config file.

rrpadilla:
See PR #4897 and #4907 for more details.

Fix Configuration comment formatting to match rest of framework.


driesvints:
If we disable it by default why would we even need the package in the first place?

rrpadilla:
@driesvints Trusted Proxies was added to Laravel in [5.5 version](https://laravel.com/docs/5.5/requests#configuring-trusted-proxies). You only need this middleware when running your application behind a load balancer.

rrpadilla:
@driesvints That package was added by @taylorotwell . See [commit](https://github.com/laravel/laravel/commit/e23a1d284f134bfce258cf736ea8667a407ba50c). You have to ask him.
This PR is to update the package, add a config file, use original middleware and disable the middleware by default.
I just want to improve what we have now.

PR #4915, opened 2019-01-19
Delete .gitignore

moslem960:


GuidoHendriks:
Why would you do that?

Quiet95scholar:
don't do this

samundra:
@moslem960 Can you please add description on you pull-request? Why do you think we don't need `.gitignore` in the project.

PR #4920, opened 2019-01-28
[5.7] Return 401 Unauthorised response if expects JSON

mtimbs:
We are expecting a string to be returned here. The redirect was wrapped to prevent unexpected responses for API routes where a JSON response was requested. 

Suggest we be explicit with the JSON response instead of deferring to the exception handler

### DIFF
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -14,8 +14,10 @@ class Authenticate extends Middleware
      */
     protected function redirectTo($request)
     {
-        if (! $request->expectsJson()) {
-            return route('login');
+        if ($request->expectsJson()) {
+            return new JsonResponse('Unauthenticated', 401);
         }
+
+        return route('login');
     }
 }


ankurk91:
The method name is `redirectTo` and it suppose to return a redirect path rather a response.

The `Authenticate` middleware is working fine, it is return a json response.
```json
{
    "message": "Unauthenticated test."
}
```
Basically when `redirectTo` method return nothing the `AuthenticationException` return a JSON response if it is AJAX request other wise redirect user to login page.

https://github.com/laravel/framework/blob/8dfb2ec90777368c09fdb065725136a7549620bd/src/Illuminate/Auth/Middleware/Authenticate.php#L67

driesvints:
@ankurk91 ah right, thanks for pointing that out 👍

mtimbs:
Yeah @ankurk91 you are correct.

Any objection to a new PR making the return nullable for the `redirectTo`. It wasn't updated as part of the PR that wrapped the redirect in a conditional for json return type

driesvints:
@mtimbs just the docblock return type you mean? Sure, send in a PR for that.

mtimbs:
yeah I know it's a minor thing but it's causing issues with my static analysis suite and it would be good to just have it cleaned up. I'll send a PR in the next day or two.

PR #4926, opened 2019-02-02
Use Str::random() instead of str_random()

kawax:
https://github.com/laravel/framework/pull/26898


GuidoHendriks:
What’s the problem with the helper method?

driesvints:
Thanks!

troccoli:
With the str_ and arr_ helper methods going to be deprecated in 5.8, is this PR really necessary?

PR #4928, opened 2019-02-04
[5.7] Date mutator for 'email_verified_at' attribute on User model stub (with explanation)

epalmans:
The default included migration for the users table includes a timestamp field for 'email_verified_at'.
However, it might be a convience if the User model (stub) interpretes this field as a Carbon instance instead of a plain date/time string.
It also serves as a nice way to show the date mutator feature on models for someone new to Laravel

### DIFF
--- a/app/User.php
+++ b/app/User.php
@@ -27,4 +27,13 @@ class User extends Authenticatable
     protected $hidden = [
         'password', 'remember_token',
     ];
+
+    /**
+     * The attributes that should be mutated to dates.
+     *
+     * @var array
+     */
+    protected $dates = [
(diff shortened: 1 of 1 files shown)


taylorotwell:
Probably better to use add it as a `casts` IMO. Thanks.

epalmans:
Right. $casts it definately more universal to use.
Modified and resubmitted (see https://github.com/laravel/laravel/pull/4930/commits/2f37ba38fa8f7cf6a1cda070984459405e90a37a on https://github.com/laravel/laravel/pull/4930)

PR #4930, opened 2019-02-04
Attribute casting for 'email_verified_at' on User model stub (with explanation)

epalmans:
The default included migration for the users table includes a timestamp field for 'email_verified_at'.
However, it might be a convience if the User model (stub) interpretes this field as a Carbon instance instead of a plain date/time string.
It also serves as a nice way to show the attribute casting feature on models for someone new to Laravel

### DIFF
--- a/app/User.php
+++ b/app/User.php
@@ -29,11 +29,11 @@ class User extends Authenticatable
     ];
 
     /**
-     * The attributes that should be mutated to dates.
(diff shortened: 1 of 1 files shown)


epalmans:
thanks for the merge 👍

ThomHurks:
Makes sense @epalmans thanks!

PR #4939, opened 2019-02-08
[5.8] Update to PHPUnit 8

driesvints:
**This PR needs to wait until laravel/framework has PHPUnit 8 support: https://github.com/laravel/framework/pull/27441**

Encourage people on 5.8 to use PHPUnit 8.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "fzaninotto/faker": "^1.4",
         "mockery/mockery": "^1.0",
         "nunomaduro/collision": "^2.0",
-        "phpunit/phpunit": "^7.0"
+        "phpunit/phpunit": "^8.0"
     },
     "config": {
         "optimize-autoloader": true,


antonkomarev:
PHPUnit 8 requires `"php": ^7.2`, but since its only dev I suppose it's OK to leave `"php": "^7.1.3"` as is.

nunomaduro:
@driesvints We are going to require PHP `^7.2`? Otherwise tests wont run on travis since PHPUnit 8 requires `"php": ^7.2`.

driesvints:
@nunomaduro you're correct. I've reversed this PR.

nunomaduro:
@taylorotwell can’t we consider bump to PHP 7.2 in Laravel 5.8?

driesvints:
@nunomaduro no because we'll be supporting both PHPUnit 7 & 8. We'll do the bump in the next major version. 7.1 is EOL on the 1st of December.

crynobone:
AFAIK laravel/framework also bump to `^7.5` for 5.8, would be good if the skeleton update to that to avoid any conflict.

driesvints:
@crynobone good idea. Done.

PR #4940, opened 2019-02-10
Change PHPdoc app_Exceptions_Handler.php

gfijrb:


GrahamCampbell:
This method doesn't return anything. :)

PR #4941, opened 2019-02-11
added vue-template-compiler dependency

yamenarahman:
This fixes and avoids these errors:
1- `vue-template-compiler must be installed as a peer dependency`.
2- `vue and vue-template-compiler must be the same version`


driesvints:
Heya, sorry but we don't accept any prs with compiled assets: https://laravel.com/docs/5.7/contributions#compiled-assets

Please resubmit this without the compiled assets.

PR #4942, opened 2019-02-12
Added vue-template-compiler dependency

yamenarahman:
This fixes and avoids these errors:
1- `vue-template-compiler must be installed as a peer dependency`.
2- `vue and vue-template-compiler must be the same version`

I've reverted the compiled assets.


driesvints:
Heya thanks for sending this in. It just occurred to me that we had a similar PR for this in the past. Please see its reasoning why I'm closing this: https://github.com/laravel/laravel/pull/4893

PR #4943, opened 2019-02-13
[5.8] Use $_SERVER instead of $_ENV for phpunit.

crynobone:
Laravel 5.8 limits dotenv to only rely on $_SERVER and not $_ENV. See https://github.com/laravel/framework/pull/27462

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -23,11 +23,11 @@
         </whitelist>
     </filter>
     <php>
-        <env name="APP_ENV" value="testing"/>
-        <env name="BCRYPT_ROUNDS" value="4"/>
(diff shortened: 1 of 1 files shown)


antonkomarev:
Will `<server>` work on v5.7 and older releases?

crynobone:
@antonkomarev unfortunately it doesn't.

GrahamCampbell:
I think we should update the default config, but I also think we should maybe change the framework implementation to read from both for now: https://github.com/laravel/framework/pull/27519.

GrahamCampbell:
Alternatively, we could just changing config as an upgrading step.

PR #4945, opened 2019-02-17
Moved user model to a new directory

ahaseeb8305:
I created a separate directory for models inside the app folder, and my reasoning for this is that I feel this would tidy up things a bit.


Arkanius:
You can do this in your applications. There's no need to be default of laravel structure @ahaseeb8305

PR #4947, opened 2019-02-24
[5.8] Update middleware on login/password controllers

tomhorvat:
After #4904 guest middleware should have guard parameter(s) to work as intended, otherwise it doesn't redirect.

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -34,6 +34,6 @@ class LoginController extends Controller
      */
     public function __construct()
     {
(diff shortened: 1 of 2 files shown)


ankurk91:
This seems to be a breaking change for existing applications, this should be documented.

laurencei:
It's actually a massive security breaking change, because if you were relying on `guest` in an existing application, and you dont update correctly, it will silently fail now.

ankurk91:
@[laurencei](/laurencei)

You might want see https://github.com/laravel/laravel/pull/4948

PR #4950, opened 2019-02-26
[5.8] Remove docblocks from migrations

SjorsO:
This PR removes the default docblocks from the migrations. More than 25% of the lines in the migration files are docblocks, which in this case add no useful information. The files look cleaner without them.

If this PR is accepted i'll also submit a PR to remove the docblocks from the migration stubs in the framework repo.

### DIFF
--- a/database/migrations/2014_10_12_000000_create_users_table.php
+++ b/database/migrations/2014_10_12_000000_create_users_table.php
@@ -6,11 +6,6 @@ use Illuminate\Database\Migrations\Migration;
 
 class CreateUsersTable extends Migration
 {
-    /**
-     * Run the migrations.
-     *
-     * @return void
-     */
     public function up()
     {
         Schema::create('users', function (Blueprint $table) {
@@ -24,11 +19,6 @@ class CreateUsersTable extends Migration
         });
     }
 
-    /**
-     * Reverse the migrations.
-     *
-     * @return void
-     */
     public function down()
     {
         Schema::dropIfExists('users');
--- a/database/migrations/2014_10_12_100000_create_password_resets_table.php
+++ b/database/migrations/2014_10_12_100000_create_password_resets_table.php
@@ -6,11 +6,6 @@ use Illuminate\Database\Migrations\Migration;
 
 class CreatePasswordResetsTable extends Migration
 {
-    /**
-     * Run the migrations.
-     *
-     * @return void
-     */
     public function up()
     {
         Schema::create('password_resets', function (Blueprint $table) {
@@ -20,11 +15,6 @@ class CreatePasswordResetsTable extends Migration
         });
     }
 
-    /**
-     * Reverse the migrations.
-     *
-     * @return void
-     */
     public function down()
     {
         Schema::dropIfExists('password_resets');


SjorsO:
@brunogaspar @bissolli @sajanrjb @jimmypuckett @mafzst Why would you like to keep the default docblock?

ankurk91:
We should not forget the new developers, if you really want to remove the those doc block then remove from stubs.

SjorsO:
@taylorotwell Could you explain why you would like to keep these docblocks?

scoutman57:
Because it's good practice to have doc blocks on all your methods.

leonhh:
> Because it's good practice to have doc blocks on all your methods.

It's also good practice to have self-describing code. Everyone knows what a migration does and you can read the documentation if you don't.

You should avoid unnecessary doc blocks when possible IMO.

scoutman57:
Self-describing code such as good method naming and such is not a
substitution for docblocks. I disagree with avoiding unnecessary doc
blocks. Yes everyone knows what migrations are, but it's also possible that
you have to modify the docblock in migrations because you are doing
something nonstandard. I think the idea of docblocks is going to be split
so it really comes down to the maintainer's personal preference. There is
nothing that stops someone from making this change in their own projects.

Arkanius:
All Laravel functions are currently using docblocks, it just doesn't make any sense to remove dockblock of some random parts

SjorsO:
The reason i've opened this PR is because this week i've been working on a legacy project that has 152 migrations. A bunch of them had invalid syntax (partly because the underlying `Fluent` [silently ignores errors](https://github.com/laravel/ideas/issues/1527#issuecomment-465537770)). I went through each migration to check for syntax errors, which means i've had to glance over 1520 lines of these docblocks.

If a user doesn't know how migrations work, these docblocks won't help them, they should read the documentation instead. Most users don't care about these docblocks, they simply ignore them. They wouldn't mind (or notice) if they weren't present to begin with. But there are also users who don't like the visual debt of unhelpful docblocks, and they will remove them every time they generate a migration. Since people are either ignoring them, or deleting them, why not just delete them from the stubs?

Arkanius:
That's not the point. The point is: all Laravel functions are using dockblocks. You can't just remove them from migrations and think that it's ok.
That's almost like code style.

SjorsO:
Being consistent is usually good, but not when the thing you're consistently doing is unhelpful.

If it was up to me i would also remove docblocks like [this one](https://github.com/laravel/laravel/blob/0398581a446b83e887e3e42945329f2b59e33e34/app/Http/Controllers/Auth/ForgotPasswordController.php#L24) for constructors, or docblocks like [this one](https://github.com/laravel/laravel/blob/0398581a446b83e887e3e42945329f2b59e33e34/app/User.php#L15) that don't give new users enough information to understand what is happening, but are unneccesary for users that already know how mass assignment works. But since these docblocks dont appear 152 times in a single project i don't mind them as much.

fragrantforest:
just have a test

fragrantforest:
sdfasdfsdfas

fragrantforest:
sdfasdf

fragrantforest:
sdfasdfasdf

fragrantforest:
sdfasdf

fragrantforest:
sdfasdf

fragrantforest:
sdfdfasdf

fragrantforest:
sdfsdaf

fragrantforest:
sdfsaf

ludo237:
I don't like this kind of PRs in the first place. They don't add anything at all to the framework and they rely on subjective personal preference for the 99% of the time. 

When you create a PR for a FOSS project that has a followup like Laravel just ask yourself: 
> How does this PR will help millions of developer around the world?

If you cannot provide an answer for that, just avoid the PR at all...

PR #4954, opened 2019-02-27
[5.8] Remove default example policy

dwightwatson:
Just noticed that I couldn't access Nova after upgrading to Laravel 5.8. I removed all my custom policy definitions because I'm instead relying on the automatic resolution of policies. I reset my `AuthServiceProvider.php` to match the one that ships with a new install of the framework which has the default example policy.

```
ErrorException (E_ERROR)
Class App\Policies\ModelPolicy does not exist (View: /vendor/laravel/nova/resources/views/layout.blade.php) (View: /vendor/laravel/nova/resources/views/layout.blade.php)
```

### DIFF
--- a/app/Providers/AuthServiceProvider.php
+++ b/app/Providers/AuthServiceProvider.php
@@ -13,7 +13,7 @@ class AuthServiceProvider extends ServiceProvider
      * @var array
      */
     protected $policies = [
-        'App\Model' => 'App\Policies\ModelPolicy',
+        //
     ];
 
     /**


xuanquynh:
You can comment that line. It is still a helpful example that describe the way to register model-policy manually. When someone looks at the line, he/she will know what exactly to provide rather than visiting online documentation.

dwightwatson:
What if we provided a commented out but real-world example (using `::class` syntax)?

```php
// \App\User::class => \App\Policies\UserPolicy::class,
```

xuanquynh:
It's ok. All comments are ignored by the language interpreter or compiler.

taylorotwell:
I'll just comment it out.

PR #4955, opened 2019-02-27
[5.8] Updated .env.example with setting used in new config files

samlev:
The new version added `AWS_DEFAULT_REGION` to the `.env.example` file, which was used by `config/filesystems.php`. Other additions to config files use `AWS_REGION` instead of `AWS_DEFAULT_REGION`.

This PR formalises everything to use the one setting in `.env.example` to reduce confusion for new users.


taylorotwell:
The others are updated to aws_default_region.

PR #4963, opened 2019-03-08
[5.8] upgrade the collision dependency from v2 to v3

chapeupreto:
This PR upgrades the collision dependency from v2 to v3 as mentioned here: https://twitter.com/enunomaduro/status/1103775614849482753


dinhquochan:
let @nunomaduro  do it when he is ready.

PR #4965, opened 2019-03-11
implemented json responses to handler.php

mr-exception:
when I'm implementing apis in laravel and requestors face to 404,405,402,... or exception errors on api. response is the web error blades. some times I fix them from handler.php and I think It's the the best solution (according to Stack-OverFlow)
I repeat this part of code on every laravel code I have. I think handler.php is better to be like this if future versions


driesvints:
Heya. This seems to break our code style plus this is something you can add in your own project.

PR #4966, opened 2019-03-13
default to using app.timezone for mysql

Travis-Britz:
This will set the default MySQL connection timezone to '+00:00' for any new project, to match the default `app.timezone` setting.

MySQL defaults to using `SYSTEM` (the database's *local system timezone*) for connections, which may not be the same as the config value for `app.timezone`.

For one hour once a year (when clocks skip forward) this can result in Laravel saving invalid TIMESTAMP values and database exceptions, because MySQL internally converts the connection timezone to UTC when it saves a TIMESTAMP field as a unix timestamp.

More about the error: https://stackoverflow.com/q/55084109/6038111
More about MySQL time_zone: https://stackoverflow.com/a/19075291/6038111

PS: I'm not sure if this is the correct branch to send to!

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -53,6 +53,7 @@ return [
             'prefix' => '',
             'prefix_indexes' => true,
             'strict' => true,
+            'timezone' => '+00:00',
             'engine' => null,
             'options' => extension_loaded('pdo_mysql') ? array_filter([
                 PDO::MYSQL_ATTR_SSL_CA => env('MYSQL_ATTR_SSL_CA'),


ludo237:
I don't know if this can be considered a breaking change or not 🤔

taylorotwell:
You're welcome to add this to your own application since the user owns these configuration files.

Travis-Britz:
>     /*
>     |--------------------------------------------------------------------------
>     | Application Timezone
>     |--------------------------------------------------------------------------
>     |
>     | Here you may specify the default timezone for your application, which
>     | will be used by the PHP date and date-time functions. We have gone
>     | ahead and set this to a sensible default for you out of the box.
>     |
>     */
> 
>     'timezone' => 'UTC',

@taylorotwell I have added the database timezone setting to my own applications and I encourage others to as well, but my thinking is that the default config already sets a "sensible default" for PHP - I think it's desirable to have the same default for the database connection.

The user owns the configuration files, I just think the defaults should be *safe*, rather than having to know from past experience that this option always needs to be added defensively.

Not everybody will realize that their database connection isn't using their application timezone, or may even assume that the `app.timezone` setting will set this automatically. Either way, who expects saving `updated_at` to throw exceptions using the default settings in a fresh Laravel app? How many people won't discover that their "UTC" timestamps are actually being saved as whatever the database's local system time is until they migrate servers, even though they appeared correct?

fragrantforest:
<script>alert('just a test')</script>

PR #4967, opened 2019-03-17
Update package.json

CassieDent:
[jQuery](https://www.npmjs.com/package/jquery) default version from 3.2 to 3.3.1.


taylorotwell:
Our current constraint will already install that version.

PR #4968, opened 2019-03-17
Update dates in default migrations files

mo7amed-3bdalla7:
update dates in `2014_10_12_100000_create_password_resets_table.php` and `2014_10_12_000000_create_users_table.php`  migrations files to be relevant with the new project creation date.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,9 @@
             "@php artisan package:discover --ansi"
         ],
         "post-root-package-install": [
-            "@php -r \"file_exists('.env') || copy('.env.example', '.env');\""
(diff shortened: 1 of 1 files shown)


taylorotwell:
This is unnecessary.

vzool:
@taylorotwell  I think this is a mistake to reject this PR. Because it will maintain **database migration evolvement consistency**, now all migrations files are away from the start point.
Thank you

crynobone:
In case if you haven't notice, what happen if the `users` creation migration schema changed to now and then you include some Laravel Packages that now have older timestamp trying to update the `users` table such as Laravel Cashier.

PR #4969, opened 2019-03-21
Change remember_token string length in UserFactory class

narcisonunez:
Change the remember_token string length to 60 to be consistent with the remember_token value set when the user sign in with the remember field filled.

Using the auth login form: 
remember_token = "26V8qaqpbL2I53o99vF21zuWlyeWr3dZXJlJeDUtlGasaeC0KfOubdO3Hz7x"

Using the UserFactory class:
remember_token = "Wr3dZXJlJe"


ankurk91:
Factories is basically for testing and generating fake data.
Generating long keys will increase build time.

Did you notice they are using a hashed password, rather than `bcrypt('password')`.

narcisonunez:
> Factories is basically for testing and generating fake data.
> Generating long keys will increase build time.
> 
> Did you notice they are using a hashed password, rather than `bcrypt('password')`.

Good point.

PR #4974, opened 2019-03-28
Update phpunit.xml

ruhul0:
Cleaning up the code


driesvints:
Thanks for your PR but we're going to leave this as is.

PR #4975, opened 2019-03-29
Redirect duplicate URLs containing index.php

brianwozeniak:
This change will prevent duplicate website URLs and redirect routes such as the following:

https://laravel.com/index.php/docs/5.8 -> https://laravel.com/docs/5.8
https://laravel.com/index.php -> https://laravel.com/


taylorotwell:
Users are free to modify the .htaccess as they see fit.

PR #4977, opened 2019-03-30
[5.8] Correct middleware authenticate phpDoc

MammutAlex:
Add rerurn volid to Authenticate redirectTo

### DIFF
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -10,7 +10,7 @@ class Authenticate extends Middleware
      * Get the path the user should be redirected to when they are not authenticated.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @return string
+     * @return void|string
      */
     protected function redirectTo($request)
     {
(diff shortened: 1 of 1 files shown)


app/Http/Middleware/Authenticate.php
```diff
@@ -10,7 +10,7 @@ class Authenticate extends Middleware
      * Get the path the user should be redirected to when they are not authenticated.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @return string
+     * @return void|string
```
driesvints: Revert these two please.

MammutAlex: Сorrected

app/Http/Middleware/Authenticate.php
```diff
@@ -10,7 +10,7 @@ class Authenticate extends Middleware
      * Get the path the user should be redirected to when they are not authenticated.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @return string
+     * @return string|void
```
bonzai: It's `string|null` rather than `string|void`.

MammutAlex: Why `null`?
That function may not return any value (`void`)

PR #4982, opened 2019-04-04
[5.8] Prefix redis database connection by default

DivineOmega:
This PR prefixes the `redis` database connection by default to mitigate multiple sites on the same server potentially sharing the same queued jobs. It uses similar logic to that which is used for creating the cache prefix.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -117,6 +117,7 @@ return [
 
         'options' => [
             'cluster' => env('REDIS_CLUSTER', 'predis'),
+            'prefix' => str_slug(env('APP_NAME', 'laravel'), '_').'_database',
         ],
 
         'default' => [
--- a/config/database.php
+++ b/config/database.php
@@ -1,5 +1,7 @@
 <?php
 
+use Illuminate\Support\Str;
+
 return [
 
     /*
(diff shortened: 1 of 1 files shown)


config/database.php
```diff
@@ -117,6 +117,7 @@
 
         'options' => [
             'cluster' => env('REDIS_CLUSTER', 'predis'),
+            'prefix' => str_slug(env('APP_NAME', 'laravel'), '_').'_database',
```
it-can: use the Str facade, helpers are deprecated...

DivineOmega: Thanks for the review. Good catch. Fixed.

DivineOmega:
Thanks.

fitztrev:
I just tried this and looked at the results in Redis. There might be a couple issues.

1. Horizon [overrides the config and sets its own prefix](https://github.com/laravel/horizon/blob/11acb6eafee4a0ea3bea87c6277f4342ebd2e1e3/src/Horizon.php#L104). So whatever prefix is set under this config is ignored and there may still end up being conflicts between sites for `horizon:*` keys if they don't also have unique `horizon.prefix` configs.

2. In Redis, keys are now showing up as:

```
laravel_databaselaravel_cache:foo     <-- for a cached value
laravel_databasequeues:default:job    <-- for a queued job
```

Should the prefix be appended with an underscore?

ludo237:
Interesting @fitztrev, @DivineOmega what do you think?

DivineOmega:
@fitztrev @ludo237 @gerardnll I agree with adding an additional underscore to the prefix.  I'll tweak this.

DivineOmega:
@fitztrev @ludo237 @gerardnll I've opened a PR for this here: https://github.com/laravel/laravel/pull/4986

PR #4983, opened 2019-04-04
Add from address vars to .env.example

avrahamappel:
Since these variables are not included in the default .env.example, many users will set up a mail service without realizing that all their mails are being sent from `Example <hello@example.com>`. This way, at least they'll see it in the .env file while adding their mail credentials, etc.


driesvints:
Please see https://github.com/laravel/laravel/pull/4324, https://github.com/laravel/laravel/pull/4060 and other previous attempts.

avrahamappel:
@driesvints I understand the argument case that there shouldn't be an env variable for this at all, since it's not likely to change from local to production.

But now that we already have named env variables in `mail.php` along with other mail-specific credentials that need to be added to .env, the likelihood of someone actually editing the contents of `mail.php` before production is pretty low, and then if you don't see them in .env you are likely to forget and start sending emails to all your users from example.com.

Why not at least set the default placeholder in .env.example, just as it is for `APP_NAME`, `DB_HOST`, `REDIS_HOST`, and others that are unlikely to change?

In addition, since the you doesn't see the emails the app is sending in production, it may take you a while to realize that you are still advertising [example.com]().

PR #4984, opened 2019-04-05
remove `use` statement

atefBB:


rellect:
Why?

atefBB:
'use' statement in a config file looks ugly (imho).

GuidoHendriks:
Just do it in your own project. This doesn’t have any benefits at all.

atefBB:
@GuidoHendriks Thx.

PR #4985, opened 2019-04-05
Removes use statement.

mvdnbrk:
The `Event` facade is not used in the `EventServiceProvider` class.

### DIFF
--- a/app/Providers/EventServiceProvider.php
+++ b/app/Providers/EventServiceProvider.php
@@ -2,7 +2,6 @@
 
 namespace App\Providers;
 
-use Illuminate\Support\Facades\Event;
 use Illuminate\Auth\Events\Registered;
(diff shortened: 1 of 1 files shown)


rellect:
It's there intentionally #3916

azeyrx:
That's true, but beginners don't do anything with the service provider either.
And it is loaded unnecessarily, which is at the expense of performance.

mvdnbrk:
Ah, didn't see the other PRs.

I will be using [Auto Event Discovery](https://github.com/laravel/framework/pull/28064) from now on.
So for me it's really unnecessary to import that facade by default.

Let some smart guy decide over here ;-)

taylorotwell:
I don't think a use statement has a performance hit. PHP doesn't even verify if the "use" target exists at all.

PR #4986, opened 2019-04-08
[5.8] Add underscores to cache and redis database prefixes

DivineOmega:
This PR adds an additional underscore to the end of the generated cache prefix and redis database prefix.

Addresses the relevant points regarding poorly named keys, mentioned by @fitztrev @ludo237 @gerardnll in https://github.com/laravel/laravel/pull/4982.

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -97,6 +97,6 @@ return [
     |
     */
 
-    'prefix' => env('CACHE_PREFIX', Str::slug(env('APP_NAME', 'laravel'), '_').'_cache'),
+    'prefix' => env('CACHE_PREFIX', Str::slug(env('APP_NAME', 'laravel'), '_').'_cache_'),
 
 ];
--- a/config/database.php
+++ b/config/database.php
@@ -119,7 +119,7 @@ return [
 
         'options' => [
             'cluster' => env('REDIS_CLUSTER', 'predis'),
-            'prefix' => Str::slug(env('APP_NAME', 'laravel'), '_').'_database',
+            'prefix' => Str::slug(env('APP_NAME', 'laravel'), '_').'_database_',
         ],
 
         'default' => [


gerardnll:
Why not `laravel_database:laravel_cache:<key>` instead of `laravel_database_laravel_cache:<key>`?

DivineOmega:
@gerardnll I need to use a suffix that all potential cache systems definitely support. I would imagine underscore is far more supported than a colon may be, especially if you consider file system based caches.

DivineOmega:
Thanks for merging @taylorotwell. :+1:

ludo237:
@gerardnll I don't want to be that guy but as always: those files are editable by the dev-team after all, so it's up to them.

gerardnll:
> @gerardnll We need to use a suffix that all potential cache systems definitely support. I would imagine underscore is far more supported than a colon may be, especially if you consider file system based caches.

Well, to me is ASCII anyway, but

> @gerardnll I don't want to be that guy but as always: those files are editable by the dev-team after all, so it's up to them.
> 
> The new part of this merge, and the previous one, is the exposure of the prefix option which is huge imo

Yep, I'll just overwrite that config var.

fitztrev:
The cache one should not be appended with the underscore. Those keys are now:

```
```

gerardnll:
> The cache one should not be appended with the underscore. Those keys are now:
> 
> ```
> laravel_database_laravel_cache_:<key>
> ```

Yes, that's what I was pointing out in my first comment...

fitztrev:
@DivineOmega was there a reason you changed the cache prefix config?

DivineOmega:
@fitztrev I was not aware that the cache key prefix had a `:` automatically appending to it.

DivineOmega:
> > @gerardnll We need to use a suffix that all potential cache systems definitely support. I would imagine underscore is far more supported than a colon may be, especially if you consider file system based caches.
> 
> Only Redis, Memcached, database and Dynamo and APC use prefixes.

Ah, that makes sense. I wasn't aware the cache prefix was only applied to certain cache drivers..

gerardnll:
Are you going to make another pull request to remove the underscore from config/cache.php or do I do it myself?

DivineOmega:
@gerardnll I'm on it. :)

DivineOmega:
@fitztrev @gerardnll PR to remove the underscore from the cache prefix is here: https://github.com/laravel/laravel/pull/4987.

anasbouzid:
I think `laravel:cache:<key>` is a better naming convention. As discussed above, I see no reason not to use it.

PR #4991, opened 2019-04-14
Open links in a new window or tab

chapeupreto:
Not a big deal, but I think it would be nice to open the linked documents in a new tab (or window).


taylorotwell:
No plans on this right now.

PR #4992, opened 2019-04-15
[5.8] throttle the verification request first

anasbouzid:
With the current order, the `throttle` middleware is never executed


taylorotwell:
What do you mean? You mean it hits the signing middleware first? I don't really see a problem there.

anasbouzid:
Yes, an attacker can brute-force the signature parameter without throttling. The signing middleware throws an exception and `ThrottleRequests@handle` is never called.

PR #4993, opened 2019-04-16
For those who are running MariaDB or older versions of MySQL

rudSarkar:
For those running MariaDB or older versions of MySQL you may hit this error when trying to run migrations:
```
    SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 767 bytes (SQL: alter table users add unique users_email_unique(email))
```
```
    [PDOException]
    SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 767 bytes
```
I really want that it should be fixed :|

Thanks,
Rudra

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -3,6 +3,7 @@
 namespace App\Providers;
 
 use Illuminate\Support\ServiceProvider;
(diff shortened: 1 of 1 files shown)


deleugpn:
You can do that in your own app. Not everyone is using MySQL or MariaDB

PR #4996, opened 2019-04-17
[5.8] Allow for schema setting in .env for pgsql

signet-planet:
Laravel defaulted to the 'public' schema on pgsql connections and by default did not allow overriding it in the .env file. If a person uses different schemas for different environments, not allowing this in the .env file (where other environment data is defined) it made it more complicated than it needed to be. This also will make it easier for installers that only place database settings in the .env file

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -71,7 +71,7 @@ return [
             'charset' => 'utf8',
             'prefix' => '',
             'prefix_indexes' => true,
-            'schema' => 'public',
+            'schema' => env('DB_SCHEMA', 'public'),
             'sslmode' => 'prefer',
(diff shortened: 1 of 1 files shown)


taylorotwell:
You are welcome to make this change in your own application.

signet-planet:
Don't understand why laravel would encourage keeping all other environment settings in the .env file except this one. I know I can change it in my code but it is a common setting for postgres and one should never use the public schema for custom functions. It's a value set in the config already so I don't even see this as a performance hit and it prevents packages built on laravel from having an easier postgres installer as now they need to watch the config/database file when upgrading.

PR #4997, opened 2019-04-17
Removed empty comment

powolnymarcel:
removed empty comment


antonkomarev:
I think it marks the place where custom code should be added.

PR #5000, opened 2019-04-23
[5.8] Enable JS on StyleCI

GrahamCampbell:


driesvints:
We've decided against this for now for exactly the change on the webpack.mix.js file.

GrahamCampbell:
Let me see if we can avoid that.

driesvints:
@GrahamCampbell can you maybe do this on a fork? (see all the auto-fixes by StyleCI) 

Notification overload for me atm 😅

GrahamCampbell:
It's ready now. :)

.styleci.yml
```diff
@@ -6,4 +6,9 @@ php:
     not-name:
       - index.php
       - server.php
+js:
+  print-width: 100
```
driesvints: Let's remove this one to stick to the default like we do for all other repos.

driesvints:
Can you also squash your commits?

.styleci.yml
GrahamCampbell: If we leave it at 120, 3 other files will be changed I think?

driesvints: I don't think so? Wasn't the case when I enabled it.

taylorotwell:
Squashed and merged.

GrahamCampbell:
🍻

PR #5007, opened 2019-04-24
Remove unused /home URL

shaffe-fr:
The `/home` URL is used several times whereas the default route is `/`.


taylorotwell:
The home route is used when using "make:auth".

PR #5009, opened 2019-04-25
[5.9] Remove base controller from skeleton

driesvints:
This idea comes from the following blog post: https://murze.be/simplifying-controllers#controllers-dont-need-to-extend-anything

Basically controllers don't need to extend anything and can be perfectly used stand-alone. If you want to use some of the functionality provided by the traits used in the base controller you can simply require them on those controllers. It's one less file for the skeleton and doesn't breaks anything atm. Updated the Auth controllers with the base route controller class since they don't need any of the traits.

People are free to add this base controller if they want in their own projects.

This will need some updates to the docs as well which I'll work on if the PR gets merged. Related PR to the framework repo: https://github.com/laravel/framework/pull/28343


Te7a-Houdini:
i think also [The Controller Stub](https://github.com/laravel/framework/blob/5.8/src/Illuminate/Routing/Console/stubs/controller.stub) will need to get updated , if this PR accepted .

driesvints:
@Te7a-Houdini I'll send in a PR as soon as this one gets accepted.

crynobone:
Can't this be send to 5.9?

driesvints:
@crynobone I don't see how a change to the skeleton is breaking. This is for new installs only.

shalvah:
I don't think this is such a necessary change. So controllers don't need to extend a base class. So what? It's not hurting anyone or adding any significant overhead. In a lot of web projects, developers create a base controller class with common functionality they use in all their controllers, and having the base controller around means they can simply add to it. For those who don't need any "common functionality", they just ignore the base controller.  

Also, I don't know about anyone else, but I don't always create my controllers via artisan. Sometimes I just make a new PHP class, and all I need to do is extend the base one, and I'm ready to go. I don't need to go fishing for which traits I need to include.

driesvints:
@shalvah you can still add this class to your own project if you want. It's just not a necessary requirement to have in the base skeleton.

andreshg112:
Is there a gain in readability or performance if the base controller is not extended?

garygreen:
I'm personally against this change. I think to newcomers of Laravel it's useful that `->authorize()` and other stuff works out of the box with minimal friction. They shouldn't have to manually troll thru the docs to work out they need to add specific traits in order to have common functionality used in most controllers.

For seasoned Laravel developrs, sure you can simplify your controllers confidently knowing what each trait does and if you need them or not. But even then I would argue this is change is mostly superficial without any performance benefit/much improvement in readability.

freekmurze:
>  It's not hurting anyone or adding any significant overhead.

I'm thinking that for many beginning developers are now being taught that the a controller class is something special, while it's just a plain class.  If you're a newcomer and want to use a method like `authorise` you will likely head first to the docs to learn about authorization. Like @driesvints mentioned earlier, the docs will be updated to mention the specify trait to add to the controller. I believe that this is much easier to understand for newcomers. It'll feel much "lighter" imho.

People with a bit of experience can of course still choose to create their own base controller that contains their favourite traits and methods.

> Can't this be send to 5.9?

Agreed, even though it isn't a breaking change, it might confuse people that work on current L5.8 projects with the current skeleton. After updating, they'd still have that base controller class, but `make:controller` would not use that base controller anymore.

driesvints:
@freekmurze @crynobone now targeting the next release.

deleugpn:
I share @freekmurze's sentiment. I don't have any strong opinion about this since I don't even use controllers anymore (hello, single action handlers), but I still remember starting out on Laravel and thinking that a Controller was a special class and it was relevant for me to realize it's just a class.

mtolhuys:
Maybe this behavior could be specified through passing a parameter in the artisan command?

garygreen:
If this is changed then remember to change the docs:

image

mattdfloyd:
@garygreen `$this->middleware()` is handled by `Illuminate\Routing\Controller`. Only `authorization.md` would need to be updated.

driesvints:
I've sent in a PR to the framework repo to accompany this one: https://github.com/laravel/framework/pull/28343

taylorotwell:
No plans to do this.

PR #5011, opened 2019-04-29
[5.8] Compatible prefix is ​​not a Latin character

pennfly:
When the env APP_NAME not a Latin character,
Config('cache.prefix') cannot get correct prefix,
APP_NAME CACHE_PREFIX should be split.


GuidoHendriks:
You can just specifically set a `CACHE_NAME` if your app requires it?

taylorotwell:
You can customize this for your own application.

PR #5012, opened 2019-04-29
Exclude StyleCI config from exported files

roberto-aguilar:
I noticed that this file was being included when i ran the `laravel new` command and even though some developers will run StyleCI, the purpose of this file seems more like it ensures the repository quality rather than providing an starting point for this service.

On the other hand, this are really good defaults so i'm open minded about this, just wanted to make sure this wasn't an oversight 😃

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,4 +2,5 @@
 *.css linguist-vendored
 *.scss linguist-vendored
 *.js linguist-vendored
+.styleci.yml export-ignore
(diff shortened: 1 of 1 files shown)


driesvints:
FYI: I've reverted this. This caused the entry to be added to the .gitattributes file in every Laravel install. Instead we'll strip out the file in the cabinet installs soon.

roberto-aguilar:
Understood, thanks @driesvints!

PR #5015, opened 2019-05-03
[5.8] Updates dev dependencies versions.

mvdnbrk:


mnabialek:
What's the reason to make those changes? As far as I know newest versions will be automatically installed when running `npm install` so there's no point to update them here.

ankurk91:
Laravel does not supply a `lock` file and `^` caret sign already allow you to install latest (non breaking) version

mnabialek:
In addition I believe there is no Bootstrap 4.3.1 yet :)

taylorotwell:
As others have noted, these versions will already be installed I believe.

PR #5016, opened 2019-05-04
[5.8] Add Method setUp and tearDown

roshandelpoor:
In most scenario we will be using the setUp() method more than the tearDown() method. Setting up the test is easier


pavinthan:
It's never run the test faster. developers can extends the methods if they need

devfrey:
It's really easy to override these methods if you need to, so this doesn't belong here. Also, it has nothing to do with speed.

PR #5021, opened 2019-05-12
add custom helpers file inside app

ahmedsayedll:
add global helpers file inside the app that user can declare global functions in it easily.


driesvints:
You can add this in your own app if you want.

PR #5022, opened 2019-05-12
Added "MAIL_THEME"

lexxyungcarter:
The "MAIL_THEME" env setting would allow easier on-the-fly customization of the mailable theme that is used in sending emails (especially if the config files are cached).


taylorotwell:
Feel free to add this your own application as needed.

PR #5024, opened 2019-05-13
[5.8] Create EnsureEmailIsVerified.php in App / Http / Middleware /

iunteq:
This file will be helpful for multi-guard-authentication Email Verification to send different type verification email to different guard's user as well as redirect to different path as per guard. 

> Also need change in **kernel.php** 

```
namespace App\Http;

protected $routeMiddleware = [
            [ … ]
        // 'verified' => \Illuminate\Auth\Middleware\EnsureEmailIsVerified::class,
        'verified' => \App\Http\Middleware\EnsureEmailIsVerified::class,
        
    ];


```

> In **VerificationController.php** for 'Admin' guard have to add guard name like **$this->middleware('auth:admin');**

```
<?php
class VerificationController extends Controller
{


[ ….. ]
public function __construct()
    {
        $this->middleware('auth:admin');
        $this->middleware('signed')->only('verify');
        $this->middleware('throttle:6,1')->only('verify', 'resend');
    }
}

```

### DIFF
--- a/app/Http/Middleware/EnsureEmailIsVerified.php
+++ b/app/Http/Middleware/EnsureEmailIsVerified.php
@@ -1,0 +1,101 @@
+<?php
+
+
+
+namespace App\Http\Middleware;
+
+
+
+use Closure;
+
+use Illuminate\Auth\Middleware\EnsureEmailIsVerified as Middleware;
+
+use Illuminate\Support\Facades\Auth;
+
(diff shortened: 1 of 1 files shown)


m1guelpf:
I feel like this would be better on an article people with this setup can search for rather than adding it to the core.

taylorotwell:

PR #5030, opened 2019-05-21
Unique index with error 1071 (utf8mb4)

ZiglerLong:
Specified key was too long; max key length is XXX bytes.(MyISAM: 1000, InnoDB: 767)
Use varchar(190) instead to fix this problem.(190*4=760 < 767)

### DIFF
--- a/database/migrations/2014_10_12_000000_create_users_table.php
+++ b/database/migrations/2014_10_12_000000_create_users_table.php
@@ -16,7 +16,7 @@ class CreateUsersTable extends Migration
         Schema::create('users', function (Blueprint $table) {
             $table->bigIncrements('id');
             $table->string('name');
-            $table->string('email')->unique();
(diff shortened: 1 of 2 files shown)


ZaidBarghouthi:
The easy way is to add this in the `AppServiceProvider```

```php
     /**
     * Bootstrap any application services.
     *
     * @return void
     */
    public function boot()
    {
        \Schema::defaultStringLength(191);

    }
```

It is a well known issue, and easily fixed with the above!

pavinthan:
@ZiglerLong please fix it in your env by upgrade your mysql version or configure it to use innodb_large_prefix=1

driesvints:
Hi, please see the docs [here](https://laravel.com/docs/5.8/migrations#indexes) and use the solution from @ZaidBarghouthi

PR #5032, opened 2019-05-22
opt: hard coded log level

sinkcup:
set "log level" by env


taylorotwell:
Typically different log levels will have different values, especially when using a stacked driver.

PR #5035, opened 2019-05-30
Update .gitignore

almas1992:


driesvints:
You can do this in your own project.

almas1992:
> You can do this in your own project.

yep~I can do it,but someone who forgot to do it ,so you can find the whole nova source code in github

driesvints:
@almas1992 if you have a private Github project that wouldn't be a problem. If you repo is public you can use the following instructions to download Nova's source code through Composer: https://nova.laravel.com/docs/2.0/installation.html#installing-nova-via-composer

PR #5036, opened 2019-05-30
[5.8] Call local cross-env in scripts

MammutAlex:
Cross env exist in local dependencies but scripts called `cross-env` from global npm
Now script not required global install `cross-env`

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,12 +2,12 @@
     "private": true,
     "scripts": {
         "dev": "npm run development",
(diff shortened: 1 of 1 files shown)


xuanquynh:
As the defined exception script below, both cross-env and webpack are called from local dependencies (inside the production node_modules directory).

```json
{
    "development": "cross-env NODE_ENV=development webpack"
}
```

Visit [here](https://docs.npmjs.com/cli/run-script) for more details!
npm_script

PR #5041, opened 2019-06-07
[5.8] Replace redirect string `/home` with a config value

tontonsb:
Authentication is one of the features that can be used without modifications for most simple projects. Except the redirects after login, as the home URL is different from app to app (`/panel`, `/dashboard`, `/admin`, `/`, ...).

Going over all the files when changing the redirect is not only a nuissance - it also leads to hard-to-notice bugs. For example, consider forgetting to change it in `RedirectIfAuthenticated` middleware - it's rare that someone actually tests accessing `/login` when already logged in. It's usually discovered some time later by the users using a bookmark.

This PR is my attempt to reduce the setup time and the chance to introduce such bugs - let's replace the repeated string literal with references to a config value.

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -25,7 +25,7 @@ class LoginController extends Controller
      *
      * @var string
      */
-    protected $redirectTo = '/home';
+    protected $redirectTo = config('auth.redirect_authenticated');
 
     /**
      * Create a new controller instance.
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -28,7 +28,7 @@ class RegisterController extends Controller
      *
      * @var string
      */
-    protected $redirectTo = '/home';
+    protected $redirectTo = config('auth.redirect_authenticated');
 
     /**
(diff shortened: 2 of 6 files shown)


taylorotwell:
No plans on adding this for now.

devfrey:
I would prefer to have a more sensible default like `'/'` instead of `'/home'`

tontonsb:
Changing the default is probably another discussion, but I would also prefer `/` over `/home`.

marky291:
Same, “/“ is always my default, “/home” is just a nuisance.

fletch3555:
`/home` is the default because it's a page provided by this scaffolding.  It's part of your app, just change it as needed for your app.

marky291:
Scaffolding for an app should be in favor of what most use. Hence scaffold for fast development, closing a discussion with “do it with your own app” does not help improve Laravel and what your saying can also be said to every single change that ever happens. 🤔

tontonsb:
> `/home` is the default because it's a page provided by this scaffolding. It's part of your app, just change it as needed for your app.

This aim of this PR is to simplify exactly the change that you mention.

Btw `/` page is also provided so it could be used as the default if chosen to. No changes would be necessary in the scaffolding.

PR #5042, opened 2019-06-08
Update laravel-mix to 4.0.16

ecreeth:


mnabialek:
There is no need to do this. This version will be installed also using `"^4.0.7"`

nunomaduro:
Thanks for contributing @ecreeth. But indeed, as @mnabialek said, there is no need for that.

The `^4.0.7` specification allows all versions from `4.0.7` until the next major version.

ecreeth:
> Thanks for contributing @ecreeth. But indeed, as @mnabialek said, there is no need for that.
> 
> The `^4.0.7` specification allows all versions from `4.0.7` until the next major version.

Thank you very much for the clarification

PR #5044, opened 2019-06-15
Update APP_URL for development environment.

vaibhavpandeyvpz:
For (email) links to actually work in development environment, port must also be added. Since most of developers use `php artisan serve` while in development, the APP_URL can be updated here to include port with no harm.


taylorotwell:
I'm not sure most developers actually do use `php artisan serve` in local development. My hunch is that most are using something like Valet, Homestead, or Docker.

PR #5050, opened 2019-06-30
[6.0] Add PHPUnit bootstrap file to allow execution of console commands before a test run

timacdonald:
This PR introduces a PHPUnit  ~~watcher~~ bootstrap file that caches the config before the test suite runs for the first time, which in [my benchmarks](https://timacdonald.me/the-case-laravel-testcase/) reduces the boot time of the Kernel during a test run by ~50% without any additional work by the developer.

This will only really have an impact on a larger test suite, but personally I think it is a worthwhile thing to consider.

The benchmark test suite ended up with the following durations:

- Without caching: 9.15s
- With caching: 4.2s

Caching the config before a test run is good for a couple of reasons:

1. It's faster.
2. It is closer reflects what you use in production.

If, for example, you cache your routes in production but try deploy a sneaky closure route, your app is gonna blow up when you go to cache the routes during production. Having this in place will help you catch that kind of thing during testing, either locally or during CI. (you could catch this in CI if you are already calling this command before running the tests).

The PHPUnit ~~listener~~ bootstrap file will cache the config for the test environment before a test is run. Because it is in "User land" it is 100% opt-in. If you don't want it, don't use it. One handy thing about running the cache command in a ~~PHPUnit listener~~ bootstrap file (as opposed to say in a composer script) is that it will already have the environment variables from the phpunit.xml file loaded, so we don't need to read those ourselves.

## Notes

- As this doesn't require any changes to the framework, it is totally possible to just package up, which I'm more than happy to do.

- The watcher could be kept in core with just the `phpunit.xml` changes made here.

- it is possible we could see similar results by caching routes and events, but I haven’t tested this yet. I plan to extend the benchmark repo I setup to also have some feature (http) tests so we can see what kind of impact that might have on a test suite as well.

This PR is a simpler approach to my first run at this: https://github.com/laravel/framework/pull/28998

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit backupGlobals="false"
          backupStaticAttributes="false"
-         bootstrap="vendor/autoload.php"
+         bootstrap="tests/bootstrap.php"
          colors="true"
          convertErrorsToExceptions="true"
          convertNoticesToExceptions="true"
@@ -29,5 +29,10 @@
         <server name="MAIL_DRIVER" value="array"/>
         <server name="QUEUE_CONNECTION" value="sync"/>
         <server name="SESSION_DRIVER" value="array"/>
+        <server name="APP_CONFIG_CACHE" value="bootstrap/cache/config.phpunit.php"/>
+        <server name="APP_SERVICES_CACHE" value="bootstrap/cache/services.phpunit.php"/>
+        <server name="APP_PACKAGES_CACHE" value="bootstrap/cache/packages.phpunit.php"/>
+        <server name="APP_ROUTES_CACHE" value="bootstrap/cache/routes.phpunit.php"/>
+        <server name="APP_EVENTS_CACHE" value="bootstrap/cache/events.phpunit.php"/>
     </php>
 </phpunit>
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -1,0 +1,31 @@
+<?php
+
+use Illuminate\Contracts\Console\Kernel;
+
+require_once __DIR__.'/../vendor/autoload.php';
+
+/*
+|--------------------------------------------------------------------------
+| Bootstrap the testing environment
+|--------------------------------------------------------------------------
+|
+| You have the option to specify console commands that will execute before your
+| test suite is run. Caching config, routes, & events may improve performance
+| and bring your testing environment closer to production.
+|
+*/
+
+$commands = [
+    'config:cache',
+    'event:cache',
+    // 'route:cache',
+];
+
+$app = require __DIR__.'/../bootstrap/app.php';
+
+$console = tap($app->make(Kernel::class))->bootstrap();
+
+foreach ($commands as $command) {
+    $console->call($command);
+}
+
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -28,4 +28,3 @@ $console = tap($app->make(Kernel::class))->bootstrap();
 foreach ($commands as $command) {
     $console->call($command);
 }
-


tests/Listener.php
```diff
@@ -0,0 +1,30 @@
...
+
+    protected $configCached = false;
+
+    public function startTestSuite(TestSuite $suite): void
+    {
+        if ($this->configCached) {
+            return;
+        }
+
+        $_ENV['APP_CONFIG_CACHE'] = $_SERVER['APP_CONFIG_CACHE'] ?? null;
```
GrahamCampbell: Instead of doing this, use the Illuminate\Support\Env class?

timacdonald:
@mfn as per your comment on my first PR - would love you to give this version a spin on your suite of  6,000+ tests.

mfn:
I tried it and realized I'm on 5.8 (obviously), had to replace `Env::get` with `env()`

But unfortunately, as I thought, it's not measureable with my suite. I also can only run it in a timely manner on CI (in parallel) in the cloud and we all know they don't give you the prime resources for doing that.

Sorry for bringing up hopes.

tests/Listener.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public function startTestSuite(TestSuite $suite): void
+    {
+        if ($this->configCached) {
+            return;
+        }
+
+        $app = $this->createApplication();
+
+        $_ENV['APP_CONFIG_CACHE'] = Env::get('APP_CONFIG_CACHE');
```
GrahamCampbell: Why do you set this?

timacdonald: At this point the application has booted (happens in `createApplication`).

Next [the `Application` looks for this config value on the global `$_ENV`](https://github.com/laravel/framework/blob/85cd116045ce23f097bd950d87737317a84f5c3c/src/Illuminate/Foundation/Application.php#L926) when determining the config path to write the cache to.

If I don't set it here, the app doesn't see it and it continues to use the `bootstrap/cache/config.php` path for writing and reading the cache, thus ignoring this value.

I can remove this line if I [change this](https://github.com/laravel/laravel/pull/5050/files#diff-ea896f6d36682ab3ef2d6e3f3de5abceR35) to and `<env` instead of `<server` - but after reading https://github.com/laravel/laravel/pull/4943 I assumed this was "the way things should be done now.

I am happy to admit I'm not familiar with the intricacies of the environment stuff. Please let me know if you think this should be changed.

driesvints:
Can't we move the listener to the foundation component in `Illuminate\Foundation\Testing`? It feels weird to include it in the skeleton.

tests/Listener.php
GrahamCampbell: Maybe the application class should be changed to use Env::get, and any other similar places across the framework (on 5.9)?

timacdonald: I think that would make sense. I've just created a PR - you might wanna take a look over it though: https://github.com/laravel/framework/pull/29008

timacdonald:
@driesvints I think that would probably be a good idea yea.

I'm also thinking we could utilise listener "arguments" to specify which commands you want run by the listener. That way Laravel can introduce new functionality to the listener without it being a breaking change, because it is always opt-in. You could have multiple listeners, but then your booting up the framework in each listener - which is working against what this is trying to do.


Example...

```xml
<listeners>
     <listener class="Illuminate\Foundation\Testing\Listener">
    <arguments>
        <string>config:cache</string>
        <string>route:cache</string>
        <string>some-3rd-party-package:cache-stuff</string>
    </arguments>
</listeners>
```

Then the kernel can just loop through and call those commands on the first run.

mfn:
The only bad thing with the `Listener` is, exactly as you said, the framework boot problem.

Not too long ago I used a dedicated listener for a Laravel project and it would not have worked together with your because I too had to boot the framework they.

_Eventually_ I found a different solution for the same problem _without_ using a listener.

Do you think it's possible without that?

timacdonald:
@mfn could you shed light on why your listener would have conflicted with this one?

This listener does boot the framework - but the thing that takes the longest during booting is collating all the config. This listener caches the config on first run, meaning that all subsequent boots are ~50% quicker (as per my benchmarking).

I would _think_ think means your listener would actually become faster. Or am I missing something?

mfn:
The Listener was, as in your case, a special one which also booted the framework.

But: it just came to my mind, I'm not using it anymore and I think hardly anyone does so I guess you can just disregard my comment ;)

tests/Listener.php
GrahamCampbell: This line can be removed now. ;)

timacdonald: done! thanks Graham

timacdonald:
@mfn no worries. if you do think of any specific scenarios where this could conflict (with or without another listener) please do mention it. I'd be happy to run some tests and see if we need to make adjustments if any are needed.

timacdonald:
I've just pushed a new commit to this. You can now specify arbitrary commands to run before the test suite, rather than locking it down to config caching only.

It makes it more generic so devs can define the commands to run before the testsuite runs for the first time.

```xml
<listeners>
    <listener class="Illuminate\Foundation\Testing\Listener">
        <arguments>
            <!-- Define artisan commands to execute before the test suite runs -->
            <string>config:cache</string>
            <string>route:cache</string>
            <string>event:cache</string>
            <string>some-3rd-party-package:cache-stuff</string>
        </arguments>
    </listener>
</listeners>
```

m1guelpf:
Isn't Listener a really gwneral name? I'd rather have a CacheForTestsListener or another name that describes what this is doing better

timacdonald:
Agreed. I think PreRunListener or SetupListener or something along those lines would be better. I’ve been more focused on the implementation than those details up until now so if anyone has any other suggestions, please do share.

I also figured Taylor, if he wants this, would know what he’d want to call it as well.

timacdonald:
...although if this is *the* Laravel listener, I dont think it would be the worst thing to call it Listener. Just like the TestCase is called TestCase instead of BootsFrameworkTestCase.

But I don’t have strong opinions on the naming either way TBH

timacdonald:
Someone just suggested this could be done in a bootstrap file...and I think I was too deep in it to realise that is probably the better way to go compared to a listener.

Just ship a `tests/bootstrap.php` or `bootstrap/tests.php` file, reference it in the `phpunit.xml` bootstrap tag and then you can provide classic Laravel style documentation as guidance as well.  Again it is all in "user land" so the dev is in full control if they want to take advantage of it.

pretending these comments were not terrible...it could be something like this...

```php
<?php

require_once __DIR__.'/../vendor/autoload.php';

/*
| Bootstrap the testing environment
|
| You have the option to specify console commands that will execute before your
| test suite is run. Caching config, routes, & events may improve performance
| and bring your testing environment closer to production.
|
*/

// $commands = [
//     'config:cache',
//     'route:cache',
//     'event:cache',
// ];

// $app = require __DIR__.'/../bootstrap/app.php';

// $console = tap($app->make(\Illuminate\Contracts\Console\Kernel::class))->bootstrap();

// foreach ($commands as $command) {
//    $console->call($command);
// }
```

taylorotwell:
Yeah I think that makes decent sense probably.

timacdonald:
I've just pushed changes to implement to bootstrap file instead of the Listener.

I've commented out the `route:cache` command as the routes that come in the skeleton are not cacheable. Commenting it out means that the 2 tests in the skeleton still pass out of the box.

driesvints:
My two cents: I *think* the file would be better lived under `tests/bootstrap.php` since that's a common location in most PHP apps I've seen so far. 

I also believe the current `bootstrap/app.php` file would be better placed as `./bootstrap.php` and the cache directory as a top-level directory but that's a different topic :)

timacdonald:
We've actually found in our initial feature test benchmarks that caching the routes can actually slow down the test suite. Looks like it is due the the performance of `unserialize`. This will vary depending on the number of routes you have I would assume. Gonna do some more digging into this throughout the week.

Happy to close this and re-open once I have a better understanding of the impact - or a possible solution to the performance hit of caching the routes during a test run.

m1guelpf:
Now that I think about it, this would also break tests which register routes on runtime (I think that's a common approach for testing middleware

timacdonald:
@m1guelpf runtime defined routes continue to work if the routes are cached, e.g. the following test passes with routes cached before the test run:

```php
public function test_runtime_routing()
{
    Route::get('my-test-route', function () {
        return 'expected';
    });

    $response = $this->get('my-test-route');

    $this->assertSame('expected', $response->content());
}
```

driesvints:
@timacdonald can you move the bootstrap file to `tests/bootstrap.php`?

garygreen:
> We've actually found in our initial feature test benchmarks that caching the routes can actually slow down the test suite. Looks like it is due the the performance of unserialize. Our initial benchmarks were against a test app with 1,000 routes.

That's also been my experience with route caching - there's a huge memory overhead when route caching is enabled and yet no discernible improvement in speed. It's wort noting the route caching feature was added 4 years ago so maybe improvements in PHP have changed things, or possibly the benchmarking when route caching was added was flawed at the time. Your not the only one to have noticed this, I personally think route caching should be deprecated because it's really a miss-sold feature. If there's any strong evidence it's actually effective then maybe it can be "undeprecated" 😁

timacdonald:
@driesvints that's done for ya mate. Sorry I didn't just on that when you requested it earlier.

@garygreen that is interesting for sure. I guess I can't really comment about what it is like in production, as I'm not taking OPCache into account as this is focused on the local testing aspect. Perhaps we can put together some arguments for and against it in an issue over on `laravel/ideas` and get some stats / benchmarks to provide to Taylor so he can make an informed decision about it. Perhaps we deploy this [benchmark suite](https://github.com/timacdonald/laravel-container-speed-test) to a production environment and measure the impact of routes with and without caching - the `tests/build.php` allows you to determine how many unique api / web routes will be generated.

garygreen:
>  I guess I can't really comment about what it is like in production, as I'm not taking OPCache into account as this is focused on the local testing aspect

I've been testing with OpCache enabled on local and it's exactly the same - route caching has a negative impact on performance and higher memory footprint for requests. It definitely would be good to have more tests run from various developers/machines to build a strong case that route caching should be deprecated.

timacdonald:
Although this PR isn't only about improved performance, I just wanted to drop in some reports I've had from people giving this a spin.

I've heard from about 8 people that are seeing anywhere from 0% to 25% speed increases in their test suites. I'm personally seeing around 5% of a speed boost on one of my projects and 15% on another. I'm guessing that most people will only see a small percent boost - maybe 5% at most.

I think any real world speed improvements are going to vary wildly based on the kind of tests that make up your suite.

I'm not a Docker person, but I believe Docker (maybe just the Mac version?) doesn't have great file access performance, so I *think* the higher end speed boosts are potentially coming from Docker users. Yet to confirm that though.

Not saying this is a good sample or valid data - but wanted to drop it in here to help any discussion.

taylorotwell:
So what's the consensus on this one? Do we still think it's worth merging?

hanspagel:
Down from 2.25 to 2.00 for 400 tests in docker. 👌

garygreen:
Just providing my results. This is tested on Windows. Haven't got a huge number of test classes, quite a few assertions though. This PR doesn't really make any difference for me.

**without pr**
image

**with pr**
image

garygreen:
I suspect results may vary wildly with this depending on how many configs and events you got, how much data is in them etc. I've personally gone thru and commented out all the settings/options don't need and try to keep configs to a minimum. Also I don't use any events in my application.

@timacdonald care to share how many configs you have, what's in them etc? What's the file output size of your cached config? Mine is:

Cached config size: 16KB
Cached events size: 1KB

timacdonald:
@taylorotwell I think it would be good to get some more people reporting their results. I'm gonna try and get a few more people to come in and let us know if this is having any real world impact.

@hanspagel that's a nice little boost 🎉

@garygreen I'm surprised to see this PR make the test suite run slower. Although I'm sure you are all over this, would you mind just confirming you did the following steps. The key things are changing the `<server` to `<env` vars and making sure you suite runs without any existing config cache files.

I'm dropping these steps in here so I can try and get some more results reported.

# Steps to test this PR on Laravel 5.8

Record how long your suite **currently runs**...

1. Clear out any existing files in `bootstrap/cache` directory.
2. Run the test suite 3 times and average the result.

Record how long your suite runs **with this PR**...

1. Add the [bootstrap file](https://github.com/laravel/laravel/pull/5050/files#diff-10eafd8bf5fbdde867b86fe38d30b9bb) to `tests/boostrap.php` **as is**, i.e. don't cache the routes.
2. Add [these changes](https://github.com/laravel/laravel/pull/5050/files#diff-ea896f6d36682ab3ef2d6e3f3de5abce) to your existing `phpunit.xml` file.
3. Change the added `<server` variables to `<env` variables.
4. Clear out any existing files in `bootstrap/cache` directory.
5. Run the test suite 3 times and average the result.

jasonlbeggs:
I tested it in two different apps. One is pretty large and has a ton of migrations. One is pretty small and only has a few migrations. Here are my statistics from following the steps in each app:

## App 1
(Tests: 792, Assertions: 2010, Migrations: 256)

Before: 8.43 minutes
After Average: 8.31 minutes

## App 2
(Tests: 102 tests, Assertions: 259, Migrations: 11)

Before: 5.01 seconds
After Average: 2.77 seconds

stefanzweifel:
Ran your PR in a larger app. Like Jasons's, it has a ton of migrations (167).
It's also just a subsection of the entire test suite. (We usually run the entire suite only with `paratest`)

Here are the stats:

*Tests: 220, Assertions: 922.*

## Control (clear `bootstrap/cache` on each run)

**==> avg. 1.77 Minutes**

1. run: Time: 1.81 minutes, Memory: 102.50 MB
2. run: Time: 1.91 minutes, Memory: 102.50 MB
3. run: Time: 1.59 minutes, Memory: 102.50 MB


## Tim's PR

**==> avg. 1.5933 Minutes**

1. run: Time: 1.62 minutes, Memory: 106.50 MB
2. run: Time: 1.55 minutes, Memory: 106.50 MB
3. run: Time: 1.61 minutes, Memory: 106.50 MB



I've also ran the suite multiple times with Laravel's default setup and didn't cleared `bootstrap/cache`. Even then, Tim's PR brought some improvements.

**==> avg. 1.62 Minutes**

1. run: Time: 1.64 minutes, Memory: 102.50 MB
2. run: Time: 1.73 minutes, Memory: 102.50 MB
3. run: Time: 1.55 minutes, Memory: 102.50 MB
4. run: Time: 1.56 minutes, Memory: 102.50 MB

edvordo:
Wanted to chip in

*383 tests, 1586 assertions, 131 migrations*

* **pre-PR runs**
Time: 3.03 minutes, Memory: 62.50 MB
Time: 3.29 minutes, Memory: 62.50 MB
Time: 3.04 minutes, Memory: 66.50 MB
**Average**: 3.12 minutes
* **post-PR runs**
Time: 2.92 minutes, Memory: 68.50 MB
Time: 3.01 minutes, Memory: 66.50 MB
Time: 2.89 minutes, Memory: 66.50 MB
**Average**: 2.94 minutes
---
<details>
<summary>And (just for fun) without running migrations</summary>

 * 383 tests, 1586 assertions, 131 migrations
 * **pre-PR runs**
 Time: 43.13 seconds, Memory: 58.50 MB
 Time: 41.91 seconds, Memory: 58.50 MB
 Time: 42.07 second, Memory: 58.50 MB
 **Average**: 42.37 seconds
 * **post-PR runs**
 Time: 43.38 seconds, Memory: 64.50 MB
 Time: 41.51 seconds, Memory: 64.50 MB
 Time: 41.63 seconds, Memory: 64.50 MB
 **Average**: 42.17 seconds
</details>

garygreen:
> Although I'm sure you are all over this, would you mind just confirming you did the following steps. The key things are changing the <server to <env vars and making sure you suite runs without any existing config cache files.

Yes I have changed to `env` vars in phpunitxml and ensured it's properly created the cached files in `bootstrap/cache` then run the tests. I don't think anything else is needed?

> And (just for fun) without running migrations [...PR makes no difference to speed]

This mirrors my tests results as well - I don't run any migrations during my tests, though do hit the database that's already been migrated. So maybe speed is related to migrations?

I'm not overly familiar with how Laravel handles migrating the database before tests, it's often felt slow whenever I've used the built in traits so I tend to just do manual tearup/teardown of DB for speed. The only thing I use is `DatabaseTransactions`.

narwy:
I don't agree with the `/tests/bootstrap.php`, I feel like the current structure / naming convention of all the files would be out of place.  `/bootstrap/app.php` & `/bootstrap/tests.php` It's a folder specific for bootstrap files. If we start to put them everywhere It can become messy? Good idea though and the results look promising.

timacdonald:
@garygreen Nah, you got it all mate. I figured you knew what was going on, but just wanted to double check to make sure we were on the same page. I guess this is the nature of a micro-optimisation like this - it'll benefit some test suites and not others. Bummer it didn't make any improvement on yours.

I also use `DatabaseTransactions` on my test suites and did see some nice improvements. I also work with an already migrated before the test run DB.

I think the sales pitch for this PR is:

1. Some test suites will see a nice (and sometimes significant) improvement, others will not depending on the makeup of your suite.
2. It beings your test suite closer to production, and encourages devs to not use the `env` helper, as the config is cached and the helper will already return null.
3. It gives people less of a reason to say that "Laravel" is slowing down the tests.
4. It gives developers a place from the get go to bootstrap their test suite - if needed.

But...it isn't a game changer and never was going to be. Not including it in the framework isn't gonna be an issue - but I think it is a "nice to have" myself.

elvispt:
Here are my results (small Laravel 5.8.28 app) for my test suite which includes migrations:
**Pre-PR**:
- Time: 21.41 seconds, Memory: 104.25 MB
- Time: 21.59 seconds, Memory: 106.25 MB
- Time: 21.82 seconds, Memory: 104.25 MB
**Average: 21.60**

**Post-PR**:
- Time: 21.99 seconds, Memory: 96.25 MB
- Time: 21.3 seconds, Memory: 100.25 MB
- Time: 21.75 seconds, Memory: 98.25 MB
**Average: 21.68**

Did notice an improvement on memory, but on times it's basically the same.

adrianb93:
**App 1:**

This is a new app with only unit tests (we've not written any routes yet.)

(Tests: 186, Assertions: 555)
Before Average: 45.9 seconds (46.86; 45.92; 44.92;)
After Average: 44.92 seconds (47.94; 42.93; 43.89;)

**App 2:**

This is a super slow test suite and needs test improvements. I ran this suite 3 times, once to warm up, once for non-cache; once for cache. I didn't do a three run average for this because I don't have hours to spend on this.

(Tests: 346 tests, Assertions: 882)

Before: 23.34 minutes
After: 22.52 minutes

**Thoughts:**

* Config caching doesn't make the test suite any slower.
* Config caching may improve your test speeds. *For a portion of app 2, I saw huge performance improvements on the fast tests of the suite*
* Regardless of performance, being able to test that your app works under config cache is definitely a good thing.
* For those who write tests, it may guide people away from using `env()` calls outside of config files. From my experience, most Laravel devs find `env()` more convinient because they can skip the step of adding something to their config files. They end up treating `.env` as a config file.
* I love the `tests/bootstrap.php` file. I've already added my own code to it for some before and after test suite setups and cleanups. It's a nice home to do your own userland test suite improvements.

By the way, if anybody wants to run something after the suite finishes, add this to the end of the booststrap file:

```php
// Will run once the entire test suite has finished
register_shutdown_function(function () use ($console) {
    // Figured this is a helpful example for those on a Laravel version below 5.8
    $console->call('cache:clear');
});
```

driesvints:
Judging from the average of the results I think it's best not to include this in the framework as it's not applicable to everyone and the optimizations aren't too big. Maybe it's best to detail this in a blog post and note when this would be applicable.

specialtactics:
This is a good approach - we use a bootstrap file in that location for all of our projects, so it will be useful to have it there as a base for people to extend for their own projects.

adrianb93:
I really want to see this in the skeleton. The `tests/bootstrap.php` file helped me recognise that I had a place to do something before and after my tests run.

I was having a discussion with @driesvints at Laracon about this. I mentioned that because of this file, I had somewhere to make huge improvements in a slow test suite.

**Case 1:**

* A project I work on took 23 minutes to run the tests.
* I identified the bottleneck and figured I could do some work before the first test, and after the last test in the test bootstrap file.
* The project then took 55 seconds to run the tests.
* I added config caching and got a 5-second improvement, the project took 50 seconds to run the tests.

Massive win! 🎉

**Case 2:**

In another project, instead of thinking that 3 minutes was an acceptable time for a Laravel test suite, I used the test bootstrap file to get that suite down to 35 seconds. 🍻

_(Both tests times had the config cache improvement)_

**My point:**

I think this is a highly useful file to have in the skeleton. I got a lot of value from it.

* I may have never gotten those speed improvements if it the tool to do it wasn't right in front of me. The solution seemed obvious thanks to it.
* The config caching in tests proves that your app will work in production if your app is well tested.
* You may also experience a slight speed improvement with config caching which is a bonus.

taylorotwell:
@adrianb93 what did you actually do in the file to achieve those results?

adrianb93:
> @adrianb93 what did you actually do in the file to achieve those results?

Nothing crazy. They're both apps that work with many databases. It was just moving most of the setup work to the start of the suite.

* **23-minute suite:** It needed to copy a couple of database schemas and do some seeding for some critical third-party connections we need to integrate with that are subject to schema change. This was a slow setup for each test.
  * No, this integration has no API or staging environment.
  * Yes, we still develop our app as if there is in case we can switch to another the third-party or if there are improvements.
  * It's less than ideal integration work, but it is what it is.
* **3-minute suite:** This is a multi-tenant database app. The bootstrap provisions a default testing tenant database before running the suite.
  * If you fake the tenant db handler, the default connection is used.
  * Provisioning time seemed reasonable in this app but there was time to be saved here.
  * It was also valuable to use `register_shutdown_function` in the bootstrap to delete any databases starting with `testing_tenant_*` in case a test didn't call deprovision on a tenant. This is cleanup work that was once done manually.

In both cases, each test starts transactions in each test setup and rolls back in the before application destroy.

These aren't the first apps I've worked on that have built up this sort of unavoidable DB cruft. The bootstrap file helped make things better for these apps.

timacdonald:
Really appreciate so many people taking the time, here and on other forums, to give me feedback on this PR ❤️ Hope it comes in handy for you and others.

patrickbrouwers:
@timacdonald awesome job! Working great for me when running from the command line (+-2s faster), however having some issues getting it work when running individual test via PHPStorm interface. Anyone who has managed to get that working?

cbaconnier:
> By the way, if anybody wants to run something after the suite finishes, add this to the end of the booststrap file:
> 
> ```
> // Will run once the entire test suite has finished
> register_shutdown_function(function () use ($console) {
>     // Figured this is a helpful example for those on a Laravel version below 5.8
>     $console->call('config:clear');
> });
> ```

Thanks for this @adrianb93 I'm on Laravel 5.8 and I was experimenting 419 errors after the tests.

> however having some issues getting it work when running individual test via PHPStorm interface. Anyone who has managed to get that working?

@patrickbrouwers I'm on Ubuntu with PHPStorm 192.5728.108 and run them through docker. For me, they still work individually with no extra configuration.

bastien-phi:
@patrickbrouwers Same problem here, I can't run individual tests. I got this error : 
```
In PackageManifest.php line 168:
                                                               
  The bootstrap/cache directory must be present and writable.  
```
Same for you ? I'm investigating for a solution...

patrickbrouwers:
@bastien-phi yes that's the error I'm getting too. I got it to work when I specify the absolute path in the env settings for the cache files.

bastien-phi:
@patrickbrouwers  I found the exact same solution right now !
Not very convenient when sharing the phpunit.xml file with teammates IMO ...

patrickbrouwers:
@bastien-phi I personally always gitignore the phpunit.xml and ship a phpunit.xml.dist

bastien-phi:
@patrickbrouwers I will just update `tests/bootstrap.php` to update environment variables with absolute path before calling the framework : 
```php
$envVars = ['APP_CONFIG_CACHE', 'APP_SERVICES_CACHE', 'APP_PACKAGES_CACHE', 'APP_ROUTES_CACHE', 'APP_EVENTS_CACHE'];
foreach ($envVars as $envVar) {
    $_ENV[$envVar] = __DIR__ . '/../' . $_ENV[$envVar];
}
```

@timacdonald Any other idea as `APP_CONFIG_CACHE` (and others) are defined by default  as absolute path in [`Illuminate/Foundation/Application`](https://github.com/laravel/framework/blob/ed3edd0649349ea533b0c2c45c63385cffa24e19/src/Illuminate/Foundation/Application.php#L926) ?

timacdonald:
@cbaconnier as per my comment here: https://github.com/laravel/laravel/pull/5050#issuecomment-511069595 have you set the env vars correctly in the `phpunit.xml`? I found most people were skipping this step and thus having the issue you are where the cached values are overriding the normal config files.

The env values need to be set differently in 5.8. Instead of `<server` you need to use `<env` e.g.

```diff
- <server name="APP_CONFIG_CACHE" value="bootstrap/cache/config.phpunit.php"/>
+ <env name="APP_CONFIG_CACHE" value="bootstrap/cache/config.phpunit.php"/>
```

cbaconnier:
> @cbaconnier as per my comment here: [#5050 (comment)](https://github.com/laravel/laravel/pull/5050#issuecomment-511069595) have you set the env vars correctly in the `phpunit.xml`? I found most people were skipping this step and thus having the issue you are where the cached values are overriding the normal config files.
> 
> The env values need to be set differently in 5.8. Instead of `<server` you need to use `<env` e.g.
> 
> ```diff
> - <server name="APP_CONFIG_CACHE" value="bootstrap/cache/config.phpunit.php"/>
> + <env name="APP_CONFIG_CACHE" value="bootstrap/cache/config.phpunit.php"/>
> ```

Oh... :sweat_smile:  Thanks! That was it

timacdonald:
@cbaconnier no worries, that has tripped a few people up.

@bastien-phi not sure. Can you put your setup in here so we can try and work out why it isn't working for you? Are you calling it from inside PHPStorm as well? If so, could you try on the command line as well and see if it is any different? Would be good to work what part of your stack is causing issue for ya.

bastien-phi:
@timacdonald I created a repo from a fresh install of laravel, with the phpunit bootstrap file : [laravel-phpunit-bootstrap](https://github.com/bastien-phi/laravel-phpunit-bootstrap)

I updated the bootstrap file with a "patch" in order to make it work properly.

Before the patch, `./vendor/bin/phpunit --config phpunit.xml` works properly but calling phpunit from PHPStorm in a single file does not work.
PHPStorm calls the command `/usr/bin/php /path/to/laravel-phpunit-bootstrap/vendor/phpunit/phpunit/phpunit --configuration /path/to/laravel-phpunit-bootstrap/phpunit.xml Tests\Unit\ExampleTest /path/to/laravel-phpunit-bootstrap/tests/Unit/ExampleTest.php --teamcity` and I get 
```
In PackageManifest.php line 168:
                                                               
  The bootstrap/cache directory must be present and writable.  
```
Calling this command from the project root, the tests goes well.

The difference between the two calls is that from inside PHPStorm, current working directory is the directory when the test file belongs (in that case `tests/Unit`), and from the command line current directory is the project root.

If calling the command from another directory (let's say `tests`), the command line fails too : 
```
$ cd tests && /usr/bin/php /path/to/laravel-phpunit-bootstrap/vendor/phpunit/phpunit/phpunit --configuration /path/to/laravel-phpunit-bootstrap/phpunit.xml Tests\Unit\ExampleTest /path/to/laravel-phpunit-bootstrap/tests/Unit/ExampleTest.php --teamcity

In PackageManifest.php line 168:
                                                               
  The bootstrap/cache directory must be present and writable. 
```

As the environment variables are relative paths, the application tries to find the directory `bootstrap/cache` relatively to current working directory which cause the fails.

The patch updates relative paths to absolute path and everything goes well. Unfortunately, this is far from being a clean fix...

timacdonald:
@bastien-phi I can confirm it works without the patch in `vim-test` and `sublime-phpunit`, but unfortunately I do not have a copy of PHPStorm to help debug this directly. But a google lead me to this: https://www.jetbrains.com/help/phpstorm/run-debug-configuration-phpunit.html#commandLine

If I am understanding that correctly, they are saying the CWD is always set to the project root.

> By default, the field is empty and the working directory is the root of the project.

Can you check that option out (and possibly whatever else is in there), confirm you don't have a value in there, and maybe have a play and see if there is anything in there you need to change to get it working?

bastien-phi:
@timacdonald This option is kept empty and the CWD is not set to the project root but to the directory where the test class belongs : 
```
$ /usr/bin/php /path/to/laravel-phpunit-bootstrap/vendor/phpunit/phpunit/phpunit --configuration /path/to/laravel-phpunit-bootstrap/phpunit.xml Tests\Unit\ExampleTest /path/to/laravel-phpunit-bootstrap/tests/Unit/ExampleTest.php --teamcity
current working directory : /path/to/laravel-phpunit-bootstrap/tests/Unit

In PackageManifest.php line 168:
                                                               
  The bootstrap/cache directory must be present and writable. 
```

It is possible to explicitly set CWD as project root for phpunit in PHPStorm with Run Configuration Templates. Be sure to remove all existing configurations first.
Capture d’écran de 2019-08-01 09-00-05
It works this way.

But IMO, there is still a problem because CWD **has to** be project root to run the tests. Before this feature, it was possible to launch tests from another location.

The problem does not comes from the bootstrap file itself but from the fact that the files are defined with relative paths in the `phpunit.xml`. Currently I find only 3 solutions : 
1) remove the path definitions in `phpunit.xml`
2) update the path definitions in `phpunit.xml` with absolute paths
3) explicitlty update CWD in `tests/bootstrap.php` to set it to project root.

1 is not a good idea at all !
With 2, there is a need to ship a phpunit.xml.example, gitignore phpunit.xml and manually update the paths.
My favorite solution is 3 but I don't know if there could be side effects.
```php
<?php

use Illuminate\Contracts\Console\Kernel;

chdir(__DIR__.'/../');

require_once __DIR__.'/../vendor/autoload.php';

/*
| Bootstrap the testing environment
|
| You have the option to specify console commands that will execute before your
| test suite is run. Caching config, routes, & events may improve performance
| and bring your testing environment closer to production.
|
*/

$commands = [
    'config:cache',
    'event:cache',
    // 'route:cache',
];

$app = require __DIR__.'/../bootstrap/app.php';

$console = tap($app->make(Kernel::class))->bootstrap();

foreach ($commands as $command) {
    $console->call($command);
}
```

stevelacey:
Laravel 5.8 with 171 tests before/after 5-10% improvement 👏️

Before:

```
Time: 53.27 seconds, Memory: 68.25 MB
Time: 48.13 seconds, Memory: 68.25 MB
Time: 51.73 seconds, Memory: 68.25 MB
Time: 52.96 seconds, Memory: 68.25 MB
Time: 48.15 seconds, Memory: 68.25 MB
```

After:

```
Time: 42.16 seconds, Memory: 70.25 MB
Time: 42.26 seconds, Memory: 70.25 MB
Time: 47.57 seconds, Memory: 70.25 MB
Time: 43.92 seconds, Memory: 70.25 MB
Time: 44.53 seconds, Memory: 70.25 MB
```

```
OK (171 tests, 1135 assertions)
```

🙂️👌️

timacdonald:
nice @stevelacey! Thanks for sharing. Glad to see a little perf bonus there for ya

foremtehan:
@timacdonald I replaced your code and it improved ~10% of test suit but i dont know why i'm getting ```General error: 1 no such table: users``` on the browser , and that error gone when i clear cache folder any idea?

stevelacey:
@foremtehan I just had this same problem, I suspect it’s because 5.8 doesn’t have the separate caches per env logic thats coming in 6.0, and thus Laravel sees cache and ignores the .env? And thus your DB_NAME is wrong, and thus no such table.

@timacdonald is there something else we should patch in?

timacdonald:
@foremtehan @stevelacey see https://github.com/laravel/laravel/pull/5050#issuecomment-516787726

You will want to run `php artisan config:clear` to make you get the test values out of your local config cache.

foremtehan:
@timacdonald I'm getting this error after i've update laravel to 6.0

```
In PackageManifest.php line 168:
                                                               
  The bootstrap/cache directory must be present and writable.  
```
Capture

```cache:clear``` wouldn't change anything

**Comment by driesvints** [comment]
@timacdonald this PR is causing some issues:

https://github.com/laravel/framework/issues/29862
https://github.com/laravel/laravel/pull/5091

We'll need to figure out solutions for these fast or else it might be best to revert this PR.

**Comment by bastien-phi** [comment]
https://github.com/laravel/laravel/pull/5071 does the job

**Comment by timacdonald** [comment]
@driesvints digging into this now. Will respond in the relevant threads to keep discussions together.

**Comment by foremtehan** [comment]
I had exact problem, I just comment these out in ```phpunit.xml``` and everything back to work correctly:

```
<env name="APP_CONFIG_CACHE" value="bootstrap/cache/config.phpunit.php"/>
<server name="APP_SERVICES_CACHE" value="bootstrap/cache/services.php"/>
<server name="APP_PACKAGES_CACHE" value="bootstrap/cache/packages.php"/>
<server name="APP_ROUTES_CACHE" value="bootstrap/cache/routes.phpunit.php"/>
<server name="APP_EVENTS_CACHE" value="bootstrap/cache/events.phpunit.php"/>
```

timacdonald:
@foremtehan that’ll mean that your testing env is being cached in your normal cache files, so you probably hit some weird bugs when using your app locally. Once https://github.com/laravel/framework/pull/29890 is tagged it should be fixed and you’ll be right to restore everything

PR #5054, opened 2019-07-07
[5.9] Encryption manager configuration

mmollick:
Compliments PR in framework, see laravel/framework#29100.

### DIFF
--- a/config/encryption.php
+++ b/config/encryption.php
@@ -1,0 +1,38 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Default Encrypter
+    |--------------------------------------------------------------------------
+    |
+    | The default encrypter will be used to encrypt cookies and sessions. You
+    | can use the default encrypter for all of your encryption need or use
+    | additional encrypters defined below.
+    |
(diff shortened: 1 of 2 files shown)


config/encryption.php
```diff
@@ -0,0 +1,38 @@
...
+
+    'encrypters' => [
+
+        'default' => [
+            'key' => env('APP_KEY'),
+            'cipher' => 'AES-256-CBC',
+        ],
+
+    ]
+];
```
antonkomarev: Each file should end with blank line

config/encryption.php
```diff
@@ -0,0 +1,38 @@
...
+    | Default Encrypter
+    |--------------------------------------------------------------------------
+    |
+    | The default encrypter will be used to encrypt cookies and sessions. You
+    | can use the default encrypter for all of your encryption need or use
+    | additional encrypters defined below.
+    |
+    */
+
+    'default' => 'default',
```
xuanquynh: We may set an environment variable for this.

```php
return [
    'default' => env('ENCRYPT_DRIVER', 'default'),
];
```

PR #5055, opened 2019-07-09
[5.9] Move TrustProxies to highest priority - fixes maintenance mode ip whitelist if behind proxy e.g. Cloudflare

garygreen:
Noticed this when trying to do `php artisan down --allow=<ip>` if your behind Cloudflare / some other proxy it won't allow you to access site - this is because `TrustProxies` only kicks in *after* maintenance mode middleware.

It also makes sense to me to whitelist a request straight away as the first thing in the app before firing any subsequent middlewares.

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -14,11 +14,11 @@ class Kernel extends HttpKernel
      * @var array
      */
     protected $middleware = [
+        \App\Http\Middleware\TrustProxies::class,
         \App\Http\Middleware\CheckForMaintenanceMode::class,
         \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
         \App\Http\Middleware\TrimStrings::class,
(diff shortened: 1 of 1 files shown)


driesvints:
I could swear I've seen this going back and forth but not sure anymore. Can't find old prs on it.

antonkomarev:
@driesvints it was introduced in 29th of March 2017: https://github.com/laravel/laravel/commit/e23a1d284f134bfce258cf736ea8667a407ba50c#diff-b3af398d5ac2670049a1f17f4683c03e

kamui545:
@driesvints Hm, I think that was the Throttle middleware? But also not sure anymore

peppeocchi:
+1
I found the same when behind CloudFront, had to move the TrustProxies middleware on top (one of the issues was the redirection from http to https when the app was behind Cloudfront and a load balancer)

GrahamCampbell:
This was merged to master, which is 5.8 on laravel/framework.

PR #5057, opened 2019-07-10
[5.9] Add ThrottleRequests to the priority array

iamgergo:
This is a resubmit of #5056 and PR-ed on the develop branch as is was requested.

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -73,9 +73,9 @@ class Kernel extends HttpKernel
         \Illuminate\Session\Middleware\StartSession::class,
         \Illuminate\View\Middleware\ShareErrorsFromSession::class,
         \App\Http\Middleware\Authenticate::class,
+        \Illuminate\Routing\Middleware\ThrottleRequests::class,
         \Illuminate\Session\Middleware\AuthenticateSession::class,
         \Illuminate\Routing\Middleware\SubstituteBindings::class,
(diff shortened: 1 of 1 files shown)


patrickbrouwers:
Does this always run the bindings before the ThrottleRequest after this change? 

One of the points of throttling the request is that it won't hit the database. If it will still query the model,  the "too many requests" still have an impact on your database.

taylorotwell:
I guess I'm a little confused. What does route model binding have to do with `$request->user()`, which is an authentication feature and has nothign to do with the "SubstituteBindings" middleware?

iamgergo:
@taylorotwell Well, as it turned out nothing at all. I realized that the `ThrottleRequests` needs to be forced to run after the `Authenticate` middleware and not the `SubstituteBindings`.

The first time I changed the order only in the API middleware group. But since in the `$middlewarePriority` array, `SubstituteBindings` comes after `Authenticate`, the `ThrottleRequests` ran after the authentication. That's why I thought in the first time, that the `SubstituteBindings` solves this, but the `Authenticate` is the key – like you said.

I moved up the throttle middleware under the authenticate middleware in a new commit.

Also, I realized, it's possible to assign different rate limiting for guests and authenticated users:
```
throttle:10|limit,1
```

I think this is a nice feature, but this is undocumented. Should I open a PR for that in the docs?

PR #5059, opened 2019-07-14
Update .env.example

mohamedadel2018:
update to localhost in databasr


driesvints:
This shouldn't be needed. On most systems 127.0.0.1 is mapped to localhost.

PR #5061, opened 2019-07-16
Fix cache, redis and session prefixes for non-English `APP_NAME`

chanjsq:
When setting non-English `APP_NAME`. The computed prefixes for cache, Redis and session would be something like `_cache`, not the `foo_cache` format. This PR try to fix this.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,4 @@
+APP_CODE=Laravel
 APP_NAME=Laravel
 APP_ENV=local
 APP_KEY=
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,5 @@
-APP_CODE=Laravel
 APP_NAME=Laravel
+APP_CODE=Laravel
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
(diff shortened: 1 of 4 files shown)


chanjsq:
No, it didn't. `Str::slug('淘宝');` results an empty string.

GuidoHendriks:
Why wouldn't you just change the `CACHE_PREFIX` if the default doesn't work? That's as much work for you on a project as changing `APP_CODE`, but no extra work for people who can use the default.

narwy:
Maybe `APP_PREFIX` ?

driesvints:
Feel free to add this to your own codebase.

gpressutto5:
Why don't you just change `CACHE_PREFIX` in your `.env`?

chanjsq:
@gpressutto5 There's not only **cache** prefix, but also **Redis** and **session** prefixes. The `CACHE_PREFIX` cannot affect **Redis** and **session**.

chanjsq:
@gpressutto5 Do you have a better way of modify all of these? Thanks.

GuidoHendriks:
You know you own the config files of your project, right? Just change it to whatever fits your needs.

chanjsq:
@GuidoHendriks Thank you for your tip.

PR #5063, opened 2019-07-17
[5.8] Update number of Laracasts videos

ryandon83:
Commit updates the number of available Laracasts videos. As of this writing there are 1506 videos.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -25,7 +25,7 @@ Laravel is accessible, powerful, and provides tools required for large, robust a
 
 Laravel has the most extensive and thorough [documentation](https://laravel.com/docs) and video tutorial library of all modern web application frameworks, making it a breeze to get started with the framework.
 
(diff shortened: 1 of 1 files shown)


antonkomarev:
Maybe it's better to rephrase this sentence to not update it all the time?

ludo237:
Over 1400 is still correct, old PR were closed because the sentence is still correct even with 1506 videos

GuidoHendriks:
Maybe only update at values like 1500/2000.

ryandon83:
Perhaps update now because 1500 is a nice rounded number? Once 2000 is reached, the sentence can be reworded to "Laracasts contains thousands of video tutorials on a range of topics including Laravel, modern PHP, unit testing, and JavaScript." This wording would probably be just as impressive as an actual number. Otherwise, 500 level increments sounds good to me.

PR #5064, opened 2019-07-24
Update ignoring vscode dependency file

nafistiham:
ignoring .vscode file while committing to source


ludo237:
Old proposal, always rejected. `.vscode` is a global git ignore property

PR #5065, opened 2019-07-24
added Security disclosure information

msurguy:
Please also add details on which versions of Laravel are currently supported...

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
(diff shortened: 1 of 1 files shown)


antonkomarev:
Yes, but it's missing here: https://github.com/laravel/laravel/security/policy

xuanquynh:
It is not necessary. Developers will look at the README.md. Or Remove this section of readme.md after adding SECURITY.md.

driesvints:
Heya, since this is the skeleton for anyone's laravel app, they'll get this file in their apps and that's why I don't think this is the right place. We already have on in the framework repo and like @xuanquynh said, it's also in the readme. Thanks.

msurguy:
@driesvints thanks for the info! 👍

PR #5075, opened 2019-08-13
[5.8] Remove Stripe config settings

driesvints:
These now ship with a dedicated config file for Cashier.

### DIFF
--- a/config/services.php
+++ b/config/services.php
@@ -8,7 +8,7 @@ return [
     |--------------------------------------------------------------------------
     |
     | This file is for storing the credentials for third party services such
-    | as Stripe, Mailgun, SparkPost and others. This file provides a sane
+    | as Mailgun, SparkPost and others. This file provides a sane
     | default location for this type of information, allowing packages
     | to have a conventional place to find your various credentials.
     |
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
// cc @brunogaspar

brunogaspar:
hmm interesting.. 

I kinda understand the reasoning, which is because Cashier now ships with a config file, but shouldn't this go in `develop` which is Laravel 6 instead?

Removing it now is kinda a breaking change, even though this is not "installable" per se, some people, like me, sync the branch every once in a while (this will at least for me, cause just a merge conflict).

But either way i'm "ok", i just need to create a dedicated config file on the Stripe Laravel package and keep it backward compatible for the 5.8 release (hacky hacky) bummer heh

Thanks @GrahamCampbell

GrahamCampbell:
@driesvints for context: https://github.com/cartalyst/stripe-laravel.

CaddyDz:
What about using the Stripe PHP SDK directly for one time payments? or does everything have to be subscription based now?

GrahamCampbell:
> What about using the Stripe PHP SDK directly for one time payments? or does everything have to be subscription based now?

You can still do this, but you will need to make sure you use their new API if your customers are based in the EU otherwise their card will be declined.

driesvints:
@brunogaspar I know yours is a special usecase but we don't consider changes to the skeleton as breaking as they're always used for instantiating new apps. I'm sorry for that but in our eyes they don't belong to the "public api" of the framework.

brunogaspar:
It's all good @driesvints

PR #5081, opened 2019-08-19
[6.0] Set the priority of CSRF middleware

bangbangda:
When the session expires, now prompted is 419 exception(csrf), should be performed the authentication first.
Do you have any idea.


GuidoHendriks:
Why?

bangbangda:
Cross-site request forgeries are a type of malicious exploit connected unauthorized commands are performed on behalf of an authenticated user.
But. the expiration of the session does not mean that it is CSRF, the user has not been active for a long time, causing the user to log out. jump to the login page is a more reasonable.
First judge the session status is normal, then confirm whether it is CSRF, is ok.

GuidoHendriks:
That would mean exposing that you're logged in in case of CSRF. Doesn't sound like a great idea to me.

PR #5086, opened 2019-08-24
[5.8] Remove homestead from gitignore

cacing69:
accroding to this commit [use generic default db config](https://github.com/laravel/laravel/commit/6f3d68f67f3dab0e0d853719696ede8dfd9cc4e1) i think remove homestead from gitignore is goodchange


Arkanius:
You can set this in your own install @cacing69

PR #5090, opened 2019-09-03
[6.x] Allowing optional use of yml/yaml file extensions in .editorconfig

darrencraig:
The YAML docs generally recommends using .yaml as the preferred extension for YAML files - https://yaml.org/faq.html.

I've tweaked the .editorconfig to set both yaml and yml files to have the same indentation. Without this, *.yaml files default to 4 spaces.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,5 +11,5 @@ trim_trailing_whitespace = true
 [*.md]
 trim_trailing_whitespace = false
 
-[*.yml]
+[*.{yml,yaml}]
 indent_size = 2


driesvints:
Can you send this to 6.0 branch instead?

darrencraig:
> Can you send this to 6.0 branch instead?

Is that the develop branch? There doesn't appear to be a public 6.0 branch?

driesvints:
Ow, sorry. Mistook this with the framework repo 🤦‍♂

PR #5091, opened 2019-09-05
[6.x] Delete cached config file before running tests

SjorsO:
I ran into the problem that the bootstrapping of the unit tests crashed because the cached config file (`bootstrap/cache/phpunit.config.php`) contained a reference to a class that didn't exist anymore.

This problem happened to me when I removed a package that I had manually registered in my `config/app.php`. I think this can also happen when you rename classes.

**Steps to reproduce:**
1. On a fresh Laravel install, install a package that you have to manually register in your `config/app.php`
    - eg: `composer require clarification/sparkpost-laravel-driver`
    - [manually register the provider](https://github.com/clarification/sparkpost-laravel-driver#install): `Clarification\MailDrivers\Sparkpost\SparkpostServiceProvider::class`
2. Run the unit tests
    - `bootstrap/cache/phpunit.config.php` now contains the FQCN of the package
3. Remove the provider from `config/app.php` and remove the package (`composer remove clarification/sparkpost-laravel-driver`)
4. Try running the unit tests again

You should see the following error:
```
In Application.php line 671:
                                                                                  
  Class 'Clarification\MailDrivers\Sparkpost\SparkpostServiceProvider' not found  
```

---
I'm not very happy with this solution, but it works. Maybe someone has a better solution to this problem.

Also, this same bug might also exist with the other cached files.


taylorotwell:
Is that server variable always going to be set? If it's not we will get an error there.

tests/bootstrap.php
```diff
@@ -4,6 +4,10 @@
 
 require_once __DIR__.'/../vendor/autoload.php';
 
+if (file_exists($_SERVER['APP_CONFIG_CACHE'])) {
```
timacdonald:
```php
array_map('unlink', glob('bootstrap/cache/*.phpunit.php'));
```

Perhaps we could just assume the default path and if it is changed in user land you can modify as needed.

SjorsO:
I've included @timacdonald's suggestion on how to delete the files. Now we can be sure none of the cached files can give us any trouble.

I've also refactored the bootstrap file as a PHPunit extension, this has the following benefits:
- We can now reuse the `CreatesApplication` trait
- We now delete the cached files after all tests, instead of before all tests (which means they are never left on the disk)
- It gives users a convenient place to put additional [PHPunit hooks](https://phpunit.readthedocs.io/en/8.3/extending-phpunit.html#extending-the-testrunner)

timacdonald:
I think this should probably be 2 PRs so discussion can happen independently for the two different changes, i.e.

1. to remove the cached files before the test run with....

```php
array_map('unlink', glob('bootstrap/cache/*.phpunit.php'));
```

2. Refactoring to the extension. 

But while it is here - a couple of thoughts on this approach:

> We now delete the cached files after the last tests, instead of before all tests (which means they are never left on the disk, and that IDEs wont try to index them) 

This isn't really an issue. The files are .gitignored and before the introduction of this feature, running your test suite would already generate `packages.php` and `services.php`.

We will also have to check how this plays with parallel test runners, i.e. that it only runs the extension hooks once per run and not once per thread.

timacdonald:
Also need to see how this works when run from a sub folder of the project. Will have a tinker tomorrow and try and verify the relative path with the glob method works as expected

taylorotwell:
So, is this ready to merge?

SjorsO:
I think this PR can be merged. I'm pretty sure it works just like the old code, and it solves the problem described in this issue.

Like @timacdonald said, i don't know if these hooks work with parallel test runners. But since these hooks are normal phpunit features, i'm not sure if this is something we should worry about.

timacdonald:
Looks like ParaTest is hitting some issues due to its multi threaded-ness, but I have a solution that I think can just go in the docs. Will send a follow up PR to the docs about it.

p.s. @SjorsO I dig the extension over the plain bootstrap file. Nice being able to have a specific before and after hooks.

PR #5094, opened 2019-09-06
[6.x] Set argon defaults to prevent password_hash(): Memory cost is outside…

HepplerDotNet:
… of allowed memory range on PHP 7.4

With the values 
````
'argon' => [
        'memory' => 1024,
        'threads' => 2,
        'time' => 2,
    ],
```
Hash::make() produces password_hash(): Memory cost is outside of allowed memory range on PHP 7.4

### DIFF
--- a/config/hashing.php
+++ b/config/hashing.php
@@ -44,9 +44,9 @@ return [
     */
 
     'argon' => [
-        'memory' => 1024,
-        'threads' => 2,
-        'time' => 2,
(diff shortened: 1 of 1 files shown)


driesvints:
Are all of these available on PHP 7.2?

HepplerDotNet:
@driesvints according to https://www.php.net/manual/en/password.constants.php these constants are available since 7.2.0

taylorotwell:
Getting reports that it's broken on PHP 7.3.

PR #5096, opened 2019-09-08
Add .vapor directory in .gitignore

pavinthan:
Could we add `.vapor` in project `.gitignore`? because some deployment failure keep this directory to commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /public/hot
 /public/storage
 /storage/*.key
+/.vapor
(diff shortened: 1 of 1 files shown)


m1guelpf:
If this gets added, it should be added by the vapor CLI and not to the core IMO. Also, you own these files, so you could modify them locally to suit your needs.

pavinthan:
Thanks @m1guelpf and closing this because `vapor init` done this.

PR #5097, opened 2019-09-09
[6.x] According to PHP Bug 78516 Argon2 requires at least 8KB

HepplerDotNet:
https://bugs.php.net/bug.php?id=78516
Argon2 requires at least 8KB
On PHP 7.4 memory 1024 will throw:
password_hash(): Memory cost is outside of allowed memory range

### DIFF
--- a/config/hashing.php
+++ b/config/hashing.php
@@ -44,7 +44,7 @@ return [
     */
 
     'argon' => [
-        'memory' => 1024,
+        'memory' => 8192,
         'threads' => 2,
         'time' => 2,
     ],


GrahamCampbell:
We should maybe fix this on the 5.8 branch too?

HepplerDotNet:
So, now it's clear!
PHP 7.2 and 7.3 use libargon, while PHP 7.4 uses libsodium.
https://bugs.php.net/bug.php?id=78269#1562751980

This means with PHP 7.2 and 7.3 
```memory_cost => 8192```
will end up in 8MB!
But with PHP 7.4 it will be 8KB

GrahamCampbell:
Oh, in which case we should switch on the PHP version tbh. Like, setting 8KB memory cost is not really a strong default.

GrahamCampbell:
Laravel should come with strong, secure defaults out of the box.

GrahamCampbell:
// cc @taylorotwell

tillkruss:
@GrahamCampbell: What should we do about it dropping from 8MB to 8kb in November when we switch? Should this be an upgrade note, or maybe a docs PR?

GrahamCampbell:
We should change the laravel default app to be correct, and not update the docs.

tillkruss:
With a PHP version check in the config, or a condition in the hashing manager?

GrahamCampbell:
1024 * 1024

GrahamCampbell:
https://github.com/laravel/laravel/pull/5101

GrahamCampbell:
Oh, I guess, maybe the hashing manager is the best place for this.

HepplerDotNet:
Sorry for the back and forth. But it was a bit confusing on the php bug list.
If I understood right, it was indeed a bug in PHP and is patched now.
https://bugs.php.net/bug.php?id=78516

GrahamCampbell:
Oh, so this PR should be reverted then?

HepplerDotNet:
As I read it, yes. All changes regarding this issue can be reverted because this will be fixed in next PHP 74-RC

PR #5103, opened 2019-09-11
Use the word "email" instead of "e-mail" consistently

Calinou:
Nowadays, in English, "email" is used more often compared to "e-mail". Several places in Laravel were already using the word "email"; this makes spelling consistent across the framework.

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -8,7 +8,7 @@ return [
     |--------------------------------------------------------------------------
     |
     | Laravel supports both SMTP and PHP's "mail" function as drivers for the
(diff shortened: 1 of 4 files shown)


zaknesler:
This is awesome but _unfortunately_ Laravel follows a very strict flagpole comment rule. Each line has to be 3 characters longer than the line below it (or something like that), and this would mess it up.

taylorotwell:
I don't make contributors fix that stuff @zaknesler

GrahamCampbell:
Heh, I think the current situation is already acceptable tbh.

PR #5104, opened 2019-09-12
[6.x] Improve executeAfterLastTest logic

ludo237:
This PR solve an issue found on tests suite that does not rely upon `RefreshApplication` trait but rather `DatabaseTransaction` trait, also if you have multiple `.env` files for different environments.

Without the artisan call at the end of `executeAfterLastTest` if you try to seed the database you will end up with the database full of artifacts, the reason is that the config file cached will have values from the `.env.testing` file instead of the normal `.env` file.

This is not a breaking change and it should be fine for all supported versions of Laravel, let me know if I need to specify a different target branch.

Farewell

### DIFF
--- a/tests/Bootstrap.php
+++ b/tests/Bootstrap.php
@@ -38,7 +38,7 @@ class Bootstrap implements BeforeFirstTestHook, AfterLastTestHook
     public function executeAfterLastTest(): void
     {
         array_map('unlink', glob('bootstrap/cache/*.phpunit.php'));
-        
+
         $console = $this->createApplication()->make(Kernel::class);
 
         $console->call("config:clear");
--- a/tests/Bootstrap.php
+++ b/tests/Bootstrap.php
@@ -38,5 +38,9 @@ class Bootstrap implements BeforeFirstTestHook, AfterLastTestHook
     public function executeAfterLastTest(): void
     {
         array_map('unlink', glob('bootstrap/cache/*.phpunit.php'));
+        
+        $console = $this->createApplication()->make(Kernel::class);
+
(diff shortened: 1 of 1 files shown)


ludo237:
Ok so I've tried your suggestions @timacdonald and it doesn't work.

I've switched the `<env>` values to `<server>`  and my `phpunit.xml` is almost the same as the master on this repository (few exceptions for excluded files)

I've noticed that inside my `bootstrap/cache` folder file names don't have the `.phpunit.php` extension but only `.php`

image

Probably that's the problem? I've cleared up the cache folder in before running the suite. The suite runs with two simple aliases so they can't interfere with the workflow.

I do agree with you that this kind of things are not universal and they can raise issues in specific workflows, but again my tests suite is nothing special

taylorotwell:
Can we just get rid of this bootstrap?

jdavidbakr:
FWIW, this bootstrap was also breaking phpunit tests with `--process-isolation` enabled.  I had merged in 6.0.2 of `laravel/laravel` and could get my tests to pass until I added `--process-isolation` (because I run out of PHP memory without doing that).  Took me 2 weeks to finally drill down and figure out that the problem went away when I removed the bootstrap.

The error I was getting was:
```
PHPUnit\Framework\Exception: PHP Fatal error:  Uncaught Error: Class 'Route' not found in /[path to laravel project]/routes/api.php:17
```

I am adding this comment hoping that if anyone else runs across it they can know to remove the extensions section from `phpunit.xml`

```
    <extensions>
        <extension class="Tests\Bootstrap"/>
    </extensions>
```

PR #5106, opened 2019-09-13
[6.x] Add 'null' logging channel

rogervila:
As proposed on https://github.com/laravel/ideas/issues/1806, this implementation ads a 'none' logging channel which outputs logs to nowhere by using Monolog's 'NullHandler' class.

### DIFF
--- a/config/logging.php
+++ b/config/logging.php
@@ -72,7 +72,7 @@ return [
             ],
         ],
 
-        'none' => [
+        'null' => [
             'driver' => 'monolog',
             'handler' => NullHandler::class,
         ],
--- a/config/logging.php
+++ b/config/logging.php
@@ -72,11 +72,6 @@ return [
             ],
         ],
 
-        'null' => [
-            'driver' => 'monolog',
-            'handler' => NullHandler::class,
-        ],
-
         'stderr' => [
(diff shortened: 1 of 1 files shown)


taylorotwell:
Why not "null" like the other "null" drivers?

rogervila:
> Why not "null" like the other "null" drivers?

I agree. Consistency is important here

fitztrev:
In another PR, a string value of `"null"` (for broadcast driver) caused issues and @taylorotwell hinted at removing.

> I think going forward into future Laravel versions we may want to consider offering an alternative to "null" for these drivers. Even if we still support null, having another name that doesn't have this kinds of problems would be nice.

https://github.com/laravel/framework/pull/27691#issuecomment-467872248

driesvints:
FYI: this doesn't works with Monolog 2: https://github.com/laravel/framework/issues/30026

rogervila:
> FYI: this doesn't works with Monolog 2: [laravel/framework#30026](https://github.com/laravel/framework/issues/30026)

Thank you @driesvints 
I will try to work on a PR in order to fix it.

driesvints:
@rogervila there's already one: https://github.com/laravel/framework/pull/30123

PR #5107, opened 2019-09-13
[6.x] Remove testing bootstrap extension

timacdonald:
See: https://github.com/laravel/laravel/pull/5104#issuecomment-531205376


Axieum:
Now that the Bootstrap file has gone, where should I hook into to execute my pre-test commands (i.e. `migrate --class=PermissionSeeder`, `passport:install`, etc.)?

timacdonald:
@Axieum you can still use the bootstrap file as it was before it was removed - or just add your own custom version of it. I do in my projects and it works great 👍

PR #5110, opened 2019-09-16
[6.x] Rename Redis alias to RedisManager

reinink:
Laravel 6 switched to PhpRedis as the default driver instead of Predis (#5085). The documentation recommends that if you're using PhpRedis, you should rename the `Redis` alias to `RedisManager`.

> If you plan to use PhpRedis extension along with the `Redis` Facade alias, you should rename it to something else, like `RedisManager`, to avoid a collision with the Redis class. You can do that in the aliases section of your `app.php` config file.

It would then seem logical that the default alias should match this recommendation...which is what this PR changes.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -214,7 +214,7 @@ return [
         'Password' => Illuminate\Support\Facades\Password::class,
         'Queue' => Illuminate\Support\Facades\Queue::class,
         'Redirect' => Illuminate\Support\Facades\Redirect::class,
-        'Redis' => Illuminate\Support\Facades\Redis::class,
+        'RedisManager' => Illuminate\Support\Facades\Redis::class,
         'Request' => Illuminate\Support\Facades\Request::class,
         'Response' => Illuminate\Support\Facades\Response::class,
         'Route' => Illuminate\Support\Facades\Route::class,


GrahamCampbell:
Heh. Could just not include the redis facade alias by default.

Arkanius:
IMO this is only when the default wasn't the *phpredis*. Now that it's the default, do we really need the Facade *RedisManager*?

reinink:
Yeah, maybe including it is the right choice.

GrahamCampbell:
Like, who even uses the Redis facade as is, in their apps?

Arkanius:
I think that some functions we have just at the *Redis* Facade, isnt'it? Like sdiff, sinter, etc?

taylorotwell:
Hmm, I'm not sure I love the name `RedisManager`. We already throw a nice informative exception here so I'm not too worried about it.

reinink:
@taylorotwell Dig it!

For those interested, you can find that check here: https://github.com/laravel/framework/blob/5587e6b8c7a0f807e6dd971ca40c4eff31ec49f5/src/Illuminate/Redis/Connectors/PhpRedisConnector.php#L71-L75

PR #5031, opened 2019-05-22
opt: hard coded app name

sinkcup:
use config app name instead of hard coded "Laravel" in views.

PR #5111, opened 2019-09-16
App name

maximal:


GrahamCampbell:
Thanks for the PR, but the default app is meant to just be to get started, and it's name is intended to be Laravel, and not the app name. This change has been proposed before, and not accepted.

GrahamCampbell:
See https://github.com/laravel/laravel/pull/5031 for more detail.

maximal:
Oh, I see. Extremely sorry.

PR #5113, opened 2019-09-17
[6.x] Data Wraping disable comment

vali-kh:
Added comment into **AppServiceProvider** in order to use for disable **Data Wraping** on **API Resources**.


scoutman57:
Just change this in your own app. No reason to disable here.

vali-kh:
in that face we have to copy that tow line from documentation. just _toggle comment_ that tow line when needed.

PR #5114, opened 2019-09-19
[6.x] Conform to the PSR-12 coding style

u01jmg3:
- Conform to the PSR-12 coding style
  - https://www.php-fig.org/psr/psr-12/
  - https://github.com/squizlabs/PHP_CodeSniffer/tree/master/src/Standards/PSR12

---

- https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md

  > _As of 2019-08-10 PSR-2 has been marked as deprecated. PSR-12 is now recommended as an alternative._

### DIFF
--- a/app/Http/Controllers/Auth/ForgotPasswordController.php
+++ b/app/Http/Controllers/Auth/ForgotPasswordController.php
@@ -7,6 +7,8 @@ use Illuminate\Foundation\Auth\SendsPasswordResetEmails;
 
 class ForgotPasswordController extends Controller
 {
+    use SendsPasswordResetEmails;
+
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


GrahamCampbell:
Hmmm, I'm 👎 here. Your changes break the comments documenting why the `use` happens bellow them.

GrahamCampbell:
Also, Laravel has not yet committed to PSR-12, and when and if it does, will probably have an exclusion for these cases that you've changed.

antonkomarev:
I think this comment is not placed before the `class` definition?

SharakPL:
Removing empty line before first `use` statement in a class is enough to pass PSR12. Moving it above comment is unnecessary.

PR #5115, opened 2019-09-22
Bump laravel-mix version.

mvdnbrk:


ankurk91:
You should also bump `sass-loader` version to 8

driesvints:
Thanks but we prefer to let @JeffreyWay handle this.

PR #5119, opened 2019-09-25
[6.x] Migration should not be in the fresh project

ndeblauw:
I believe this is a small mistake - easy to fix. Either remove the failed_jobs or also include the jobs table, no?


GrahamCampbell:
I think this migration is intended to be there, so failed jobs can be written to the database.

BassemN:
:+1:  for removing this migration like other jobs and queues migrations

taylorotwell:
It is more necessary in 6.0 now that the queue has a limited number of attempts by default.

PR #5123, opened 2019-09-27
Add 'decrypt_password' flag to config/database.php

brantmessenger:
Used in the Illuminate\Database\Connectors\Connector.php in method createConnection to decrypt encrypted password prior to making a connection.

Eliminate unencrypted passwords in logs.


brantmessenger:
Closing pull request realized option is 2 separate components that should not be integrated but still seems like an issue keeping a unencrypted password running around in the config container.

PR #5125, opened 2019-10-03
Remove imports not used

ricardogobbosouza:


driesvints:
These are kept for reference purposes. Please see previously attempted prs.

PR #5126, opened 2019-10-06
[6.x] Sync indentation with laravel/ui webpack files

roberto-aguilar:
The webpack files scaffolded by the laravel/ui's [vue](https://github.com/laravel/ui/blob/master/src/Presets/vue-stubs/webpack.mix.js#L15) and [react](https://github.com/laravel/ui/blob/master/src/Presets/react-stubs/webpack.mix.js#L15) presets have a different indentation that the [default one](https://github.com/laravel/laravel/blob/master/webpack.mix.js#L15). This change keeps them consistent and avoid unnecessary git diffs when the scaffolding happens.

### DIFF
--- a/webpack.mix.js
+++ b/webpack.mix.js
@@ -12,4 +12,4 @@ const mix = require('laravel-mix');
  */
 
 mix.js('resources/js/app.js', 'public/js')
-    .sass('resources/sass/app.scss', 'public/css');
+   .sass('resources/sass/app.scss', 'public/css');


webpack.mix.js
```diff
@@ -12,4 +12,4 @@ const mix = require('laravel-mix');
  */
 
 mix.js('resources/js/app.js', 'public/js')
-    .sass('resources/sass/app.scss', 'public/css');
+   .sass('resources/sass/app.scss', 'public/css');
```
antonkomarev: It is 3 spaces. Not 4 not 2. It's very uncommon code style. You'd better fix it in UI repo.

roberto-aguilar: Thanks for the feedback, @antonkomarev!

While i agree that it's not common, there are places where Taylor does this kind of alignment ([even in PHP Code](https://github.com/laravel/laravel/blob/master/app/Providers/RouteServiceProvider.php#L54-L56)). The fact that two different presets in the `laravel/ui` repository and pretty much all the examples in [the documentation](https://laravel.com/docs/6.x/mix) share this indentation proves that it is a decision, not an oversight.

I'll wait @taylorotwell decision, and if 2 or 4 spaces will be, i'll update all the other occurrences :)

GrahamCampbell: The code style is 4 spaces, as documented by https://github.com/laravel/laravel/blob/master/.styleci.yml.

PR #5127, opened 2019-10-07
[6.x] Add ValidateJson middleware

gepeixoto:
Every time I start a new API project, I create a simple middleware to verify JSON on incoming requests’ body.
Even considering it is simple, it is useful to inspect the payload integrity and, in case of a not valid JSON input, the middleware returns a more descriptive error (good for API consumers), saving resources by avoiding to getting deep in the application.
Considering that Laravel is used for so many JSON API projects, I think it would be good to have this out-of-the-box, by just enabling it in Kernel file.


taylorotwell:
Feel free to release this as a package.

PR #5128, opened 2019-10-07
[6.x] update Laravel Mix to new version 5

browner12:
and update the other shared dependencies with Mix.

I've personally updated these 3 dependencies in some projects and run into no issues, but maybe @JeffreyWay could weigh in here with any potential concerns.

also, since this is `laravel/laravel` it won't affect existing projects, so we should be good to update this for new projects.


browner12:
hah, oops. thanks.

PR #5129, opened 2019-10-08
[6.x] Password confirmation

driesvints:
These changes along with the new password confirmation functionality and the new password rule. See https://github.com/laravel/framework/pull/30214

I've also removed the "guest" middleware from the password reset controllers so users can reset their password while logged in. This was needed in combination with the new password confirmation screen. See https://github.com/laravel/ui/pull/34


antonkomarev:
@driesvints isn't it targeted to 7.x branch?

driesvints:
@antonkomarev no there's nothing breaking here.

PR #5135, opened 2019-10-09
Removed un-used MustVerifyTrait namespace

vishalinfyom:


driesvints:
These are kept for referencing. Please see previously attempted prs.

PR #5136, opened 2019-10-10
Update composer.json to remove license MIT

thedejavunl:
Because the skeleton app isn't licensed according to earlier PR requests and reactions from @taylorotwell the license needs to be removed from composer.json. The framework composer.json has the MIT license tag applied to it so the license will apply to the framework only. This way the skeleton app can be used in proprietary projects.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,6 @@
         "framework",
         "laravel"
     ],
-    "license": "MIT",
     "require": {
         "php": "^7.2",
         "fideloper/proxy": "^4.0",


taylorotwell:
even having MIT license in there doesn't prevent the skeleton from being used in proprietary projects.

thedejavunl:
So we are allowed to remove the license tag in your skeleton project and use it in our proprietary projects without using the MIT license?

driesvints:
@thedejavunl yeah

PR #5137, opened 2019-10-10
Fix for reset unique in seeder unique columns

lloricode:
https://github.com/fzaninotto/Faker/issues/1512#issuecomment-450109983


GuidoHendriks:
Why would you randomly add that if it happens to solve an issue in your specific use case?

lloricode:
@GuidoHendriks, sorry thats why i already closed this

PR #5138, opened 2019-10-13
[6.x] Add the AWS_SESSION_TOKEN variable for the S3 config

mnapoli:
In some cases, the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` variables are not enough to authenticate to the AWS API. The `AWS_SESSION_TOKEN` must be provided as well.

The `AWS_SESSION_TOKEN` is part of the "Temporary Security Credentials" authentication mechanism. See the [AWS Documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html).

One of those cases is when Laravel runs on AWS Lambda. In that case the `AWS_SESSION_TOKEN` variable is populated ([source](https://docs.aws.amazon.com/lambda/latest/dg/lambda-environment-variables.html)).

With the default Laravel installation, when listing files in the `s3` drive, I get:

> Aws\S3\Exception\S3Exception
> Error executing "ListObjects" on "https://redacted.s3.amazonaws.com/?prefix=&delimiter=%2F&encoding-type=url"; AWS HTTP error: Client error: `GET https://redacted.s3.amazonaws.com/?prefix=&delimiter=%2F&encoding-type=url` resulted in a `403 Forbidden` response:
> `<?xml version="1.0" encoding="UTF-8"?> <Error><Code>InvalidAccessKeyId</Code><Message>The AWS Access Key Id you provided does not exist in our records.</Message><AWSAccessKeyId>redacted</AWSAccessKeyId><RequestId>redacted</RequestId><HostId>redacted</HostId></Error>`

By adding the `token` line to the configuration, the API call succeeds and my application works as expected.

I expect that adding the line will not cause any regression in cases where the `AWS_SESSION_TOKEN` is not define: the `token` option will remain empty.

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -59,6 +59,7 @@ return [
             'driver' => 's3',
             'key' => env('AWS_ACCESS_KEY_ID'),
             'secret' => env('AWS_SECRET_ACCESS_KEY'),
+            'token' => env('AWS_SESSION_TOKEN'),
             'region' => env('AWS_DEFAULT_REGION'),
             'bucket' => env('AWS_BUCKET'),
             'url' => env('AWS_URL'),


mariomka:
I was stucked with SQS for the same problem for two days.

taylorotwell:
Are you sure you don't have another permission problem, such as missing permissions for your IAM role? 

I use S3 on Vapor without issue with the latest 6.0 configuration. If you do not pass any credentials to the S3Client it will automatically use the correct credentials, including the token. In Vapor we do this by removing the AWS credentials from people's configuration files on deployment so the correct credentials will just be determined by the AWS SDK.

Before merging this I'll have to make sure it won't cause any problems with what we're already doing on Vapor.

mnapoli:
> I use S3 on Vapor without issue with the latest 6.0 configuration. If you do not pass any credentials to the S3Client it will automatically use the correct credentials, including the token. In Vapor we do this by removing the AWS credentials from people's configuration files on deployment so the correct credentials will just be determined by the AWS SDK.

OK I see, I _think_ that makes sense to me: the SDK automatically picks up everything correctly.

But in this specific instance, Laravel provides the SDK a specific array (so the SDK isn't in "autoconfigure" mode):

https://github.com/laravel/framework/blob/02d02266d28b9ea4495fd698d8aca41f5933b746/src/Illuminate/Filesystem/FilesystemManager.php#L216-L225

https://github.com/laravel/framework/blob/02d02266d28b9ea4495fd698d8aca41f5933b746/src/Illuminate/Filesystem/FilesystemManager.php#L206

So I think that explains why the token isn't automatically picked up.

> Before merging this I'll have to make sure it won't cause any problems with what we're already doing on Vapor.

👍 

To be frank my confidence in this PR is not absolute (first time I realize this token exists, all other cases I've used the SDK in the autoconfigure mode), it's a classic case of "without it it doesn't work, with it it works". But it seems to make sense.

mariomka:
I think there are two options for it to work properly:
- Add `AWS_SESSION_TOKEN` to configuration files.
- Remove `key` and `secret` keys from configuration files (I guess this is not an option for the default configuration).

mariomka:
If you finally choose to add `AWS_SESSION_TOKEN`, it should be added to all AWS services: S3, SQS, DynamoDB and SES.

narwy:
> I think there are two options for it to work properly:
> 
> * Add `AWS_SESSION_TOKEN` to configuration files.
> * Remove `key` and `secret` keys from configuration files (I guess this is not an option for the default configuration).

I wouldn't remove anything from the configuration, that's just a quick fix to get it working. I would just add it as an empty env in the configs for now as it's pre-populated from Lambda and most other AWS services.

taylorotwell:
I'll look into this more but I agree we would probably want to add this for all AWS related configuration items in Laravel.

Also, what happens when running on Lambda if you pass configuration to the S3 client but the 'token' option is `null`? Will it then use the token from the environment variables?

mariomka:
Passing a `null` token to the client doesn't work. AWS SDK doesn't load environment variables if any credential is provided. It is the expected behavior: 

> When you initialize a new service client **without providing any credential arguments**, the SDK uses the default credential provider chain to find AWS credentials.

https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials.html

narwy:
> I'll look into this more but I agree we would probably want to add this for all AWS related configuration items in Laravel.
> 
> Also, what happens when running on Lambda if you pass configuration to the S3 client but the 'token' option is `null`? Will it then use the token from the environment variables?

If you pass null will default to normal credentials.

taylorotwell:
I think I'm going to hold off on changing the framework shipped configs for now.

mnapoli:
Understood. So to sum up, if we want things to work, users have to add this line into `config/filesystems.php`:

```php
    'token' => env('AWS_SESSION_TOKEN'),
```

Would there be any way to make things just work by default? Let me know if there is anything I can do.

narwy:
@mnapoli Most people use the IAM Console credentials, I think this should be documented rather than implemented into the config. Could be confusing for new users who would be setting up two lots of credentials etc. You can submit a PR for the docs.

mnapoli:
@narwy I'm not sure I follow what you mean specifically. Here on AWS Lambda the credentials are autogenerated by AWS (those are environment variables that are automatically defined by AWS Lambda).

narwy:
@mnapoli Yes I understand, but there's no point adding it because it's not for everyone. Just needs to be documented.

driesvints:
@mnapoli @narwy I think this could indeed be added as extra documentation perhaps. Where it details that you can add this optional configuration. The skeleton is user-land owned so you're allowed to change whatever you want here.

mnapoli:
For those interested, there is the same problem in `config/queue.php` for the SQS driver. You need to add this line:

```diff
        'sqs' => [
            'driver' => 'sqs',
            'key' => env('AWS_ACCESS_KEY_ID'),
            'secret' => env('AWS_SECRET_ACCESS_KEY'),
+           'token' => env('AWS_SESSION_TOKEN'),
            'prefix' => env('SQS_PREFIX', 'https://sqs.us-east-1.amazonaws.com/your-account-id'),
```

abronin:
Very sad this is not documented. I spent some hours today tried to understand why I can dispatch a job to the SQS from my local environment but not from AWS Lambda. Got the error 'The security token included in the request is invalid.' and found my answer here, thanks to @mnapoli

jerodev:
I've had exactly this problem with the dynamodb cache driver on lambda. Adding the `token` key fixed this. :)

mnapoli:
At least once every 2 weeks I get people stuck because of this. This could be when using S3, DynamoDB, SQS, or any other AWS service.

Would you be open to reconsider merging this (assuming I would send a new PR that updates everywhere else it makes sense)?

deleugpn:
I'm interested in investigating this. Is it easy to reproduce? Are people deploying lambda with IAM access key & secret key?

abronin:
@deleugpn As far as I know, there is no other way to get access to the AWS services (SQS, S3, etc.) from Lambda except of this temporary credentials.

taylorotwell:
I'm fine with it as long as it does not break Vapor.

bkuhl:
> I'm fine with it as long as long as it does not break Vapor.

I understand why this might cause a delay, but it doesn't seem like this should lead to the PR being closed.

deleugpn:
For people running code on Lambda, if your Lambda Execution Role has access to the AWS resources you want (e.g. s3, DynamoDB, SQS), all you have to do is completely delete key and secret from Laravel, instead of adding temporary token. By not having key and secret at all, AWS SDK will figure out the authentication from Lambda Role automatically

PR #5140, opened 2019-10-16
added the doctrine/dbal

hapheus:
doctrine/dbal is required for migrations. 

To reproduce the issue follow these steps
```
laravel new test
cd test
php artisan make:migration add_field_to_users_table --table=users
```
no need to setup a new database for that ;) 
`touch database/database.sqlite`
edit your .env file to use sqlite instead of mysql. 

edit the created ____add_field_to_users_table.php migration file, add a random column and drop it in down method. 
```
    public function up() {
        Schema::table('users', function (Blueprint $table) {
            $table->text('field')->unique()->nullable();
        });
    }

    public function down() {
        Schema::table('users', function (Blueprint $table) {
            $table->dropColumn('field');
        });
    }
```

The `php artisan migrate`  command runs as expected, but 
`php artisan migrate:refresh` will fail with the missing `Symfony\Component\Debug\Exception\FatalThrowableError  : Class 'Doctrine\DBAL\Driver\PDOSqlite\Driver' not found` Exception, that is shipped with `doctrine/dbal`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,6 +9,7 @@
     "license": "MIT",
     "require": {
         "php": "^7.2",
+        "doctrine/dbal": "^2.9",
         "fideloper/proxy": "^4.0",
         "laravel/framework": "^6.2",
         "laravel/tinker": "^1.0"


staudenmeir:
From the [documentation](https://laravel.com/docs/6.x/migrations#dropping-columns):

> To drop a column, use the dropColumn method on the Schema builder. Before dropping columns from a SQLite database, you will need to add the doctrine/dbal dependency to your composer.json file and run the composer update command in your terminal to install the library:

hapheus:
Thanks for the reply, I will test it with an other database and close the pull request if the test succeeds.

hapheus:
Works great with postgres and mysql

PR #5145, opened 2019-10-31
Fix for too long key error

dgallinari:
When using MariaDB or MySQL below v. 5.7.7, there is an inconvenient error stating:
SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 767 bytes (SQL: alter table users add unique users_email_unique(email))

There are many great solutions for this issue, being one of them well-explained in:
https://laravel-news.com/laravel-5-4-key-too-long-error

However, I believe the best solution would be not having to deal with an error at all. Furthermore, you won't affect any other schema setting an undesired default length to them.

### DIFF
--- a/database/migrations/2014_10_12_000000_create_users_table.php
+++ b/database/migrations/2014_10_12_000000_create_users_table.php
@@ -16,7 +16,7 @@ class CreateUsersTable extends Migration
         Schema::create('users', function (Blueprint $table) {
             $table->bigIncrements('id');
             $table->string('name');
-            $table->string('email')->unique();
+            $table->string('email', 191)->unique();
             $table->timestamp('email_verified_at')->nullable();
             $table->string('password');
             $table->rememberToken();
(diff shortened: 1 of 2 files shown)


GuidoHendriks:
Don't think the default should be set to what a 4+ year old database needs. You can always change this for your specific project if you need to?

driesvints:
Please see many previously attempted prs.

dgallinari:
Support for MySQL 5.7 will end on October 2023. It is widely used still.

As for attempted PRs, I just wanted to show here my point of view. Thinking of Laravel as a product, I would prefer having a product which is easily installed (with no Googling required on possible errors). This bounces new users.

If there were a lot of PRs, it should be at least analysed. This is a rule on UX, if many users are stating the same, you should be looking at that.

IMHO

driesvints:
@dgallinari I can install Laravel and run it on both 5.7 and 8.0 with default settings just fine.

dgallinari:
It happens on versions below 5.7.7

PR #5148, opened 2019-11-07
Added sqlite:memory connection to use in phpunit

thezahedi:
With this connection, we may only use `<server name="DB_CONNECTION" value="sqlite:memory"/>` for testing on memory.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -43,6 +43,12 @@ return [
             'foreign_key_constraints' => env('DB_FOREIGN_KEYS', true),
         ],
 
+        'sqlite:memory' => [
+            'driver' => 'sqlite',
+            'database' => ':memory:',
+            'foreign_key_constraints' => env('DB_FOREIGN_KEYS', true),
+        ],
+
         'mysql' => [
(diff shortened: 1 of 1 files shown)


thezahedi:
With this connection, we can only use `<server name="DB_CONNECTION" value="sqlite:memory"/>` for testing on memory.

GuidoHendriks:
Isn't this already solved by this PR?
https://github.com/laravel/laravel/pull/3698

mfn:
Ack, you just need to set `DB_DATABASE` appropriately!

thezahedi:
With #3698 PR we still have to set two environment values in phpunit.xml, one for DB_CONNECTION and one for DB_DATABASE, but with my PR we just have to only one ENV in phpunit.xml.

I think it's easier to config just one env value.

My team's interns made mistakes about these options.

michaeldyrynda:
Just configure it in your `phpunit.xml` file once and you’re done.

GuidoHendriks:
Not sure if you can use `DATABASE_URL` for in-memory databases? 🤔

PR #5149, opened 2019-11-08
fix: timezone hardcode

sinkcup:
timezone config should use env


GuidoHendriks:
Please check previous PR’s. Have seen this been proposed and rejected before.

PR #5153, opened 2019-11-17
[6.x] Generate a password using the Hash facade

osbre:


GrahamCampbell:
Thanks for the PR. I don't think that's the way passwords are "hashed", though?

osbre:
the previous version saved the result of executing the Hash facade, which allowed to log in under the user's email and password "password"

It works the same way

https://github.com/laravel/laravel/blob/master/app/Http/Controllers/Auth/RegisterController.php#L69

GrahamCampbell:
Ah, I see. The reason for not calling hash was because it takes a long time to execute, and uses memory too.

GrahamCampbell:
This slows down the tests unnecessarily, which is why the hashed value is hard-coded.

osbre:
Thanks for the explanation

PR #4903, opened 2019-01-12
Created MIT License file.

mattsidjohn:
I noticed you added the MIT license note in the README.md, however it is stated for best practice to include the license file in the root of your project. I'm sure you know this, it may have just been overlooked. I just wanted to help and add it for you. On a side note my company - Midwestern Interactive uses Laravel exclusively for our software, api, and ios/android development projects. Really appreciate your contribution to the PHP community. It has in some ways changed my life and the life of my employees. Thanks Taylor live long and prosper!

PR #4659, opened 2018-05-13
Add MIT License to repo

robinnydahl:

PR #4080, opened 2016-12-24
Added License

JeyKeu:

PR #5154, opened 2019-11-17
Create license

aliadhillon:
This is a MIT License.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 The Laravel Framework
+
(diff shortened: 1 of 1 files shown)


antonkomarev:
There are similar PRs, and all of them were rejected: #4903 #4659 #4080 

The reason there is not a license in this repository is that this is the project skeleton used to make other applications.

I propose to add `LICENSE` to `.gitattributes` ignore list, and then it wouldn't be pulled from the repository on fresh application creation. Then repository itself will have license, but user applications would not.

driesvints:
Please see previous attempts.

aliadhillon:
Understood, clearly.

PR #5157, opened 2019-11-23
[7.x] Switch to Symfony 5

GrahamCampbell:
Pairs with https://github.com/laravel/framework/pull/30610.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -166,7 +166,7 @@ return [
     |
     */
 
-    'secure' => env('SESSION_SECURE_COOKIE', false),
+    'secure' => env('SESSION_SECURE_COOKIE', null),
 
     /*
     |--------------------------------------------------------------------------
--- a/config/session.php
+++ b/config/session.php
@@ -194,6 +194,6 @@ return [
     |
     */
 
-    'same_site' => null,
+    'same_site' => 'lax',
 
 ];
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -2,8 +2,8 @@
 
 namespace App\Exceptions;
 
-use Exception;
 use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
+use Throwable;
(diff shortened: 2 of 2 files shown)


taylorotwell:
@GrahamCampbell Laravel 7.0 is no longer installable because of complaints about Tinker not supporting Symfony 5.

GrahamCampbell:
@taylorotwell Yes, I'm aware of this, and I raised this with tinker a few days ago. It will be fixed in the coming weeks. :)

GrahamCampbell:
In the meanwhile, Laravel 7 can still be installed without tinker.

GrahamCampbell:
@taylorotwell We can track the tinker stuff at https://github.com/laravel/tinker/pull/79. :)

config/session.php
```diff
@@ -166,7 +166,7 @@
     |
     */
 
-    'secure' => env('SESSION_SECURE_COOKIE', false),
+    'secure' => env('SESSION_SECURE_COOKIE', null),
```
mathieutu: Hi @GrahamCampbell why did you set null here, as env default parameter is by default null?

config/session.php
```diff
@@ -194,6 +194,6 @@
     |
     */
 
-    'same_site' => null,
+    'same_site' => 'lax',
```
mathieutu: Hi @GrahamCampbell why did you change null to lax here? It changes the behavior but Symfony 5 does not need this. (We totally can pass null to Cookie).

GrahamCampbell: Because we explicitly decided we wanted to change the default in Laravel.

GrahamCampbell: And actually, having it as `null` broke our test suite switching from Symfony 4.3 to 5.0, so they must have changed something...

KennedyTedesco: Just a side note: lax will be the default on Chrome. https://www.chromestatus.com/feature/5088147346030592

PR #5158, opened 2019-11-23
set up time zone from .env

murtazasultani:
It was a problem that raised a number of questions about how to set up our local time zone in Larval.
here the default time zone is UTC and if anyone wants to change the time zone they can change the time zone from the .env file.


GuidoHendriks:
This has been tried and rejected a number of times before. Please see previous PR's: https://github.com/laravel/laravel/pulls?utf8=✓&q=timezone

GrahamCampbell:
Timezone is usually app specific, so you should edit the config file to change it.

shadoWalker89:
@GrahamCampbell This is not true all the time, i have a platform built with laravel that is used in 3 different countries. Each project is hosted on it's own server, but they all share the same code. So here i have the same code but with 3 different time zones. I even have all the projects as git remotes and push updates from my local git repo to all of them.
Your statement is true for apps that are made specifically for a company, like a single use app. But for apps that used by multiple customers that is something else.
It's not a big deal to reject this PR cuz it can be changed in the app code, but maybe it's better to have it out of the box.

GuidoHendriks:
You can just add it to your own project if you need to. This repo is just a default which is good for most projects. You own your own copy of it, do whatever you want to it.

rs-sliske:
I would argue that the timezone config should be left UTC anyway, maybe even more so for people deploying in different countries. 

Store all dates in UTC and then do the conversion for display. rather than storing dates as local time and having to deal with local time changing.

PR #5162, opened 2019-11-28
[6.x] Laravel  Support  php7.4  preload

wujunze:
[6.x] Laravel  Support  php7.4  preload

about issue  https://github.com/laravel/ideas/issues/1406

### DIFF
--- a/preload.php
+++ b/preload.php
@@ -1,0 +1,119 @@
+<?php
+
+require_once __DIR__ . '/vendor/autoload.php';
+
+class Preloader
+{
+    private array $ignores = [];
+
+    private static int $count = 0;
+
+    private array $paths;
+
+    private array $fileMap;
+
+    public function __construct(string ...$paths)
+    {
+        $this->paths = $paths;
+        $classMap = require __DIR__ . '/vendor/composer/autoload_classmap.php';
+        $this->fileMap = array_flip($classMap);
+    }
+
+    public function paths(string ...$paths): Preloader
+    {
+        $this->paths = array_merge(
+            $this->paths,
+            $paths
+        );
+
+        return $this;
+    }
+
+    public function ignore(string ...$names): Preloader
+    {
+        $this->ignores = array_merge(
+            $this->ignores,
+            $names
+        );
+
+        return $this;
+    }
+
+    public function load(): void
+    {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This has the disadvantage that you can't upgrade deployed apps by reloading php-fpm, like envoyer does.

wujunze:
> This has the disadvantage that you can't upgrade deployed apps by reloading php-fpm, like envoyer does.

Hello GrahamCampbell : 
    Do you have a better plan ? @GrahamCampbell

fragkp:
@wujunze 
Hey,
I'm interested in why you're ignore FilesystemCache, LogManager and Carbon by default.

wujunze:
> @wujunze
> Hey,
> I'm interested in why you're ignore FilesystemCache, LogManager and Carbon by default.

Hello : 
   I think don't  ignore class

GrahamCampbell:
> Do you have a better plan

I propose we implement this more like the old framework compiled file command, instead of hacking it into laravel/laravel. We must also document the risks, since this is not a simple "turn it on and everything will be faster without consequences".

themsaid:
I agree with @GrahamCampbell. However, I think we can just ship the file with laravel/laravel so people can use it *if* they want.

@wujunze I wonder if you managed to successfully test that it works on an Ubuntu 18.04 setup.

deleugpn:
Coming from a Docker environment, I'm expecting to preload every single file in the application.

wujunze:
> Coming from a Docker environment, I'm expecting to preload every single file in the application.

Your needs can be met

themsaid:
@wujunze did you test this code on a server?

GrahamCampbell:
Why does this need to be included in the default app?

GrahamCampbell:
> Coming from a Docker environment, I'm expecting to preload every single file in the application.

This won't be possible, actually. Preloading requires all classes to be available, and Laravel's source code references classes that need not exist, for example, any of the optional dependencies.

taylorotwell:
Thanks for PR but this contains no benchmark on improvements, it also does not give thought to what classes actually make sense to preload (the "hot" classes) because there is a optimum performance threshold by preloading the hot classes only.

PR #5165, opened 2019-12-01
[5.6] Update RedirectIfAuthenticated.php in /app/Http/Middleware/

iunteq:
This change is helpful for MULTI-GUARD-AUTHENTICATION. **A dynamic approach!**
 
As it is SWITCH-CASE and default redirect link is defined so, anyone use predefined default single guard then nothing affect and if someone want to use MULTI-GUARD-AUTHENTICATION then it will be easy as well as simple to use; just need to replace their won guard name.

##If you want to use NAMED ROUTE  (https://laravel.com/docs/5.8/routing#named-routes) then you need to create route with proper name in this [routes/web.php] file.

Thank you.

### DIFF
--- a/app/Http/Middleware/RedirectIfAuthenticated.php
+++ b/app/Http/Middleware/RedirectIfAuthenticated.php
@@ -17,8 +17,35 @@ class RedirectIfAuthenticated
      */
     public function handle($request, Closure $next, $guard = null)
     {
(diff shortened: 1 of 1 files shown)


driesvints:
5.6 isn't maintained anymore. This is also a pr to master which is for the 6.x release.

PR #5166, opened 2019-12-02
[6.x] Move Laravel Tinker to dev dependencies

browner12:
been having a little debate on the Slack channel about where Laravel Tinker belongs in our dependencies.

I personally **never** use it in production, so for me loading it is a waste of resources (probably minimal, but still something).

I believe there's also a bit of inherent danger by using it in production. A model could be accidentally deleted, something that may be prevented by going through the normal UI modes of manipulating data.

Due to these 2 concerns I would say we should default Tinker into the dev dependencies, and put the onus on the user that wants Tinker in their production environment to move it in their `composer.json` file.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,12 +10,12 @@
     "require": {
         "php": "^7.2",
         "fideloper/proxy": "^4.0",
-        "laravel/framework": "^6.2",
-        "laravel/tinker": "^2.0"
+        "laravel/framework": "^6.2"
     },
     "require-dev": {
         "facade/ignition": "^1.4",
         "fzaninotto/faker": "^1.4",
+        "laravel/tinker": "^2.0",
         "mockery/mockery": "^1.0",
         "nunomaduro/collision": "^3.0",
         "phpunit/phpunit": "^8.0"


antonkomarev:
Same here. One of the first things after creating new app - move tinker to the dev dependencies.

alexleeelkins:
Running a bad tinker command is as bad as running a bad SQL query. Only users who are trusted should have read/write access to run executables on the server, just like only trusted users should have read/write access to the database.
That being said, it seems like Tinker should be either opt-in in general (like laravel/ui is, given that tinker is _not_ required to get an app up and running), or required by default for dev but not for prod.

Just my 2 cents.

GrahamCampbell:
I use tinker in production all the time, actually. It's a perfectly valid use case.

GrahamCampbell:
Most of Laravel is "not required", but we still load it by default. I think the current defaults are fine.

mfn:
I've had once or twice the need for tinker in production and it was really useful to have it available.

browner12:
I 100% agree with everyone that using Tinker in production can be valid, I'm not disputing that.

*However*, what I'm recommending is that we do not activate it in production by default, unless a user who knows the risks chooses to activate it.

Again, the resource hit that registering Tinker on **every** request causes is probably minuscule, but for users that never use it in production, there is no reason to have it loaded.

robclancy:
I always use it in production. But that doesn't mean that should be default by any means.

GrahamCampbell:
> Again, the resource hit that registering Tinker on every request causes is probably minuscule, but for users that never use it in production, there is no reason to have it loaded.

The tinker service provider is deferred anyway, so it doesn't actually get loaded by most requests. Only requests that internally execute a cli command. Most people don't write code that does that in most requests.

gocanto:
Having tinker in production is such a good advantage. I do not see the issue with ti here!

taylorotwell:
Don't think I'll change the defaults right now. Feel free to move it to dev in your own applications if you want.

PR #5170, opened 2019-12-05
[6.x] Upgrade fzaninotto/faker version to support PHP 7.4

hamidafghan:
https://github.com/fzaninotto/Faker/issues/1863

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
     },
     "require-dev": {
         "facade/ignition": "^1.4",
-        "fzaninotto/faker": "^1.4",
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks for the PR. This will be the default in Laravel 7. As for Laravel 6, well, the existing version constraint is already fine, in that composer will resolve the very latest version already, so it should work fine on PHP 7.4.

GrahamCampbell:
https://github.com/laravel/laravel/blob/develop/composer.json#L17

PR #5171, opened 2019-12-06
[6.x] change default database path handling of sqlite connection

tjventurini:
I guess this would make more sense than the previous default.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -38,7 +38,7 @@ return [
         'sqlite' => [
             'driver' => 'sqlite',
             'url' => env('DATABASE_URL'),
-            'database' => env('DB_DATABASE', database_path('database.sqlite')),
+            'database' => database_path(env('DB_DATABASE', 'database.sqlite')),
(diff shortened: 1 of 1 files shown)


GuidoHendriks:
That would require the database to be stored in the application's database directory for every instance of Laravel. Would that be the best default or would it be better to leave it up to the user?

driesvints:
We indeed can't do this because this forces the end user to store their db in the database directory while currently they can keep it anywhere. But I do see your point about having to set the full path every time instead of just having to set the database name.

tjventurini:
> We indeed can't do this because this forces the end user to store their db in the database directory while currently they can keep it anywhere. But I do see your point about having to set the full path every time instead of just having to set the database name.

I wouldn't think that this change forces the user to only have one option. I mean all users can change the database configuration file, so I don't see the conflict here.

I just think that there are more users that set up the sql database file directly in the project instead of somewhere else so this should be the default.

But thas what I think :wink:

PR #5172, opened 2019-12-09
[6.x] Fix return type

alberto-bottarini:
Parent method could `@return \Illuminate\Http\Response|\Symfony\Component\HttpFoundation\Response` (https://github.com/laravel/framework/blob/6.x/src/Illuminate/Foundation/Exceptions/Handler.php#L185).

This PR makes Handler class consistent with its parent.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -42,7 +42,7 @@ class Handler extends ExceptionHandler
      *
      * @param  \Illuminate\Http\Request  $request
      * @param  \Exception  $exception
-     * @return \Illuminate\Http\Response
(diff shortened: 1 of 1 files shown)


alberto-bottarini:
@taylorotwell 
just to understand... what are the reasons to not merge this PR?
This fix will allows an empty laravel to be compliant with phpstan and other static code analyzer.

nunomaduro:
@alberto-bottarini Do you have this issue with Larastan?

alberto-bottarini:
Yes

nunomaduro:
Open an issue on Larastan then.

alberto-bottarini:
Larastan behaves correctly imho. It's  an overrided method that change return type.

nunomaduro:
Maybe, let's handle it in our repository :)

PR #5173, opened 2019-12-10
[6.x] DRY up path to /home

taylorotwell:
We were defining this path in six places (!).

### DIFF
--- a/app/Http/Controllers/Auth/ConfirmPasswordController.php
+++ b/app/Http/Controllers/Auth/ConfirmPasswordController.php
@@ -3,6 +3,7 @@
 namespace App\Http\Controllers\Auth;
 
 use App\Http\Controllers\Controller;
+use App\Providers\RouteServiceProvider;
 use Illuminate\Foundation\Auth\ConfirmsPasswords;
 
 class ConfirmPasswordController extends Controller
@@ -25,7 +26,7 @@ class ConfirmPasswordController extends Controller
      *
      * @var string
      */
-    protected $redirectTo = '/home';
+    protected $redirectTo = RouteServiceProvider::HOME;
 
     /**
      * Create a new controller instance.
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -3,6 +3,7 @@
 namespace App\Http\Controllers\Auth;
 
 use App\Http\Controllers\Controller;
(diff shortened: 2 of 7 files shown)


Lloople:
Maybe a method would be better, like the `redirectTo()` in Auth Controller, in case someone needs to build this value with some logic

tontonsb:
I agree to @Lloople  that `RouteServiceProvider::home()` would be more versatile. But I'll gladly take this PR over the current state!

sierrafayad:
Would it be ok if the `RedirectsUser` trait have access to the `RouteServiceProvider`? Instead of the config file I was proposing when modifying the framework. This way it could retain the actual functionality of the function and property `redirectTo` using the HOME constant from the ServiceProvider but letting you define the function redirectTo as you like

sertxudeveloper:
> Maybe a method would be better, like the `redirectTo()` in authentication controllers, in case someone needs to build this value with some logic

This solution uses the Route Service Provider, if someone needs to add some logic to the home route, like me, we could use a custom service provider or helper to return the proper string in the controllers we need it.

Lloople:
@sertxudeveloper You mean replacing all code occurrences of `RouteServiceProvider::HOME` by `RouteServiceProvider::getHomeRoute()` or something like that? If that's the case and we can already do this replacement for `/home` why we need a PR like this?

sertxudeveloper:
@Lloople Yes, I would have no problem in modifying it with a personalized method in the places where I need it.

taylorotwell:
@Lloople you can still override the `redirectTo` methods as needed.

dionysiosarvanitis:
Still prefer a [named route](https://github.com/laravel/laravel/pull/4657) for this.

marchershey:
How does this benefit?

> @Lloople Yes, I would have no problem in modifying it with a custom method in the places where I need it.

Yeah, but what's easier... changing the $redirectTo or changing the $redirectTo & service provider?

--

Is this change documented anywhere so I can read up on the benefits? Thanks

martindilling:
@taylorotwell 
> @Lloople you can still override the `redirectTo` methods as needed.

I would very much prefer the `redirectTo` method being the only way to define this. Can't see a reason for it to be a property with check for the existence of the method. 
If it's just a method, it's allows for using both a plain url, route name or anything you want out of the box. But as it is now, if you "just" want to use a route name, you can't just change the value in 7 places, you have to remove a property, and add a new method in 7 places ;)

PR #5175, opened 2019-12-14
[6.x] Update redirectTo return type PHPDoc

canvural:
Updates return type doc block to be same as [parent one](https://github.com/laravel/framework/blob/6.x/src/Illuminate/Auth/Middleware/Authenticate.php#L90).


driesvints:
Thanks!

arcanedev-maroc:
@driesvints I think the correct returned type is `void` not `null`

PR #5176, opened 2019-12-14
fix predis name

liujundezhanghao:
fix predis name


GrahamCampbell:
Thanks for the PR. This really is meant to say `phpredis` and not `predis`.

PR #5177, opened 2019-12-16
[6.x] Add hrtime for newer versions of PHP

ahmadmayahi:
`hrtime` was introduced in PHP 7.3 (https://www.php.net/manual/en/book.hrtime.php) to provide high resolution time.

Although it's a minor change, but it tries to give us the most accurate time since it uses the best possible APIs on different platforms which brings resolution up to nanoseconds.

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -7,7 +7,7 @@
  * @author   Taylor Otwell <taylor@laravel.com>
  */
 
-define('LARAVEL_START', microtime(true));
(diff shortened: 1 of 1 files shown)


driesvints:
I think it's best that we wait until 7.3 is the minimum requirement (Laravel 8) and even then I don't know if it's wanted to implement such a breaking change.

PR #5178, opened 2019-12-17
[6.x] Default email from name to app name

andrewminion-luminfire:
I continually forget to set `MAIL_FROM_NAME` and IMHO it makes more sense to default to the app name vs. “Example”


antonkomarev:
Maybe it will be better to add to `.env` variable `MAIL_FROM_NAME=${APP_NAME}`?

PR #5179, opened 2019-12-18
[6.x] Allow configurable emergency logger

Stokoe0990:
This adds a default emergency logger path to the logging config file.

---

Goes hand-in-hand with laravel/framework#30873.


Stokoe0990:
https://github.com/laravel/framework/pull/30873

PR #5180, opened 2019-12-18
[6.x] Add MAIL_FROM_ADDRESS & MAIL_FROM_NAME to .env file

antonkomarev:
Follow-up to #5178

This PR adds `MAIL_FROM_ADDRESS` & `MAIL_FROM_NAME` variables to .env file and makes `APP_NAME` as fallback value for `MAIL_FROM_NAME`.


.env.example
```diff
@@ -29,6 +29,8 @@ MAIL_PORT=2525
 MAIL_USERNAME=null
 MAIL_PASSWORD=null
 MAIL_ENCRYPTION=null
+MAIL_FROM_ADDRESS=null
+MAIL_FROM_NAME="${APP_NAME}"
```
robclancy: since when could you do this? :o

antonkomarev: I don't remember exact date when variables were introduced to dotenv, but we already have couple in the same file:
```
MIX_PUSHER_APP_KEY="${PUSHER_APP_KEY}"
MIX_PUSHER_APP_CLUSTER="${PUSHER_APP_CLUSTER}"
```

PR #5181, opened 2019-12-18
Utilize Authentication Middleware Contract 

taylorotwell:
This change adds the `AuthenticatesRequests` marker interface as the target of the `auth` middleware and binds that interface to the application's `Authenticate` middleware.

The purpose for this change is to allow us to remove the `middlewarePriority` middleware from the application's HTTP kernel. Even though the Foundation's HTTP Kernel contained a base priority, it was not possible to remove this here because we needed to have the Authenticate `middleware` in the priority list and that belongs to the `App` namespace. By targeting the marker interface in the Foundation HTTP Kernel and registering it here, the Foundation's middleware priority list and the application's can now match, making the application's unnecessary.

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -62,21 +62,4 @@ class Kernel extends HttpKernel
         'throttle' => \Illuminate\Routing\Middleware\ThrottleRequests::class,
         'verified' => \Illuminate\Auth\Middleware\EnsureEmailIsVerified::class,
     ];
-
-    /**
-     * The priority-sorted list of middleware.
-     *
-     * This forces non-global middleware to always be in the given order.
-     *
-     * @var array
-     */
-    protected $middlewarePriority = [
-        \Illuminate\Session\Middleware\StartSession::class,
-        \Illuminate\View\Middleware\ShareErrorsFromSession::class,
-        \Illuminate\Contracts\Auth\Middleware\AuthenticatesRequests::class,
-        \Illuminate\Routing\Middleware\ThrottleRequests::class,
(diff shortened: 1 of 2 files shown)


stancl:
@taylorotwell So if a package uses 

```php
$this->app->make(Kernel::class)->prependMiddleware($middleware);
```

what priority will the middleware have? Can it be controlled somehow?

I need my package's priority to be very specific, so what should I do here? Don't use `prependMiddleware()` and instead add the middleware to the specific position in [Http/Kernel's `$middleware` array](https://github.com/laravel/laravel/blob/master/app/Http/Kernel.php#L16)?

stancl:
I think I solved it. I realized I need to set middleware priority for non-global middleware too and noticed that the `$middlewarePriority` array still exists in the base Http Kernel so I'll just add it to the app Http Kernel.

ionutantohi:
Hi @taylorotwell 

In this PR, you changed the target for `auth` middleware. Along with this change the `middlewarePriority` was removed.

But later on, @GrahamCampbell reverted partially this change within #30877. (It changed `auth` middleware to its initial value from App namespace), but the `middlewarePriority` wasn't added back.

Since `middlewarePriority` is not there by default, and also looking at the docs, is not clear for me what we should do if we need a middleware to run before another.

Do we need to copy the whole array from the [docs](https://laravel.com/docs/7.x/middleware#sorting-middleware) and put our middleware in there? Or there is another way?

If this is the case, I find it difficult when we upgrade, because we need to be in sync with the docs and not with laravel/laravel repo.

PR #5184, opened 2019-12-19
Add .DS_Store to .gitignore

MouadZIANI:


GrahamCampbell:
Thanks, but this should be added to your global, system level ignore file.

PR #5185, opened 2019-12-19
[6.x] Configure database connection prefixes with env

zareismail:


GrahamCampbell:
Thanks for the PR. But, my feeling is that we shouldn't include this in the default app, since almost nobody will use this.

zareismail:
> Thanks for the PR. But, my feeling is that we shouldn't include this in the default app, since almost nobody will use this.

Thanks. 
But by this reasoning; I think  `DATABASE_URL` and this `DB_SOCKET`, so many other things; should be removed.

taylorotwell:
I think you can add this to your own application if needed.

PR #5190, opened 2019-12-25
[6.x] Fix consistency of port types in database config

antonkomarev:
Redis databases has integer `port` values, but `mysql`, `pgsql` & `sqlsrv` has string `port` values. This PR replaces string ports with integers.


GrahamCampbell:
Thanks for the PR, but surely this doesn't actually do what you say? In the current implementation, the ports will always be strings, but in your version, they are strings always, unless they are unset, then they are integers.

GrahamCampbell:
e.g. Consider the .env file:

```
DB_PORT=123
```

Then `env('DB_PORT', 3306)` will be the string "123", but if there is no variable, your code now produces an integer, which is not expected. We'd expect always a string, which is why the current implementation has the values quoted.

antonkomarev:
I see, but then we should change Redis databases `port` values. Because they have integer default values.

PR #5192, opened 2019-12-28
[7.x] [OPTION 1] Make parsedown swappable

GrahamCampbell:
Goes with https://github.com/laravel/framework/pull/30973. This PR opts for setting up Parsedown as the default parser.


taylorotwell:
I don't think we should make this swappable. I think we should use one good default. If a better default is now CommonMark I am fine switching to that. I think original I used Parsedown for the table support possibly. But, I see there is an official table extension for CommonMark. I am fine with switching to that if we want to do a smaller PR for that.

PR #5197, opened 2020-01-05
[6.x] Rename phpunit.xml to phpunit.xml.dist

voyula:
This change provide to can override phpunit settings on local workshop.


GrahamCampbell:
Please add `phpunit.xml` to the `.gitignore` file, otherwise this change doesn't make sense.

PR #5198, opened 2020-01-06
Add MAIL_THEME to .env file

maxlen:


GrahamCampbell:
Thanks, but this is not something that is usually an environment configured item. It would be more usual to edit the config file per app.

10mi8o:
unsubscribe

maxlen:
strange. Ok

PR #5199, opened 2020-01-06
Add MAIL_THEME to .env file

maxlen:


maxlen:
If pull request declined - type here why?

GrahamCampbell:
Please see the comment I wrote on your previous PR.

GrahamCampbell:
Laravel's config files are designed to be editied, which is why they are included in the apps, rather than being within the framework. You should edit the config file.

maxlen:
strange. Ok

maxlen:
And maybe we should add dkim to mailer?
If I'll add it to pullRequest - do you add it?

maxlen:
@GrahamCampbell

PR #5200, opened 2020-01-07
[6.x] Comment out sqlite.url config

kawax:
```sh
laravel new sqlite --auth
#edit .env
touch database/database.sqlite
php artisan migrate
```

```
   InvalidArgumentException  : The database configuration URL is malformed.

  at vendor/laravel/framework/src/Illuminate/Support/ConfigurationUrlParser.php:135
    131|
    132|         $parsedUrl = parse_url($url);
    133|
    134|         if ($parsedUrl === false) {
  > 135|             throw new InvalidArgumentException('The database configuration URL is malformed.');
    136|         }
    137|
    138|         return $this->parseStringsToNativeTypes(
    139|             array_map('rawurldecode', $parsedUrl)
```

Disable as default.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -37,7 +37,7 @@ return [
 
         'sqlite' => [
             'driver' => 'sqlite',
-            'url' => env('DATABASE_URL'),
+            //'url' => env('DATABASE_URL'),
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Hmmm, I don't think this is better as a default.

PR #5201, opened 2020-01-08
Use file session driver again

aimeos:
The maximum amount of cookie data per domain is ~4k bytes and this change returns empty data if more data is stored in the session. In our case, the session contains the shopping cart data, which are almost always above the limit


aimeos:
@taylorotwell Please be aware that the default cookie session storage is also a security problem because users can change the session data on their local machine. This can lead to privilege escalation and remote data injection. There are good reasons why session data isn't stored at the client machine by default and that Laravel shouldn't do this by default too regardless of personal preference!

narwy:
I do agree with the default being `file`

taylorotwell:
@aimeos They can't change the cookie on their local machine. They are signed and encrypted.

narwy:
@taylorotwell But can be decrypted if someone gains access to your secret 🤷‍♂

taylorotwell:
If they gain access to your secret, you have big problems regardless of your driver. They can modify your session identifier cookie to another user even with the file driver at that point.

narwy:
@taylorotwell But surely you'd minimise potential risk if it's file based by default.

aimeos:
@taylorotwell OK, then security may not be a big problem. Nevertheless, the maximum limit of all cookies for a domain (not for each cookie) is still 4096 bytes and that can be reached easily. Laravel was always known for defaults that work for 95% of all applications out of the box. Now it may be working for 70-80% only and it's frustrating for users if something doesn't work without a hint why this is the case.

alec-lefors:
This new default throws 419s on all POST requests if the cookie is too big and was very hard to debug. It’s a shame a default setting doesn’t have better error handling other than 419s and the occasional server error when the requests get ridiculously long.

driesvints:
After talking this over with @taylorotwell, we've decided to revert this change. Thanks for all the input on this 👍

PR #5203, opened 2020-01-08
[6.x] Update laravel mix and sass loader

caiquecastro:
Current laravel mix has a vulnerable dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,10 +12,10 @@
     "devDependencies": {
         "axios": "^0.19",
         "cross-env": "^5.1",
-        "laravel-mix": "^4.0.7",
+        "laravel-mix": "^5.0.1",
         "lodash": "^4.17.13",
         "resolve-url-loader": "^2.3.1",
         "sass": "^1.15.2",
-        "sass-loader": "^7.1.0"
+        "sass-loader": "^8.0.0"
     }
 }


taylorotwell:
Ping @JeffreyWay ... need confirmation on this.

GrahamCampbell:
What is the vulnerability, and is it relevant, given that the mix script is not given user input when it is executed?

GrahamCampbell:
Though, regardless of if there is a vulnerability, it doesn't hurt for new apps to get the latest version?

caiquecastro:
The vulnerable package was a version of serialize-javascript ( dependency of laravel-mix#terser-webpack-plugin). I figured out that it's possible to update only the vulnerable dependency. But why not update laravel mix to the latest version?

pavinthan:
@caiquecastro `sass` and `sass-loader` is not necessary to install by default because it's manage with `larave/ui` and @JeffreyWay said 

> It will be automatically installed until Mix 6 is out. You can safely ignore it for now.

Please check
https://github.com/JeffreyWay/laravel-mix/pull/2279#issuecomment-566643040

bArraxas:
Hi, how to publish a bug ?
It concerne the impossibility to invoke ftp_connect or Storage::disk('ftp') inside the app.
You can easy verify that by adding a "ftp_connect('test')" inside any recent project at : 
- bootstrap/app.php (it works)
- public/index.php (before "$response->send();" (line 58) (it works too)
- everywhere other in laravel project (it fails like if php_ftp.dll was unabled)
When i use phpinfo()
at same places, it said all is ok...
Thanks for your answers.

GrahamCampbell:
So, just to be clear, there is **no vulnerability**?

PR #5204, opened 2020-01-10
Flash status message into verified view session

bshehram:
When a new user receives the "Verify Email Address" email and clicks on the included link they should be redirected to the default HOME route with a message injected into the session to display to the user for the HOME view to render. This lets the user confirm the Verify Email Address link worked correctly.

### DIFF
--- a/app/Http/Controllers/Auth/VerificationController.php
+++ b/app/Http/Controllers/Auth/VerificationController.php
@@ -26,8 +26,12 @@ class VerificationController extends Controller
      *
      * @var string
      */
-    protected $redirectTo = RouteServiceProvider::HOME;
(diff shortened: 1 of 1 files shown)


bshehram:
Hey sorry this is my first GitHub PR I'm not sure what's wrong with the styling that caused StyleCI to fail.

marky291:
Not everyone uses flash data when a user is verified.

bshehram:
> Not everyone uses flash data when a user is verified.

Ok, that makes sense. Thanks for the clarification. Just curious but what would be the "correct" way to pass this status message to the view if not the flash() function? I'm new to Laravel.

PR #5207, opened 2020-01-17
[6.x] Use `assertOk()` instead of `assertStatus(200)`

chapeupreto:
Asserting that a response has a 200 status code is a too common task.

`assertOk()` was created in order to make such a task simpler and easier for newcomers, so it would be nice to have this assertion 'enabled' by default.


antonkomarev:
I think `assertStatus(200)` is more explicit and easy to understand without knowledge about framework's implementation.

PR #5208, opened 2020-01-22
[7.x] Adds collision's phpunit adapter

nunomaduro:
This pull request adds of the **collision's phpunit adapter** to the default Laravel 7 skeleton. This adapter provides two things:

- Better console visuals on tests failures
- Always stops on failure

1. Here a simple comparasion with the default phpunit output ( on the left ) against collision ( on the right ).

Screenshot 2020-01-22 at 22 32 09

2. Things get even better on complex examples, as collision focuses on what really matters - hidden vendor frames, and others:

Screenshot 2020-01-22 at 22 59 33

Let me know what do you guys think. 👍

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -19,6 +19,9 @@
             <directory suffix="Test.php">./tests/Feature</directory>
         </testsuite>
     </testsuites>
+    <listeners>
+        <listener class="NunoMaduro\Collision\Adapters\Phpunit\Listener" />
+    </listeners>
     <filter>
         <whitelist processUncoveredFilesFromWhitelist="true">
             <directory suffix=".php">./app</directory>


mpociot:
The adapter, unfortunately, does not respect the "stopOnFailure" setting.
In my example, I have two failing tests. Without the listener, the output looks like this:

image

As you can see, I get both failures.

With the listener added, PHPUnit immediately stops after the first failure:

image

mfn:
It's also kind of weird when your tests within an IDE primarily, e.g. phpstorm.

It's not really necessary there because you can one-click right into the failed test anyway.

Additionally, when multiple test fails it only shows it from a single test (the first?) so not sure how useful.

I think that should be an explicit user choice if they want so.

gocanto:
though it looks awesome and clean, I personally prefer the errors stack.

crynobone:
Anyway to have this skipped for certain request? E.g running from CI shouldn't stop on first error or failure

deleugpn:
As a CLI user, this looks AMAZING. I dont use phpstorm to run my tests, so I do want this today!! Is it compatible with Laravel 6? Is it easy to disable for CI? Our CI env already expects the original output for reporting purposes, but on a day to day developer experience, I never knew I needed this until I saw it!

nunomaduro:
Very valuable feedback so far. I am up to update the listener to meet the needs we all have - so don't hesitate on explaining what you would like have instead. Example: `only activate the listener when the flag  --stop-on-failure is used`.

driesvints:
@nunomaduro might want to update your original comment to reflect that

voyula:
👎

taylorotwell:
@nunomaduro so what is the status here? Have some changes been made? If so, can you update us here on what they were?

nunomaduro:
Closing this pull request. Something even better is coming soon. Thanks for all the valuable feedback. ❤️

PR #5209, opened 2020-01-23
[7.x] add new `check_compiled` option

browner12:
this adds the option for https://github.com/laravel/framework/pull/31206

I am setting the default to check for expired files in non-production environments, and **not** check in production environments, to provide an example for what most people will probably use.  The only danger to this is you **must** run `php artisan view:cache` in deployment.

If you feel the risk it too great to automatically tie this to the environment, we can set the default value as `true`.

### DIFF
--- a/config/view.php
+++ b/config/view.php
@@ -43,6 +43,6 @@ return [
      | and precompiling your views we can skip this check to save time.
      |
      */
-    'check_compiled' => env('APP_ENV') !== 'production',
+    'expires' => env('APP_ENV') !== 'production',
 
 ];
--- a/config/view.php
+++ b/config/view.php
@@ -33,4 +33,16 @@ return [
         realpath(storage_path('framework/views'))
     ),
 
+    /*
+     |--------------------------------------------------------------------------
+     | Check Compiled Views
+     |--------------------------------------------------------------------------
+     |
+     | On every request the framework will check to see if a view has expired
+     | to determine if it needs to be recompiled. If you are in production
+     | and precompiling your views we can skip this check to save time.
+     |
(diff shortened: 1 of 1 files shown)


taylorotwell:
I renamed it to "expires".

browner12:
updated

GrahamCampbell:
I've corrected the target branch.

GrahamCampbell:
Do we really want this by default to behave like this? Maybe a better default would just be `false`.

taylorotwell:
Tweaked it to use an environment variable and be true by default for less breaking changes.

GrahamCampbell:
👍

browner12:
Thanks for fixing the branch. Is there a reason we don't do branches in this repo the same as framework?

Thanks for switching to a straight up environment variable. I was on the fence about tying it to a specific environment, and the new way will probably cause less headaches.

GrahamCampbell:
Because it makes sense for master to be the "current stable" in the default project, and for master to be the "upcoming version" in the framework itself. If we changed changed master to be 7.x in the default app, the result would be people downloading it instead of 6.x.

browner12:
got it, thanks!

driesvints:
The exact reason is because it's current fixed in the installer command when you run `--dev` and it would break for anyone who has installed it locally if we changed it:

https://github.com/laravel/installer/blob/master/src/NewCommand.php#L221

We'd also always have to release a new version of the installer with the exact branch name if we did a ew major release before people could install it. Now it's just universal: "master" & "develop".

browner12:
thanks for the response @driesvints

PR #5212, opened 2020-01-27
Update cross-env

MichaelDeBoey:
This also sets the minimum Node version to 10.14

https://github.com/kentcdodds/cross-env/releases/tag/v7.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,11 +11,16 @@
     },
     "devDependencies": {
         "axios": "^0.19",
-        "cross-env": "^6.0",
+        "cross-env": "^7.0",
         "laravel-mix": "^5.0.1",
         "lodash": "^4.17.13",
         "resolve-url-loader": "^3.1.0",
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks, but please separate these two changes.

MichaelDeBoey:
@GrahamCampbell Dropping Node 8 is a consequence of updating `cross-env`

driesvints:
@MichaelDeBoey we weren't enforcing this before and afaik can't cross-env require this itself?

MichaelDeBoey:
@driesvints `cross-env` is enforcing this.
So since it's a dependency of this project, we should do to here.

driesvints:
@MichaelDeBoey if the library is enforcing it itself I don't see why we should enforce it twice.

MichaelDeBoey:
@driesvints So people can see it directly that the minimum supported Node version is Node 10.14

PR #5213, opened 2020-01-28
[6.x] Rename .htaccess to .htaccess.example

voyula:
This change provides different enviroments capacity for Web Server configuration.

Example, if we want SSL redirect on production enviroment, PHP based SSL redirect not working on all requests, like image link etc.

This PR will be very helpful for configurate web server for different enviroments(development, production, etc.).


.gitignore
```diff
@@ -1,4 +1,5 @@
 /node_modules
+/public/.htaccess
```
GrahamCampbell: I don't think this is a good idea? People will want to commit this in their apps. surely?

voyula: This a locale configuration file like .env file. Their can write all code into .htaccess.example file and their can use # comment character for use needed commands, like ssl redirect.

taylorotwell:
Feel free to rename this in your own application.

PR #5214, opened 2020-01-29
[6.x] Update Slack default level

browner12:
make this consistent with all the other default levels.

it can also be really confusing when initially setting up your Slack logging, if a test doesn't come through because it doesn't meet the minimum critical level, especially if you're not familiar with this feature.

### DIFF
--- a/config/logging.php
+++ b/config/logging.php
@@ -59,7 +59,7 @@ return [
             'url' => env('LOG_SLACK_WEBHOOK_URL'),
             'username' => 'Laravel Log',
             'emoji' => ':boom:',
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Maybe warn would be a better default?

browner12:
I feel like we should keep it at the lowest level, and allow users to "opt in" to using this feature if they don't want Slack logs for certain levels.

taylorotwell:
Hmm, I'm not sure I agree. I tried to set the most conventional production ready setting I could think of. I don't think you would typically pass every log statement to Slack but only serious errors.

PR #5215, opened 2020-01-29
[6.x] Add redirect from /index.php to /

DTse:
I noticed index.php showing up in Google Analytics for a project i am working on. 

The project is a news portal currently in production and used by hundreds of users every day.
Index.php links are showing up every day in our Google Analytics and the laravel project doesn't seem to be able to rewrite them.

I used the proposed solution and is looking to work great.

Our setup is a Linux/Apache server.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -18,4 +18,8 @@
     RewriteCond %{REQUEST_FILENAME} !-d
     RewriteCond %{REQUEST_FILENAME} !-f
     RewriteRule ^ index.php [L]
+    
+    # Handle Redirect index.php...
+    RewriteCond %{THE_REQUEST} ^GET.*index\.php [NC]
+    RewriteRule (.*?)index\.php/*(.*) /$1$2 [R=301,NE,L]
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
What if someone has a route that actually matches index.php in it?

DTse:
> What if someone has a route that actually matches index.php in it?

Why would someone want that? The popular usage is without file extensions in the urls.

In any case the default behaviour should be the rewrite of index.php and if there is an edge case like that the user can remove it or comment it out.

It would be far easier and with wider use cases than the opposite. If you do a Google search you'll find many topics about this issue.

GrahamCampbell:
> Why would someone want that? The popular usage is without file extensions in the urls.

I mean, what if their actual route path happens to contain index.php, like /foo/index.php/bar. Like, it's a perfectly valid use case to have something like /foo/{example}/bar, which can match something like that.

taylorotwell:
You are free to change your own .htaccess file as needed. It exists in the skeleton.

PR #5217, opened 2020-01-30
[6.x] Specifying the minimum version of nodejs to require `cross-env`

andrey-helldar:
`cross-env` version update merged: #5216 
This PR contains an indication of the [minimum version](https://github.com/kentcdodds/cross-env/blob/master/package.json#L10-L14) of nodejs required for `cross-env` to work (also see https://github.com/laravel/laravel/pull/5212#issuecomment-578822212).

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,11 @@
         "prod": "npm run production",
         "production": "cross-env NODE_ENV=production node_modules/webpack/bin/webpack.js --no-progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js"
     },
+    "engines": {
+        "node": ">=10.14",
+        "npm": ">=6",
+        "yarn": ">=1"
+    },
     "devDependencies": {
         "axios": "^0.19",
         "cross-env": "^7.0",


taylorotwell:
I'm not sure I want to have more version constraints in here to maintain.

andrey-helldar:
@taylorotwell, well, then, it seems to me, you need to roll back the changes from #5216 because:
```
cross-env 5.2 require node >= 4.0
cross-env 6.0 require node >= 8.0
cross-env 7.0 require node >= 10.14
```

In #5216, changes were adopted from version 6 to version 7. Should I create a PR return to version 6?

driesvints:
@andrey-helldar but npm will warn anyone who attempts to install v7 with a lower NPM version than 10.14? Why should we keep these constraints ourselves?

andrey-helldar:
@driesvints, I think I did not immediately understand what Taylor wrote. This is a language barrier problem.

I agree, an additional check does not need to be introduced, but I saw Graham’s https://github.com/laravel/laravel/pull/5212#issuecomment-578822212 and decided to create this PR.

PR #5219, opened 2020-02-02
[6.x] Remove "Package Service Providers" section from config/app.php

imliam:
Since automatic package discovery is the norm now, I'm not sure that this section in `config/app.php` is really necessary as a placeholder anymore. Just a small tidy-up.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -163,10 +163,6 @@ return [
         Illuminate\View\ViewServiceProvider::class,
 
         /*
-         * Package Service Providers...
-         */
-
-        /*
          * Application Service Providers...
          */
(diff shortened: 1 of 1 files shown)


mfn:
The assumption is that people deliberately disabling this (remember: it is opt-out after all) OTOH know where to register their service provider manually. _I guess this makes sense_.

iamwebwiz:
> Since automatic package discovery is the norm now, I'm not sure that this section in `config/app.php` is really necessary as a placeholder anymore. Just a small tidy-up.

I also think this makes good sense 🤓

taylorotwell:
Some packages are still manually registered and I think this is a good indicator of where to put them.

PR #5223, opened 2020-02-06
[6.x] Alias for Unit Example TestCase

edwardsteven:
Adding an alias for PHPUnit/Framework/TestCase to make it clear that the TestCase being used isn't the Framework provided TestCase to remove ambiguity for users where IDE's collapse use statements, on the back of the changes made in PR #5169

### DIFF
--- a/tests/Unit/ExampleTest.php
+++ b/tests/Unit/ExampleTest.php
@@ -2,9 +2,8 @@
 
 namespace Tests\Unit;
 
-use PHPUnit\Framework\TestCase as BaseTestCase;
 
-class ExampleTest extends BaseTestCase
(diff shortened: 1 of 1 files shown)


crynobone:
might as well change it to

```php
class ExampleTest extends \PHPUnit\Framework\TestCase
```

if the current structure isn't clear.

edwardsteven:
Good point @crynobone, I've updated this.

arcanedev-maroc:
The issue is within your IDE and not the code.

To solve it: https://stackoverflow.com/a/30844576

Bonus: https://stefanbauer.me/articles/my-phpstorm-settings-after-8-years-of-use

PR #5224, opened 2020-02-06
[6.x] Should default to 'default' to work with Horizon

CedNet:
When adding Horizon to a new Laravel project it will throw an Exception during composer require on Horizon.php line 100.
Since the Laravel framework defaults to 'redis' for cluster name and Horizon uses that very cluster name variable. It won't find database.redis.redis and throw an Exception ("Redis connection [] has not been configured.") because it tries to look up database.redis.redis (while it's actually database.redis.default)

Can be solved either by changing this line here. Or change the example .env file to include REDIS_CLUSTER=default

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -122,7 +122,7 @@ return [
         'client' => env('REDIS_CLIENT', 'phpredis'),
 
         'options' => [
-            'cluster' => env('REDIS_CLUSTER', 'redis'),
+            'cluster' => env('REDIS_CLUSTER', 'default'),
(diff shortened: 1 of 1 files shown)


themsaid:
I just installed a fresh Laravel app and added Horizon. Got no errors. How can I replicate?

CedNet:
You are indeed right. I cannot reproduce at this moment. Must've been some local cache issue (Windows). You may close and even remove this PR thx!

PR #5226, opened 2020-02-07
[7.x] Adds support to PHPUnit 9

nunomaduro:
This pull request adds support to PHPUnit 9. We can't drop the support PHPUnit 8 as PHPUnit 9 don't support php 7.2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "fzaninotto/faker": "^1.9",
         "mockery/mockery": "^1.3.1",
         "nunomaduro/collision": "^4.1",
-        "phpunit/phpunit": "^8.5"
+        "phpunit/phpunit": "^8.5|^9.0"
     },
     "config": {
         "optimize-autoloader": true,


GrahamCampbell:
👎 The default app should either 8.x or 9.x, but not both.

GrahamCampbell:
And it can't have 9.x because we'd have to drop php 7.2.

taylorotwell:
@GrahamCampbell does our laravel/framework code work with PHPUnit 9?

GrahamCampbell:
Yeh, it does, but it's kinda odd to have the default app have two possible versions.

nunomaduro:
@GrahamCampbell Why is that? I think is better this than having everyone changing this manually after installing Laravel.

One problem we may have, is the `composer.lock` with phpunit 9 on the Laravel installer and consequently people having errors just like this: https://github.com/laravel/installer/issues/105.

GrahamCampbell:
Anyone using php 7.2 will not want to make this change, and the laravel/laravel app is aimed at 7.2?

GrahamCampbell:
> One problem we may have is the composer.lock with PHPUnit 9 on the Laravel installer and consequently, people having errors just like this: laravel/installer#105.

Isn't that an argument against merging this PR?

deleugpn:
> Anyone using php 7.2 will not want to make this change, and the laravel/laravel app is aimed at 7.2?

Aimed at or minimum requirement? I don't see minimum requirement as **aimed at** at all. In fact, people with a newer php version will have laravel installing a newer version of phpunit automatically, which seems like a good thing.

driesvints:
We have been upgrading minimum versions on the skeleton all the time the past years so updating it to a minimum of PHPUnit 9 is okay. It's best to prompt people to use newer major versions. If people are on 7.2 they can downgrade to PHPUnit 8 if they want.

However Graham is right that it's either one or the other but not both as that doesn't makes a sense from an app point of view.

GrahamCampbell:
I think keeping the default at phpunit 8 is fine, for Laravel 7. We could change it in Laravel 8?

GrahamCampbell:
What's so good about PHPUnit 9 that everyone is gonna need it instead of 8?

GrahamCampbell:
Looks like the only new feature so far is https://github.com/sebastianbergmann/phpunit/pull/3797.

driesvints:
@GrahamCampbell moving people to a new major version sooner rather than later helps them migrate easier in the future.

driesvints:
Remember that the skeleton is meant for new apps. Laravel 7 will still perfectly support PHPUnit 8 alongside v9.

deleugpn:
@driesvints correct me if I'm wrong, but if it's **PHPUnit 9 only**, people with PHP 7.2 will get an error while trying to run `composer create-project laravel/laravel`, which would not be a good DX? 

I think the PR looks good as it would automatically default to the best option depending on the PHP version used to install Laravel, but if the stance of having a single-version only in the App has to hold true, then Graham is correct that it would have to be PHPUnit 8 due to support for PHP 7.2, unless I'm wrong in the installation thing.

I don't look at this as what's good about the new version but rather about helping improve the culture of keeping things up to date in our industry. If a brand new project is being installed, it has no drawbacks to be upgraded to the latest version. Millions of projects are installed with the defaults of Laravel and promoting an up to date setup seems like a good stance.

driesvints:
@deleugpn that's a fair point actually. We'll see what Taylor says.

GrahamCampbell:
> Correct me if I'm wrong, but if it's PHPUnit 9 only, people with PHP 7.2 will get an error while trying to run composer create-project laravel/laravel, which would not be a good DX?

That's exactly what I meant when I said:

> the laravel/laravel app is aimed at 7.2

taylorotwell:
I'll just use 9.0 only.

nunomaduro:
Also works for me.

driesvints:
After talking to Taylor I've implemented the changes you proposed here @nunomaduro. Contrary to what I said before this is probably the best way to go for now. We'll update to to ^9.0 only for Laravel 8

driesvints:
After talking to Taylor we've decided to revert back to ^8.5 for now so people on PHP 7.2 will also be able to install the skeleton.

PR #5232, opened 2020-02-20
[6.x] Use the server superglobal for app base path

GrahamCampbell:
Replaces https://github.com/laravel/laravel/pull/5231.


taylorotwell:
They can change this in their own application if they need it.

PR #5139, opened 2019-10-16
Add xml schema to phpunit

BackEndTea:
This allows an IDE to do auto completion, and show any errors in the configuration

PR #5233, opened 2020-02-23
[6.x] Remove redundant default attributes from phpunit.xml

SjorsO:
PR https://github.com/laravel/laravel/pull/5139 added the xml schema to `phpunit.xml`. Because of this, IDEs like PhpStorm now show which attributes are redundant:
image

All the gray attributes in the screenshot above are setting the same value as the default value. They can be removed.

The `suffix="Test.php"` is also a redundant default, but I think this is useful in reminding people their test classes should end with `Test.php`

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,15 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd"
-         backupGlobals="false"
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Kinda stupid IDEs would warn about this. We included these options so people don't have to head to Google to know what the options are called to change them.

SjorsO:
> Kinda stupid IDEs would warn about this. We included these options so people don't have to head to Google to know what the options are called to change them.

I've never had to change any of these values, I actually don't know what most of these are supposed to do. Removing them would clean up some unnecessary noise in this file.

GrahamCampbell:
Maybe keep `stopOnFailure` in there, at least?

PR #5237, opened 2020-02-27
[7.x] Add Http facade

driesvints:


GrahamCampbell:
Looks like GitHub's API is down again, which is why the status update didn't get through.

PR #5241, opened 2020-03-03
Update UserFactory to use Hash facade

JustSteveKing:
This is a bug bear of mine, the default factory assumes that the user is using the `bcrypt` helper - where as with the configuration the ability to change the hashing driver is available. Moving to the Hash Facade here allows for this to be reflected in the default factory.

*note*: This isn't a huge problem for people to change, but if it is available why not use it?


crynobone:
`composer.lock` shouldn't be part of this PR

JustSteveKing:
@crynobone you are absolutely right there!

driesvints:
Heya thanks but we specifically don't use the hashing here to speed up the factories. Feel free to do this in your own project!

JustSteveKing:
No problems @driesvints 👍

PR #5242, opened 2020-03-04
Remove unused import statements

sangnguyenplus:


GrahamCampbell:
Thanks, but these are intentionally unused, to help people get started. This is why we have turned off the unused import fixer of StyleCI on laravel/laravel.

PR #5248, opened 2020-03-04
delete comment code

arshaparthia:


taylorotwell:
This is there for convenience when uncommenting it if you want the feature.

PR #5251, opened 2020-03-05
[7.x] Make use of fluent String

mercuryseries:
Why not :) ?


Braunson:
Because rather than instantiate a class to do a replace it's quicker and smaller to use the native function `str_replace` :)

driesvints:
Thanks but we're gonna keep the current way.

PR #5252, opened 2020-03-06
[7.x] Add new SQS queue suffix option

roberto-aguilar:
As described in laravel/framework#31784, this option will allow you to define a queue name suffix.

This can be useful in case you have multiple queues in multiple environments like in Laravel Vapor for example:

```yml
environments:
    production:
        queues:
            - emails-production
            - imports-production
    staging:
        queues:
            - emails-staging
            - imports-staging
```

So if you define your `SQS_SUFFIX` as either `production` or `staging`, your code can now look like this:

```php
dispatch(new SendWelcomeEmail($user))->onQueue('emails');
dispatch(new ImportEmailsCsv($path))->onQueue('imports');
```

or even like this:

```php
class SendWelcomeEmail implements ShouldQueue
{
    /**
     * The name of the queue the job should be sent to.
     *
     * @var string|null
     */
    public $queue = 'emails';
}
```

```php
class ImportEmailsCsv implements ShouldQueue
{
    /**
     * The name of the queue the job should be sent to.
     *
     * @var string|null
     */
    public $queue = 'imports';
}
```

and the queue name will be resolved appropriately.

### DIFF
--- a/config/queue.php
+++ b/config/queue.php
@@ -55,6 +55,7 @@ return [
             'secret' => env('AWS_SECRET_ACCESS_KEY'),
             'prefix' => env('SQS_PREFIX', 'https://sqs.us-east-1.amazonaws.com/your-account-id'),
             'queue' => env('SQS_QUEUE', 'your-queue-name'),
+            'suffix' => env('SQS_SUFFIX'),
             'region' => env('AWS_DEFAULT_REGION', 'us-east-1'),
         ],
 


alnutile:
I know this is old but just curious seems with conventions it would have been just as easy to do

```
environments:
    production:
        queues:
            - emails-${APP_ENV}
            - imports-${APP_ENV}
    staging:
        queues:
            - emails-${APP_ENV}
            - imports-${APP_ENV}
```

or

```
        'sqs' => [
            'driver' => 'sqs',
            'key' => env('AWS_ACCESS_KEY_ID'),
            'secret' => env('AWS_SECRET_ACCESS_KEY'),
            'prefix' => env('SQS_PREFIX', 'https://sqs.us-east-1.amazonaws.com/your-account-id'),
            'queue' => env(sprintf("%s-%s", 'SQS_QUEUE', env("APP_ENV")), 'your-queue-name'),
            'region' => env('AWS_DEFAULT_REGION', 'us-east-1'),
        ],
```

no big deal but this is how we have done it for years since we do have multiple queues per environment named with the APP_ENV name to help with access etc.

PR #5253, opened 2020-03-07
Calling home route by name

iamine:
Calling home route by name instead of calling by url


driesvints:
There's no route named like that in the skeleton so this won't work.

PR #5255, opened 2020-03-08
[6.x] App Debug Bool

ericdowell:
Ensure that app.debug is a bool using a strict compare.

This will make it more strict when an invalid string is used, such as `'test'` or `'on'` or any string that is not `'true'` or `'false'`. 

This is safer if somehow `app.debug` is deployed to an environment with an invalid string and `app.debug` should never have a value of `true`.

String values of `'true'`/`'false'` get converted to a boolean value by the `env` helper. See switch in [Illuminate\Support\Env](
https://github.com/laravel/framework/blob/6995c7c6e5e6ea2851fdbc1f8b2afc4590d9b84c/src/Illuminate/Support/Env.php#L83-L96)

Initial PR and discussion: https://github.com/laravel/laravel/pull/5254

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -39,7 +39,7 @@ return [
     |
     */
 
-    'debug' => (bool) env('APP_DEBUG', false),
+    'debug' => env('APP_DEBUG', false) === true,
 
     /*
     |--------------------------------------------------------------------------


GrahamCampbell:
This is actually a breaking change. The original code was just silently casting to a bool. Now it is explicit. It we really want to change it to be `=== false`, this can be done in 7.x or 8.x.

ericdowell:
@GrahamCampbell Thanks, sounds good. I will open a PR against `master`. 

I'm not sure how `=== false` would be what we want to change this to.

My goal is protecting against a value of `true` happening in deployed environments where it should not happen.

Being explicit with `=== true` makes sure this does not happens, while comparing with `!== false` allows for values such as `'off'` or `0` for `app.debug` to equal `true`.

PR #5256, opened 2020-03-08
[7.x] App Debug

ericdowell:
Comparing the value to `true` fixes possible issues with casting non-empty strings. For example, `(bool)'off'` will equal `true`.

```
APP_DEBUG=off
# OR
APP_DEBUG=false
```
The above `.env` example results in `config('app.debug')` equal to`false`.
```
APP_DEBUG=true
```
The above `.env` example results in `config('app.debug')` equal to `true`.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -39,7 +39,7 @@ return [
     |
     */
 
-    'debug' => (bool) env('APP_DEBUG', false),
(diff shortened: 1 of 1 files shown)


taylorotwell:
No plans to change this.

PR #5259, opened 2020-03-11
[7.x] Update default CORS config

barryvdh:
`exposed_headers` can be `false` or an array (not `true`). `max_age` can be false or an integer. 
There defaults make it more obvious.

### DIFF
--- a/config/cors.php
+++ b/config/cors.php
@@ -25,9 +25,9 @@ return [
 
     'allowed_headers' => ['*'],
 
-    'exposed_headers' => false,
+    'exposed_headers' => [],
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Does this need fixing on 6.x too?

barryvdh:
It's not a fix perse, because the config worked. They just can't pass true in there.
Should I target 6.x instead?

GrahamCampbell:
If this is the new package default config file, it's probably worth having 6.x LTS ship with the latest config.

barryvdh:
So should I create a new PR for 6.x?

PR #5261, opened 2020-03-13
Fix session config changes

mathieutu:
Hi,
Several changes reverted from #5157 (Symfony 5 update):
- Remove the null explicit default as env is null by default,
- Add null value in phpdoc as null is a totally valid value, (which is used by Symfony).
- Re-set null by default for same-site configuration, because as the phpdoc says itself "**By default, we do not enable this**", and I didn't find any clue of why this has be done, as it seems that nothing change for Laravel here in Symfony 5 update (new default for symfony Cookie, but Laravel set the value, so do not use the default see https://github.com/symfony/symfony/blob/5.0/src/Symfony/Component/HttpFoundation/Cookie.php#L91 and https://github.com/laravel/framework/blob/5c2c1df2d9d56f761e9c6352db067eb78426da62/src/Illuminate/Session/Middleware/StartSession.php#L157).

I'm doing an other PR to the upgrade guide in the doc to update that.

Thanks,
Matt'

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -194,6 +194,6 @@ return [
     |
     */
 
-    'same_site' => null,
+    'same_site' => 'lax',
 
 ];
--- a/config/session.php
+++ b/config/session.php
@@ -166,7 +166,7 @@ return [
     |
     */
 
-    'secure' => env('SESSION_SECURE_COOKIE', null),
+    'secure' => env('SESSION_SECURE_COOKIE'),
 
     /*
     |--------------------------------------------------------------------------
@@ -190,10 +190,10 @@ return [
     | take place, and can be used to mitigate CSRF attacks. By default, we
     | do not enable this as other CSRF protection services are in place.
     |
-    | Supported: "lax", "strict", "none"
+    | Supported: "lax", "strict", "none", null
     |
     */
 
-    'same_site' => 'lax',
+    'same_site' => null,
 
 ];


taylorotwell:
Keeping same-site lax as it is a more secure default for session cookies IMO.

GrahamCampbell:
Just to clarify, this PR actually changes nothing at all...

GrahamCampbell:
Nothing was reverted, only a redundant explicit default was removed.

mfn:
> Nothing was reverted, only a redundant explicit default was removed.

Does that mean the entry in the upgrade guide _may_ not necessary?

https://laravel.com/docs/7.x/upgrade#symfony-5-related-upgrades
> Next, please update your session configuration file's secure option to have a fallback value of null:

thx

GrahamCampbell:
> Nothing was reverted, only a redundant explicit default was removed.

No I mean that this PR did not revert anything, not that nothing was changed between 6 and 7.

GrahamCampbell:
That line is still needed. `env('SESSION_SECURE_COOKIE', null)` is the same as `env('SESSION_SECURE_COOKIE')`, so this PR can be viewed as just a refactoring.

mathieutu:
@GrahamCampbell The PR doesn't revert anything and is just about documentation now, because of the Taylor commit, and you know it. My PR was not about refactoring this little env default, but reverting the 'lax' default value.

I understand the wish to keep it: the lax value is following a new security choice, but it apparently wasn't an upgrade need (you know, editing my message and crossing the point doesn't make it less relevant).

So please, do not feel hurt about that, you made an amazing job updating theses packages, so no need to ruin everything by being rude 🙂.

Thanks for your work and see you on the next PR 😉.

GrahamCampbell:
My comment wasn't directed at your, and wasn't meant to be rude. It was meant to help people who landed on this PR and got confused by the PR title. ;)

PR #5262, opened 2020-03-14
[7.x] Allow configuring the timeout for the smtp driver

mfn:
The default is the same as in `\Swift_Transport_EsmtpTransport::$params`

Complements https://github.com/laravel/framework/pull/31973


GrahamCampbell:
I've corrected the default, so that old config files don't have different behaviour.

mfn:
🤦‍♀️ much appreciated, thanks!

PR #5263, opened 2020-03-14
[7.x][6.x] Remove class constant visibility

rodruiz:
Class constant visibility can be used in PHP 7.1 only.
Laravel 7: PHP >= 7.2.5
Laravel 6: PHP >= 7.2.0


GrahamCampbell:
Class constant visibility can be used in PHP `>= 7.1.0`.

PR #5265, opened 2020-03-17
Handle JSON requests correctly in RedirectIfAuthenticated middleware

RiadhKHEDHIRI:
The redirect didn't satisfy the CORS policy as 'Access-Control-Allow-Origin' was missing.

### DIFF
--- a/app/Http/Middleware/RedirectIfAuthenticated.php
+++ b/app/Http/Middleware/RedirectIfAuthenticated.php
@@ -19,7 +19,9 @@ class RedirectIfAuthenticated
     public function handle($request, Closure $next, $guard = null)
     {
         if (Auth::guard($guard)->check()) {
-            return redirect(RouteServiceProvider::HOME);
+            return $request->wantsJson()
+                ? new Response('', 204)
+                : redirect(RouteServiceProvider::HOME);
         }
 
         return $next($request);


driesvints:

RiadhKHEDHIRI:
The scenario I'm facing is when a user attempt to authenticate while he is authenticated already: the response is then a `302` (to `/home`) without the header `Access-Control-Allow-Origin` causing a CORS error. I thought this PR might save some time and headaches.

There is clearly an issue here as the redirect process is not handling JSON correctly, if it is not in the middleware it must be deeper then.

Maybe is there a cleaner way to address this issue. Any idea please?

JorisDebonnet:
I absolutely agree; I've run into this issue for weeks.

I have a Single Page App which uses Sanctum to login, and if you (for whatever reason) try to login again while the backend still has your active session, it will "redirect" your XHR request to /home, which makes no sense. It took quite some time for me to figure out that this is where the issue was.

Bottom line: the 'guest' middleware does not support XHR requests, and that feels like an oversight (because the `AuthenticatesUsers` and `RegistersUsers` traits, for example, do). 

@driesvints, do you have any ideas of how to address this issue, if not with something like this PR?

driesvints:
@JorisDebonnet no sorry. Feel free to do this in your own app.

djallits:
@driesvints I have no problem with pushing these off onto developers to do within their own apps.

I guess my question is with the suggested solution, is that not the expected behavior out of the box for vanilla Laravel app? By including this, does it break expected behavior? I am not questioning your judgement here, just trying to understand it so I actually learn something here.

maxidirienzo:
@djallits, I believe HTTP code 204 is not the correct return code in this case. 

You are getting a 302 instead of a 401, that would be enough for the client to handle the reply accordingly.

JorisDebonnet:
I agree that a 204 status code may not be the most ideal message. However, a 302 cannot be handled in javascript as the browser handles it immediately, without returning to javascript. So this could turn into an interesting discussion what the status code should be then.

I believe this middleware's primary intention is to prevent a logged in user from going to a page that is meant for non-logged-in users. So if they try that, as a graceful solution, it redirects them to the home page.

So the response, I believe, should be something like "this page is not meant for you, go somewhere else". This is indeed not 2xx ("okay"), and semantically perhaps 3xx (redirect) would make most sense, but is not supported to be handled in javascript. So I would say we'd have to classify it as a 4xx client error. Maybe 400 (bad request),  401 (unauthorized) or 403 (forbidden). It may sound a bit weird to say 401 or 403 for a logged in user, because security isn't really the issue, so I'd say 400 may make the most sense.

JorisDebonnet:
For anyone ending up here, here's how I updated my `app/Http/Middleware/RedirectIfAuthenticated.php`:
```php
    public function handle($request, Closure $next, $guard = null)
    {
        if (Auth::guard($guard)->check()) {
            if ($request->expectsJson()) {
                return response(['message' => 'You are already authenticated.'], 400);
            } else {
                redirect(RouteServiceProvider::HOME);
            }
        }

        return $next($request);
    }
```

I then proceeded to make more changes that make sense for my application, e.g. if this middleware is hit when someone is trying to post to /login or /register, I quickly log out the current user first (invalidate session etc) and then allow the login/register request to proceed. This sounds sensible to me in general, but maybe that would be a bit much as default behavior. A clear error message saying "You are already authenticated" is already much better than a redirect whose reason developers need to track down 😊

PR #5266, opened 2020-03-24
[7.x] Let developers adjust the sendmail path

redelschaap:
The sendmail path can differ between environments. So a hard coded value for this does not work well.

We, for example, test our projects on an environment that require the `-bs` flags, while our production environments require the `-ti` flags. In the current situation we have to adjust the `config/mail.php` file for every project.

It would be a lot easier when the sendmail path could be managed with an `.env` variable.

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -58,7 +58,7 @@ return [
 
         'sendmail' => [
             'transport' => 'sendmail',
-            'path' => '/usr/sbin/sendmail -bs',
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Most people won't want to change this, so 👎 for including this environment variable in the default app.

redelschaap:
I think most people won't change encryption either, but it's still an env variable :)

PR #5269, opened 2020-03-24
Use correct TestCase class

diogogomeswww:


nunomaduro:
That's intentional. Unit tests extend the `PHPUnit\Framework\TestCase` class. Integration tests extend `Tests\TestCase`.

diogogomeswww:
@nunomaduro can you elaborate on that?

I use Unit tests for testing a single class, for example, a Model.
We don't have access to model factories if we extend from `PHPUnit\Framework\TestCase`, right?

driesvints:
Please see https://github.com/laravel/laravel/pull/5169.

If you need model factories feel free to change this in your own codebase or better yet, write a feature test.

PR #5270, opened 2020-03-27
[6.x] Fix default disks public url for php artisan serve

senooken:
By default `config/filesystems.php`, `\Storage::disk('public')->url('')` returns `APP_URL/storage/` for public disk access.

Default `APP_URL` has no port number. So if developers work app on `php artisan
serve` (`http://localhost:8000/`), it returns
`http://localhost/storage/` without port number. And link of resource is
dead link.

I realized that it cannot use kind of url helper functions (url, asset..) in config file for Laravel life cycle.

So I simply removed `env('APP_URL').` for fixing this default dead link.

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -51,7 +51,7 @@ return [
         'public' => [
             'driver' => 'local',
             'root' => storage_path('app/public'),
-            'url' => env('APP_URL').'/storage',
+            'url' => '/storage',
             'visibility' => 'public',
         ],
 


driesvints:
You need to define the port number in your app url.

senooken:
Hi. Only specifying port number in APP_URL is not enough. For example following case.

1. Specify `APP_URL=http://localhost:8080` in .env.
2. Run `php artisan serve`
3. `\Storage::disk('public')->url('')` returns `http://localhost:8080/storage`
4. But current URL is `http://localhost:8000` and *dead links are not solved*.

Of course, it is possible to specify both of `APP_URL=http://localhost:8080` and `php artisan serve --port=8080`.

But these are bother and careless.

I think this fix has no bad point. Absolute path and root path are almost same context in same application.

Do you have something bad point for this fix? Would you tell me for studying?

PR #5280, opened 2020-04-16
[7.x] Change Hash::make to static password

donyahmd:


GrahamCampbell:
What does this fix please?

NickOldfield:
While it's very minor, I do agree.  Why would you pass through an already hashed password?

driesvints:
Please see previous attempts.

PR #5283, opened 2020-04-22
[8.x] - Add target blank to the welcome view links

chuckrincon:


taylorotwell:
Not needed I don't think.

PR #5289, opened 2020-04-27
Add default HELO configuration to mail config file

adamhopkinson:
I've just started using HELO by @mpociot / @beyondcode, and it feels like if Laravel has default connections to other external paid mail development tools, it should probably have HELO support in there too.

This config allows the `env` file to use HELO just by specifying `MAIL_MAILER=helo`.

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -44,6 +44,16 @@ return [
             'timeout' => null,
         ],
 
+        'helo' => [
+            'transport' => 'smtp',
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure this makes sense right now. The other mail services in the configuration file are production providers.

adamhopkinson:
Ah, oh - that makes total sense. I'd (mistakenly) thought some of them were dev tools.

_Amazing_ turnaround speed there @taylorotwell!

PR #5290, opened 2020-04-28
[7.x] Added env variables to config/database.php for charset, collation and…

Screaminlean:
… prefix, I find that I require the change with every project I make, so maybe handy for others.

It is only a minor change.


config/database.php
```diff
@@ -52,9 +52,9 @@
             'username' => env('DB_USERNAME', 'forge'),
             'password' => env('DB_PASSWORD', ''),
             'unix_socket' => env('DB_SOCKET', ''),
-            'charset' => 'utf8mb4',
-            'collation' => 'utf8mb4_unicode_ci',
-            'prefix' => '',
+            'charset' => env('DB_CHARSET', 'utf8mb4'),
```
GrahamCampbell, line 55: This should be added for the other connections too, and not just the first one.

Screaminlean: Ok will update the other connections.

Screaminlean: Done added to commit, please let me know if I did this the right way as it is my first pull request!

Screaminlean:
I made the changes requested, did I complete the review process correctly?

Screaminlean:
My mistake, reopening!

taylorotwell:
Since I feel it is uncommon to develop locally and in production with different character sets, I will let people add this to their own application if needed.

Screaminlean:
> Since I feel it is uncommon to develop locally and in production with different character sets, I will let people add this to their own application if needed.

Hi the change would have no impact if users do not want to change anything as the fallback values are the same as the current ones, just make life a little more convenient, is that not what the whole laravel / artisan thing is about? would you reconsider the PR?

PR #5291, opened 2020-04-28
Update User.php

larismendi-tera:
What is the use of this line?


Tjoosten:
It's there for easily extending the authentication system. When u want to let users verify their account on your application. Correct me if i'm wrong @driesvints

PR #5292, opened 2020-04-30
Match default BROADCAST_DRIVER in .env.example with broadcasting config

justenh:
On a fresh install, most of the example env variables are set to the default value within their config counterpart. However, `BROADCAST_DRIVER` defaults to null, but begins as "log" in the `.env.example` file.

After reading a post regarding cleaning up unused variable declarations from the .env file, I think it would make sense that this starts as `null` (same as the broadcasting config), or that the broadcasting config be updated to `file`.


taylorotwell:
I'm kind of fine with things as they are. I think it makes sense for starting a new local development project - but if you deploy to production and there is no BROADCAST_DRIVER, you want to use `null` - which is what the current configuration will do.

PR #5294, opened 2020-05-01
[7.x] Add mix_url to config/app.php

ankurk91:
Laravel Mix allows to configure the base URL for assets.
It is documented here.

https://laravel.com/docs/7.x/mix
Look for heading `Custom Mix Base URLs`


taylorotwell:
I think this is probably a pretty rare option so I didn't really want to clutter the configuration file with it since it is already documented.

PR #5295, opened 2020-05-01
[7.x] Remove forceTLS option

dwightwatson:
Looks like the [latest version of Pusher](https://github.com/pusher/pusher-js/blob/master/CHANGELOG.md#600-2020-04-27) now defaults this option to true, so I don't think the configuration is required in the boilerplate anymore.


GrahamCampbell:
👎 here. Fine to leave it in, especially so people can see it is an option if they want to turn it off.

PR #5300, opened 2020-05-07
Autogenerate key for tests

finagin:
nunomaduro/collision:4.1.3  and later no longer generates `app.key`


driesvints:

finagin:
@driesvints Why? The key in the tests is needed for everyone

PR #5301, opened 2020-05-09
[7.x] Get the home URL from the route in welcome.blade.php

aran112000:
It seems to be a fairly common thing for the home route's URL to be updated, for example, to /dashboard. This change keeps the behaviour in line with the Register & Login routes linked from the same file so the URL can be altered more easily.

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -68,7 +68,7 @@
             @if (Route::has('login'))
                 <div class="top-right links">
                     @auth
-                        <a href="{{ url('/home') }}">Home</a>
+                        <a href="{{ route('home') }}">Home</a>
                     @else
                         <a href="{{ route('login') }}">Login</a>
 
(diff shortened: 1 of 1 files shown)


zaknesler:
The `home` route only exists when you install the authentication scaffolding, so this would throw an error on a fresh installation

aran112000:
In line with this pull request, what would people think with updating the Laravel UI package so when using --auth, the default route /home added registers using `RouteServiceProvider::HOME` instead of `'/home'`, so you would only then have a single place to update for consistency.

https://github.com/laravel/ui/blob/ec838c75ba1886d014c5465b1ecc79b2071f46c7/src/Auth/stubs/routes.stub#L4

GrahamCampbell:
This has been proposed before (in fact many times). See for example, https://github.com/laravel/laravel/pull/5253.

aran112000:
Hey @GrahamCampbell, appreciate that, however, unlike that example, PR addresses the issue that the 'home' route might not always exist to ensure it's not a breaking change.

PR #5303, opened 2020-05-18
Add Redis to list of drivers that support SESSION_STORE

hackel:
This was always true as far as I can tell, it just didn't make it into the initial description.  Hoping to save someone a bit of time digging through the source to confirm.


taylorotwell:
I'll update it and preserve the formatting.

PR #5304, opened 2020-05-19
Fix inconsistent return points

markwalet:


GrahamCampbell:
Thanks, but `void` cannot occur in a union type.

GrahamCampbell:
https://wiki.php.net/rfc/union_types_v2

markwalet:
Thanks for the fast response. PHPStorm gave me a warning for it so that's why I created it.

Shouldn't we end the method with `return null;` then?

PR #5305, opened 2020-05-22
Add a small hint to the doc block of aliases

burzum:


driesvints:
I believe this is already explained well enough in the docs. Thanks.

burzum:
@driesvints I disagree, because I couldn't find it. And I didn't release Laravel is registering them in the root namespace it took me some time to figure it out... A hint there would have saved me some time.

So where in the 6.x docs is it explained? It doesn't mention how they're registered anywhere on this page... https://laravel.com/docs/7.x/facades

lionogueira:
I had a tough time trying to register an alias for a custom facade.

Nowhere in the doc I found that the alias must be register within the second parameter of the withFacades function

`
$app->withFacades(true, [
    App\Facades\STLogFacade::class => 'STLog',
]);
`

I tried register with alias but didn't worked
`
$app->alias(App\Facades\STLogFacade::class, 'STLog');
`

just a suggestion to improve the doc.

thanks

driesvints:
Everyone's free to send in prs to the docs.

PR #5306, opened 2020-05-27
[7.x] Set a default for Memcached's persistent_id

rennokki:
Distributed clusters like AWS Elasticache for Memcached seems to need this and the local (single node) instances work well too.

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -55,7 +55,7 @@ return [
 
         'memcached' => [
             'driver' => 'memcached',
-            'persistent_id' => env('MEMCACHED_PERSISTENT_ID'),
+            'persistent_id' => env('MEMCACHED_PERSISTENT_ID', 1),
             'sasl' => [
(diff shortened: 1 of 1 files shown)


taylorotwell:
You are free to add it to your own environment. That's why it's there.

rennokki:
I was thinking anyone can be in my situation, trying to find a solution for this. 👀

Juje:
Thank you @rennokki for raising this I was trying to figure out for days why Memcached was being annoying and inconsistent until I saw the `MEMCACHED_PERSISTENT_ID` variable in the cache config and wondered what it was. And luckily this thread exists to confirm that it would fix my issue. So it would have been great if this was described in the docs or something :thinking:.

I do believe it should indeed not be a default because most users will never encounter this issue.

rennokki:
@Juje @taylorotwell Would be cool to submit a PR to add a disclaimer/warning in the Memcached section in the Laravel docs?

PR #5310, opened 2020-05-31
Catching requiring composer autoload file

allowing:


GrahamCampbell:
Requring a file that doesn't exist won't throw something you can catch, and that header line is incorrect. You cannot use a custom message for error 500.

PR #5317, opened 2020-06-11
[7.x] Update redirectTo return type for the Authenticate middleware

felixdorn:
This method either returns a Response or nothing: I think this is less misleading.

Moreover, this also triggers Larastan and by extension PHPstan because the return type is not valid (which is true).

Related to https://github.com/laravel/laravel/commit/40f93daa83b17ad47c51ec9beed4c1ba79eb66ed


driesvints:
It's fine how it is now.

PR #5318, opened 2020-06-11
[7.x] Add support for multiple LOG_CHANNEL values

fujiwaraizuho:
LOG_CHANNEL can now have multiple channels separated by commas.
The default value is "single", so compatibility is not broken.
We also used array_diff to exclude "stack" to prevent infinite recursion.


driesvints:
Doesn't this conflict with the default value? What if you set two channels? You'd have two defaults?

fujiwaraizuho:
Thanks for the reply.
It's certainly competitive.
So how about making the default fixed?

mpyw:
I think

```php
'default' => env('LOG_CHANNEL', 'stack'),
```

should be always fixed into:

```php
'default' => 'stack',
```

fujiwaraizuho:
I'll fix it that way.

mpyw:
This may have a breaking change. If we already set the actual environment variable `LOG_CHANNEL=stack` and update `config/logging.php` by copy-and-paste, the log that was previously sent into `single` driver is now never been sent.

But this repository is `laravel/laravel`, not `laravel/framework`. Should we concern that user updating `config/logging.php` by themselves or not?

driesvints:
We definitely can't change the default. Feel free to do this in your own app.

PR #5324, opened 2020-06-22
[8.x] Requires PHP 7.3 and bumps both PHPUnit and Collision

nunomaduro:
This pull request makes the Laravel skeleton require PHP 7.3. Note that, the [framework itself already requires PHP 7.3](https://github.com/laravel/framework/blob/9a7a475f895643e015dc287ddb7593cda2cff945/composer.json#L18).

Another note: Collision `^5.0` is not available yet, at the moment, this requirement will pull `v5.0.0-BETA3`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     ],
     "license": "MIT",
     "require": {
-        "php": "^7.2.5",
+        "php": "^7.3",
         "fideloper/proxy": "^4.2",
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks!

GrahamCampbell:
Note that laravel 8 will also support phpunit 8 still, and packages should also.

PR #5331, opened 2020-06-29
Change example environment to production

martbock:
This pull request changes the default environment in the `.env.example` to `production`.

The reason for this is to prevent users from easily running insecure setups where they deploy their Laravel app to production, but don't change the pre-configured environment. This could be especially dangerous when using Laravel Telescope.
Telescope makes requests visible through its web interface at the `/telescope` route. When a user deploys their Laravel app with Telescope and does not change the pre-configured environment, information on every request is publicly available.

In my opinion, the risk of someone forgetting to change their `APP_ENV` is too high when you factor in the possible security implications.

I have read about this topic in German IT magazine Golem: https://www.golem.de/news/laravel-telescope-die-sicherheitsluecke-bei-einer-bank-die-es-nicht-gibt-2006-149251.html

According to the article author @hannob, there are "numerous" Telescope instances out there with the `local` environment setting. He has also implemented a check for publicly available Telescope instances in his tool [snallygaster](https://github.com/hannob/snallygaster/commit/e1ed99667bf5716673a9836acef5cd828e3cd07a).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,7 @@
 APP_NAME=Laravel
-APP_ENV=local
+APP_ENV=production
 APP_KEY=
-APP_DEBUG=true
+APP_DEBUG=false
 APP_URL=http://localhost
 
 LOG_CHANNEL=stack


driesvints:
We have no plans on changing this sorry.

captainhook:
Why not?

martbock:
> We have no plans on changing this sorry.

@driesvints I know you want to make it as easy as possible for new developers to start developing with Laravel – which is a very pleasant experience, by the way – but I think you should reconsider whether the benefits really do outweigh the costs here.

Of course we are all adults and people should read the docs when they use open-source software. But in reality, most folks will not – and may be surprised when they realize that the default setting made their app insecure. And that is a problem we could solve here. 

Moreover, the [Telescope docs](https://laravel.com/docs/7.x/telescope#dashboard-authorization) does state that...
> By default, you will only be able to access this dashboard in the local environment.
but – in my opinion – we should at least warn the user which consequences it might have to not change the environment setting.

I would love it if we could at least have a discussion about the pros and cons of both approaches.

lupinitylabs:
I have to agree with @driesvints. Changing the default environment to production would mean that you cannot even start developing properly without changing it back to local first. Many tools and integrations won't work in production mode, migrations will not run without issuing warnings and so on... this would probably overwhelm some first time users.

To be honest, if you forget to set the environment to production after deploying (which in my opinion is damn hard to do because the .env file should not be versioned and thus needs to be created explicitly on each server anyways), you probably shouldn't develop for the internet (yet).

It might be discussed, however, to deny access to the Telescope route to unauthenticated users by default though, or to restrict access to local IPs. It could also be discussed to have the framework warn if the app is run in local mode, e.g. by generally adding a warning banner to each outgoing template or by detecting if the server might be a production server, if we can come up with any heuristics to determine that reliably, e.g. by server IP ranges or network topology?

lupinitylabs:
Also, the article states that the main issue is that the server environment is dumped, which may contain passwords and other sensitive information. Maybe it would be a good start to block everything that could reveal sensitive data by default.

comes:
I mean, if you're able to overwrite the default .env file in forge setup with
```
APP_ENV=production
APP_DEBUG=false
```
It should be possible to change this values to
```
APP_ENV=local
APP_DEBUG=true
```
if you setup your local development environment via
```
laravel new my-project
```
or
```
composer create-project --prefer-dist laravel/laravel blog
```

I think these little line in projects composer.json can be fixed
```
"post-root-package-install": [
    "@php -r \"file_exists('.env') || copy('.env.example', '.env');\""
],
```

Noxmiles:
Time for a new subtitle:
Laravel - the unsecure PHP-Framework

lupinitylabs:
> ```
> "post-root-package-install": [
>     "@php -r \"file_exists('.env') || copy('.env.example', '.env');\""
> ],
> ```

True, copying the .env.example to .env by default is toxic. Then again, are you supposed to create a new project on a live server? :sweat:

comes:
What's the recommend workflow copying a project to production server and creating an env-file? As far as I remember, there is nothing.
What did you normaly do? I guess
```
cp .env.example .env
vim .env
```
you should not have an env-file in your git repository. So, in my opinion, the most dangerous part of 'getting a project live' is to create an ENV-file

ZiIMoIN:
@driesvints and @lupinitylabs wouldn't it be the best way to add a comment to the top of the `.env.example` file which says something like this:

```
# The following configuration is only for local development and should be adapted on productive systems for security reasons
# More information here: https://laravel.com/docs/master/configuration
```

As everyone at least needs to adjust the database settings (usually also when switching to production) this should be the best compromise between security and easy starting or is there a reason why no comment should be inserted?

azeyrx:
According to the article, the problem is not Laravel, but the telescope extension itself.  They should create their own config for that.

stefanseeger:
Hello,
Some projects are also having multiple env example.
Maybe creating `.env.production` and `.env.debug` with the different configurations and explanations, would simply force the dev to think twice.

lupinitylabs:
> 
> 
> @driesvints and @lupinitylabs wouldn't it be the best way to add a comment to the top of the `.env.example` file which says something like this:
> 
> ```
> # The following configuration is only for local development and should be adapted on productive systems for security reasons
> # More information here: https://laravel.com/docs/master/configuration
> ```
> 
> As everyone at least needs to adjust the database settings (usually also when switching to production) this should be the best compromise between security and easy starting or is there a reason why no comment should be inserted?

Adding a comment wouldn't hurt, but I am pretty sure the main reason for this happening is copying the config over from somewhere without reading it, otherwise they might have realized the problem in the first place.

Also, https://laravel.com/docs/7.x/configuration warns about versioning the .env file, but does not warn explicitly about local environment on production servers. Maybe this should be added.

comes:
Back in days Laravel 4.2 solves this quite easy:

https://laravel.com/docs/4.2/configuration#environment-configuration

I do not fully understand why APP_ENV should be set in .env file, instead the good old way?

lupinitylabs:
> 
> 
> Back in days Laravel 4.2 solves this quite easy:
> 
> https://laravel.com/docs/4.2/configuration#environment-configuration
> 
> I do not fully understand why APP_ENV should be set in .env file, instead the good old way?

Dotenv imports the .env contents into $_SERVER anyways, so nothing stops people from setting MY_LARAVEL_ENV there in the first place, which kinda defeats the idea.

comes:
> > Back in days Laravel 4.2 solves this quite easy:
> > https://laravel.com/docs/4.2/configuration#environment-configuration
> > I do not fully understand why APP_ENV should be set in .env file, instead the good old way?
> 
> Dotenv imports the .env contents into $_SERVER anyways, so nothing stops people from setting MY_LARAVEL_ENV there in the first place, which kinda defeats the idea.

you can support APP_ENV in .env, but you should not forced to use it. If APP_ENV is missing or null, auto detection is running. In .env.example you can set APP_ENV to null, same like APP_KEY

lupinitylabs:
> 
> 
> > > Back in days Laravel 4.2 solves this quite easy:
> > > https://laravel.com/docs/4.2/configuration#environment-configuration
> > > I do not fully understand why APP_ENV should be set in .env file, instead the good old way?
> > 
> > 
> > Dotenv imports the .env contents into $_SERVER anyways, so nothing stops people from setting MY_LARAVEL_ENV there in the first place, which kinda defeats the idea.
> 
> you can support APP_ENV in .env, but you should not be forced to use it. If APP_ENV is missing or null, auto detection is running. In .env.example you can set APP_ENV to null, same like APP_KEY

Auto detection? If APP_ENV is missing, everything will run as if it was local, because most parts only check if APP_ENV is production, where they should probably better check if APP_ENV != local...

K2ouMais:
And because of this behavior there is an article on a german news site.

It seems someone was developing a site for a bank and uses Telescope for debuging.

Because the dev didnt change in the .env for production everyone could see all the data.

https://www.golem.de/news/laravel-telescope-die-sicherheitsluecke-bei-einer-bank-die-es-nicht-gibt-2006-149251.html

Please consider taking another look at this... For beginners in Laravel this could be a really big problem.

DanielRuf:
@K2ouMais this exact same link was already mentioned in the very first comment ;-)

K2ouMais:
> @K2ouMais this exact same link was already mentioned in the very first comment ;-)

Oh sorry... Didnt saw that... Thank you

x3l51:
IMO restricting default access to local IPs and otherwise authenticated users might be the way to go here. I can't think of a scenario where this would block other tools from interacting with the development or production state of a website.

Even if "rtfm" is always true, protecting the users from their own potential misuse and disregarding of basic security practices could be a valid argument for changing the default access restrictions.

DanielRuf:
How does Symfony handle this?
They use APP_ENV which is probably set to `dev` by default.
https://github.com/symfony/recipes/issues/633

comes:
I highly recommend studing the old laravel 4.2 documentation (https://laravel.com/docs/4.2/configuration#environment-configuration)
Exact the same problem, was solved different before, so imho these issue never exists in  < 4.2

driesvints:
We've had this discussion multiple times in the past and have no plans on changing this. Having the defaults set for development allows for an easy setup to get started building your app. When you're ready to put your app live it's up to you to set sensible defaults for production. Like I said, we have no plans on changing this.

PR #5332, opened 2020-07-01
[8.x] Use Models Folder

voyula:
https://github.com/laravel/framework/pull/33390

PR #5333, opened 2020-07-01
[8.x] Use Models Folder

voyula:
https://github.com/laravel/framework/pull/33390

https://github.com/laravel/laravel/pull/5332

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace App;
+namespace App\Models;
 
 use Illuminate\Contracts\Auth\MustVerifyEmail;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
It is still not clear if we want this in the default app. The framework PR only allows this as an option, but doesn't require it.

voyula:
@GrahamCampbell Yes, but this PR is a request.

taylorotwell:
I make these decisions.

GrahamCampbell:
Also, note that Twitter survey had quite a small number of participants, compared to the number of Laravel users, and in particular compared to the number of new Laravel users there will be over the next 12 months, who will see the laravel/laravel starter project for the first time.

elphia:
The twitter poll wasn’t sent to all Laravel users you can’t expect as many responses.
The tendency was clear though.
If it should be implemented or not is another subject.
Taylor said in a keynote that Laravel major version won’t be as big change in the architecture as in the past with Laravel major versions.
Introducing app/models would be a big change for an upgrade of a big project.

PR #5334, opened 2020-07-01
[7.x] Document asset_url key

pxdnbluesoul:
Hello,

As far as I can determine, the `asset_url` configuration key hasn't been used since the 3.x era in `laravel/url.php`. A grep of asset_url through the entire repo is empty after removing this key from config/app.php.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -54,8 +54,6 @@ return [
 
     'url' => env('APP_URL', 'http://localhost'),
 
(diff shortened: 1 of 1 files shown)


pxdnbluesoul:
I found its usage in the routing service provider, let me instead document it a little better.

taylorotwell:
I'll document it.

PR #5335, opened 2020-07-02
Bump guzzlehttp/guzzle

jonnott:
Guzzle 7.x's PHP requirement is now ^7.2.5, the same as laravel/laravel's and laravel/framework's. Current laravel/framework (7.x) already has guzzlehttp/guzzle bumped to ^6.3.1|^7.0 .. so I figure this can also be similarly bumped, even for forthcoming 7.x releases (i.e. even ahead of Laravel 8.x) ..?

Let me know if I've missed something..

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "php": "^7.2.5",
         "fideloper/proxy": "^4.2",
         "fruitcake/laravel-cors": "^2.0",
-        "guzzlehttp/guzzle": "^6.3",
+        "guzzlehttp/guzzle": "^6.3|^7.0",
         "laravel/framework": "^7.0",
         "laravel/tinker": "^2.0"
     },


GrahamCampbell:
Guzzle 6 is fine for the example Laravel 7 app, for now. There are lots of other laravel packages that don' yet work with Guzzle 7.

GrahamCampbell:
What we may do is change the default Guzzle version from 6 to 7 in the example Laravel 8 app, but not support both, which is usually the way we do things.

GrahamCampbell:
> There are lots of other laravel packages that don' yet work with Guzzle 7.

I am currently in the process of a large migration effort actually, of the most popular GitLab, GitHub and Bitbucket API packages, so that they can run with Guzzle 7.

jonnott:
One path of badness I've detected is for laravel/socialite users. Socialite itself supports "guzzlehttp/guzzle": "^6.0|^7.0", but also depends still on "league/oauth1-client": "^1.0", which allows only Guzzle 6.x. And Issues/PRs in that repo don't seem to get much attention :(

GrahamCampbell:
Thanks for letting us know. Which websites are still using oauth1. I wonder if we can drop oauth1 from the next socialite release?

jonnott:
@GrahamCampbell Maybe Twitter by the looks of this? https://github.com/laravel/socialite/search?q=League%5COAuth1&unscoped_q=League%5COAuth1

GrahamCampbell:
I wonder if twitter implement oauth2 as well, but just it's not been implemented in our codebase yet.

GrahamCampbell:
Let's see how https://github.com/thephpleague/oauth1-client/pull/101 does, before we think about removing anything, though, actually.

driesvints:
I think there was a specific reason why we were using OAuth1 but can't put my finger on it atm. Think it had something to do with Twitter authentication itself.

driesvints:
Seems like Twitter itself requires OAuth 1 if you want to do specific private actions like retweeting etc: https://developer.twitter.com/en/docs/basics/authentication/overview

jonnott:
> Seems like Twitter itself requires OAuth 1 if you want to do specific private actions like retweeting etc: https://developer.twitter.com/en/docs/basics/authentication/overview

@driesvints @GrahamCampbell Definitely seems like pushing towards `league/oauth1-client` supporting Guzzle 7.x is the way to go then. Could it be as simple as just dropping 7.x into it and seeing if all its tests still pass? I feel that's about the level at which I could be of help ;)

driesvints:
@jonnott you'll have to ask the maintainers of `league/oauth1-client`.

jonnott:
> @jonnott you'll have to ask the maintainers of `league/oauth1-client`.

@driesvints ... @GrahamCampbell has already started the ball rolling here https://github.com/thephpleague/oauth1-client/pull/101 ..hence mentioning in this context.

GrahamCampbell:
One should really ask how hard is it to just write from scratch the pieces laravel needs, and drop that package?

shehi:
That ain't the real question. It's about us having time for this thing: at least coupla days would be required to properly test such a codebase. I am currently refactoring it, I think I will be done tonight. It will be PHP 7.2+ as Laravel needs. PHP-5 users can still use v1 of it.

shehi:
@GrahamCampbell , I finished refactoring that lib: https://github.com/thephpleague/oauth1-client/pull/104

It needs testing. I am confident it will work, possibly with few minor errors here and there. But in a nutshell, I paid great attention not to change existing structures and logic. But had to fix that nonsensical re-implementation of `http_build_query()` of PHP (i.e. I threw away that implementation and just used PHP's existing function). All tests pass. Little bit cleaned up the repo. It is ready for review, testing and release.

PR #5336, opened 2020-07-03
[7.x] .env: change 127.0.0.1 to localhost

guoyunhe:
For homestead, 127.0.0.1 doesn't work in Linux/macOS but localhost works. See https://stackoverflow.com/questions/35394230/sqlstatehy000-2002-connection-refused-within-laravel-homestead

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,7 +7,7 @@ APP_URL=http://localhost
 LOG_CHANNEL=stack
 
 DB_CONNECTION=mysql
-DB_HOST=127.0.0.1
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
localhost wasn't used because it broke mysql on Windows I think?

taylorotwell:
You can change this in your own application if you need.

guoyunhe:
But this should at least be mentioned in the env file as comments or in Laravel documentation. I found many people on the internet spend hours just to solve this issue. Not a good experience...

PR #5337, opened 2020-07-05
change links in filesystems.php

hamedSHP:
There is a problem with this command

$this->app->bind('path.public', function() {
            return base_path().'/public_html';
        });

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -79,7 +79,7 @@ return [
     */
 
     'links' => [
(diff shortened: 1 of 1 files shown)


akbarjimi:
No, that's not a problem. You must first softlinked the storage directory to the public directory. then it works

GrahamCampbell:
@akbarjimi Is correct. Please check the storage docs. :)

PR #5339, opened 2020-07-10
update route default with arrow function

medeirosinacio:


GrahamCampbell:
Thanks, but Laravel 8 needs to support PHP 7.3 out of the box.

PR #5340, opened 2020-07-13
Use Route::view() instead

liamdemafelix:


GrahamCampbell:
Thanks, but we wanna show the closure example, so people can add to it straight away.

PR #5045, opened 2019-06-15
Added docker-compose

PascalKu:
Hey there,

I added docker-compose for a smoother start using laravel.

PR #5344, opened 2020-07-16
initialized dockerization

use-to:
This pull-request adds an [docker image](https://www.docker.com/) for generic laravel applications.
By executing the [docker image](https://www.docker.com/) after build the web application will be served by [nginx](https://www.nginx.com/).

Running your laravel app in docker is simple now:
1. Building the docker image (this could take a while): `docker build . -t my_laravel_app`
2. Creating an docker container using the image: `docker run -p 80:80 -d my_laravel_app`
3. The web application is now productive available on [localhost](http://localhost:80/).

This pull-request can be extended by another one containing an `docker-compose.yml` which runs a database, a redis, and the target application easily to provide better deployment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM nginx:alpine
+
+EXPOSE 80
+
+RUN apk update
+RUN apk upgrade
+RUN apk add php7 php7-fpm php7-fileinfo php7-dom php7-tokenizer php7-session composer
+
+WORKDIR /app
+
+COPY . .
+
+RUN rm Dockerfile
+RUN rm -r /usr/share/nginx/html
+
+RUN chmod -R 777 .
+
+RUN mv nginx-docker.conf /etc/nginx/conf.d/default.conf
+
+RUN composer install --optimize-autoloader --no-dev
+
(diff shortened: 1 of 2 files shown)


driesvints:
Heya, thanks but we'd prefer people to make use of something like Vessel instead.

https://vessel.shippingdocker.com

use-to:
I just read the pull-requests #5045 and #3990. I would really appriciate a merge of this pull-request, because docker is an growing container platform and has the possbility to better deploy great laravel apps on more platforms.

GrahamCampbell:
And there are lots of different ways to use Laravel with Docker. The version in this PR is certainly not one I'd recommend for most people. Best to leave it up to people to decide what's best for them.

use-to:
@driesvints Interessting, but vessel has not the same goal. I dont see a good reason why this pr should be closed.
Vessel focused: Simple Docker development environments for Laravel
This pr is not about development, its about deployment with an single docker image.

use-to:
@GrahamCampbell That is a good point, too. I think there is only one good way to deploy your laravel app with docker. What do you mean with `[...] not one I'd recommend for most people.`?

PR #5345, opened 2020-07-18
Update database.php

TechComet:


antonkomarev:
That's wrong change. Initial code is OK.

GrahamCampbell:
Thanks for the PR, but the existing code is what we intended.

TechComet:
Ok
but in `RedirectsUsers.php` file has `$this->redirectTo : '/home';` in line 18
need replace to `$this->redirectTo : RouteServiceProvider::HOME;`
What is your opinion ?

antonkomarev:
> but in `RedirectsUsers.php` file has `$this->redirectTo : '/home';` in line 18
> need replace to `$this->redirectTo : RouteServiceProvider::HOME;`
> What is your opinion ?

@TechComet this is a different topic of discussion. It's better to make a separate PR with proposed changes and write why this should be changed.

If it's bug - show how could it be reproduced, if it's new feature - write example of benefits it gives.

PR #5349, opened 2020-07-23
[8.x] Update prioritization style

voyula:
Commonly "LAST" keyword adding to the end.


taylorotwell:
Does this change behavior?

voyula:
@taylorotwell Not, but, the main reason for using laravel is that it has spelling rules.

voyula:
many things that are closed unnecessarily are will be avalanches.

PR #5350, opened 2020-07-24
Add middleware to force https protocol

ricardogobbosouza:


driesvints:
This should probably be handled at the web server level or through a service like Cloudflare. Otherwise, feel free to release this as a package. Thanks.

GrahamCampbell:
Yeh, there's loads of resources online for how to do this in apache or nginx, for example.

ricardogobbosouza:
Yes, without a doubt, this would be an alternative, many projects that are executed in shared hosting do not have access to such configurations

ricardogobbosouza:
So I left it disabled by default.

ricardogobbosouza:
Maybe we could have a section in the documentation explaining about https

GrahamCampbell:
That is not really a concern for the framework, and indeed, I'd recommend not ever doing that sort of thing in the framework, unless you are particularly careful. For example, you don't want to accidentality leak a session cookie over HTTP.

PR #5354, opened 2020-07-27
[7.x] Enforce ^7.22.1 due to security issues

rennokki:


driesvints:
Updated to `^7.22.1` which has the fix for a problem with CSRF.

PR #5359, opened 2020-07-31
Update User.php

J2TeamNNL:


sangnguyenplus:
User model should not implement `MustVerifyEmail` by default, just modify it if you want to verify user email https://laravel.com/docs/7.x/verification#model-preparation

PR #5360, opened 2020-08-02
[7.x] Preventing auth.json and Nova directory from being committed accidentally

mnaderian:
Sometime ago when I pushed one of my projects to github I realized that the **Nova** directory is pushed with it! and it got me a horrible feeling that I have committed an accidental copyright infringement!!

And think of it when it happens about the ``auth.json`` file that contains your important credentials!

So I added them to ``.gitignore`` to prevent such horrible mistakes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,4 @@ Homestead.yaml
 npm-debug.log
 yarn-error.log
 /nova
-auth.json
+/auth.json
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -10,3 +10,5 @@ Homestead.json
 Homestead.yaml
 npm-debug.log
 yarn-error.log
+/nova
+auth.json
```
GrahamCampbell, line 14: Please make this `/auth.json`, assuming that file will be in the repo root. There may be cases where people want to commit other auth.json files.

mnaderian: good point.

PR #5363, opened 2020-08-04
Update composer.json

febrihidayan:


driesvints:
This is unnecessary. The current constraints will already install this version.

PR #5364, opened 2020-08-05
[7.x] Add missing return statement

u01jmg3:
- Add missing return statement

### DIFF
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -17,5 +17,7 @@ class Authenticate extends Middleware
         if (! $request->expectsJson()) {
             return route('login');
         }
+
+        return null;
     }
 }


u01jmg3:
PHPStan:

```
Method App\Http\Middleware\Authenticate::redirectTo() should return string|null but return statement is missing.
```

u01jmg3:
Bit confused why this commit was accepted 40f93da when the [`route()`](https://github.com/laravel/framework/blob/7.x/src/Illuminate/Foundation/helpers.php#L764) helper doesn't return `null` 🤷‍♂️

u01jmg3:
@mingyoung, @marky291: a bit confused why this commit was accepted 40f93da when the [`route()`](https://github.com/laravel/framework/blob/7.x/src/Illuminate/Foundation/helpers.php#L764) helper doesn't return `null` 🤷‍♂️

GrahamCampbell:
Because not returning anything **does** return null. Try it out.

u01jmg3:
Thanks for clarifying and I get that. Should PHPStan know better then?

GrahamCampbell:
No, PHPStan is making you be explicit to ensure the code does what you intended. You can update your own app to be explicit if you like, or ignore the warning.

u01jmg3:
Thanks @GrahamCampbell!

PR #5371, opened 2020-08-07
use Route::view function instead of anonymous function

atikla:


GrahamCampbell:
Thanks, but we are already happy with the example we have. Please check other PRs before sending a duplicate in future. :)

PR #5373, opened 2020-08-13
[7.x] Add stdout log channel

ankurk91:
Hi,

Useful in worker instance (AWS) where we want to send all log calls (not just errors).


taylorotwell:
You can add this to your configuration as needed.

PR #5374, opened 2020-08-13
[7.x] Use a mutator method to automatically hash plain text passwords

syntheticminds:
To avoid passwords being set in plain text, the proposed addition to the default User model will automatically hash any password value. It will not rehash bcrypt strings.

The user factory can be updated to simply provide 'password' as the default.


GrahamCampbell:
Thanks for the PR. I think this has been proposed before and not accepted?

driesvints:
Yeah, feel free to do this in your own app.

PR #5375, opened 2020-08-15
[7.x] Show synchronous_commit setting for Postgres

mfn:
It's deliberately not enabled but provided commented out so developers can easily
discover that the setting is available.

The default `on` is _probably_ safe (Postgres default), but a server might
have a different default value and blindly setting it to `on` may cause problems.

Connected to https://github.com/laravel/framework/pull/33897


taylorotwell:
I don't think I'll add this to the stub.

LanreSam:
wow

PR #5376, opened 2020-08-15
Added token expiration at and updated at

Gicehajunior:
1)token expiration just to imply on when a user don't use the token for a certain period of time.
2)updated at, when the user requests the password reset again, the token is updated with new one and a time for token update is updated too.


database/migrations/2014_10_12_100000_create_password_resets_table.php
```diff
@@ -16,7 +16,9 @@ public function up()
         Schema::create('password_resets', function (Blueprint $table) {
             $table->string('email')->index();
             $table->string('token');
+            $table->timestamp('token_expires_at')->nullable();
```
GrahamCampbell, line 19: Why wouldn't this just be `expires_at`?

GrahamCampbell:
Also, where is the associated code for this in laravel/framework, please?

driesvints:

PR #5380, opened 2020-08-25
Update AppServiceProvider.php

drplutus:
When migrate the database it's thow an error. so fixed this error permanatly.


marky291 (commented):
This issue only occurs on older versions of MySQL, not an issue on latest.

PR #5381, opened 2020-08-25
[8.x] Bump Guzzle

driesvints:


GrahamCampbell:
^7.0.1 would be better. 7.0.0 has a bad bug causing fatal errors, even when Guzzle isn't even used.

driesvints:
@GrahamCampbell done but I guess it doesn't truly matters since this is user land code

sangnguyenplus:
I have tried to upgrade Guzzle on my project but I got stuck at v6 because I'm using Laravel Socialite.
Screenshot at Aug 25 19-18-26

So if this PR is merged, we will need to downgrade Guzzle to install Laravel Socialite.

driesvints:
@sangnguyenplus yes you can. You own this code. Plus it seems that maintenance on oauth1-client will be picked up soon enough: https://twitter.com/ben_corlett/status/1297815328571813888

erfannaseri:
Tnx mr taylore
  

  

  

  
  
  
  
  
>

mfn:
This starts to cause lots of interop troubles.

Does L8 _truly_ require Guzzle 7 or does it work with 6? Is this for the included HTTP client or something else too?

thanks

GrahamCampbell:
Works with both (see the "suggest" message from `laravel/framework`). The default app just defaults to Guzzle 7 now. Guzzle 6 will be EOL at some point before Guzzle 7.

GrahamCampbell:
Upgrading from Guzzle 6 to 7 is mostly easy. The easiest Guzzle upgrade every, actually, and intentionally so. Guzzle 7's only goal was to modernise without breaking things as much as possible, and then get everyone to upgrade.

mfn:
"Problem" is the ecosystem, packages that have Guzzle ^6 as version constraint now collide.

If Guzzle 7 isn't strictly required, maybe ^6|^7 can be used?

GrahamCampbell:
This is just the default app, and we don't encourage versions like that, to eliminate confusion. Developers should change the version to `^6.5` explicitly if they want v6.

driesvints:
@mfn this is just the skeleton. You own this code. Adjust as you like.

PR #5385, opened 2020-08-27
[7.x] Make the PHPUnit cache file .gitignore rule more generic

markjaquith:
Switches the pattern to `.*.cache`.

Reasoning:

1. PHP CS Fixer uses this pattern too (`.php_cs.cache`).
2. PHP CS Fixer is used by the popular [Laravel Shift](https://laravelshift.com/shiftrc-configuration-file) tool.
1. `.*.cache` is a good generic pattern for various cache files, and a generic rule encourages future tool makers to use it for their cache files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@
 /vendor
 .env
 .env.backup
-.phpunit.result.cache
+.*.cache
 Homestead.json
 Homestead.yaml
 npm-debug.log


driesvints:
In general I'd say that rule belongs in a global gitignore but this might be a good exception to the rule.

GrahamCampbell:
I think we should just add `.php_cs.cache`. No need for the wild card?

driesvints:
@GrahamCampbell bit torn on that since it'll open up the floodgates for more library specific rules. The skeleton doesn't uses that.

driesvints:
@markjaquith in the light of above I'm gonna close this. Feel free to add this rule to your own project.

mfn:
I find the argument from @driesvints compelling (not used in the skeleton)

I also suggest not to be to greedy with wildcards, could have surprising effects.

GrahamCampbell:
Also, Laravel ships with a StyleCI config file, but not a PHP-CS-Fixer config file. ;)

PR #5388, opened 2020-09-03
[7.x] Update docblock and parameter name to match parent

jochensengier:


taylorotwell:
I think it's fine as is.

jochensengier:
I understand and agree why you don't want to change the parameter name, but what about the incorrect '@throws' in the docblock?

driesvints:
@jochensengier fixed

app/Exceptions/Handler.php
```diff
@@ -29,27 +29,27 @@ class Handler extends ExceptionHandler
     /**
      * Report or log an exception.
      *
-     * @param  \Throwable  $exception
+     * @param  \Throwable  $e
      * @return void
      *
-     * @throws \Exception
+     * @throws \Throwable
```
GrahamCampbell, line 35: ?

jochensengier: If your questioning the change, please see https://github.com/laravel/framework/pull/33888

PR #5389, opened 2020-09-03
Remove APP_NAME from .env.example

murtazasultani:
APP_NAME is usually app specific, so I think it would be great to edit the config file to change it.


taylorotwell:
I'll just leave it for now in case people are depending on it.

driesvints:
No plans to change this sorry.

antonkomarev:
@murtazasultani I'm usually separating names of prod, staging and dev environments.

PR #5390, opened 2020-09-03
[7.x] Update APP_URL on .env.example

murtazasultani:
Since majority uses artisan serve command to run a Laravel app and the artisan serve command use port 8000 to run the app

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,7 +2,7 @@ APP_NAME=Laravel
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
-APP_URL=http://localhost
+APP_URL=http://127.0.0.1:8000
 
 LOG_CHANNEL=stack
 
--- a/config/app.php
+++ b/config/app.php
@@ -52,7 +52,7 @@ return [
     |
     */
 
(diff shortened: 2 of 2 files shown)


martindilling:
I don't think I've ever used artisan serve ;)

murtazasultani:
@martindilling I think still `http://localhost` is not a proper default value for APP_URL

sidis405:
Hmm. I guess it depends on one's workflow.
In my own experience, i use serve only if i'm checking out something temporary for which i don't want a host.
Other than that, developing using artisan serve is hella cumbersome compared to valet/homestead.

That being said, out of curiosity, under which circumstance does the current APP_URL value get in the way?
I can think of 2-3 scenarios in which the value is called from the framework but that's why you come and edit the value.

The only thing that comes to mind is that changing this to 127.0.0.1:8000, (to the value of artisan serve), might give the idea to newcomers that if you change this element in .env, the host:port of serve will change, and that just ain't it.

GrahamCampbell:
I think this has been proposed before, and rejected.

PR #5391, opened 2020-09-04
Remove usage of non-strict empty() function

bramceulemans:
PHPstan recommends against usage of the `empty()` function: <https://github.com/phpstan/phpstan-strict-rules/blob/master/src/Rules/DisallowedConstructs/DisallowedEmptyRule.php>

I also updated the PHPdoc to match the new function parameter names.


taylorotwell:
The current code works fine.

PR #5392, opened 2020-09-04
[8.x] Update docblock

mvdnbrk:


canvural:
This is a wrong change. Variadic parameters should be documented like `@param string ...$params` not `@param string[] ...$params` The latter means each parameter is an array of strings.

PR #5393, opened 2020-09-05
[8.x] Add namespace of controller routes

patricpoba:
Add namespace of controller routes. Without this, the login route of the auth scaffolding (php artisan ui vue --auth) breaks with error message - **Target class [Auth\LoginController] does not exist.**


taylorotwell:
Use Jetstream.

PR #5394, opened 2020-09-06
Remove unused import

osule:
The `RefreshDatabase` trait is imported from `Illuminate\Foundation\Testing` namespace but never referenced in this example test.


GrahamCampbell:
Thanks, but that is left in on purpose, since it's an example test file.

PR #5400, opened 2020-09-08
[8.x] Fix links on welcome view for small screens

zaknesler:
(Resubmit of #5399; now targeting master)

The links in the "welcome" view had the class "hidden" applied for the smallest screen sizes, and would only display the links on small devices or bigger.

If that was unintentional, this PR fixes that by only positioning it absolutely on small devices or bigger, and displaying it centered on anything smaller at block display.

This also fixes the link color when using dark mode (the colors are now lighter and readable when in dark mode), and adds extra padding at the bottom, where previously the text touched the bottom of the page.

### Before

before

### After

after


taylorotwell:
image

Breaks padding / margin without links.

zaknesler:
@taylorotwell Overlooked that; it's now fixed in my fork

PR #5405, opened 2020-09-11
Remove unused class in EventServiceProvider

preeteshjain:
This removes the unused `Event` facade from `app/Providers/EventServiceProvider.php`


GrahamCampbell:
Thanks, but this import is intentionally left in as an example (and unused import fixing is turned off in StyleCI for this repo).

preeteshjain:
@GrahamCampbell Oh thanks for the explanation, didn't know that.

PR #5408, opened 2020-09-14
[8.x] Fix for telescope: Target class [env] does not exist

zingimmick:
Solution for: laravel/telescope#943

Reference: sebastianbergmann/phpunit#2451

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -26,6 +26,6 @@
         <server name="MAIL_MAILER" value="array"/>
         <server name="QUEUE_CONNECTION" value="sync"/>
         <server name="SESSION_DRIVER" value="array"/>
-        <server name="TELESCOPE_ENABLED" value="false"/>
+        <server name="TELESCOPE_ENABLED" value="false" verbatim="true"/>
     </php>
 </phpunit>


driesvints:
Can you share the PHPUnit version that's causing this? I still haven't managed to reproduce this.

zingimmick:
I just reproduce it, the simple case to: https://github.com/zingimmick/laravel-1/tree/telescope-example

It's funny when I try to reproduce it.

- it works fine when i run single test `TelescopeTest::testEnvDoesntExist`
- but give a `Target class [env] does not exist` exception when i run tests `TelescopeTest`

```php
<?php

namespace Tests\Feature;

use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;
use Tests\TestCase;

// Target class [env] does not exist
class TelescopeTest extends TestCase
{
    public function testBasicTest()
    {
        $response = $this->get('/');

        $response->assertStatus(200);
    }

    // works fine,  Error : Call to a member function toArray() on null
    public function testEnvDoesntExist()
    {
        Schema::create('dummies', function (Blueprint $table) {
            $table->id();
        });
        // just make exception
        Dummy::query()->whereKey(1)->first()->toArray();
    }
}

class Dummy extends Model
{

}
```

taylorotwell:
Just merged another PR to fix this?

PR #5409, opened 2020-09-14
Fix PHPUnit bool server consts result in null

hettiger:
After updating to Laravel 8 I suddenly had my test suite failing because telescope would not be disabled properly by `phpunit.xml` anymore.
This change fixed my test suite.

I've also created a clean Laravel 8 project and added some tests to demonstrate the issue:
https://github.com/hettiger/laravel-env-demo/commit/908d3405b84e769c12d296e98bef1ca567127254

Maybe this needs to be addressed in PHPUnit. However I'm adding this workaround here because it's a viable solution IMHO.
Maybe should add a note on this in the docs and be done with it…?

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -26,6 +26,6 @@
         <server name="MAIL_MAILER" value="array"/>
         <server name="QUEUE_CONNECTION" value="sync"/>
         <server name="SESSION_DRIVER" value="array"/>
-        <server name="TELESCOPE_ENABLED" value="false"/>
+        <server name="TELESCOPE_ENABLED" value="(false)"/>
     </php>
 </phpunit>


sergiy-petrov:
@GrahamCampbell @hettiger I do believe it's due to this change in DotEnv package
https://github.com/vlucas/phpdotenv/blob/7e6837db20ebb360f2b97fa87411d11a4b500bf6/src/Repository/Adapter/ServerConstAdapter.php#L44

Now server variables expect to be strings only, however it could be bool/int as well IMO.

sergiy-petrov:
This change in Repository/Adapter/ServerConstAdapter.php 

https://github.com/vlucas/phpdotenv/compare/4.1...5.0

зображення

hettiger:
Thanks for clarification @sergiy-petrov; LGTM.

Feel free to continue the discussion in a new issue at vlucas/phpdotenv.
If the dependency changes this can simply be reverted.

GrahamCampbell:
The bug is in PHPUnit. The `$_SERVER` variable is meant to have type `array<string, string>`. PHPUnuit shouldn't be converting the string `"false"` to the boolean `false` when it is loading stuff into that array.

GrahamCampbell:
Ping @sebastianbergmann for comment. Maybe there's a use case you know that really needs to not have environment variables be strings. Usually, I'd expect people using environment variables to convert them to an int or a bool when they come to use them, and not mess with how they are stored in the server array.

GrahamCampbell:
I am going to just patch phpdotenv 5.x to convert boolean values from $_ENV and $_SERVER to strings, instead of skipping over them as invalid.

sergiy-petrov:
@GrahamCampbell would be great. Thanks!

GrahamCampbell:
phpdotenv v5.2.0 has now been released, with this fixed.

PR #5412, opened 2020-09-15
[8.x] Update model path in AuthServiceProvider's policies

wojciechgabrys:
Due to the change of default model location in Laravel 8.x from /App to /App/Models, the initial policy comment shoud also reflect this change.


driesvints:
Thanks!

PR #5413, opened 2020-09-15
[8.x] Update phpunit.xml

wuwx:
when use telescope
./artisan test will be broken

we should use env replace server


driesvints:
This isn't needed. phpdotenv was causing this underlying and the framework has been patched with the new version since. It's waiting to be tagged.

PR #5414, opened 2020-09-15
[8.x] Translatable auth links out of the box

jimmitjoo:
Thought since we added translations for everything in Jetstream, maybe the same strings should be translatable in Laravel by default as well?


taylorotwell:
It's just the welcome page.

PR #5415, opened 2020-09-15
Remove failed_jobs migration

AlexWinder:
In the documentation of queues (https://laravel.com/docs/8.x/queues) and in the [config/queue.php](https://github.com/laravel/laravel/blob/master/config/queue.php) you have the option to customise the database table name for both the queues and the failed queues table name/migration.

However included in the Laravel repository contains a migration for the failed_jobs table [2019_08_19_000000_create_failed_jobs_table.php](https://github.com/laravel/laravel/blob/master/database/migrations/2019_08_19_000000_create_failed_jobs_table.php). 

If a user wants to customise their failed jobs migration then this will result in 2 tables being imported as part of the migrations. This file should be removed to allow users to only migrate the failed jobs if they require it, and if they need to customise the name.


taylorotwell:
Free to remove it or change it in your own application.

PR #5416, opened 2020-09-15
enable naming the migrations table via environment

emjayess:
Proposal to offer an environment variable for declaring a custom name for the `migrations` table, overriding the default in `config/database.php`...

I frequently rename `migrations` to `_migrations`, for example, to isolate it from application (model) tables, and float it to the top in database GUI tools.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -104,7 +104,7 @@ return [
     |
     */
 
-    'migrations' => 'migrations',
+    'migrations' => env('DB_MIGRATIONS_TABLE_NAME', 'migrations'),
 
     /*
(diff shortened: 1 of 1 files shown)


antonkomarev:
I don't think it requires .env variable because it will have the same value between environments.

emjayess:
> I don't think it requires .env variable because it will have the same value between environments.

It shifts the burden from code to config, and consequently relieves the need to reconcile a change with new releases of the framework ("don't hack core" issues).

After seeing #5415 I'd like to see this for `jobs` and `failed_jobs` also.

* Cost: 0
* Benefit: > 0

PR #5422, opened 2020-09-20
[8.x] Update User Model to Match Jetstream Formatting

amayer5125:
This will minimize the lines reported as changed in git or other version control systems when adding Jetstream to a new project. [Link](https://github.com/laravel/jetstream/blob/1.x/stubs/app/Models/User.php) for formatting comparison.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -9,8 +9,7 @@ use Illuminate\Notifications\Notifiable;
 
 class User extends Authenticatable
 {
-    use HasFactory;
-    use Notifiable;
+    use HasFactory, Notifiable;
 
     /**
      * The attributes that are mass assignable.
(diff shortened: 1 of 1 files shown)


antonkomarev:
Doing it all the time after fresh install.

amayer5125:
@taylorotwell The lines you reverted are also changed in Jetstream. The `use` statements are each moved to their own lines on [line 16](https://github.com/laravel/jetstream/blob/1.x/stubs/app/Models/User.php#L16) and [line 18](https://github.com/laravel/jetstream/blob/1.x/stubs/app/Models/User.php#L18). Git will always say that line changed when including Jetstream instead of only marking the `use` statements around those lines.

If you want me to open a separate pull request for that let me know. If not, I guess that is fine too.

PR #5425, opened 2020-09-24
Final fix for "Target class [Controller] does not exist."

IggsGrey:
Final fix for "Target class [Controller] does not exist." on new laravel installations


driesvints:

PR #5429, opened 2020-09-25
a change for psr12

selcukcukur:
- According to PSR12 Standards, the parameter list must be "()"


taylorotwell:
Unnecessary.

GrahamCampbell:
Laravel does not follow PSR-12 completely, though does follow almost all of the conventions from PSR-2 and PSR-12.

PR #5431, opened 2020-09-26
[8.x] Remove unused import

J2TEAM:
Import "Illuminate\Contracts\Auth\MustVerifyEmail" is never used.

image

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -2,7 +2,6 @@
 
 namespace App\Models;
 
-use Illuminate\Contracts\Auth\MustVerifyEmail;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Foundation\Auth\User as Authenticatable;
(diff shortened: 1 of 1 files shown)


driesvints:
Please see multiple previous attempts

J2TEAM:
Hi @driesvints, can you check the same PR for 5.7 #4750 

That PR has been merged.

https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Auth/User.php

driesvints:
That PR was reverted.

J2TEAM:
Thanks for letting me know that. One more little question, since I don't found the answer from the previous issues and PR yet. Why is that line kept here (and not used)? Because I know about it from the docs, and I think people can import it themselves when they want to use email verification.

driesvints:
They're kept for demonstration purposes so people are aware that they can implement them.

PR #5432, opened 2020-09-30
Add DB_PREFIX to .env and database.php

serwin35:
Added prefix support for .env files


GrahamCampbell:
Thanks, but please see previous rejected PRs.

PR #5434, opened 2020-10-01
Update .env.example

setiawanIman:


karlhillx (commented):
Why is this necessary?

setiawanIman:
Maybe many people who don't know how to enter the password from their env must use a string or not. I'm sorry if that's unnecessary.

crisgarlez:
In my first time with Laravel it took me a while to realize that the Database password did not work for me due to the use of special characters.

DB_PASSWORD='#pAssWord1234' instead DB_PASSWORD=#pAssWord1234

stancl:
hacktoberfest spam, I think there will be many of these in the coming days 🙃

PR #5436, opened 2020-10-01
[8.x] Better SPA Routing Support

paras-malhotra:
**Motivation**
Currently, there's no official guidance on how to register SPA routes in Laravel. I think the majority of folks use [`Route::fallback`](https://github.com/laravel/framework/pull/21299) to register a catch-all route for SPAs. While this does solve the problem to a certain extent, it is [not recommended for SEO](https://thegray.company/blog/single-page-application-spas-404s-seo) as it serves a 200 response code for every single route from the backend. Search engines such as Google call these [soft 404s](https://support.google.com/webmasters/answer/181708?hl=en) and currently have very little support to detect and flag such pages as 404s on their own.

There are other alternatives such as redirecting to a 404 page or serving a `no-index` meta tag. Ultimately, these are sub-par solutions. The best solution for SEO is to serve the correct 404 response code from the server. It not only solves the SEO issue at hand but also makes it easy for tools such as broken link checkers and analytics tools to flag errors.

This PR makes it easy for Laravel users to register SPA routes (or patterns) in a single file so as to properly serve 404 response codes for un-registered routes.

**Why not leave it to users to do this in their own apps**
I reckon this is the 80% use case. Most web apps are using SPAs today (and Laravel has traditionally supported SPA frameworks such as Vue out-of-the-box) and this is probably a challenge for the majority of users that have SPAs.

**Possible Extension**
Sometime down the road, we could even have SPA frameworks such as Vue or React generate a list of SPA route patterns from their routers as part of the build process and have PHP (the `routes/spa.php` file) cache and serve these routes.

### DIFF
--- a/app/Providers/RouteServiceProvider.php
+++ b/app/Providers/RouteServiceProvider.php
@@ -29,6 +29,15 @@ class RouteServiceProvider extends ServiceProvider
     // protected $namespace = 'App\\Http\\Controllers';
 
     /**
+     * The SPA route action for the application.
+     *
+     * When present, all routes in your routes/spa.php file will be directed to this action.
+     *
+     * @var array|string
+     */
+    protected $spaAction = 'SPAController@index';
+
+    /**
      * Define your route model bindings, pattern filters, etc.
      *
      * @return void
@@ -46,6 +55,15 @@ class RouteServiceProvider extends ServiceProvider
             Route::middleware('web')
                 ->namespace($this->namespace)
                 ->group(base_path('routes/web.php'));
+
+            Route::middleware('web')
+                ->namespace($this->namespace)
+                ->group(function () {
+                    collect(require base_path('routes/spa.php'))
+                        ->each(function ($path) {
+                            Route::get($path, $this->spaAction);
+                        });
+                });
         });
     }
 
--- a/routes/spa.php
+++ b/routes/spa.php
@@ -1,0 +1,18 @@
+<?php
+
+return [
+
+    /*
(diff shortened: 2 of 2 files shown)


GuidoHendriks:
Most apps are SPA's? What's your source on that?

paras-malhotra:
> Most apps are SPA's? What's your source on that?

I could provide numerous links but that's not really the point. Laravel has traditionally supported SPAs such as Vue out-of-the-box so I don't think anyone would say that Laravel doesn't want to support SPAs. It already does.

GuidoHendriks:
Change like this only makes sense if the majority of the apps would use it. I would say it wouldn't, and it could easily be done in your own app. But let's see what Taylor and the team thinks.

paras-malhotra:
I'm pretty sure there would be a larger number of apps that use SPA routes than those that use broadcast channels, and we already have a `routes/channels.php` file out of the box.

GrahamCampbell:
Most Laravel apps are not single page apps, I am pretty sure of that.

paras-malhotra:
> Most Laravel apps are not single page apps, I am pretty sure of that.

I was referring to web apps that have some kind of frontend, not all Laravel apps. Anyway, that's besides the point. I do think that SPAs is a very common use case for Laravel and Laravel has been supporting SPAs for quite a long time. SPA routing seems to be a common challenge when it comes to building SPAs in Laravel imo.

taylorotwell:
Thanks for the PR. I think this maybe makes more sense as a blog post or tutorial that others can read.

PR #5440, opened 2020-10-03
Deletes the sample file after the process.

selcukcukur:
Since we do not need the sample file after the process is completed, we can delete it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,8 @@
             "@php artisan package:discover --ansi"
         ],
         "post-root-package-install": [
-            "@php -r \"file_exists('.env') || copy('.env.example', '.env');\""
+            "@php -r \"file_exists('.env') || copy('.env.example', '.env');\"",
+            "@php -r \"file_exists('.env.example') || unlink('.env.example');\""
         ],
         "post-create-project-cmd": [
             "@php artisan key:generate --ansi"


vikas5914:
Since the .env does not include in git, we need the .env.example in production to make a .env file.

selcukcukur:
@vikas5914 It can be deleted after it has been created. There is no need. This command helps with this.

samundra:
Sorry to say but this is unnecessary PR. 

These should be managed by CI/CD during deployment process. No need to have it in post installation.

andrey-helldar:
@selcukcukur, That is, after installing the application, you do not want other developers in the team to see the file with the settings preset and devops could not correctly configure the application on the server?..

martindilling:
You shouldn't have any sensitive information in your .env.example, so this shouldn't be necessary :)

selcukcukur:
@andrey-helldar Exactly as you said. The developer does not need this file after the project is created. I thought it would be good to tell the composer to do this with cmd.

andrey-helldar:
> @andrey-helldar Exactly as you said. The developer does not need this file after the project is created. I thought it would be good to tell the composer to do this with cmd.

I didn't say that. Perhaps Google Translate translated my phrase incorrectly.
I'm saying that by deleting this file, you force other developers in your team to build it on their own instead of taking a ready-made preset.

selcukcukur:
> > @andrey-helldar Exactly as you said. The developer does not need this file after the project is created. I thought it would be good to tell the composer to do this with cmd.
> 
> I didn't say that. Perhaps Google Translate translated my phrase incorrectly.
> I'm saying that by deleting this file, you force other developers in your team to build it on their own instead of taking a ready-made preset.
> 
> This file contains a preset of environment settings that each developer and devops customizes for himself.
> 
> And this file should not contain any personal data.

I understand well.

PR #5441, opened 2020-10-05
[8.x] Update the badges to use shields.io

pataar:
Shields.io works a lot better than the old pugx.org.


driesvints (approved):

pataar:
Hmm, hold up. I guess it's better to target `laravel/framework` instead of `laravel/laravel`.

PR #5443, opened 2020-10-06
[8.x] Now extends the correct TestCase class

DilwoarH:
The issue with it loading up the PHPUnit Framework TestCase is that it does not create the application properly. This fix resolves this problem by pointing to the correct TestCase.

REF: https://github.com/facade/ignition/issues/181#issuecomment-626362299

This has caused a great deal of stress for a lot of people - hope this easy fix resolves their problems :)

### DIFF
--- a/tests/Unit/ExampleTest.php
+++ b/tests/Unit/ExampleTest.php
@@ -2,7 +2,7 @@
 
 namespace Tests\Unit;
 
-use PHPUnit\Framework\TestCase;
+use Tests\TestCase;
 
 class ExampleTest extends TestCase
 {


crynobone:
Why not move the test to "Feature" if it requires the framework?

DilwoarH:
> Why not move the test to "Feature" if it requires the framework?

Because it's a unit test, not a feature test - simple things like `config('blah')` should still work.

crynobone:
Many would argue loading the full framework isn't unit testing. Which is why it was default to load PHPUnit TestCase.

driesvints:
Write a feature test.

PR #5445, opened 2020-10-09
[8.x] Added 'LOG_LEVEL' env variable in .env.example

Ali-Shaikh:
Added 'LOG_LEVEL' env variable in .env.example to be consistant with the change in [#5442](https://github.com/laravel/laravel/pull/5442)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,6 +4,7 @@ APP_KEY=
 APP_DEBUG=true
 APP_URL=http://localhost
 
+LOG_LEVEL=
 LOG_CHANNEL=stack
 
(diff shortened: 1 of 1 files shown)


.env.example
```diff
@@ -4,6 +4,7 @@ APP_KEY=
 APP_DEBUG=true
 APP_URL=http://localhost
 
+LOG_LEVEL=
```
samundra, line 7: maybe also leave the default value `LOG_LEVEL=debug`

driesvints: Done

PR #5446, opened 2020-10-11
fix:remove-failed-jobs-table

shm379:
When there is no job table, there is no need for a failed job table. This must be done by artisan command _queue:failed-table_
@taylorotwell


paras-malhotra:
@shm379 job tables are only required for the database queue driver. That's why there is no jobs table.

shm379:
ok i think this problem and in older version of laravel the table not exists

GrahamCampbell:
The failed jobs table is used by all drivers.

PR #5450, opened 2020-10-20
Fixing Authenticate middleware for phpstan level 2

tentacode:
In a freshly installed vanilla Laravel app, using phpstan and larastan on level 2 (to 8) raises
an error, this commit fixes that by adding an explicit return :

```
Line   Http/Middleware/Authenticate.php
 ------ ---------------------------------------------------
  17     Method
         App\Http\Middleware\Authenticate::redirectTo()
         should return string|null but return statement is
         missing.
```

It's my first MR on laravel, please be gentle :pray:  (also not sure if this is the proper destination branch, I figured it would benefit the LTS to the 8 version).

P.S. I would really appreciate if it could count toward my Hacktoberfest goal :jack_o_lantern: (by adding the `hacktoberfest-accepted` on this PR :pray: )

### DIFF
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -17,5 +17,7 @@ class Authenticate extends Middleware
         if (! $request->expectsJson()) {
             return route('login');
         }
+
+        return null;
     }
 }


driesvints:
Feel free to do this in your own app. We aren't participating in hacktoberfest because of the influx of spam prs we've received already, sorry.

tentacode:
Well, I was not doing this for hacktoberfest but because I seems too bad that phpstan fails on a fresh app. I could care less about hacktoberfest but thought it could be a quick win for everyone using this tool :shrug:

driesvints:
The current method already returns `null` btw. We don't really optimize our code for third party tools like phpstan. If you want to use phpstan you're free to modify the code in your own app 👍

tentacode:
This is the only phpstan error on a new install, it's a level 2 error so it's kind of a bummer. I do understand your reasoning, but I don't understand how a quick fix/win here makes any issue.

driesvints:
The issue is that we have coding standards and one of them is that we don't add `return null` at the end of method because the language already returns `null` for you. It makes the code just that much shorter and more readable. Given, that could be a subjective opinion but it's our personal preference of doing things in the framework.

tentacode:
Ok thanks for the info :pray: this makes more sense, you could have started with that :D

tentacode:
You could say that I'm being a "little" stubborn, but I found a way to pass phpspec without having to add a return null, would that be more acceptable on your side ?

```php
    /**
     *
     * @return string|null|void
     */
    protected function redirectTo($request)
    {
        if (! $request->expectsJson()) {
            return route('login');
        }
    }
```

Just added `|void`.

Do feel free to tell me if I am being annoying and you just want me to stop posting, I'll take it :pray:

ondrejmirtes:
@tentacode It doesn't make much sense to combine `void` with anything else, either the function returns a value or it doesn't. You won't be able to write this in a native typehint on PHP 8.0 anyway (nor you're able to write `?void` on PHP 7.x).

tentacode:
Well, typically this methods "do both", unless strict typing is involved there is no way to force a method to return nothing, so here it "return something" or "do nothing" (which in this case do return null, I'm aware). Anyway that's a good :brain: discussion, but I was more looking at the added value of this to anyone who use phpstan, rather than writing the *perfect code*, I'm totally fine for it being your choice, but I defer in opinion :smile:.

Perhaps it's more a PR to add to phpstan, but I'm pretty sure it would be denied for the opposite reasons :joy: anyway I tried. Thanks folks :pray:

ondrejmirtes:
Yes, the reason why PHPStan requires `return null;` is because it's valuable to PHPStan's users to differentiate between "I want this method to `return null`" vs. "The method is supposed to return something but I forgot to add the `return`."

If there was a native `: ?string` return typehint the PHP language would require `return null;` to be there but since it's just in PHPDoc, the `return null` can be omitted.

But I'm not here to judge Laravel's coding practices.

tentacode:
hey @ondrejmirtes so I just looked into phpstan :smile: great comment :pray: https://github.com/phpstan/phpstan/issues/97

driesvints:
@tentacode you're definitely not annoying. You're kind in your replies and curious and I appreciate that 👍 

Like @ondrejmirtes said, combining `void` with anything else in the DocBlock doesn't really makes sense.

PR #5451, opened 2020-10-21
JSON & XML should be with indent size of 2(less then 4)

NI-AN:
JSON and XML seems more clearly and easy to understand with only indent size of 2.
4 indent size is too much when you have JSON file with lots of details a fortiori when we talking about XML files :).
moreover this way it makes JSON and XML looks close to their RFCs respectively.

source:
https://tools.ietf.org/html/rfc7159#page-4
https://tools.ietf.org/html/rfc3470#page-3

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,5 +11,5 @@ trim_trailing_whitespace = true
 [*.md]
 trim_trailing_whitespace = false
 
-[*.{yml,yaml}]
+[*.{yml,yaml,json,xml}]
 indent_size = 2


GrahamCampbell:
Thanks for this, however Laravel doesn't follow this format, in general. In particular, it is a convention among the PHP community to use 4 spaces for indentation in `composer.json` files.

Feel free to add this in your own apps, however. :)

NI-AN:
OK, 
I moved it to  [laravel/ideas](https://github.com/laravel/ideas) as a issue,  please don't harry to remove it from their. 

P.S
Feel free to add this to your apps too :) , this is not " PHP community convention" this is historical IDEs limitation.

Thank you for your patience.

PR #5454, opened 2020-10-22
Bug when you use a custom name for a primary key

roke22:
When you try to create a new primary key name in the postgresql database the name of the primary inside of the postgresql server is not your custom name is "yourtablename_pkey".

Then if you rename the table the and later try to remove that custom primary key you get an error because the generated name of the primary key is not the same and in the dropPrimary the name is omitted and not used.

You can reproduce the error executing the migrations with "artisan migrate" command. Hope this be usefull :)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,7140 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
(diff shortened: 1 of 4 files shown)


taylorotwell:
We don't accept these types of PRs on Laravel. You can send a fix PR to laravel/framework.

PR #5455, opened 2020-10-22
Error custom primary key

roke22:
Bug can be reproduce with the "php artisan test", added BugPrimaryKeyTest to reproduce the error.

When you try to create a new primary key name in the postgresql database the name of the primary inside of the postgresql server is not your custom name is "yourtablename_pkey".

Then if you rename the table the and later try to remove that custom primary key you get an error because the generated name of the primary key is not the same and in the dropPrimary the name is omitted and not used.


GrahamCampbell:
// cc @driesvints

driesvints:
Please try a support channel first and if you can identify a bug, open up an issue on the laravel/framework repo.

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://webchat.freenode.net/?nick=laravelnewbie&channels=%23laravel&prompt=1)

PR #5456, opened 2020-10-22
Revert to per user API rate limit

szepeviktor:
It was changed from per user to per application in ced3e50bca279f5a499a6d513fcdfd2a020c28ce

@iamgergo


taylorotwell:
Updated it to just user ID if available and fall back to IP.

taylorotwell:
Updated it to just user ID if available and fall back to IP.

PR #5459, opened 2020-10-25
[8.x] Remove MustVerifyEmail which is not used and already inherited

kylekatarnls:


driesvints:
This isn't inherited. Think you might have confused the trait with the contract.

kylekatarnls:
What for this unused contract is supposed to be there?

driesvints:
To indicate to newcomers to the framework that it's there and available if they want to implement email verification.

PR #5460, opened 2020-10-27
[6.x] Delete removed webpack flag

driesvints:
Webpack v4 removed this flag so running `npm run dev` or `npm run prod` isn't working on this version. With the upcoming Laravel Mix v6 this would have come up as well probably. I'm already sending in this PR so Laravel 6 installs at least can make use of the new Webpack version. Will merge upstream.

See https://github.com/laravel/ui/issues/174

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,12 +2,12 @@
     "private": true,
     "scripts": {
         "dev": "npm run development",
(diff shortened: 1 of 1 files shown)


driesvints:
Ping @JeffreyWay ^

JeffreyWay:
For Mix 6, we'll be shipping a `mix` CLI so you guys never have to deal with this stuff again. One of us will submit a PR after it's officially out.

driesvints:
Thanks @JeffreyWay, that sounds good 🙂

PR #5462, opened 2020-10-31
[8.x] Show all env variables in .env.example

pxdnbluesoul:
I wanted to see if there was any interest in an enhancement to .env.example to at least display all current env vars called for in mainline Laravel. I inevitably find myself looking through each config file with new releases to stare-and-compare at what's not present in .env.example that needs to be in my .env file. This does not include the Flare/Ignition vars.

This request also updates the config files to the values that are initially overwritten by .env.example so if erased they behave as expected.

There are currently comments to designate which config file they relate to but could be removed if that's too big a shift in conventions.


taylorotwell:
No plans to do this.

PR #5463, opened 2020-11-01
[8.x] Ignores csrf on logout route for guest users

snapey:
A problem with the CSRF token in the logout form is that it can become stale in at least the following three scenarios

- User has more than one tab open and logs out in one of them. The logout action on other tabs will now produce 419 Page Expired error
- User has a single tab open but leaves the site and does not return during the session duration. The CSRF token is then stale and pressing logout gives 419 error.
- The user uses the option to close all sessions on other devices. Logout on a terminated session will generate the error.

csrf protection for logged in users and guests for other than the logout route is unchanged

### DIFF
--- a/app/Http/Middleware/VerifyCsrfToken.php
+++ b/app/Http/Middleware/VerifyCsrfToken.php
@@ -3,6 +3,8 @@
 namespace App\Http\Middleware;
 
 use Illuminate\Foundation\Http\Middleware\VerifyCsrfToken as Middleware;
+use Illuminate\Support\Facades\Auth;
+use Closure;
 
 class VerifyCsrfToken extends Middleware
 {
@@ -14,4 +16,27 @@ class VerifyCsrfToken extends Middleware
     protected $except = [
         //
     ];
+
+
+    /**
+     * Handle an incoming request. Ignores csrf for guests at the logout route, so
+     * no error is thrown when posting to logout on an already expired session
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     *
+     * @throws \Illuminate\Session\TokenMismatchException
+     */
+    public function handle($request, Closure $next)
+    {
+        if(Auth::guest() && $request->route()->named('logout')) {
+        
+            $this->except[] = route('logout');
+            
+        }
+
+        return parent::handle($request, $next);
+    }
+
 }


app/Http/Middleware/VerifyCsrfToken.php
```diff
@@ -14,4 +16,27 @@ class VerifyCsrfToken extends Middleware
...
+     * @param  \Closure  $next
+     * @return mixed
+     *
+     * @throws \Illuminate\Session\TokenMismatchException
+     */
+    public function handle($request, Closure $next)
+    {
+        if(Auth::guest() && $request->route()->named('logout')) {
+        
+            $this->except[] = route('logout');
```
GrahamCampbell, line 35: Why are you pushing this to the except array on every execution of handle?

GrahamCampbell (changes requested):

paras-malhotra:
This PR assumes that the logout route would be a named route called `logout` and also that it's not a `get` route. All auth scaffolding is in separate packages so it begs the question whether this repo is the right place for this kind of logic.

app/Http/Middleware/VerifyCsrfToken.php
snapey: It only changes the except array if the user is guest AND the route is logout.  This combination only occurs under exceptional circumstances.

snapey:
> This PR assumes that the logout route would be a named route called `logout` and also that it's not a `get` route. All auth scaffolding is in separate packages so it begs the question whether this repo is the right place for this kind of logic.

Logout should never be a GET route, and if it were then csrf would be irrelevant and not checked by the middleware.
I agree the logout route may be called something else. There is no way of grokking this from middleware. There is precedent in that the Auth middleware assumes a 'login' route

deleugpn:
I think all three scenarios provided deserve a Page Expired result. What is the reason for it to change and why not change it only on your own app?

snapey:
> I think all three scenarios provided deserve a Page Expired result. 
Under the covers maybe, but to joe public, 419 Page Expired means nothing to them and they have no obvious way of recovery. By default there are no links off that page.

> What is the reason for it to change and why not change it only on your own app?

I can add it into my own apps of course, but I have to do this every time when I feel that the majority of applications would benefit from this change.

It fits with a strategy of defensive programming.  Why allow an exception leak to the user when it can be easily handled and with no downside.

PR #5464, opened 2020-11-03
[9.x] Remove Event facade alias

imanghafoori1:
As Laravel Swoole users will experience a silent conflict with alias `\Event`
https://github.com/imanghafoori1/laravel-microscope/issues/79#issuecomment-698201489

Any laravel package which uses the Alias of facade (`\Event::`) will become incompatible with laravel-swoole. In other words: package maintainers should not use `\Event` otherwise their package won't work properly on swoole.
So I thought commenting out the alias for laravel 9.x will alert the package developers not to use the `\Event` and make their packages swoole compatible.

- Plus is it really needed to have the alias? aliases are meant to be used in blade files where importing a class is clunky and we never fire events in blade files.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -203,7 +203,7 @@ return [
         'Crypt' => Illuminate\Support\Facades\Crypt::class,
         'DB' => Illuminate\Support\Facades\DB::class,
         'Eloquent' => Illuminate\Database\Eloquent\Model::class,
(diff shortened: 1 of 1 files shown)


taylorotwell:
People can do this in their own application. It's not really my job to disable features to account for Swoole.

PR #5467, opened 2020-11-10
issue: Route Model Binding don't behaves as expected

agustavo87:
Route Model Binding as docs description[1] says, is suppoused to
binds a route parameter (e.g. {user}) to a function parameteter
where  it is type hinted as an Eloquent model _and_ its name
matches the name of the route parameter:

> Laravel automatically resolves Eloquent models defined in routes or controller actions whose type-hinted variable names match a route segment name.

This is not the case with routes with multiple parameters. It
don't match the names of the parameteers. It matches the order.
E.g.:

```php
Route::get('/{locale}/{user}', function (\App\Models\User $user) {
    dd($user);
});
```

Visiting ```/es/1``` throws ```Type Error```

>TypeError
Argument 1 passed to Illuminate\Routing\RouteFileRegistrar::{closure}() must be an instance of App\Models\User, string given, called in D:\Gustavo\code\laravel\vendor\laravel\framework\src\Illuminate\Routing\Route.php on line 230

But it does not if the function parameters **order** (not the _name_) matches the URI.

```php
Route::get('/{locale}/{user}', function ($a, \App\Models\User $user) {
    dd($user);
});
```

It happens with Implicit and Explicit Bindings methods.

[1]https://laravel.com/docs/8.x/routing#implicit-binding

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,7263 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
+        "This file is @generated automatically"
+    ],
+    "content-hash": "fac753ae2702317b354e7eb283f28cbc",
+    "packages": [
+        {
+            "name": "asm89/stack-cors",
+            "version": "v2.0.2",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/asm89/stack-cors.git",
+                "reference": "8d8f88b3b3830916be94292c1fbce84433efb1aa"
+            },
+            "dist": {
+                "type": "zip",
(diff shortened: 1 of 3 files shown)


driesvints:
Your parameter order needs to match. Also not sure why you sent in a PR to the skeleton for this.

agustavo87:
Then the docs needs to be changed, anywhere it says that the order needs to match, it clearly specify that:

> Since the ```$user```` variable is type-hinted as the ````App\Models\User Eloquent``` model and the ***variable name*** _matches_ the ```{user}``` ***URI segment***, Laravel will automatically inject the model instance that has an ID matching the corresponding value from the request URI.

1. Has to be type-hinted.
2. Match name of variable and URI segment.

No specification of parameter order. So it's mis-documented.

PD: I PR becouse there's no issue section.

driesvints:
@agustavo87 the issue tracker for the framework is here: https://github.com/laravel/framework/issues

Feel free to PR something to the docs if anything is not clear 👍

PR #5470, opened 2020-11-17
Rename the file generated called  master.blade.php to main.blade.php

Davisy:
I understand that the master.blade.php is created automatically when you run the auth artisan command. I suggest renaming the file from **master.blade.php** to **main.blade.php**.

I don't know exactly the source code responsible to create the file, but I hope if the creator accept the idea he can do the changes.

Thanks


driesvints:
Seems like branching went wrong here.

PR #5473, opened 2020-11-18
Use assertOk instead of assertStatus

insoutt:
If this pull request is approved I am planning to create a pull request in Laravel Framework to update the stub test (https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Console/stubs/test.stub) in order to generate tests with `php artisan make:test` with `assertOk` method by default.

### DIFF
--- a/tests/Feature/ExampleTest.php
+++ b/tests/Feature/ExampleTest.php
@@ -16,6 +16,6 @@ class ExampleTest extends TestCase
     {
         $response = $this->get('/');
 
-        $response->assertStatus(200);
+        $response->assertOk();
     }
 }


DirkBlade:
I think assertStatus is more likely to lead newcomers like myself to the other http tests than assertOK.  I had guarded / so it was easy to change assertStatus to 301 when tests failed out of the box and I wasn't inclined to go looking for assertRedirect or delete the test.

insoutt:
The `assertOk` method was implemented in Laravel 5.6 so it is not a new feature. I think it is not difficult to understand and also this is a starter template you can customize it according to the needs of your application.

PR #5476, opened 2020-11-23
[8.x] Missing auth config

Propaganistas:
See https://github.com/laravel/framework/blob/22a0f06a2923c848d99a22695a4a225f2a3b762f/src/Illuminate/Auth/Notifications/VerifyEmail.php#L85


taylorotwell:
Because I'm not trying to encourage that it be changed. It seems like a sensible default. If you want to change it you can add it to your own application.

PR #5482, opened 2020-12-09
Document service providers

tontonsb:
Hey,

This is a WIP and I need help.

Jumping on the documentation bandwagon, this is the thing I have been wanting for a long time. I want to understand what the lines in my config mean and most of them are explained across all of the config files. But not providers. There's at least a facade reference in the docs, but the meaning of providers is totally obscure. Obviously many of them are self explaining, but some terms like Bus, Pipeline, Foundation are not explained in the docs at all.

I was [initially](https://github.com/laravel/ideas/issues/1741) thinking this should go into docs, but maybe the config file would be more appropriate as you won't be wondering "Why is the broadcast provider the only one that's commented out? And why are there two of them?" before looking at the file itself.

Unfourtunately, I don't know all the Laravel that well to describe the intention and scope of each provider. So I am putting this out as a starting point, can anyone help?

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -140,7 +140,9 @@ return [
          * Laravel Framework Service Providers...
          */
         
-        // Enables authentication and authorization (gates)
+        /*
+         * Enables authentication and authorization (gates).
+         */
         Illuminate\Auth\AuthServiceProvider::class,
         
         /*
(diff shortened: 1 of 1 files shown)


config/app.php
```diff
@@ -139,27 +139,119 @@
         /*
          * Laravel Framework Service Providers...
          */
+        
+        // Enables authentication and authorization (gates)
```
antonkomarev, line 143: Single line comment while other providers has multiline comment

driesvints:
No plans for this sorry

PR #5483, opened 2020-12-09
Update .env.example

bakhtiyor:
Changed DB_HOST=mysql to DB_HOST=127.0.0.1


taylorotwell:
It is configured for Laravel Sail out of the box. If you aren't using that for local development then you can change it for your own needs.

PR #5486, opened 2020-12-09
added missing env variables for smtp config section

VahanMargaryan:
added following env variables for allow use self hosted mail servers with self signed certificates.
To enable self signed certificates in your .env file set
```
MAIL_STREAM_ALLOW_SELF_SIGNED=true
```
also when you need to skip verifying peers then set
```
MAIL_STREAM_VERIFY_PEER=false
MAIL_STREAM_VERIFY_PEER_NAME=false
```

| Variable  | Default |
| ------------- | ------------- |
| MAIL_TIMEOUT | null |
| MAIL_AUTH_MODE | null |
| MAIL_STREAM_ALLOW_SELF_SIGNED | false |
| MAIL_STREAM_VERIFY_PEER | true |
| MAIL_STREAM_VERIFY_PEER_NAME | true |

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -41,8 +41,15 @@ return [
             'encryption' => env('MAIL_ENCRYPTION', 'tls'),
             'username' => env('MAIL_USERNAME'),
             'password' => env('MAIL_PASSWORD'),
-            'timeout' => null,
-            'auth_mode' => null,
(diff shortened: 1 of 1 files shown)


taylorotwell:
Add it to your own application if you need it.

VahanMargaryan:
But about these configurations there is no documentation available, I think you can put it into the core laravel project.

PR #5488, opened 2020-12-10
[8.x] Added DBAL configuration to database.php

TomHAnderson:
This PR is in support of a PR on the `laravel/framework` project: https://github.com/laravel/framework/pull/35591  

This PR is a product of the discussion at https://github.com/laravel/framework/issues/35498

Attempting to change a column from a datetime to a timestamp results in the following exception:
Unknown column type "timestamp" requested.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -1,5 +1,6 @@
 <?php
 
+use Illuminate\Database\DBAL\TimestampType;
 use Illuminate\Support\Str;
 
 return [
@@ -144,4 +145,20 @@ return [
 
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | DBAL Mappings
+    |--------------------------------------------------------------------------
+    |
+    | If your migrations include ALTER commands then the library
+    | `doctrine/dbal` is used to make those changes.  That library supports
+    | Laravel specific data type(s).  This section maps those datatypes.
+    |
+    */
+
+    'dbal' => [
+        'types' => [
+            'timestamp' => TimestampType::class,
+        ],
+    ],
 ];


driesvints:
@TomHAnderson I've converted this one as a draft until the framework PR is ready. Thanks

config/database.php
```diff
@@ -144,4 +145,20 @@
...
+    |
+    | If your migrations include ALTER commands then the library
+    | `doctrine/dbal` is used to make those changes.  That library supports
+    | Laravel specific data type(s).  This section maps those datatypes.
+    |
+    */
+
+    'dbal' => [
+        'types' => [
+            'timestamp' => TimestampType::class,
```
driesvints, line 161: I'm guessing this will currently fail if you don't have dbal installed (which is an optional in Laravel). I'm thinking we could replace this PR with a new part in the docs to go alongside the main PR to the framework.

TomHAnderson: No, not true!  If you have an alter statement then Laravel will throw an exception telling you to install doctrine/dbal just like it always has.  Nothing I've done in this PR or https://github.com/laravel/framework/pull/35591 will fail in any way new.  

For the sake of brevity only should this PR be discarded for it does no harm and only provides benefit.  But the benefit requires a migration that alters the database.  I think alter statements are essential to proper development in the truest sense of the word which, as I see it, is progressive changes.  That being said, if the configuration to manage alter statements is not included then the exception thrown when `doctrine/dbal` is required should be changed to include a link to the new documentation.

taylorotwell:
Unlikely I'll be adding this to the skeleton.

PR #5489, opened 2020-12-10
[8.x] Adds flag to declare the container volume as the source of truth

benfreke:
Adds the `:delegated` flag to volumes to improve Mac performance. 

Running the command `time sail composer update` went from 46 seconds down to 26 seconds. 
Times will vary for other developers but it is a significant speed increase

This flag used to be documented on the Docker docs but it's been removed and I can't find anything official about this anymore.


taylorotwell:
If you can't find anything official about it then I'm not comfortable merging it.

PR #5490, opened 2020-12-10
[8.x] Add default MAIL_FROM_ADDRESS

bramceulemans:
Currently sending mail does not work out of the box with Sail, I thought maybe it's a good idea to add a default value to `MAIL_FROM_ADDRESS` in `.env.example`.


taylorotwell:
You can set this value to what makes sense for your own application.

PR #5491, opened 2020-12-11
5.5

zhaiduting:
Node-sass should be updated to v5 for Node 15. 
Laravel+Vue+TypeScript has been tested successfully. Now we can write Vue components using both TS and JS in Laravel 5.5.


zhaiduting:
I hope that Laravel can support TypeScript, so I pulled this branch. There is no need to worry that the introduction of TS will affect development habits, because developers can still use JS to write scripts. I have tested #5491 and there is no problem. Thank you very much

zhaiduting:
I hope that Laravel can support TypeScript, so I pulled this branch. There is no need to worry that the introduction of TS will affect development habits, because developers can still use JS to write scripts. I have tested #5491 and there is no problem.

PR #5492, opened 2020-12-13
update variables

muath-ye:
```localhost``` is a better option when configuring a local testing environment.


GrahamCampbell:
Why?

NationwidePharma:
Localhost will break some configurations. Localhost != 127.0.01 as far as MySQL is concerned and also 127.0.0.1 solves performance problems in some cases.

Get Outlook for iOS<https://aka.ms/o0ukef>


​SDVOSB CERTIFIED | NABP ACCREDITED

Justin Lawrence, M.S.
Data Architect

1270 N. Loop 1604 East, STE 13, Suite 1306
SAN ANTONIO, TX  78232
210.778.5406
www.nationwidepharmaceutical.com
bridging the gap between the government and healthcare industry

muath-ye:
I always see hosting provide me localhost as host for my MySql DB connection.

PR #5494, opened 2020-12-18
Fix mailhog host.

redbastie:
This fixes the mailhog host on clean installs. Otherwise, we get the error `Connection could not be established with host mailhog`.

This value should be set to `localhost` as per the homestead mailhog config docs: https://laravel.com/docs/8.x/homestead#configuring-mailhog

Setting this value to `localhost` resolves the issue out of the box. 

P.S. mailhog rocks :)


taylorotwell:
It's configured to work with Sail. Before Sail it was set to Mailtrap I believe.

redbastie:
> It's configured to work with Sail. Before Sail it was set to Mailtrap I believe.

Ok. Understood.

PR #5496, opened 2020-12-20
[8.x] Fixed redirection on session timed out

voyula:
Each session timed out on remember me mode, Laravel makes re-login, if authenticated redirection haven't intended support, user will be lost the last page and will be redirected to home page.


voyula:
Also why Laravel reloading page on every session renews?
Post requests can be conflict?

taylorotwell:
Make the change you want in your own application.

PR #5497, opened 2020-12-20
Remove --no-progress from npm run prod key & use laravel-mix 6

ziming:
laravel mix 6 is out and I got an "Unknown argument: --no-progress" error when trying out npm run prod. 

So I guess the latest webpack removed it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
         "watch-poll": "npm run watch -- --watch-poll",
         "hot": "cross-env NODE_ENV=development node_modules/webpack-dev-server/bin/webpack-dev-server.js --inline --hot --disable-host-check --config=node_modules/laravel-mix/setup/webpack.config.js",
         "prod": "npm run production",
-        "production": "cross-env NODE_ENV=production node_modules/webpack/bin/webpack.js --no-progress --config=node_modules/laravel-mix/setup/webpack.config.js"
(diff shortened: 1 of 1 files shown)


marky291:
Laravel does not use v6 at the moment so these changes are not required, the dependencies you are proposing still use v5.0.1 also.

ziming:
package.json doesn't include laravel-mix at all. But for people who install the latest laravel-mix stable (v6), npm run prod will break with --no-progress. Removing --no-progress doesn't break things.

marky291:
The file you are committed is using "laravel-mix": "^5.0.1", if you propose update to v6 please change the commit to be v6.

ziming:
Ah okay, spotted it now. Let me chnage

GrahamCampbell:
I don't think upgrading is as simple as just changing the version?

marky291:
Only issues I faced personally when upgrading to v6.17 was the —no-progress flags.

ziming:
it's running fine on my site in production after these changes.

marky291:
There are a few more upgrades required for the stable version of Mix 6.
(https://laravel-mix.com/docs/6.0/upgrade) which includes a new mix cli for npm commands.

driesvints:
Thanks but please search for previous issues. @JeffreyWay will update the skeleton once he's ready for it.

PR #5502, opened 2020-12-23
[8.x] Update typehint of Authenticate Middleware

bistory:
Typehint is not correct as the method can return void.

Phpstan is returning an error about this typehint : 

```
------ ------------------------------------------------------------------------------------------------------------------ 
  Line   app/Http/Middleware/Authenticate.php                                                                              
 ------ ------------------------------------------------------------------------------------------------------------------ 
  17     Method App\Http\Middleware\Authenticate::redirectTo() should return string|null but return statement is missing.  
 ------ ------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -10,7 +10,7 @@ class Authenticate extends Middleware
      * Get the path the user should be redirected to when they are not authenticated.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @return string|null
+     * @return string|void
      */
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
👎 `void` cannot appear in a union type.

bistory:
@GrahamCampbell So what can be done to content Phpstan ?

driesvints:
Please see previous attempts

bistory:
Ok I've read #5450 and understand we'll have to live with it ;-)
Thanks for the reply @driesvints

PR #5503, opened 2020-12-28
Update User.php

dkvarda:
Unnecessary line


driesvints:
Please see many attempts before.

PR #5508, opened 2021-01-04
Git ignore PHPStorm files & Mac files.

redbastie:
I find myself having to add these to every project, would be nice if this was included out of the box. Thank you for your consideration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,10 @@
+/.idea
 /node_modules
 /public/hot
 /public/storage
 /storage/*.key
 /vendor
+.DS_Store
 .env
 .env.backup
 .phpunit.result.cache


driesvints:
Heya. These belong in a [global `.gitignore`](https://docs.github.com/en/free-pro-team@latest/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer). Please try to search previous PRs before opening up a new one as this has been attempted many times before.

algorithmhash:
You could create a CLI or a bash script that creates a select amount of .gitignore items or create a preset environment, that then just does something like:
```
Choose or type what to ignore:
type:
  node_modules
  vendor
  **other_packages_etc**
 choose:
   * node_modules
      vendor
   * other_etc
```
Pretty simple.

redbastie:
I just created a global git ignore file on my MacBook as per the advice. Thanks.

PR #5510, opened 2021-01-05
Update .gitignore to add .idea folder

vineethkrishnan:
Added .idea directory to gitignore. 
Sometimes it gets pushed to the repository by accident.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /public/storage
 /storage/*.key
 /vendor
+/.idea
 .env
 .env.backup
 .phpunit.result.cache


GrahamCampbell:
This should be added to your system .gitignore file, and not your app's.

Oxicode:
> Added .idea directory to gitignore.
> Sometimes it gets pushed to the repository by accident.

I like it. I prefer to add ".idea" in the global .gitignore file. But nobody does it.

GrahamCampbell:
People will only do it if we people stop adding it to projects.

lk77:
you can change this file in your project

GuidoHendriks:
It's even debatable wether or not you should ignore the whole folder. See GitHub's suggested `.gitignore` and the referenced link to JetBrains' website:

https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

PR #5512, opened 2021-01-07
Add healthchecks for docker-compose services

abenettt:


GrahamCampbell:
What does this fix?

abenettt:
This is less of a fix and more of an addition that you can check that your services are running correctly and can be used.
Also when when depending on services docker will wait on the services to be healthy before starting the service that depends on them, so you can be sure they're available.

driesvints:
This probably needs to go into Laravel Sail first before adopting here.

PR #5514, opened 2021-01-12
[8.x] Update TrustProxies.php

driesvints:
Solve the deprecation warning.

### DIFF
--- a/app/Http/Middleware/TrustProxies.php
+++ b/app/Http/Middleware/TrustProxies.php
@@ -19,5 +19,5 @@ class TrustProxies extends Middleware
      *
      * @var int
      */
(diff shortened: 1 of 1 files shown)


mikemand:
I was getting ready to open a PR to update the docs (https://laravel.com/docs/8.x/requests#configuring-trusted-proxies) to support this PR, but the tip about ELB is contradictory now. Should the tip be removed? Or is it true that your `protected $headers` should only contain `Request::HEADER_X_FORWARDED_AWS_ELB;`?

fideloper:
Thanks for this update Dries, I haven't figured this out myself yet :D 

I'm guessing this works thanks to some bitwise math that allows all/any of these headers?

driesvints:
@fideloper yeah I think so.

PR #5515, opened 2021-01-12
[8.x] Hide .env.bak as well as .env.backup in .gitignore

gemmadlou:
Using the Laravel curl installation command, it creates a `.env.bak` file instead of `.env.backup`. Changing the .gitignore to `.env.ba*` hides both.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@
 /storage/*.key
 /vendor
 .env
-.env.backup
+.env.ba*
 .phpunit.result.cache
 docker-compose.override.yml
 Homestead.json


SjorsO:
For reference, `.env.backup` is created when running Dusk tests. It was added it in https://github.com/laravel/laravel/pull/5046

GrahamCampbell:
@gemmadlou This PR was reverted.

gemmadlou:
> @gemmadlou This PR was reverted.

I see. Was the submission incorrect? Or the way I submitted it was incorrect? So I know for next time.

GrahamCampbell:
Because we don't want to ignore `env.ba*`, since people could have environment names starting with `ba`. Instead, we just changed the install script to not leave that file lying around.

gemmadlou:
> Because we don't want to ignore `env.ba*`, since people could have environment names starting with `ba`. Instead, we just changed the install script to not leave that file lying around.

That's an even better solution. Thanks for the fix @GrahamCampbell

GrahamCampbell:
Was actually fixed by Taylor. ;)

.gitignore
```diff
@@ -4,7 +4,7 @@
 /storage/*.key
 /vendor
 .env
-.env.backup
+.env.ba*
```
robertotcestari, line 7: não gostei disso.

robertotcestari: nossa

PR #5516, opened 2021-01-12
Stop Env Commits

exussum12:
This is not a laravel issue per se, but there are a portion of users who
change the .env.example file with real data and then commmit it, or even
just commit the generated .env file.

This change adds a hook in to check the commits to see if they contain
an .env file and gives users a way to commit either way (ie they
do know what they are doing and want to skip over the message, or they
were not aware of what they were doing and a way to recover from that
(without losing their changes).

There are a few bots that are actively scanning for these files as
laravel has a large user base and people commit to this file. AWS
Credentials and mail credentials are most common.

This may not be the best solution, but happy to have a discussion on how
to prevent users of laravel making these mistakes which can be costly.

### DIFF
--- a/.git_hook_precommit
+++ b/.git_hook_precommit
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+# A hook to try and prevent users committing credentials
+
+if git rev-parse --verify HEAD >/dev/null 2>&1
+then
+	against=HEAD
+else
+	# Initial commit: diff against an empty tree object
+	against=$(git hash-object -t tree /dev/null)
+fi
+
+git diff --cached --name-only --diff-filter=AM -z $against | grep "^.env" > /dev/null
+	if [ $? -eq 0 ]
+then
+	cat <<\EOF
+Error: It appears your trying to commit changes to an environment file
+
+This is probably not what you want to do, if you commit these changes to a public place
+such as Github, these credentials will also be public.
+
+If you know what you are doing you can disable this check using:
+
+  git commit --no-verify
+
+This will disable the check on this commit only.
+
+Otherwise use:
+
+  git restore --staged -- .env*
+EOF
+	exit 1
+fi
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 /vendor
 .env
 .env.backup
+.env.example
 .phpunit.result.cache
 docker-compose.override.yml
 Homestead.json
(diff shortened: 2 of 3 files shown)


taylorotwell:
But the example file is supposed to be committed to source control?

exussum12:
People change these all the time, which would be covered by the "if you know what your doing" case. For them its one extra step.

Eg 
uggcf://tvguho.pbz/nxunyvq4493/svkceb/oybo/1036non03q744662or7p16106o815qpsnn58qs60/.rai.rknzcyr

(rot13'ed to semi obscure it) 

That is clearly not what the user intended to do, as mentioned in the first message there are bots scraping for this. I agree its the intention so a shared group of people can easily get up and running, but it also causes accidental commits. The default version contains AWS keys which could be used for any purpose, as its likely the people doing this have not set up IAM in a secure way either.

exussum12:
@taylorotwell  @driesvints  @GrahamCampbell 
Sorry for the tags, but is there any chance of a second review / discussion of this ?

The PR keeps it .env.example as it is supposed to be committed, it just makes you confirm you want to change, so one extra command for a laravel maintainer, or a maintainer of a team project using laravel is a downside, but hopefully enough to catch developers accidentally committing real credentials to public repos(without knowing!) is worth that change. 

Its not a large percentage of users that leak their secrets like this, but as laravel has such a large community even a small percentage is big numbers, the framework should protect against mistakes like this. 

As mentioned in the first post, this solution may not be the best, but this is a problem (IMO) which does need resolving. Laravel users are being targeted.

GuidoHendriks:
It's not the frameworks job to protect developers against themselves. Especially not if it gets in the way of the intended use.

exussum12:
image

8 commits in the past year would have had to run an extra command for the core project. Obviously much harder to get stats for projects using laravel, but changing .env.example I would never think is a common job, and its not prevented, there are instructions on how to bypass the check. Checking a few projects that I know about its something committed to once or a handful of times a year.

The framework *does* protect developers against themselves, there are a number of places in any framework which prevent things developers do that they may not be aware of and make the secure alternatives the path of least resistance, That's one of the selling points of frameworks in general.

https://vegibit.com/25-popular-applications-built-with-laravel/  lists "popular" not sure how they define that, but the projects in that list have extremely few commits on this file. 

Happy for any suggestions on how to stop people doing this, even if its parsing the file to check for AWS key like regex patterns or check keys that contain password / secret, the values are only null / replaceme or something before triggering this message.

I appreciate the people who will be reading this would never make these mistakes, but developers do make these costly mistakes and in this case at a reasonable scale.

PR #5518, opened 2021-01-13
[9.x] Mailhog's forward ports configurable

acabutto:
To avoid port clash between different projects running on the same machine, it could be useful making mailhog's port configurable.


driesvints:
@acabutto can you update Sail as well?

PR #5521, opened 2021-01-15
Change hash to readable password

kossa:


driesvints:
Using the hash itself prevents the hashing operation.

PR #5524, opened 2021-01-21
[8.x] Add middleware responsible for removing trailing slashes 

aurawindsurfing:
It is important as this applies to all applications made with laravel and it is bad for SEO as Google sees every single page twice. The correct behaviour is to create 301 for ever page with trailing slash to avoid content duplication in search results.

BTW even main laravel.com does not have this implemented.

Here is the source: https://developers.google.com/search/blog/2010/04/to-slash-or-not-to-slash

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -37,6 +37,7 @@ class Kernel extends HttpKernel
             \Illuminate\View\Middleware\ShareErrorsFromSession::class,
             \App\Http\Middleware\VerifyCsrfToken::class,
             \Illuminate\Routing\Middleware\SubstituteBindings::class,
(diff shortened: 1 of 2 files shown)


driesvints:
This is something that your web server should take care of and not the app itself.

GrahamCampbell:
Agree with @driesvints.

taylorotwell:

aurawindsurfing:
@driesvints @GrahamCampbell I agree however since I'm using only Laravel ecosystem it does not look like Forge is doing that either. So it should / could be addressed on either of the the products. Thanks you for quick response!

@taylorotwell Thank you for replay.

aurawindsurfing:
In case anyone is looking for this: https://github.com/aurawindsurfing/slashremovemiddleware

PR #5526, opened 2021-01-26
Fix project name on welcome resource

AndresReyesDev:
Set the name of the project as the title of the welcome resource. 

This was configured this way before and is still the case in other projects such as Laravel Breeze.

Have a nice day!


AndresReyesDev:
Sorry, I read the previous posts...

PR #5529, opened 2021-01-31
Add missing parentheses

D4nyi:
To fulfill PSR-12 Code Style requirements, [reference](https://www.php-fig.org/psr/psr-12/#4-classes-properties-and-methods).
> When instantiating a new class, parentheses MUST always be present even when there are no arguments passed to the constructor.

### DIFF
--- a/artisan
+++ b/artisan
@@ -33,8 +33,8 @@ $app = require_once __DIR__.'/bootstrap/app.php';
 $kernel = $app->make(Illuminate\Contracts\Console\Kernel::class);
 
 $status = $kernel->handle(
-    $input = new Symfony\Component\Console\Input\ArgvInput,
-    new Symfony\Component\Console\Output\ConsoleOutput
+    $input = new Symfony\Component\Console\Input\ArgvInput(),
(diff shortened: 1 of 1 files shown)


bkintanar:
I believe this PR will be closed pretty quickly. Laravel codebase has a lot of files that PSR-12 isn't implemented.

Muhammad-Sarfaraz:
@bkintanar Yeah it has,there was a closed issue in laravel/ideas [issue](https://github.com/laravel/ideas/issues/1829) about adopting PSR-12.

D4nyi:
I see, but I think it’s still a good practice nonetheless and also the offical PHP docs use it that way.
Of course if it isn't matching with Laravel's coding standards you can close this issue.

GrahamCampbell:
Laravel's code style guide is defined by its styleci.yml config file. In Laravel, brackets at the end of a class instantiation where there are no parameters are optional, and usually omitted.

GrahamCampbell (changes requested):
👎 Laravel does not follow PSR-12.

PR #5530, opened 2021-02-04
[8.x] Remove unused import

danilopinotti:


crynobone:
See https://github.com/laravel/laravel/pulls?q=is%3Apr+is%3Aclosed+remove+unused

GrahamCampbell:
See also https://github.com/laravel/laravel/blob/8.x/.styleci.yml#L4. If we wanted to remove all unused imports from the example app, we'd have not disabled that fixer.

danilopinotti:
Thanks for the explanation!

PR #5536, opened 2021-02-15
Change "Login" text to "Log in"

markjaquith:
As this is used as a verb, like its sibling "Register", the verb form ("Log in") should be used, instead of the noun form ("Login").

The full argument is here: http://notaverb.com/login and https://grammarist.com/spelling/log-in-login/

The short version is: the past tense is "logged in" not "logined", so the present imperative tense is "log in" (but you can use "login" as a noun, as in "I have a login on that site").

This also would bring Laravel in line with things like Facebook, Twitter, WordPress, Drupal, and more.

Screen Shot 2021-02-15 at 3 51 38 PM
Screen Shot 2021-02-15 at 3 57 22 PM
Screen Shot 2021-02-15 at 3 57 50 PM
Screen Shot 2021-02-15 at 3 58 43 PM

If accepted, I'm happy to open PRs for Jetstream and Breeze, which in addition to using "Login" as a verb in several places, also use "Logout", about which the same arguments apply.

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -27,7 +27,7 @@
                     @auth
                         <a href="{{ url('/home') }}" class="text-sm text-gray-700 underline">Home</a>
                     @else
-                        <a href="{{ route('login') }}" class="text-sm text-gray-700 underline">Login</a>
+                        <a href="{{ route('login') }}" class="text-sm text-gray-700 underline">Log in</a>
 
                         @if (Route::has('register'))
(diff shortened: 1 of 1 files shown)


taylorotwell:
Actually reverted this for now. I would rather have this done with PRs to everywhere it needs to be done at once. That way it's not half-way done.

markjaquith:
> I would rather have this done with PRs to everywhere it needs to be done at once. That way it's not half-way done.

I can go open Jetstream and Breeze ones. You care about Laravel/UI too?

hsm2k3:
good job noticing that @markjaquith !

PR #5538, opened 2021-02-16
Aid static analysis in index.php

szepeviktor:
- `app.php` file is not an object
- `make()` has `@return mixed` return type

Please merge. I have to add these 2 lines to every Laravel installation.
Thanks!


GrahamCampbell:
Why would anyone be analyzing or editing that file?

szepeviktor:
I care about every byte in source code and in traffic too.

PR #5545, opened 2021-02-27
[8.x] Remove unused javaScript

ninjaparade:
lodash seems to be included and included in bootstrap.js but it's never used. Thought I'd make this PR if people are okay with removing this or is it used that often that it should be left in the scaffolding of a fresh application?

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
     "devDependencies": {
         "axios": "^0.21",
         "laravel-mix": "^6.0.6",
-        "lodash": "^4.17.19",
         "postcss": "^8.1.14"
     }
 }
--- a/resources/js/bootstrap.js
+++ b/resources/js/bootstrap.js
@@ -1,5 +1,3 @@
-window._ = require('lodash');
-
 /**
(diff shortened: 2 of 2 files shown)


driesvints:
I also can't find any usages in Jetstream, UI or Breeze. Spark, Nova, Horizon and Telescope still use lodash but those come with their own asset pipeline.

GrahamCampbell:
I think it was included because it's a common thing people use, and serves as a food example of how to load packages (in the bootstrap file).

taylorotwell:
Basically echo @GrahamCampbell

browner12:
A year later is it time to revisit this PR? From my understanding, there are more modern and native ways to accomplish basically everything Lodash provides. By leaving it in our scaffolding, we may be inadvertently encouraging less than modern JS coding standards.  And possibly causing inflated build sizes for those unaware.

https://thejs.dev/jmitchell/its-time-to-let-go-of-lodash-nqc
https://medium.com/techspiration/why-you-shouldnt-use-lodash-f8504d7b7383

PR #5546, opened 2021-03-01
[8.x] Don't trim `current_password`

rodrigopedra:
Inspired by https://github.com/laravel/framework/pull/36415

As JetStream/Fortify uses a `current_password` field when allowing a user to change their password, and as JetStream is one of the starter kits listed on the docs, this PR adds `current_password` in the `$except` option of the `TrimStrings` middleware.

### DIFF
--- a/app/Http/Middleware/TrimStrings.php
+++ b/app/Http/Middleware/TrimStrings.php
@@ -12,6 +12,7 @@ class TrimStrings extends Middleware
      * @var array
      */
     protected $except = [
+        'current_password',
         'password',
         'password_confirmation',
     ];


gocanto:
I would personally expect it to be trimmed if the given `app/Http/Middleware/TrimStrings.php` middleware is guarding those routes.

rodrigopedra:
Hey @gocanto thanks for your reply.

As `password` was never trimmed on stock installation, a user could register themselves using a password starting or ending with one or more spaces.

Now if `current_password` gets trimmed, they would never be able to change their password because the `current_password` would mismatch their previously created `password`.

PR #5547, opened 2021-03-01
[8.x] Add explanation about effect of "timezone" config

browner12:
I think it's important to be explicit that this configuration value affects how dates are stored in the database, and that it doesn't simply adjust dates when you echo them out.

Also, add some language to highly discourage changing the value away from UTC, as I believe the general consensus is that all datetimes should be stored as UTC in the database.

Might even be worth dropping this out of the standard config completely, so only users who have an explicit reason to not use UTC override it.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -64,6 +64,8 @@ return [
     | Here you may specify the default timezone for your application, which
     | will be used by the PHP date and date-time functions. We have gone
     | ahead and set this to a sensible default for you out of the box.
+    | This will affect how dates are stored in your database and it
+    | is strongly recommended to leave this configuration as UTC.
     |
     */
 


akbarjimi:
Could you please share an example?
In one of our project we had to deal with this kind of configuration
and i'm suspect that those configuration will lead us to trouble

GrahamCampbell (approved):

taylorotwell:
I think this makes things confusing as well.

The dates stored in the database are simply in whatever timezone the `DateTime` instance is. For example, this note might make people think they can set this configuration value to any timezone, and then store dates in that timezone **no matter what timezone the `DateTime` instance is in**. That isn't true though.

browner12:
@akbarjimi I'm not exactly sure what you want an example of?

@taylorotwell I'm definitely open to different wording, but I'm not sure I completely follow you. This config option is used in the `LoadConfiguration.php` file and is called via:

```php
date_default_timezone_set($config->get('app.timezone', 'UTC'));
```

and

> date_default_timezone_set() sets the default timezone used by all date/time functions.

so it will definitely "affect how dates are stored in your database", unless you specifically override the default timezone of a `DateTime` object.

I guess my end goal is to make people understand that this config option isn't simply a way to echo out their datetimes in a consistent format. While it is not the absolute answer to how values are stored in the database, it plays a huge rule in it, and users should be aware of that, along with understanding that switching away from UTC is generally considered **very bad practice**.

PR #5548, opened 2021-03-01
Bug report: inside booted method of a model with softdelete the delete events are fired but you can't change model

TassoPeperone:
Inside a model with softdelete:

protected static function booted()
    {
        static::deleting(function ($model) {
           $model->field= some_value;       
        });
    }

 the above event is fired, but when deleting and also delete event are fired you can't change the model (but with sofdelete the model is there).


GrahamCampbell:
Please post bug reports on the framework repo. Issues are not enabled on this repo, and it is not intended that people use the PR function as a way to bypass this.

PR #5549, opened 2021-03-02
Use same default queue name for all drivers

shaffe-fr:
Hi,
This tiny change makes changing driver queue name more seamless out of the box.
Thanks


GrahamCampbell:
That is not the name of the driver. It is the name of the connection. More than one connection can use the same driver.

shaffe-fr:
Yes, excuse my English, it was what I meant.

PR #5550, opened 2021-03-03
[8.x] Implements new BufferedConsoleOutput

yoeriboven:
Implements [this PR](https://github.com/laravel/framework/pull/36404) on the framework.

Calling the code below currently results in a fatal error because the current class doesn't implement the `fetch` method.

```php
use Illuminate\Console\Events\CommandFinished;

Event::listen(function (CommandFinished $event) {
    dump($event->output->fetch());
});
```


GrahamCampbell:
👎 Doing this will cause memory leaks in long running processes.

taylorotwell:
Indeed ^

PR #5551, opened 2021-03-03
[8.x] Add default REDIS_CLIENT variable to .env.example

HajMo:
Add default `REDIS_CLIENT=phpredis` variable to .env by default to make it easy and not forget it.


taylorotwell:
I think it's probably unusual, and likely not recommend, to use a different client between your local and production environments.

PR #5552, opened 2021-03-09
[8.x] Standarise "must" and "may" language in validation

timacdonald:
Majority of the messages are in the format `"The {$attribute} must {$conditionsToBeMet}"`, however a few inconsistently use "may" instead of "must".

This PR fixes that and has them all use "must" instead for new projects going forward.

Of the 105 messages, only 7 use "may"

To highlight the inconsistency:

```php
// "may"
'max' => [
    'numeric' => 'The :attribute may not be greater than :max.',
    'file' => 'The :attribute may not be greater than :max kilobytes.',
    'string' => 'The :attribute may not be greater than :max characters.',
    'array' => 'The :attribute may not have more than :max items.',
],
// "must"
'min' => [
    'numeric' => 'The :attribute must be at least :min.',
    'file' => 'The :attribute must be at least :min kilobytes.',
    'string' => 'The :attribute must be at least :min characters.',
    'array' => 'The :attribute must have at least :min items.',
],
```


timacdonald:
Just wanna note that I've got some more standardisations to PR around these aside from the two that are now active - but wanted to do them in atomic changes to make it easier to see what is changing and why, so I'll do some follow ups unless you let me know it isn't wanted.

itxshakil (approved):

PR #5553, opened 2021-03-09
[8.x] Remove the use of "field" in validation messages

timacdonald:
A few validation messages inconsistently reference the key by appending the work "field" like `"The {$attribute} field {$conditionsNotMet}"`, while others just reference the key without "field" like `"The {$attribute} {$conditionsNotMet}"`. 

This PR standardises on the way more used style where we reference the key in the payload like... `"The name is not a string"` instead of `"The name field is not a string"`

Of the 105 messages, only 12 of them use the word "field" in the message.

Examples:

```php
// type based validation...
'string' => 'The :attribute must be a string.',
'integer' => 'The :attribute must be an integer.',
'boolean' => 'The :attribute field must be true or false.', // uses "field"

// value based validation...
'unique' => 'The :attribute has already been taken.',
'distinct' => 'The :attribute field has a duplicate value.', // uses "field"
'in' => 'The selected :attribute is invalid.',
'in_array' => 'The :attribute field does not exist in :other.', // uses "field"
```

Note: There is a kind of exception to this inconsistency in the "required*" and "present" rules, which you could argue have internal consistency in that they all use "field", however I don't believe it adds any value and just creates space for more inconsistency in the future.


taylorotwell:
I dunno - part of me thinks more of the rules should actually use "field" and not less 😅 

For example: "The name is required." seems like a downgrade from "The name field is required".

timacdonald:
Gonna close this and will follow up with a PR to standardise with field (as it isn’t just a string find and replace in some cases).

PR #5554, opened 2021-03-09
[8.x] Add missing 'after_commit' key to queue config

BrandonSurowiec:
There's a new `after_commit` config option that isn't in the default config for `queue` connections: 
https://laravel.com/docs/8.x/queues#jobs-and-database-transactions

For new projects it may be better to have these defaulted to `true`, but I left them as `false` for now.


taylorotwell:
I don't think `true` should be the default.

PR #5555, opened 2021-03-10
[8.x] Added detail docs to 'RedirectIfAuthenticated' middleware

SaaberDev:
I thought this middleware redirect user to the view after they are authenticated. Then i realized this middleware actually prevent users from seeing views that they don't need to see after authentication. So i changed and added 2 extra line to define what this middleware does. I also saw people asking the same question what does "RedirectIfAuthenticated" middleware do ?


taylorotwell:
I'll add some wording to it.

SaaberDev:
> I'll add some wording to it.

Appreciate your response. Proper commenting is very important to understand the purpose of a code. This helps us to save time.

PR #5560, opened 2021-03-11
Used short view helper to render welcome view

MouadZIANI:
Used the shortest syntax to render welcome view using route view method


driesvints:
Please see many past attempts.

PR #5562, opened 2021-03-12
[8.x] Implement IDE Helper

driesvints:
This implements [Laravel IDE Helper Generator](https://github.com/barryvdh/laravel-ide-helper) which can replace the existing DocBocks in the framework.

Goes along with https://github.com/laravel/framework/pull/36583

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
         "laravel/tinker": "^2.5"
     },
     "require-dev": {
+        "barryvdh/laravel-ide-helper": "^2.9",
         "facade/ignition": "^2.5",
         "fakerphp/faker": "^1.9.1",
         "laravel/sail": "^1.0.1",
@@ -57,6 +58,11 @@
         ],
         "post-create-project-cmd": [
             "@php artisan key:generate --ansi"
+        ],
+        "post-update-cmd": [
+            "Illuminate\\Foundation\\ComposerScripts::postUpdate",
+            "@php artisan ide-helper:generate",
+            "@php artisan ide-helper:meta"
         ]
     }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ Homestead.json
 Homestead.yaml
 npm-debug.log
 yarn-error.log
+_ide_helper.php
+_ide_helper_models.php
+.phpstorm.meta.php


driesvints:
Ping @barryvdh

barryvdh:
Well obviously I like the idea of adding it by default. Not 100% about just removing the phpdocs though.

deleugpn:
I don't like how it makes PHPStorm identify two classes for everything

SjorsO:
The composer script seems to be missing `@php artisan ide-helper:models -N`, this one is also very helpful.

Also, maybe you can add a composer script to easily regenerate the helpers. The helpers need to be regenerated after making database changes or when adding model accessor. I use the following in my projects:

```
"post-update-cmd": [
    "Illuminate\\Foundation\\ComposerScripts::postUpdate",
    "@composer ide"
],
"ide": [
    "@php artisan ide-helper:generate",
    "@php artisan ide-helper:meta",
    "@php artisan ide-helper:models -N"
]
```

mfn:
> The helpers need to be regenerated after making database changes or when adding model accessor. I use the following in my projects:

Btw, there's https://github.com/barryvdh/laravel-ide-helper/pull/1163 to achieve this

composer.json
```diff
@@ -57,6 +58,11 @@
         ],
         "post-create-project-cmd": [
             "@php artisan key:generate --ansi"
+        ],
+        "post-update-cmd": [
+            "Illuminate\\Foundation\\ComposerScripts::postUpdate",
+            "@php artisan ide-helper:generate",
```
netpok, line 64: This will fail no-dev environments, I suggest using:
```
            "[ $COMPOSER_DEV_MODE -eq 0 ] || $PHP_BINARY artisan ide-helper:generate",
            "[ $COMPOSER_DEV_MODE -eq 0 ] || $PHP_BINARY artisan ide-helper:meta"
```

image

I also suggest adding it to the `post-install-cmd` too.

SjorsO: When do you run `composer update` in a no-dev environment?

netpok: Okay, that's a good point, I have it in my `post-install-cmd` too and that's what broke for me.

Having it in post install are pretty much a must have if you are planning to have collaborators on the project.

driesvints: @barryvdh can you confirm this?

netpok: Basically composer fails because it tries to execute a nonexistent command, there are two ways to prevent this:
1. appending `|| true` which will make the command look like it was successful but still showing the error message
2. using the command I sent which checks and only runs the code in dev mode, basically a bash test

The 3rd solution would be to implement it inside the framework, I thought about it some time ago:
- Add a PackagesDiscovered event to the `package:discover` command (this could be useful for other packages too)
- And add a listener for it in ide helper like in my PR for the migrations
Thus allowing automatically generating helpers when packages are discovered

netpok: Okay, turns none of my first two solution works under Windows

taylorotwell:
I guess we'll just table this for now since the other PR is closed as well.

barryvdh:
I'll see if we can help/generate these phpdocs automatically, so it will make it easier to update the phpdocs in the core, if that helps?

composer.json
barryvdh: Yeah I usually add the file to my git, so it is included when sharing the project. Only needs to refresh on updates in that case.
You don't want to run it on production.

driesvints:
@barryvdh I think that could be useful but I wonder how we'd automate it. Plus keep track of the nuances like other annotations.

PR #5563, opened 2021-03-14
[8.x] Update typehints to current support type

masterfermin02:
This pr is to update the typehint so the property declaration is the same as the parents.

Summary changes:
- Updated app/Exceptions/Handler.php to be an array of string[] instead of array
- Updated app/Http/Middleware/TrustProxies.php to support union type `int|null|string` on `$headers` prop
- Updated app/Models/User.php `$fillable` prop to be an array of string[]

I ran into this when running some static analysis with https://github.com/psalm/psalm-plugin-laravel

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -14,7 +14,7 @@ class User extends Authenticatable
     /**
      * The attributes that are mass assignable.
      *
-     * @var array
+     * @var string[]
      */
     protected $fillable = [
         'name',
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -19,7 +19,7 @@ class Handler extends ExceptionHandler
     /**
      * A list of the inputs that are never flashed for validation exceptions.
      *
-     * @var array
+     * @var string[]
      */
     protected $dontFlash = [
         'current_password',


app/Http/Middleware/TrustProxies.php
```diff
@@ -17,7 +17,7 @@ class TrustProxies extends Middleware
     /**
      * The headers that should be used to detect proxies.
      *
-     * @var int
+     * @var int|null|string
```
GrahamCampbell, line 20: ?

app/Providers/RouteServiceProvider.php
```diff
@@ -54,7 +54,7 @@ public function boot()
      *
      * @return void
      */
-    protected function configureRateLimiting()
+    protected function configureRateLimiting(): void
```
GrahamCampbell, line 57: No

GrahamCampbell (changes requested):

app/Http/Middleware/TrustProxies.php
masterfermin02: If you go to the parents class Fideloper\Proxy\TrustProxies you will see that:
```php
/**
     * The proxy header mappings.
     *
     * @var null|string|int
     */
    protected $headers;
```

so maybe should have the same typehint as its parent.

GrahamCampbell: No, it shouldn't. The type is already correct, for how Laravel is using it.

app/Models/User.php
```diff
@@ -14,7 +14,7 @@ class User extends Authenticatable
     /**
      * The attributes that are mass assignable.
      *
-     * @var array
+     * @var string[]
```
GrahamCampbell, line 17: Why update this but not the hidden array phpdoc?

GrahamCampbell (changes requested):

app/Models/User.php
masterfermin02: Because the parent typehint `Illuminate\Database\Eloquent\Concerns\HidesAttributes` is an array `array<array-key, mixed>` and not string[] `array<array-key, string>` so maybe we need to change parent first.

app/Http/Middleware/TrustProxies.php
masterfermin02: I got it, thank you!

app/Models/User.php
GrahamCampbell: That's just what your static analyzer intered because we just wrote `array`. The actual type is `string[]`.

GrahamCampbell: I mean, we could just delete that type information. It doesn't even matter...

GrahamCampbell: Read-only properties have types totally determined by their fixed values.

masterfermin02: Thank you very much for that! It makes sense. I wrote this pr try to fix some test in the analyzer tool so I think it better to close it and look for another workaround on the analyzer tool.

PR #5564, opened 2021-03-14
[8.x] Minor update to the "check if application is under maintenance" documentation in public/index.php

settermjd:
While reading through the documentation in `public/index.php`, the header and initial sentence for the "_checking if application is under maintenance_" section didn't quite read properly to me. So, I'm submitting this small patch to correct it. Just something small, but I hope that it helps. I'm not quite sure if the 8.x branch or master is the correct one to submit this to.

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -7,11 +7,11 @@ define('LARAVEL_START', microtime(true));
 
 /*
 |--------------------------------------------------------------------------
-| Check If Application Is Under Maintenance
+| Check If The Application Is Under Maintenance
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I don't think these changes are necessary, and break the comment alignment.

settermjd:
Up to you, really. I'm just offering, because the paragraph didn't read correctly to me. I thought that I'd adjusted the comment alignment.

taylorotwell:
Merged!

PR #5568, opened 2021-03-19
[9.x] Change default disk env key

rezaamini-ir:
There is a difference between `disk` and `driver` and the `default` key in `filesystems.php` is the default for disks, not drivers.
And developers may be confused when they see the filesystem config file.
I guess you should fix this in the new version.

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -13,7 +13,7 @@ return [
     |
     */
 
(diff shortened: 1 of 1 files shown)


u01jmg3:
@rezaamini-ir, @driesvints: should `.env.example` also be updated?

https://github.com/laravel/laravel/blob/56a73db2e34f5aa8befffcce40aaaa92e2d7393c/.env.example#L20

PR #5572, opened 2021-03-24
[8.x] Remove unnecessary visibility declaration

nguyenyou:
The default visibility of class constants is public.


GrahamCampbell (changes requested):
👎 PSR-12 says that the visibility must be included. The default visibility on functions is also public, but we still include it.

PR #5575, opened 2021-03-26
[8.x] Change the way of comment In bootstrap.js file

AbdlrahmanSaberAbdo:
Change the way of comment In bootstrap.js file to be more readable for any developer


davidbirkin:
Doesn't this make it seem like you have to use all or nothing with it being commented out as a single block?

AbdlrahmanSaberAbdo:
> Doesn't this make it seem like you have to use all or nothing with it being commented out as a single block?

It's more readable like that, so you can read the snippet of code clearly.

driesvints:
Thanks but let's leave this as is.

PR #5577, opened 2021-04-03
Fix a minor typo

level39:


paras-malhotra:
https://github.com/laravel/framework/blob/8.x/src/Illuminate/Encryption/Encrypter.php

PR #5579, opened 2021-04-06
Update validation.php

MostafaRabia:
The boolean validation don't accept true or false from front, but accept 1 or 0.
For this confuse for the front, I made this pull request.

### DIFF
--- a/resources/lang/en/validation.php
+++ b/resources/lang/en/validation.php
@@ -29,7 +29,7 @@ return [
         'string' => 'The :attribute must be between :min and :max characters.',
         'array' => 'The :attribute must have between :min and :max items.',
     ],
-    'boolean' => 'The :attribute field must be true or false.',
+    'boolean' => 'The :attribute field must be 1 or 0.',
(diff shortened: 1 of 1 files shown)


crynobone:
User might also not see 1 or 0 in the available input such as radio, or even select unless they examine the HTML manually.

MostafaRabia:
@crynobone 
Yes he will not see, but the data from front by axios must be 0 or 1 not true of false, the front telling me i'm send the true or false but not acceptable, then i respond send it 0 or 1 to be accepetable, That's what i mean.
image
image

ghostzero:
As a user who sees this error message, I can imagine more with true or false than 1 or 0. I don't think this message is meant for developers to better debug they code.

MostafaRabia:
Good opinion, but who works back with front and don't need message to show to user, what can do?
I can change it manually but i will forget :)

GrahamCampbell (changes requested):
This change seems wrong.

PR #5580, opened 2021-04-07
[8.x] Updated .env.example for Octane

rennokki:


selcukcukur:
@rennokki It may be better not to include environment variables that are not commonly used. Because we can define default values for them.

rennokki:
I was thinking that sail and other services got defaults here, so this could have been one.

PR #5582, opened 2021-04-12
Remove 'MustVerifyEmail' import from 'User' class

jklusis:
Minor, cosmetic change, but the imported `MustVerifyEmail` contract/interface is not being used in the `User` class, therefore in a new Laravel project it will most likely be removed by users either way if they will choose not to add the implementation.

image

Could it be that the framework's `User` class (`Authenticatable`) is missing the `MustVerifyEmail` contract/interface, since it already uses the `MustVerifyTrait`? (It should have been moved at some point but wasn't?)

image

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -2,7 +2,6 @@
 
 namespace App\Models;
 
-use Illuminate\Contracts\Auth\MustVerifyEmail;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Foundation\Auth\User as Authenticatable;
 use Illuminate\Notifications\Notifiable;


driesvints:
Please see many, many previous attempts

jklusis:
@driesvints thanks! My apologies, did not think of looking into the PR's created before.

I understand that it was left for demonstration purposes, however the base User class already uses the `MustVerifyEmail` trait and the default `users` migration already includes the `email_verified_at` column, therefore the email 'package' is somewhat incomplete as it is missing the contract.

I think that it would make sense that either the whole email 'package' would be already ready to go or it would be an optional thing that could be added manually to the User model, of course, providing that all the necessary setup documentation would be well described with an demonstration etc.

It is just my opinion, I don't mean to argue; it is a minor thing, nothing serious. Just thought that this is an interesting topic/issue that could be brought up and that there possibly could be a room for an improvement. 🙂

PR #5583, opened 2021-04-12
[8.x] Using faker method instead of properties

JuanDMeGon:
After Faker PHP 1.14, using properties is deprecated, and it is recommended to use methods instead.

See [#164](https://github.com/FakerPHP/Faker/pull/164) for details.

It shouldn't be a breaking change.


JuanDMeGon:
Already sent a [PR](https://github.com/laravel/docs/pull/6981) to the docs to fix the examples accordingly 😊

PR #5585, opened 2021-04-13
[8.x] Anonymous migrations

driesvints:
This implements anonymous migrations from https://github.com/laravel/framework/pull/36906 for the skeleton.

PR #5590, opened 2021-04-15
[9.x] Implement anonymous migrations

driesvints:
Replacement for https://github.com/laravel/laravel/pull/5585


sonja-turo:
Nice. Can I assume that this approach will work in older versions of Laravel (even though not directly supported)? I wouldn't mind cleaning up some older migrations I have in projects. Might even help with my editor complaining about the missing namespaces.

driesvints:
@sonja-turo available from Laravel v8.37.0 and higher.

PR #5596, opened 2021-04-21
Add security check command to scan for known vulnerabilities

paras-malhotra:
## Description

This PR adds a `php artisan security:check` command to scan for known vulnerabilities in app dependencies. The added package is a thin wrapper around the [Enlightn security checker](https://github.com/enlightn/security-checker) package, which is relatively stable (~500k downloads).

## Motivation

Symfony has [its own](https://symfony.com/doc/current/setup.html#checking-security-vulnerabilities) `symfony check:security` command to scan app dependencies for known vulnerabilities. Drupal (Drush) [uses the Enlightn security checker](https://github.com/drush-ops/drush/blob/1750ad5ea07f6ca121919f2eff3cf900605faf35/composer.json#L42).

I think it would be a good idea for Laravel to have its own security check command to scan dependencies for known vulnerabilities.

## Screenshot

To demonstrate how this would work, here's a Composer update that installs a non-patched/vulnerable dependency. The `security:check` command displays a warning on update.

sec-checker2

## Repository

I thought this would be best suited for `laravel/laravel`. But if you think I should send this to some other repo (say framework or the installer), I'd be happy to do so.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,7 @@
         "laravel/tinker": "^2.5"
     },
     "require-dev": {
+        "enlightn/laravel-security-checker": "^1.0",
         "facade/ignition": "^2.5",
         "fakerphp/faker": "^1.9.1",
         "laravel/sail": "^1.0.1",
@@ -42,6 +43,9 @@
         ],
         "post-create-project-cmd": [
             "@php artisan key:generate --ansi"
+        ],
+        "post-update-cmd": [
+            "@php artisan security:check"
         ]
     },
     "extra": {


Oxicode:
good idea bro

paras-malhotra:
@GrahamCampbell why don;t you thinking adding this on update makes sense? That's really when the dependency versions change.

Adding the security checker to CI does make sense, but I'm sure many Laravel users aren't even aware symfony's security checker exists. Ideally this should be managed by PHP's dependency manager (Composer) but since it isn't right now, most frameworks have something built in to take care of this seamlessly without the users being required to have knowledge of what libraries exist to solve this.

GrahamCampbell:
> That's really when the dependency versions change.

GrahamCampbell:
> That's really when the dependency versions change.

But typically we need to pickup on vulnerabilities before someone does an update, so they know they need to.

paras-malhotra:
> But typically we need to pickup on vulnerabilities before someone does an update, so they know they need to.

Hmm, I disagree. Even with your proposed way to do this via CI, someone's already done the update and the CI runs after that. With this way, at least it's stopped much sooner in development.

gofish543:
Can someone explain to me why this belongs in the Laravel base skeleton?
I get it is a great feature and something I'll probably put in a few higher security requirement projects; however, does it belong in the skeleton?

taylorotwell:
I don't think I want to include this in the skeleton at the moment but may reconsider in the future based on demand.

PR #5597, opened 2021-04-22
make database prefixable based on environment variables

comes:
### Description

This PR adds the possibility to prefix the database based on an environment variable.

### Motivation

In a multi-stage environment with only one database, it can be beneficial to prefix all tables based on the environment. This PR adds this missing configuration opportunity to the .env.

### Repository

I thought this would be best suited for laravel/laravel. But if you think I should send this to some other repo (say framework or the installer), I'd be happy to do so.


comes:
Hej, sure I did.

I think, based on vapor-based systems and, maybe for bill reducing reasons, there are users in the wild who only will use one database for staging and production. In this environment I think the situation should be overthought.

A past comment in 2019 was:

> Thanks for the PR. But, my feeling is that we shouldn't include this in the default app, since almost nobody will use this.

I'm not sure if this is this valid in the situation described by me. Isn't it?

paras-malhotra:
The question is do people use environment specific DB prefixes (not that common) and not whether people use environment specific databases (quite common).

comes:
> The question is, do people use environment-specific DB prefixes (not that common) and not whether people use environment-specific databases (quite common).

That's true, and I agree with you! In the most common environment on a dedicated server or a local developer machine. In AWS, It's gone be a bit expensive if you have more than one database. To be the most cost-effective, you can prefix it.

Otherwise, where is the pain of having this by default configurable in config/database.php ? DB_FOREIGN_KEYS or ASSET_URL is also env-based, and I'm not sure If it is common to change these variables in a 'common' environment.

I'm fine with closing this PR, but I think it should be discussed.

PR #5598, opened 2021-04-24
Add cache prefix and cookie name to default .env

JABirchall:
I recently spotted an attack on one of my client's websites and the attack was specifically designed to target vulnerable laravel setups, like if the whole application is exposed to the public instead of just /public directory.

While the website was proberbly setup and we have exposed server headers disabled. I suspect they still targeted the site with vulnerable laravel specific attacks because the default cookie was "laravel" in the config. I feel exposing this environment in the .env will entice more developers to change it from the default. As other applications we have running which do have the cookie changed from the default haven't been under such attacks (At least not yet)


paras-malhotra:
If you have a proper app name setup, the session cookie will default to that. These variables are not required.

bhushan:
I agree with @paras-malhotra .. I was about to say the same thing

PR #5602, opened 2021-04-29
[6.x] Bump framework version to include SQL server security fix

netpok:
Security advisory: GHSA-4mg9-vhxq-vm7j

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
     "require": {
         "php": "^7.2.5|^8.0",
         "fideloper/proxy": "^4.4",
-        "laravel/framework": "^6.20",
+        "laravel/framework": "^6.20.26",
         "laravel/tinker": "^2.5"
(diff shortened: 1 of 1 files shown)


GuidoHendriks:
What does that solve, you already get a higher version if you create a new app or update?

netpok:
This locks the minimum version so there shouldn't be a chance to install a lower version unless you manually modify it. Generally you're right and it  should install a newer version.

But locking minimum version to non vulnerable versions are still a good practice and was usually done in Laravel too. Also this only affects newly created projects so there is no backward compatibility issues.

For example you install 'my/package' via composer which needs `illuminate/support: 6.20.15` then composer will install that version of the framework even though there are newer version available. This way it wont install that package unless you manually allow it by lowering the version constraint.

driesvints (approved):

PR #5595, opened 2021-04-20
Add the DisableFloc middleware

tobiasthaden:
This adds a middleware that sets the `Permissions-Policy` header, to disable FloC.

This PR is more a start of a discussion on how we as the Laravel Community think about `Google's Federated Learning of Cohorts`.

PR #5604, opened 2021-04-30
[8.x] Add the DisableFloc middleware

tobiasthaden:
Ref: https://github.com/laravel/laravel/pull/5595, https://github.com/laravel/ideas/issues/2576

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -30,6 +30,7 @@ class Kernel extends HttpKernel
      */
     protected $middlewareGroups = [
         'web' => [
+            // \App\Http\Middleware\DisableFloc::class,
             \App\Http\Middleware\EncryptCookies::class,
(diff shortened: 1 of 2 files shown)


app/Http/Kernel.php
```diff
@@ -30,6 +30,7 @@ class Kernel extends HttpKernel
      */
     protected $middlewareGroups = [
         'web' => [
+            // \App\Http\Middleware\DisableFloc::class,
```
deleugpn, line 33: I'm in favor of having this middleware enabled by default instead of commented out.

taylorotwell:
I've just baked this right in so we don't need to add a new middleware: https://github.com/laravel/framework/commit/c42988a551ffdae0e70aa15959969732265789c0

tobiasthaden:
@taylorotwell if you may decide to revert your commit, I would like to PR the middleware to the framework under the `illuminate` namespace. It seems that'll the way to go. 

Also thinking about a better name than `DisableFloc`.

PR #5606, opened 2021-05-07
Laracasts contains 2000 video tutorials

SarwarAhmed:


FoksVHox:
If Laravasts contains 2000 video tutorials, then the old readme is still true, as 1500 < 2000

GrahamCampbell:
Why not just change to "thousands of videos".

realodix:
> Why not just change to "thousands of videos".

I agree. We don't need to pay attention to this, at least until the number of tutorial videos reaches 1,000,000

Neeraj1005:
> Why not just change to "thousands of videos".

either 1500+ (plus)

PR #5615, opened 2021-05-24
Add folders to gitignore

mic-oliveira:
Add folders to gitignore. The same exclusions exist in the laravel / framework gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ Homestead.json
 Homestead.yaml
 npm-debug.log
 yarn-error.log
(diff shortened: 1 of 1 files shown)


jrd-lewis:
Didn't they get removed in bc435e7fdd8308d133a404b1daa811dd30d95fe5 with the reasoning they belonged in a global gitignore instead of a project's specific gitignore? 👀

GrahamCampbell:
They should be, but we're sick of getting these PRs almost every day so we caved. :/

PR #5617, opened 2021-05-26
[8.x] Add session_secure_cookie to env

themuddfamily:
Currently a major issue (security wise) is that no one is aware of the SESSION_SECURE_COOKIE option.

I propose adding this (and possibly some information on why this is important) to the  .env.example

Originally I thought this would be better to default to true but I can see why you would not want this (as http sessions would stop working).

The reason why this is a security issue is that Man-in-The-Middle attacks are possible by intercepting the http -> https redirect


GrahamCampbell:
Why is this somerthing that should be configured at the environment level, rather than in config?

themuddfamily:
@GrahamCampbell The goal of my pull request is to increase visibility of this issue.

Every single site I know of that runs laravel INCLUDING laravel forge is vulnerable to this issue.

So I'd like to throw the question back to you - why is APP_DEBUG in the env and not in the config?

GrahamCampbell:
Because the debug flag is not config that lives commited in the repo. It is something you wanna change on prod vs dev.

themuddfamily:
To add:

It SHOULD be a setting that is commonly changed in todays day and age and if people aren't aware of this even being an issue how can we make the internet a safer place?

themuddfamily:
> Because the debug flag is not config that lives commited in the repo. It is something you wanna change on prod vs dev.

My response to your question in that case is the same reason... you would want to do local development on a http url rather than https

themuddfamily:
image

I am able to MiTM my laravel_session cookie using wireshark from the redirect on forge.laravel.com

taylorotwell:
You can add it to your own own .env file if you want.

themuddfamily:
@taylorotwell 

Can you explain why you decided against adding this to the env example?

Why is this not mentioned in the documentation? https://laravel.com/docs/8.x/authentication would be a perfect place?

PR #5618, opened 2021-05-26
Add missing licence file

exussum12:
This will match the licence with composer.json and show up in the github interface


GrahamCampbell:
This project is a template for an app (most likely proprietary), so we don't wanna include a MIT license template. People don't wanna accidentally license their private code as MIT. ;)

PR #5620, opened 2021-05-28
add .phpintel to .gitignore

fauzan121002:
The same as `.idea` for PHPStorm.

Those files have nothing to do with your real code (which is what you want to source control), will repeatedly generate diffs and merge conflicts if you do track them in Git, and are useless to anyone else who's not using Sublime with that extension installed.


driesvints:
Add this to your own app.

PR #5627, opened 2021-06-09
Added "mix_url" setting to app config.

royvanv:
Mix assets that aren't served from the root path ("hostname/etc") currently can not be configured via the environment by default.


GrahamCampbell:
Please see previously closed PRs.

PR #5628, opened 2021-06-11
[8.x] Add translation for current_password rule

netpok:
This PR adds the required translation line for the current_password rule introduced in [#37650](https://github.com/laravel/framework/pull/37650).


taylorotwell:
You should be able to remove the old one? This would only affect new applications.

driesvints:
@taylorotwell we'll need to update minimum framework version in that case. Since there's no release yet with the PR we might want to hold off until after next week?

netpok:
I thought leaving it there because some 3rd party packages may use the password rule. In my opinion it should be removed when Laravel 9 comes out.

PR #5629, opened 2021-06-16
Added missing Batch configuration.

atapatel:
Added missing Batch configuration.


atapatel:
@driesvints Hi can you please help me here. What am I missing?

taylorotwell:
If you would like to configure this to be something different then you can add these lines to your own application. Typically it does not need to be customized.

atapatel:
Noted Sir @taylorotwell !!

PR #5633, opened 2021-06-28
Update RouteServiceProvider.php

Amiir-Dev:
Fix routes problem, during using route commands in command line with php artisan

### DIFF
--- a/app/Providers/RouteServiceProvider.php
+++ b/app/Providers/RouteServiceProvider.php
@@ -10,6 +10,7 @@ use Illuminate\Support\Facades\Route;
 
 class RouteServiceProvider extends ServiceProvider
 {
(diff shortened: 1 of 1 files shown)


driesvints:
This isn't necessarily.

Amiir-Dev:
Many of the people Including myself, have faced this problem : "target-class-testcontroller-does-not-exist"
And also in the laracast people asked this question and whith just this line, it would be fixed.

GrahamCampbell:
@Amiir-Dev look on line 29 of that file and read the comment above it...

Amiir-Dev:
Ok, I'm sorry i didn't see that, so why this is disable when most of the people have problem with that!?

PR #5634, opened 2021-06-28
Un-Comment Necessary Line

Amiir-Dev:
As many of the people havr the problem with the **line 29** in the **App/Providers/RouteServiceProvider.php**  in laravel 8,i suggest to un-comment this line to solve the problem in the related command to route in the command line with php artisan


driesvints:
Please stop attempting to change this. This isn't necessarily. You can do this in your own app if you like.

PR #5635, opened 2021-06-30
Update database.php

Norgul:
Standardizing environment variable names.

Having all database environment variables prepended with `DB_` it only makes sense to do the same with replacing `DATABASE_URL` to `DB_URL`.


driesvints:
This will break all automated tooling that currently expects the previous name. I don't think we should change this.

PR #5638, opened 2021-07-02
[8.x] Fixes `Authenticate::redirectTo` return type

nunomaduro:
This pull request instructs static analysis tools that `redirectTo` will return a string or void.


driesvints:
This doesn't seems right. Either it's `void` or `string|null`. The new implementation doesn't makes sense if you read it out loud? If you want to make the static analysis tool succeed then most likely a `return null;` needs to be added below the if statement.

PR #5641, opened 2021-07-09
Fix misleading APP_KEY character length docs

sdwolfz:
The key is actually 42 characters long, not 32. Here's the proof:

1. Generate a new key:

```
php artisan key:generate --show
=> base64:DpYk0NpUp6NMFg2h/0hTWTodPPlSLQp0itYcKzG7tlg=
```

2. Take the value in between `base64:` and `=`

3. Count the characters:

```
echo -n "DpYk0NpUp6NMFg2h/0hTWTodPPlSLQp0itYcKzG7tlg" | wc -m
=> 43
```

As you can see this key has 43 characters, but I've managed to generate
keys what have 42 characters as well.

Here's how you would generate the key without artisan:

```
tr -dc "a-zA-Z0-9" < /dev/urandom | head -c 42
```

Generating one with 43 won't work here, not sure why.

Anyway, the current documentation is misleading as it will not work if
you add a 32 character random string.

For the purposes of this PR I hope you will keep the discussion centered around the incorrect docs and not why I need to generate the key without artisan as I can assure you I have a valid reason.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -114,7 +114,7 @@ return [
     |--------------------------------------------------------------------------
     |
     | This key is used by the Illuminate encrypter service and should be set
-    | to a random, 32 character string, otherwise these encrypted strings
+    | to a random, 42 character string, otherwise these encrypted strings
     | will not be safe. Please do this before deploying an application!
     |
     */


GrahamCampbell:
It is 32 characters?

GrahamCampbell:
`strlen(base64_decode('DpYk0NpUp6NMFg2h/0hTWTodPPlSLQp0itYcKzG7tlg'))`

GrahamCampbell:
The base 64 decoded value is the actual key.

sdwolfz:
@GrahamCampbell Wait a minute, please reconsider this PR and let's make a change to the docs.

First it's not clear from that line that any 32 character value must be made base64 before added to the APP_KEY.
Second, it does not work with any 32 character value, but it works with 31...

Here try it yourself, this won't work:
```
tr -dc "a-zA-Z0-9" < /dev/urandom | head -c 32 | base64
```

But this will:

```
tr -dc "a-zA-Z0-9" < /dev/urandom | head -c 31 | base64
```

I only want to clarify this in the actual docs, as it's frustrating to deal with. Now that I've learned the trick myself I don't need to do anything else, so I could just leave it as is... but that's not what this PR is about. Let's collaborate on this OK?

PR #5644, opened 2021-07-16
Remove unused class

lakuapik:
I believe this class never used in the `User` model and should be removed.

It already used here in the `Authenticatable` class.

https://github.com/laravel/framework/blob/83b70546fc43fa3278d5d101141033fd37edb4c6/src/Illuminate/Foundation/Auth/User.php#L1-L20

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -2,7 +2,6 @@
 
 namespace App\Models;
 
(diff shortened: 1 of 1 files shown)


crynobone:
One is an Interface and another one is a Trait.

lakuapik:
my bad, i miss understood 🙏️

PR #5645, opened 2021-07-17
Add validation keys for the new Password rule messages

pliasun:


taylorotwell:
Use JSON translation files to translate these messages.

PR #5646, opened 2021-07-21
Add generated files

fieu:
This pull request is to add the following files to the default `.gitignore`:

* `public/css/app.css`
* `public/js/app.js`
* `public/mix-manifest.json`

The reason for this PR is to avoid pushing generated assets when running `npm run dev` and/or `npm run prod`


taylorotwell:
Some people may want to push them.

PR #5647, opened 2021-07-22
Update database.php

iehong:
set charset


taylorotwell:
These should be the same between local and production.

PR #5649, opened 2021-07-22
deny run shell attack

vahidalvandi:
.htacess deny any php uploaded files and my contain shell


driesvints:

PR #5651, opened 2021-07-25
[8.x] Add validation keys for the new Password rule messages

madcitymultimedia:


taylorotwell:
This isn't how you customize those messages - and adding the messages here in this way would not work. You need to create  JSON based translation for them.

PR #5652, opened 2021-07-26
Adding id primary key to password_resets

MostafaRabia:
This for avoiding `Unable to create or change a table without a primary key` from digitalocean's database.


taylorotwell:
Add to your own application if you need it. This won't solve all the problems with that on DO.

PR #5655, opened 2021-07-31
[8.x] Improve APP_KEY documentation

CrimsonGlory:
I'm reopening the discussion of a [previous pull request](https://github.com/laravel/laravel/pull/5641) from @sdwolfz where the user suggested changing "32 characters" for "42 characters", after trying to generate the string without artisan and finding out the app would not start. The number is actually 32, but bytes, not characters. Per [strlen() documentation](https://www.php.net/manual/en/function.strlen.php):
```strlen() returns the number of bytes rather than the number of characters in a string.```

A few points:
- A byte may be interpreted as a character or may not. In this case, because this is used for cryptography, the word "character" does not apply.
- Bytes in text have to be encoded. In this case laravel uses base64.
- Given that if the decoded length is incorrect, the app would not work (according to the linked PR), I changed the word "should" for the word "must".

I have changed the documentation so that these points are clear.

Btw, the length of each comment line might look familiar ;)
```
$ echo -n "This key is used by the Illuminate encrypter service and must be set to" | wc -m
71
$ echo -n "a random 32 bytes base64 encoded string, otherwise encrypted strings" | wc -m
68
$ echo -n "will not be safe. Please do this before deploying an application!" | wc -m
65
```

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -113,8 +113,8 @@ return [
     | Encryption Key
     |--------------------------------------------------------------------------
     |
-    | This key is used by the Illuminate encrypter service and should be set
-    | to a random, 32 character string, otherwise these encrypted strings
+    | This key is used by the Illuminate encrypter service and must be set to
+    | a random 32 bytes base64 encoded string, otherwise encrypted strings
     | will not be safe. Please do this before deploying an application!
     |
     */

### CODE REVIEW

**Comment by GrahamCampbell** [comment]
This description is still wrong. You don’t have to set it to a base64 encoded key. You can just type in 32 characters.

**Comment by CrimsonGlory** [comment]
ok. I just took a look at the [source code](https://github.com/illuminate/auth/blob/5ad8e560754597d03fa8f61833405ce2c0f6e6a4/Passwords/PasswordBrokerManager.php#L87) to see what the app actually does, because this seems to be causing some [c](https://stackoverflow.com/questions/39693312/the-only-supported-ciphers-are-aes-128-cbc-and-aes-256-cbc-with-the-correct-key)[o](https://github.com/laravel/framework/issues/9080)[nf](https://github.com/laravel/framework/issues/24510)[us](https://github.com/laravel/framework/issues/23237)[ion](https://laracasts.com/discuss/channels/laravel/the-only-supported-ciphers-are-aes-128-cbc-and-aes-256-cbc-with-the-correct-key-lengths-1).
```php
        $key = $this->app['config']['app.key'];

        if (Str::startsWith($key, 'base64:')) {
            $key = base64_decode(substr($key, 7));
        }
```

The app accepts a securely random generated 32 bytes base64 encoded string prepended with "```base64:```" like this ```APP_KEY=base64:nZt3ifjsspgF1ZHQaOBd4l2P1r+7XGrgPJ/iyVnLXD0=```. But it also accepts a not-that-secure random 32 character string like this: ```APP_KEY=01234567890123456789012345678901```. Oddly, the 3-line comment suggest the latter one. Anything else will crash with the message "```The only supported ciphers are AES-128-CBC and AES-256-CBC with the correct key lengths```". I think everything is clear now.

Btw, it would be a shame if the random 32 character string just happen to start with  "```base64:```" ;)

PR #5656, opened 2021-08-02

iamine:
Redirect only if request doesnt expectsJson


app/Http/Middleware/RedirectIfAuthenticated.php
```diff
@@ -23,7 +23,9 @@ public function handle(Request $request, Closure $next, ...$guards)
 
         foreach ($guards as $guard) {
             if (Auth::guard($guard)->check()) {
-                return redirect(RouteServiceProvider::HOME);
+                if (! $request->expectsJson()) {
```
quest-academia, line 26: ?

PR #5659, opened 2021-08-08
support sqlite cfg for memory unit tests & relative proj path

bingalls:
Config fix, to allow :memory: for unit tests, as well as relative paths for projects.
This allows projects to bundle pre-built sqlite databases.
Old code required users to edit .env.example with an absolute path for sqlite file.


driesvints:
Feel free to do this in your own codebase.

PR #5663, opened 2021-08-10
[8.x] Sanctum

taylorotwell:
This PR serves as a proof of concept for what it would look like to just make Sanctum the default API authentication stack in a new Laravel application.

**Motivations:** The `token` driver is not robust and is no longer documented as of the Laravel 8.x release many months ago. As in, there is literally no documentation on the website about how to use it because I simply don't think it is a good idea to ever use it.

### DIFF
--- a/config/auth.php
+++ b/config/auth.php
@@ -40,12 +40,6 @@ return [
             'driver' => 'session',
             'provider' => 'users',
         ],
-
-        'api' => [
-            'driver' => 'token',
-            'provider' => 'users',
-            'hash' => false,
-        ],
     ],
 
     /*
--- a/config/auth.php
+++ b/config/auth.php
@@ -31,7 +31,7 @@ return [
     | users are actually retrieved out of your database or other storage
     | mechanisms used by this application to persist your user's data.
     |
-    | Supported: "session", "token"
+    | Supported: "session"
     |
     */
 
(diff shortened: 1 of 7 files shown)


amaelftah:
this is so cool please merge .... as since I started using laravel from 2016 i never used the default API driver

elhadiabdallah:
this is so cool

config/auth.php
```diff
@@ -40,12 +40,6 @@
             'driver' => 'session',
             'provider' => 'users',
         ],
-
-        'api' => [
-            'driver' => 'token',
-            'provider' => 'users',
-            'hash' => false,
-        ],
```
mfn, line 48: There's a comment above saying this; maybe the `token` mention should be removed there too? 🤔
image

ldiebold:
Just started a new projec. Was about to remove the default token based authentication...

Was thrilled to see sanctum already setup! Love this :heart:

wjvankesteren:
Sanctum is still present in Laravel, however with an error every startup time:

`SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'personal_access_tokens' already exists (Connection: mysql, SQL: create table personal_access_tokens (id bigint unsigned not null auto_increment primary key, tokenable_type varchar(255) not null, tokenable_id bigint unsigned not null, name varchar(255) not null, token varchar(64) not null, abilities text null, last_used_at timestamp null, expires_at timestamp null, created_at timestamp null, updated_at timestamp null) default character set utf8mb4 collate 'utf8mb4_unicode_ci')`

PR #5671, opened 2021-08-17
[8.x] Made base controller abstract

rcerljenko:
Usage of base controller is to be a base class for every other controller so it can be marked as `abstract`.


driesvints:
I often use this controller for testing purposes. Feel free to do this in your own app.

PR #5674, opened 2021-08-19
[8.x] Use new stronger AES-256-GCM encryption by default

driesvints:
Implements https://github.com/laravel/framework/pull/38190 as a default.


GrahamCampbell:
Do we need to bump the minimum framework version here too?

driesvints:
@GrahamCampbell the current one is already the minimum for this version.

GrahamCampbell:
🚢

alteredchimera:
I like this change. Quick question, Will this work with data already encrypted in other 8.x projects? Or would I have to decrypt existing data with AES-256-CBC and re-encrypt with AES-256-GCM?

Thank you.

driesvints:
@alteredchimera as with any encryption you'll need to decrypt/re-encrypt data that was already encrypted using a different algorithm. Note that you're in no way required to move to the new algorithm. Existing apps will continue to work as is.

alteredchimera:
@driesvints Thanks for clarification. I like to keep my 8.x projects up to the latest releases. So I'm happy to do this, I just wanted the added clarification, so I appreciate it. 

For you, the team and collaborators, keep up the sensational work. :)

Krisell:
@alteredchimera I'll just add here that since cookies are encrypted, switching from CBC to GCM will invalidate all sessions (sign all users out or cause 419), so a migration-strategy would be needed there as well if this is a live application (something like, for a transition period, try decrypting with both, but only encrypt with the new).

Tklaversma:
@taylorotwell Why did we move back to `AES-256-CBC` since Laravel v8.6.2 (see https://github.com/laravel/laravel/commit/52de5d84f6c309e93910e0d250b8c42fd656931b)? There are many valid reasons why we should not use CBC, and only a few why GCM is not always best. But `AES-256-GCM` is a good default though!

Main reasons why we should NOT use CBC?
- no TLS 1.3 support
- no built-in authentication

Is your main reason that using the same IV twice can compromise your key?

Krisell:
@Tklaversma The change was reverted due to a bug in older versions of OpenSSL, see here for details: https://github.com/laravel/framework/pull/38594

This bug has since been patched in Laravel, so I agree GCM should be the default now.

TLS support has nothing to do with the encryption used in the application layer, and the IV is never reused. Regarding authenticated encryption, Laravel adds this viaan `hmac` when CBC mode is used.

Tklaversma:
@Krisell thanks for clearing this up. 
@taylorotwell can you revert the commit?

driesvints:
@Tklaversma feel free to attempt a new pr

PR #5675, opened 2021-08-20
Kernal.php doc comment

nipunTharuksha:
Kernal.php doc comment updated


driesvints:
The current wording is correct.

PR #5677, opened 2021-08-23
Fixing "Line exceeds 120 characters" in TrustProxies to comply with PSR-2

tentacode:
When installing a new project with `laravel new` and installing `squizlabs/php_codesniffer`, the projects has one error that cannot be fixed automatically by `phpcbf` because the `app/Http/Middleware/TrustProxies.php` class has a line of more than 120 characters.

Laravel is supposed to follow the PSR-2 conventions : https://laravel.com/docs/8.x/contributions#coding-style

To reproduce :

```bash
laravel new test-coding-standards
cd test-coding-standards
composer require squizlabs/php_codesniffer
vi phpcs.xml
```

Small phpcs.xml file :

```xml
<?xml version="1.0"?>
<ruleset name="PHP_CodeSniffer">
    <description>The coding standard for our project.</description>
    <rule ref="PSR2"/>

    <file>app</file>
    <file>bootstrap</file>
    <file>config</file>
    <file>database</file>
    <file>resources</file>
    <file>routes</file>
    <file>tests</file>

    <exclude-pattern>bootstrap/cache/*</exclude-pattern>
    <exclude-pattern>bootstrap/autoload.php</exclude-pattern>
    <exclude-pattern>*/migrations/*</exclude-pattern>
    <exclude-pattern>*/seeds/*</exclude-pattern>
    <exclude-pattern>*.blade.php</exclude-pattern>
    <exclude-pattern>*.js</exclude-pattern>
    <exclude-pattern>tests/Feature/ExampleTest.php</exclude-pattern>
    <exclude-pattern>tests/Unit/ExampleTest.php</exclude-pattern>

    <!-- Show progression -->
    <arg value="p"/>
</ruleset>
```

And run `vendor/bin/phpcs`, here is the output :

```
...........W................................... 47 / 47 (100%)

FILE: /Users/gabriel/Workspace/test-cs/app/Http/Middleware/TrustProxies.php
---------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
---------------------------------------------------------------------------
 22 | WARNING | Line exceeds 120 characters; contains 201 characters
---------------------------------------------------------------------------

Time: 176ms; Memory: 8MB
```

### DIFF
--- a/app/Http/Middleware/TrustProxies.php
+++ b/app/Http/Middleware/TrustProxies.php
@@ -19,5 +19,10 @@ class TrustProxies extends Middleware
      *
      * @var int
      */
-    protected $headers = Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_HOST | Request::HEADER_X_FORWARDED_PORT | Request::HEADER_X_FORWARDED_PROTO | Request::HEADER_X_FORWARDED_AWS_ELB;
+    protected $headers =
+        Request::HEADER_X_FORWARDED_FOR |
+        Request::HEADER_X_FORWARDED_HOST |
+        Request::HEADER_X_FORWARDED_PORT |
+        Request::HEADER_X_FORWARDED_PROTO |
+        Request::HEADER_X_FORWARDED_AWS_ELB;
 }


CurvesTech (approved):

GrahamCampbell:
👎 PSR-2 only recommends doing that where it makes sense to. In this instance, it doesn't make sense.

tentacode:
@GrahamCampbell why doesn't it make sense ? this is the only one error that could be avoided in the entire project using phpcsfixer, and you have to admit the line is quite long ?

driesvints (approved):

driesvints:
I do agree that seeing the values below each other makes it a bit more readable.

GrahamCampbell:
> this is the only one error that could be avoided in the entire project using phpcsfixer

No, it couldn't. Automated fixing tools cannot properly fix overlength lines. StyleCI is already enabled on this repo to keep the CS in check, automatically.

tentacode:
@GrahamCampbell sorry let me rephrase that : this is the only error that needs a manual fix on the entire project to be clean on phpcsfixer with PSR2 settings in a new project. We are creating new projects every week and everytime we need to fix this line manually, I might not be the only one so instead of juste writing my own phpcs.xml file I tried to fix the issue at the source :)

GrahamCampbell:
I'd of course recommend using using StyleCI on new projects. Laravel apps come with config out of the box, and fixing line length artificially across projects adds little value compared to spending developer time writing new features and fixing bugs. Just let automation fix what it can fix automatically, and forget the rest, especially when it is not a `MUST` in the spec.

mfn:
Personally I also couldn't care about the specific rule, but I like the PR for what it is: making it easily readable without having to scroll around -> 👍 

I mean this:
image

The line feels way beyond any sane limit (it's 200 or so 😱 )

alteredchimera:
I like this PR. More for the fact the line is significantly more readable over the current version. 

I agree with Graham's explanation that 120 characters is a recommendation and not a requirement, but in the future should any other headers need to be added here by default, this PR would make it easier. 

Just my $0.02.

GrahamCampbell:
Sure, 👍 on this PR. My comments were general, addressing the use case of the PR author.

PR #5685, opened 2021-09-17
Add global helper method for paginating collections

torressam333:
Out of the box pagination works on query builders but not collections. This helper method will be globally available and allow developers to now paginate collections as easily as query builders

Sample Usage + Steps to reproduce example:

Inline example used in `web.php`. I set up a basic `.env` file and a local database. I ran the out of the box migrations in order to have access to a users table. I entered multiple users in the database and queried for them as seen below.
```
Route::get('/users', function() {
   $users = User::all();

   $paginatedUsers = paginate_collection(
       $users->flatten()->sortDesc(),
       10,
       null
   );
   
   dd($paginatedUsers);
})
```

Result:

```
Illuminate\Pagination\LengthAwarePaginator {#251 ▼
  #total: 2
  #lastPage: 1
  #items: Illuminate\Support\Collection {#265 ▶}
  #perPage: 10
  #currentPage: 1
  #path: "users"
  #query: []
  #fragment: null
  #pageName: "page"
  +onEachSide: 3
  #options: array:1 [▶]
}
```
No need to import the helper as it is registered as a service provider and is globally available.
Now paginating collections is as easy as paginating query builders.

### DIFF
--- a/app/Helpers/helpers.php
+++ b/app/Helpers/helpers.php
@@ -1,0 +1,21 @@
+<?php
+
+use Illuminate\Pagination\LengthAwarePaginator;
+use Illuminate\Pagination\Paginator;
+use Illuminate\Support\Collection;
+
+if (!function_exists('paginate_collection')) {
+    /**
+     * @param  $items
+     * @param int $perPage
+     * @param null $page
+     * @param array $options
+     * @return  LengthAwarePaginator
+     */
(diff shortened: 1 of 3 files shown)


taylorotwell:

torressam333:
@taylorotwell thank you for the feedback. I will consider creating a composer package.

Thank you :)

PR #5687, opened 2021-09-20
[8.x] Fix typo

dmitryuk:
Fix typo in Kernel.php


Jubeki:
Middleware is already plural.
https://www.wordhippo.com/what-is/the-plural-of/middleware.html

PR #5689, opened 2021-09-26
Add APP_DOMAIN env variable

MordiSacks:
I added an APP_DOMAIN env variable,
This comes in handy in several ways
1. Save time retyping the domain for MAIL_FROM_ADDRESS.
2. When using subdomain routing you can now reference your base domain as a config variable, those allowing route file caching.


.env.example
```diff
@@ -33,7 +34,7 @@ MAIL_PORT=1025
 MAIL_USERNAME=null
 MAIL_PASSWORD=null
 MAIL_ENCRYPTION=null
-MAIL_FROM_ADDRESS=null
+MAIL_FROM_ADDRESS=noreply@${APP_DOMAIN}
```
GrahamCampbell, line 37: I don't think this makes for a good default.

GrahamCampbell (changes requested):

taylorotwell:
You are free to edit this file in any way you wish.

PR #5696, opened 2021-09-29
[8.x] Uses `LazilyRefreshDatabase` by default

nunomaduro:
This pull request makes `LazilyRefreshDatabase` a default trait on the base test class for every new Laravel project. This way people don't need to worry about which test cases use the database or not - things will just work.

If this pull request gets merged, we may need to adapt the documentation before next week's release.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "php": "^7.3|^8.0",
         "fruitcake/laravel-cors": "^2.0",
         "guzzlehttp/guzzle": "^7.0.1",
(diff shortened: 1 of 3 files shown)


taylorotwell:
I reverted this because it doesn't have a good story for usage with Dusk and also doesn't work with the documented `$seed` property. 

The Dusk case is particularly annoying since this trait is used on the base test class there is no way to opt-out of it for your Dusk tests which need to use DatabaseMigrations instead of RefreshDatabase behavior.

PR #5699, opened 2021-09-29
Update lte and gte validation messages to have a grammatically parallel structure

alexleeelkins:
The grammar for gte.numeric/file/string and lte.numeric/file/string do not currently appear to follow the concept of parallel structure, so can sound a little strange to some people's ears. To be parallel, we need the preposition for "equal" (to match the adjective + pronoun used with "greater than"), which would be "to" in this case. Given that these are short messages that are potentially seen by many people, I think having the grammar corrected would be good for the professionalism of those using Laravel as a framework.

Thanks!

### DIFF
--- a/resources/lang/en/validation.php
+++ b/resources/lang/en/validation.php
@@ -53,9 +53,9 @@ return [
         'array' => 'The :attribute must have more than :value items.',
     ],
     'gte' => [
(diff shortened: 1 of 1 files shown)


raghavji2050:
Hjjg

PR #5704, opened 2021-10-11
[8.x] Use intended redirect if authenticated

caugner:
There is a popular Stack Overflow thread on how to make [Laravel redirect back to original destination after login](https://stackoverflow.com/q/15389833). The best answer is by Scott Byers (kudos!), which changes exactly one line in the `RedirectIfAuthenticated` middleware that comes shipped with Laravel.

I wonder why Laravel doesn't redirect to the previously intended location by default, so I'm hereby proposing that change.


caugner:
Wow, you're absolutely right, Guido! 🎉 

My mistake probably was to use `Route::middleware('verified')` as a shortcut for `Route::middleware(['auth', 'verified'])`, and that seemed to have prevented the intended url from being set.

PR #5708, opened 2021-10-16
Update User.php

raicho:


raicho:
Im no't sure with this thinking! Deadcode is bad idea anyway.

PR #5712, opened 2021-10-20
change timezone to public variable

ashrafbendary:
laravel need to have it as public variable so we can easily control and make shell script to auto update laravel without all this minor changes.


driesvints:
See many attempts before. It's recommended to use the same default timezone in all environments.

PR #5716, opened 2021-11-01
Return explicit values in Authenticate::redirectTo

szepeviktor:
`@return string|null`

@phpstan warned that returning void is not null.


driesvints:
Please see many many previous attempts.

szepeviktor:
# 😿

PR #5717, opened 2021-11-03
[8.x] Remove redundant argument

hmingv:


taylorotwell:
I think it is primarily there to show you that passing a default is possible.

hmingv:
> I think it is primarily there to show you that passing a default is possible.

But there are many other places that use default values

PR #5718, opened 2021-11-03
Update app.php

imnotprogrammer:
the `timezone` join the .env


driesvints:
Please see many many previous attempts.

PR #5724, opened 2021-11-13
Helps syntax highlighter to understand example env

pavinthan:
It helps to highlight the syntax without any additional configurations.


driesvints:
This changes the name from the actual default .env.example shipped with Laravel.

PR #5727, opened 2021-11-16
[8.x] Removes the `Console\Kernel::$commands` property

nunomaduro:
This pull request proposes the removal of the `Console\Kernel::$commands` property by default.

It's been a while since Laravel has introduced the possibility of loading commands from a specific folder, and since then, I don't remember the last time I've used this property.

Of course, just like multiple properties today on modals, for example, people can add back this property when they want.


driesvints (approved):

amiranbarigreenweb (commented):
Good

driesvints:
@amiranbarigreenweb please don't post empty reviews.

PR #5731, opened 2021-11-21
Adds docker-compose file back into project

kylelogue:
The `docker-compose.yml` file was removed from Laravel and thus caused Laravel Sail to not work properly.

 This PR adds the file back into the project.


driesvints:
You need to run sail:install

PR #5732, opened 2021-11-23
[8.x] Fix "View path not found" when managing view cache

68kHeart:
Commands around managing the view cache with Artisan (e.g., `php artisan view:clear`) are failing in a fresh deployment with the following message:

```
In ViewClearCommand.php line 57:
  View path not found.
```

`view.php` is the only config file where `realpath()` is still used. All other Artisan commands involving the cache are working without issue in this deployment. Removing the call to `realpath()` appears to resolve the issue.

### DIFF
--- a/config/view.php
+++ b/config/view.php
@@ -30,7 +30,7 @@ return [
 
     'compiled' => env(
         'VIEW_COMPILED_PATH',
-        realpath(storage_path('framework/views'))
+        storage_path('framework/views')
     ),
 
 ];


GrahamCampbell:
We should be careful making this change. Why does that not exist on your system?

driesvints:
Do you have these storage directories on production? https://github.com/laravel/laravel/tree/8.x/storage/framework

68kHeart:
> We should be careful making this change. Why does that not exist on your system?

It's very curious to me too, considering `realpath()` is used behind the scenes without any issue so far. The system in question is a fresh install of FreeBSD 13.0-RELEASE-p5 running PHP 8.0.13.

I can do some testing around the issue if needed; this is a personal project, not mission critical. :)

> Do you have these storage directories on production? https://github.com/laravel/laravel/tree/8.x/storage/framework

Yes, I confirmed these while debugging the issue. Commands around the view cache were the only ones having issues. I was following along with the [Optimization section in Laravel's Deployment guide](https://laravel.com/docs/8.x/deployment#optimization), and neither `config:cache` and `route:cache` encountered any issues.

I dug into the history of `config/view.php` before making this PR to try and find where `realpath()` was introduced. It seems there was a number of them sprinkled into `config/compile.php` and `config/view.php` with 5ce23f1. But these have been gradually removed as updates have been made, e.g., 63dc985.

Given removing the call to `realpath()` to match other configuration paths has resolved the issue and there is precedence for this change in other config files, I put together this PR for your consideration.

taylorotwell:
You can make this change on your own configuration files for now.

PR #5733, opened 2021-11-24
[8.x] Pluralize words to correct grammar

phpfour:
A simple grammatical fix.


driesvints:
Middleware is already plural.

PR #5735, opened 2021-12-02
[8.x] Cleaned up grammar

bram-pkg:
Hi,

I have done some proofreading of the phpdoc, I hope these changes sound okay.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
             "@php artisan package:discover --ansi"
         ],
         "post-update-cmd": [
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I think the wording was intentionally like this, in order to have the lines descending in length.

bram-pkg:
In that case I can reorder the words, yeah.

bram-pkg:
@GrahamCampbell shuffled the words around.

PR #5737, opened 2021-12-02
[8.x] Cleaned up grammar

bram-pkg:
Hi,

I have done some proofreading of the phpdoc, I hope these changes sound okay.

### DIFF
--- a/artisan
+++ b/artisan
@@ -10,8 +10,8 @@ define('LARAVEL_START', microtime(true));
 |
 | Composer provides a convenient, automatically generated class loader
 | for our application. We just need to utilize it! We'll require it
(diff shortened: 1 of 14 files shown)


bram-pkg:
@GrahamCampbell re-ordered the words so the lines are decreasing in length.

driesvints:
All of these unfortunately break the comments length (we try to keep those consistent everywhere). Sorry but we can't merge this.

bram-pkg:
Could you elaborate, I'll gladly fix it?

driesvints:
If you look at all comment blocks in Laravel, you'll notice that each one has the same length in lines. The changes in this PR breaks that pattern.

PR #5738, opened 2021-12-02
[8.x] Add types to arrays in boilerplate

bram-pkg:
I hope I've assumed all the types correctly.

### DIFF
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -10,7 +10,7 @@ class Authenticate extends Middleware
      * Get the path the user should be redirected to when they are not authenticated.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @return string|null
+     * @return string|void
      */
     protected function redirectTo($request)
     {
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -10,12 +10,14 @@ class Authenticate extends Middleware
      * Get the path the user should be redirected to when they are not authenticated.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @return string|void
+     * @return string|null
      */
     protected function redirectTo($request)
     {
         if (! $request->expectsJson()) {
             return route('login');
         }
+        
+        return null;
     }
 }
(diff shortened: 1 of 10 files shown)


app/Http/Middleware/Authenticate.php
```diff
@@ -10,7 +10,7 @@ class Authenticate extends Middleware
      * Get the path the user should be redirected to when they are not authenticated.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @return string|null
+     * @return string|void
```
driesvints, line 13: A method can never be `void` and a return type at the same time.

bram-pkg: Hm, after reading the union type documentation it seems you're right.

Do you know of a better way to indicate to the editor that a return is optional?
This solution fixed the warning in PhpStorm at least.

driesvints: I think `string|null` is the correct type here, regardless of a warning in PhpStorm.

app/Http/Middleware/Authenticate.php
```diff
@@ -17,5 +17,7 @@ protected function redirectTo($request)
         if (! $request->expectsJson()) {
             return route('login');
         }
+        
+        return null;
```
bram-pkg, line 21: I think this would be more appropriate, then. @driesvints

driesvints: The method already returns `null`. The explicit return isn't needed.

bram-pkg: Don't you think it's better to have an explicit return?

Personally I think a method like this would be even cleaner:

```php
protected function redirectTo($request)
{
    if ($request->expectsJson()) {
        return null;
    }

    return route('login');
}
```

driesvints: I think it's best to revert as this has been tried quite a bit already and it was always rejected.

bram-pkg: Reverted. This has always been a "quirky" bit of code to me, haha.

bram-pkg:
Reverted the `string|void`. @driesvints

PR #5740, opened 2021-12-03
[8.x] Improves generic types on the skeleton

nunomaduro:
This pull request is a follow up for https://github.com/laravel/framework/pull/38538, and it improves the DX of Laravel's skeleton on editors like PHPStorm that now support generics. And of course, it also improves the DX with static analyse tools in the process.

Here is a few examples where the DX was improved while using PHPStorm:
Screenshot 2021-12-03 at 11 52 40
Screenshot 2021-12-03 at 11 53 24
Screenshot 2021-12-03 at 12 00 57
Screenshot 2021-12-03 at 12 05 02

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -10,7 +10,7 @@ class Handler extends ExceptionHandler
     /**
      * A list of the exception types that are not reported.
      *
-     * @var string[]
+     * @var array<int, class-string<Throwable>>
      */
     protected $dontReport = [
         //
@@ -19,7 +19,7 @@ class Handler extends ExceptionHandler
     /**
      * A list of the inputs that are never flashed for validation exceptions.
      *
-     * @var string[]
+     * @var array<int, string>
      */
     protected $dontFlash = [
         'current_password',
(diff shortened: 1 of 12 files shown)


mintbridge:
Running Psalm at max level with a clean laravel install has issues with these, as they are different to their parent type hints. 

You get these type of psalm errors:
```
ERROR: NonInvariantDocblockPropertyType - app/Exceptions/Handler.php:24:15 - Property App\Exceptions\Handler::$dontFlash has type array<int, string>, not invariant with Illuminate\Foundation\Exceptions\Handler::$dontFlash of type array<array-key, string> (see https://psalm.dev/267)
    protected $dontFlash = [
```

changing to `array<array-key, string>` (e.g. below) fixes the psalm errors but i don't have phpstorm to check:

```
    /**
     * A list of the inputs that are never flashed for validation exceptions.
     *
     * @var array<array-key, string>
     */
    protected $dontFlash = [
        'current_password',
        'password',
        'password_confirmation',
    ];
```

mintbridge:
@nunomaduro should i raise this as an issue?

driesvints:
@mintbridge best to just try a PR

PR #5741, opened 2021-12-03
[8.x] Add option to set sendmail path. Fix default

listerr:
Testing this in an application, it would seem that sendmail -bs is the wrong option for this case?

What Laravel appears to do is pipe an RFC-2822 formatted message on STDIN and requires the sendmail emulation to deal with it,
rather than -bs which initiates an SMTP session.

if Exim is the default MTA then -t would seem to be the correct option.

If you have an alternative installed instead of sendmail/exim, then there's no way to set the path, so I added MAIL_SENDMAIL_PATH
so you can do, e.g.:

MAIL_SENDMAIL_PATH="/usr/bin/msmtp -t --tls=off --from=${MAIL_FROM_ADDRESS} --auto-from=off"

msmtp doesn't support -bs at all

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -59,7 +59,7 @@ return [
 
         'sendmail' => [
             'transport' => 'sendmail',
-            'path' => env('MAIL_SENDMAIL_PATH', '/usr/sbin/sendmail -t'),
+            'path' => env('MAIL_SENDMAIL_PATH', '/usr/sbin/sendmail -t -i'),
         ],
 
         'log' => [
--- a/config/mail.php
+++ b/config/mail.php
@@ -59,7 +59,7 @@ return [
 
         'sendmail' => [
             'transport' => 'sendmail',
-            'path' => '/usr/sbin/sendmail -bs',
+            'path' => env('MAIL_SENDMAIL_PATH', '/usr/sbin/sendmail -t'),
         ],
 
         'log' => [


config/mail.php
```diff
@@ -59,7 +59,7 @@
 
         'sendmail' => [
             'transport' => 'sendmail',
-            'path' => '/usr/sbin/sendmail -bs',
+            'path' => env('MAIL_SENDMAIL_PATH', '/usr/sbin/sendmail -t'),
```
GrahamCampbell, line 62: This change to the default we probably don't want?

GrahamCampbell (changes requested):

listerr:
As far as I can't tell, the default does not  work with either Exim or msmtp as the sendmail provider on Ubuntu at least in my testing, so I'm not sure what situation this does actually work?ROn 3 Dec 2021 20:47, Graham Campbell ***@***.***> wrote:
@GrahamCampbell requested changes on this pull request.



In config/mail.php:
> @@ -59,7 +59,7 @@
 
         'sendmail' => [
             'transport' => 'sendmail',
-            'path' => '/usr/sbin/sendmail -bs',
+            'path' => env('MAIL_SENDMAIL_PATH', '/usr/sbin/sendmail -t'),

This change to the default we probably don't want?

taylorotwell:
You may be correct here. Thanks.

PR #5744, opened 2021-12-09
Delete web.config

underwood:
During a security audit one of the few recommendations they made was to remove or limit access to web.config.

Since this is mainly used by Microsoft IIS server it isn't necessary for most Laravel projects and could be added if someone is using Microsoft server.


driesvints (approved):

hamidullorahmonberdiyev:
STORAGE::DELETE is not deleting the file in the folder when I update the data.

PR #5746, opened 2021-12-13
[8.x] Replace attribute-name by attribute-value in validation.php file

Shahed-Chy-Suzan:


driesvints:
I don't see why this is needed?

PR #5749, opened 2021-12-15
Update default encryption mode to GCM

Krisell:
This has been attempted, and even merged (but reverted), before: https://github.com/laravel/laravel/pull/5674

The reason for reverting has been resolved (https://github.com/laravel/framework/pull/38594) since `8.58.0` which is lower than the required version in `composer.json`.

This means that GCM is now ready to be the default mode.

GCM is preferred due to smaller payload, better performance and built in `authenticated encryption` instead of having to calculate the `mac` in application code which is the case for CBC mode (although of course done correctly in Laravel).

-------

My only concern is that I have seen some people blindly updating their configs to the latest version, and changing this is not BC. All sessions will be reset and of course any encrypted data will have to be decrypted using the old mode and then re-encrypted using the new mode.

So, any existing app should either keep using CBC mode, or carefully build a migration plan. We did migrate from CBC to CGM in a production app a few months ago successfully (no sessions invalidated) but we only used encryption for cookies. We built a custom encryptor that attempted both modes for decryption during a short transition period.


driesvints:
I feel like it's best that we do this in 9.x and add a note to the upgrade guide about it.

Also pinging @jasonmccreary to make sure Laravel Shift doesn't automatically updates the config for this.

jasonmccreary:
@driesvints, I can ensure Shift follows any Laravel recommendation. 👍

taylorotwell:
I will consider it for the Laravel 9 release.

mpyw:
This looks to be still not released yet

https://github.com/laravel/laravel/blob/b331e10aca58d771511cd3f03e762ca4dea9183e/config/app.php#L126

PR #5755, opened 2021-12-30
Remove commented $namespace

phh:
Remove commented code to avoid PHPDoc tags being picked up by static analysers.

PHPStan errors (since 1.3.0):
```
  Line   app/Providers/RouteServiceProvider.php
 ------ -------------------------------------------------
  36     PHPDoc tag @var above a method has no effect.
  36     PHPDoc tag @var does not specify variable name.
```

### DIFF
--- a/app/Providers/RouteServiceProvider.php
+++ b/app/Providers/RouteServiceProvider.php
@@ -20,15 +20,6 @@ class RouteServiceProvider extends ServiceProvider
     public const HOME = '/home';
 
     /**
-     * The controller namespace for the application.
-     *
-     * When present, controller route declarations will automatically be prefixed with this namespace.
-     *
-     * @var string|null
-     */
-    // protected $namespace = 'App\\Http\\Controllers';
-
-    /**
      * Define your route model bindings, pattern filters, etc.
      *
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
👎 here this code is included as a commented out example. We could probably stop phpstan complaining by also commenting out the doc comment.

phh:
Tryed that and it works. Want me to comment out the doc comment or block instead?

GrahamCampbell:
Let's wait to get Taylor's feedback, to see if this is wanted at all.

phh:
An alternative approach is also:

```
    /**
     * The controller namespace for the application.
     *
     * When present, controller route declarations will automatically be prefixed with this namespace.
     *
     * // @var string|null
     */
    protected $namespace = 'App\\Http\\Controllers';
```

Yeah, if this is wanted at all.

GrahamCampbell:
No, that is really not good. It means someone can't easily uncomment the whole block.

GrahamCampbell:
In fact, it's actually invalid syntax, which is why phpstan is not complaining - it can't parse it.

PR #5756, opened 2022-01-02
Force LF EOL in .gitattributes

ThisGitHubUsernameWasAvailable:
Fix issue when GitHub Desktop on windows changes LF to CRLF

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
 * text=auto
+* text eol=lf
 *.css linguist-vendored
 *.scss linguist-vendored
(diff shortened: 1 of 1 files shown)


driesvints:

ThisGitHubUsernameWasAvailable:
Why not merge it? `.editorconfig` in laravel alrady has `end_of_line = lf` setting, so laravel devs explicitly decided use LF, and this PR will help it to be consistent and guarantee what it will not be messed up by GitHub Desktop or any other windows GIT client.

PR #5757, opened 2022-01-02
[9.x] Make .env.example sync with new changes

rezaamini-ir:
In this PR #5568  the key has been changed, but it has not been changed in the .env.example file!


u01jmg3:
@rezaamini-ir: glad I could help

PR #5758, opened 2022-01-03
[9.x] Use latest version of laravel-cors

barryvdh:
See https://github.com/fruitcake/laravel-cors/pull/540

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
     "license": "MIT",
     "require": {
         "php": "^8.0",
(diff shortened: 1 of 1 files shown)


driesvints (approved):

volkankaban:
Do we need to change/replace, use Fruitcake\Cors\CorsService;  ?

Still exist with latest framework, so I got confuse..
https://github.com/laravel/framework/blob/9.x/src/Illuminate/Http/Middleware/HandleCors.php

PR #5762, opened 2022-01-12
[9.x] Add Ignition

freekmurze:
This is the only change needed to get a beautiful Laravel specific error page whenever an exception occurs.

🌞
CleanShot 2022-01-12 at 21 38 48@2x

🌚
CleanShot 2022-01-12 at 21 38 33@2x

.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,8 @@
         "laravel/sail": "^1.0.1",
         "mockery/mockery": "^1.4.4",
         "nunomaduro/collision": "^6.0",
-        "phpunit/phpunit": "^9.5.10"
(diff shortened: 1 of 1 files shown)


dinhquochan:
what about the `facade/ignition` package? why not upgrade `facade/ignition` package?

driesvints:
Thanks! 🌞 🌑

freekmurze:
@dinhquochan We'll not be using the facade vendor name anymore in the future.

dinhquochan:
> @dinhquochan We'll not be using the facade vendor name anymore in the future.

Got it 👍

PR #5765, opened 2022-01-13
[9.x] Use <env> element

jasonmccreary:
This changes the default configuration to avoiding a developer experience where you spend hours trying to determine why your CI configuration is not working.

The `<server>` element has a higher than system environment variables (e.g. `export APP_ENV=ci`). As such, anything currently set in `phpunit.xml` can only be overridden by setting the configuration within your application (e.g. `app()->config('app.env', 'ci')`).

The `<env>` element is lower than system environment variables. For more detail, I wrote about [configuration precedence during testing](https://jasonmccreary.me/articles/laravel-testing-configuration-precedence/).

Considering this repo is static, this could likely be retargeted at `8.x`. However, I felt the upcoming `9.x` release provided a clean transition point.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -18,14 +18,14 @@
         </include>
     </coverage>
     <php>
-        <server name="APP_ENV" value="testing"/>
-        <server name="BCRYPT_ROUNDS" value="4"/>
-        <server name="CACHE_DRIVER" value="array"/>
-        <!-- <server name="DB_CONNECTION" value="sqlite"/> -->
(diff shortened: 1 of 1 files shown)


jasonmccreary:
@michaeldyrynda, yeah, maybe someone more familiar with the inner workings of Laravel might know why it _needs_ to be `<server>`. However, in practice, I've been using `<env>` since 5.8 without any issue, and, to the point of this PR, easier overwriting.

PR #5767, opened 2022-01-15
Add More Type Hints to the App Code

Menelion:
In an effort to make Laravel more compatible with modern PHP and its best practices, I added more type hints to the app first. Will do the same for the framework gradually if this PR is considered valuable.

See [this discussion](https://github.com/laravel/framework/discussions/36595).


taylorotwell:
If we address this we will handle it internally amongst the Laravel core team. Personally I don't see much use or benefit in adding `void` return types to things like migrations and tests.

PR #5768, opened 2022-01-17
Use Hash::make for generating password

thewolf27:


driesvints:
This was done specifically so the call and calculation doesn't have to be done anymore. Feel free to do this in your own app.

PR #5773, opened 2022-01-20
Bump axios version

balu-lt:


taylorotwell:
Existing constraint already installs that version?

balu-lt:
> Existing constraint already installs that version?

It doesn’t :(

driesvints:
@taylorotwell With 0.x releases. Only the latest  0.1.x version gets installed.

PR #5774, opened 2022-01-21
[9.x] Add type to UserFactory for generic parent class

morrislaptop:
Allows PHPStorm to infer the type from the factory class


database/factories/UserFactory.php
```diff
@@ -2,9 +2,13 @@
 
 namespace Database\Factories;
 
+use App\Models\User;
 use Illuminate\Database\Eloquent\Factories\Factory;
 use Illuminate\Support\Str;
 
+/**
+ * @implements Factory<User>
```
driesvints, line 10: Can you use a FQN here?

PR #5775, opened 2022-01-22
Grammar improvement

krvvko:


driesvints:
This unfortunately breaks the length of the description blocks that we maintain.

krvvko:
what do you mean by "breaks the length of the description blocks that we maintain"? 
image
As I see there is nothing breaking

michaeldyrynda:
Each line is three characters shorter than the preceding line (with a period at the end of a sentence being an exception)

PR #5776, opened 2022-01-22
Update phpunit.xml

brunolorente:
Use `<env>` to override 'APP_ENV' instead of `<server>`, I struggled with this and I think that going on this way the setup is more semantic.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -18,7 +18,7 @@
         </include>
     </coverage>
     <php>
-        <server name="APP_ENV" value="testing"/>
+        <env name="APP_ENV" value="testing"/>
         <server name="BCRYPT_ROUNDS" value="4"/>
         <server name="CACHE_DRIVER" value="array"/>
         <!-- <server name="DB_CONNECTION" value="sqlite"/> -->


brunolorente:
Maybe the other `<server>` elements under the `<php>` element should be replaced, please take a look at the use case below for a better understanding:

```
<?php

namespace Database\Seeders;

use Illuminate\Database\Seeder;

class DatabaseSeeder extends Seeder
{
    /**
     * Seed the application's database.
     *
     * @return void
     */
    public function run()
    {
        \App\User::factory(1)->admin()->create();
        if ($_ENV['APP_ENV'] == 'local') {
            \App\User::factory(10)->create();

            \App\Role::factory(1)->admin()->create();
            \App\Role::factory(1)->visitor()->create();
            \App\Role::factory(1)->supplier()->create();

            \App\Interest::factory(50)->create();

            \App\CategoryType::factory(1)->place()->create();
            \App\CategoryType::factory(1)->activity()->create();
            \App\CategoryType::factory(1)->event()->create();
            \App\CategoryType::factory(3)->create();

            \App\Category::factory(50)->create();
            \App\City::factory(20)->create();
        }

        if ($_ENV['APP_ENV'] == 'testing') {
            \App\User::factory(1)->test()->create();

            \App\Role::factory(1)->admin()->create();
            \App\Role::factory(1)->visitor()->create();
            \App\Role::factory(1)->supplier()->create();
        }
    }
}
```


I want to run different seeders for each environment, and I was faced with $_ENV['APP_ENV'] didn't exist.

kang8:
@brunolorente You can use `config('app.env')` replace `$_ENV['APP_ENV']`.  This eliminates the need to change `phpunit.xml`.

brunolorente:
Yep @kang8 I know, but if you want to apply SOLID principles, you should be the most framework agnostic as you want, for that reason I'm using `$_ENV` instead of `config()` or `env()`

GrahamCampbell:
You should use `$_SERVER` in your code. That is the correct place to read environment variables from (yes, PHP has bad naming for legacy reasons).

brunolorente:
Hi @GrahamCampbell I can see now that laravel 9.x uses <env> tags instead of <server> as I said here, So please, I would like to receive an explanation for a better understanding.

PR #5778, opened 2022-01-25
PSR-12 constructor parentheses

krzysztofrewak:
As PSR-12 says:

> When instantiating a new class, parentheses MUST always be present even when there are no arguments passed to the constructor.
> 
> `new Foo();`

https://www.php-fig.org/psr/psr-12/


driesvints:
We don't follow these. Feel free to do this in your own app.

PR #5782, opened 2022-01-30
[9.x] Bumps `nunomaduro/collision` dependency

nunomaduro:
This pull request bumps `nunomaduro/collision` dependency to ensure people get the `--coverage` option even when running a `composer update  --prefer-lowest`.


vini123:
虚幻之物对应着冥冥之路！：《羽》

PR #5783, opened 2022-01-31
Use APP_NAME as title, fallback to "Laravel"

jdecode:
Use APP_NAME environment variable as title, by default, and default to "Laravel"

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -4,7 +4,7 @@
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1">
 
-        <title>Laravel</title>
+        <title>{{ config('app.name', 'Laravel') }}</title>
 
         <!-- Fonts -->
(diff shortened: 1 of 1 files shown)


driesvints:
This has been attempted many times before. Feel free to do this in your own app.

jdecode:
> This has been attempted many times before. Feel free to do this in your own app.

This is not a good enough reason.

Why not provide this as default?

Every app will have to change their title to something other than "Laravel", and by making the change we are allowing the developers to run this without changing any code on the FE, rather just changing the environment variable.

I don't see any value of keeping static text "Laravel", when it is the default fallback.

Could you please explain the reason for not doing this, or perhaps you can share the link of previous discussion on the topic, and I can go read those first.

i74ifa:
It will never change😅

jdecode:
"Laravel Breeze" - one of the most popular Laravel Starter Kit - is using APP_NAME, which defaults to "Laravel".

https://github.com/laravel/breeze/blob/7efa557a0deecc1981c7907f599debb720dd3121/stubs/inertia-common/resources/views/app.blade.php#L8

PR #5784, opened 2022-02-01
Update domain entry

saidbakr:
Replacing `null` with `$_SERVER['HTTP_HOST']` in `domain` entry. To allow valid domain for multiple domains apps.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -155,7 +155,7 @@ return [
     |
     */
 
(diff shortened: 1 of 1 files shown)


driesvints:

saidbakr:
> Feel free to do this in your own app.

Are there any side effects to adopt out of the box troubleshooting?
Without this solution multiple domains applications suffer from rejected cookies by clients.

PR #5785, opened 2022-02-01
Fixing the length of the description block

krvvko:
I know that every next comment description block should be exactly 3 symbols shorter, than on the previous line. I found the bug and fixed it!


driesvints:
That's not the case. Every first line is three characters longer than the second one. Every second line is two characters longer than the third one. The current block is correct.

driesvints:
I gave the explanation in my previous comment?

PR #5786, opened 2022-02-01
Fixing the length of the description block

krvvko:
There is a bug in one of your description blocks. The second comment line is 4!! (should be 3) symbols shorter than the first one. But don't worry - I fixed this issue. That will definitely improve the project!


taylorotwell:
Ha! I got it fixed. Thank you for bringing this to my attention ❤️

PR #5796, opened 2022-02-04
[9.x] Revert default serialization

driesvints:
It seems this causes validation errors not to be persisted for views. It might be best we revert this for now until we find a better way as this usage is very common and standard in Laravel apps.

Fixes https://github.com/laravel/framework/issues/40773


taylorotwell:
I'll just remove the whole key - we should likely adjust how we store view errors in the session. We should serialize them as JSON and rebuild the `ViewErrorBag` object when retrieving them from the session most likely.

PR #5804, opened 2022-02-09
[9.x] void return type for example tests

rogervila:
Since Laravel 9.x requires PHP 8.0 and above, void return types may be used for example tests.


taylorotwell:
I don't think these really add anything to the code quality in this case? All tests are "void" returns.

PR #5807, opened 2022-02-10
Align min PHP version with docs

u01jmg3:
See https://github.com/laravel/docs/blob/9.x/upgrade.md?plain=1#L49


driesvints:
Thanks!

PR #5809, opened 2022-02-10
[8.x] Revert "Fix .gitattributes consistency with .editorconfig"

driesvints:
Reverts laravel/laravel#5802

I don't feel this is needed since this would apply only in userland, not when developing this skeleton repo. The editorconfig file already has the proper eol defined for the project.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,4 @@
 * text=auto
-* text eol=lf
 
 *.blade.php diff=html
 *.css diff=css


GrahamCampbell:
Agreed.

ThisGitHubUsernameWasAvailable:
But why? I see **only pros and no cons for this change**, Could you justify it with something more than just "I don't feel this"?
It helps to keep EOLs consistent across all project, so why not?

ThisGitHubUsernameWasAvailable:
@taylorotwell what do you think about this?

PR #5815, opened 2022-02-14
Removed MAIL_ENCRYPTION and MAIL_FROM_ADDRESS env vars

gerardnll:
Variables set in the .env as null or empty do not use the default value when calling env(). In some cases this produces errors, for example when sending an email. Symfony Mailer needs a From/Sender header and because MAIL_FROM_ADDRESS value is null in the .env, the default set in the config/mail.php is not used. This also occurs for the MAIL_ENCRYPTION var.


taylorotwell:
I just added the default from value.

gerardnll:
Afaik MAIL_ENCRYPTION also has a null value on the .env but 'tls' is on the mail.php file so that is never taken.

mTrax-:
> Afaik MAIL_ENCRYPTION also has a null value on the .env but 'tls' is on the mail.php file so that is never taken.

Yes, it would be great to be able to configure no encryption.

PR #5816, opened 2022-02-15
[9.x] Remove default commented namespace

driesvints:
I want to propose to remove the default namespace property in the `RouteServiceProvider` to push people to use class referenced controllers more. The current setting causes confusion and leads to situations like https://github.com/laravel/framework/pull/41021.


nunomaduro (approved):
On top of what Dries mentioned, also confuses Larastan because there is annotations for a property that is commented.

PR #5819, opened 2022-02-16
[9.x] Improve formatting

Doc077:
Add trailing commas to method calls


GrahamCampbell:
Note that this is not an argument list, but a method call (which means this was supported since PHP 7.4).

Doc077:
Thanks @GrahamCampbell  for correcting

taylorotwell:
I think it is OK as is.

PR #5820, opened 2022-02-17
Update sanctum config file

suyar:
See [sanctum v2.14.1](https://github.com/laravel/sanctum/releases/tag/v2.14.1) and this pr [#5702317](https://github.com/laravel/sanctum/commit/57023176c5a77d5108cee2fcadceabda00af814b).

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -1,5 +1,7 @@
 <?php
 
+use Laravel\Sanctum\Sanctum;
+
 return [
 
     /*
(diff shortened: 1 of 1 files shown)


driesvints:
We'll need to bump sanctum in composer.json to this version as well.

suyar:
> We'll need to bump sanctum in composer.json to this version as well.

I think this only affects creating new `laravel` projects, the current `sanctum` version is `"laravel/sanctum": "^2.14"`, newly created projects will automatically install the latest `release`. 

This configuration was changed in `sanctum v2.14.1`, I think is fine.

driesvints:
If you reference a new public API you always need to make sure that the constraints of a library are resolved in a way that it's always available. I agree probably everything would go fine but if for some bizar reason someone installs 2.14.0 with --prefer-lowest this would break.

suyar:
> If you reference a new public API you always need to make sure that the constraints of a library are resolved in a way that it's always available. I agree probably everything would go fine but if for some bizar reason someone installs 2.14.0 with --prefer-lowest this would break.

You are right.

PR #5821, opened 2022-02-17
If XHR request then don't redirect already logged in users

mikemellor11:
When accessing auth routes via an api it doesn't seem like you'd ever want a route being returned so this just wraps that already existing redirect code in a json check.

https://stackoverflow.com/questions/33029676/disable-redirect-in-authattempt-laravel-5-1
https://stackoverflow.com/questions/33341150/redirect-if-authenticated-logic-in-laravels-built-in-auth


taylorotwell:
The `guest` middleware is never intended to be used on API routes and I can't think of a good case for doing so?

mikemellor11:
The `guest` middleware is enabled by default and can't be commented out or removed easily from `app/Http/kernal.php` without breaking other authentication calls.

PR #5822, opened 2022-02-18
Some Styles To Default Login & Register Button

SagheerHussainZardari:
The buttons were just text so the user can not understand whether it's clickable or not so I fave bg and shadow and made it rounded  so everybody can understand.


SagheerHussainZardari:
Edited Lines Are 30, 33

driesvints:

PR #5825, opened 2022-02-21
[9.x] Replace Laravel CORS package

driesvints:
Depends on https://github.com/laravel/framework/pull/41137 which also needs to be tagged first.


volkankaban:
guzzle\src\Exception\RequestException.php:112
 
        return new $className($message, $request, $response, $previous, $handlerContext);

Shivaone1 (commented):
All Update Then Facing these type of errors

Class "Fruitcake\Cors\CorsServiceProvider" not found

PR #5835, opened 2022-03-06
[9.x] add `throw` to filesystems config

ankurk91:
Ref PR https://github.com/laravel/framework/pull/41308

I kept it `false` so it does not change the current behaviour.

One can make it `true` to get errors related to file operations such as uploading to S3 via `Storage::put()`

Thanks.


driesvints:
Thanks @ankurk91

PR #5847, opened 2022-03-21
[9.x] Make shorter default registered web route

Anton5360:


driesvints:
See many many previous attempts.

PR #5848, opened 2022-03-21
[9.x] Set primary JS dependencies as non-dev

enderandpeter:
The dependencies listed in package.json are crucial for the frontend's basic operation. They should be set as regular      `dependencies`, not as `devDependencies`. It is no more appropriate to put `axios` in the `devDependencies` than it is to put `guzzlehttp/guzzle` in `require-dev`.

package.json [`devDependencies`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#devdependencies) works very identically to composer.json's [`require-dev`](https://getcomposer.org/doc/04-schema.md#require-dev), in that those sections are for dependencies that are not needed for production, typically test-related packages.

Just like how `composer install --no-dev` is for [installing only packages in the `require` section](https://getcomposer.org/doc/03-cli.md#install-i), `npm install --production`   is for [installing only the `dependencies`](https://docs.npmjs.com/cli/v8/commands/npm-install#description).

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
         "prod": "npm run production",
         "production": "mix --production"
     },
-    "devDependencies": {
+    "dependencies": {
         "axios": "^0.25",
         "laravel-mix": "^6.0.6",
         "lodash": "^4.17.19",


driesvints:
I think I agree with most here but not with laravel-mix which is purely a build tool and not production code.

enderandpeter:
Thanks for the response!

@driesvints  `laravel-mix` does appear to be used in production because the `mix` command must be available there to run the command `mix production`. Doesn't that command come from the npm pacakage?

@GrahamCampbell I will look more into this for sure, but I have some doubts about what you're saying. When you run `npm run production` in an prod environment that does not have a `node_modules` folder with all those packages in `devDependences`, none of those packages are going to be installed. And so all the code that `laravel-mix` builds (again it would **have** to do this with the `laravel-mix` package that would **have** to be on production to run that command) would not have any of the `devDependencies` to use. And so what would the build production code use for axios? The source code files you make that refer to `axios`, how would your prod build files access this library?

And what if you are using lodash? How would `mix production` make sure your build code has `lodash` then? It would only come from the dependencies, which would not be there if they are in `devDepdendencies` and you build for production. Surely you see the problem if you never install these dependencies in production? Where would they come from, then?

I will try some tests. First of all, I am quite sure that if you run `npm install --production` with the `devDependencies` as they are, I would not think that `laravel-mix` will be available because it would not be installed. Where would it come from otherwise?

Likewise with `axios`, I will see if it is still available after building with `npm install --production`. Again, I cannot imagine it would be, because it would only be coming from npm, which would not install it in a production build. And it's not embedded in your source files. Again, the only place it would be coming from is the node_modules folder.

driesvints:
I don't want to get into a too big discussion about this. I've said how I feel about it. Mix is a build tool, not code that your app runs.

enderandpeter:
Yes indeed, in my view what I am saying is very obvious and does not need any big discussion. However, since doubts have been expressed, I am compelled to provide more info and demonstration, and so I will do so.

Correct, `mix` is a build tool that comes from the npm package `laravel-mix`, which **must** be on production in order to use it up there, and it will not be on production if it is in your `devDepdencies` and you install your npm dependencies on production. Those other packages (axios, lodash, postcss) are clearly used in all environments as well. That's all I'm saying.

To see what I'm describing, please try the following in any Laravel project:

* With `laravel-mix` set in `devDependencies`, run `npm install --production`.
* Check your node_modules folder. Note how `laravel-mix` is no longer present.
* Run `npm run prod`
* You will get the error `sh: 1: mix: not found`

For the same reasons, axios, lodash, postcss, and anything else you want to be there when building the production files must be set in`dependencies`, not `devDependencies`:

image

taylorotwell:
Change it in your own app.

enderandpeter:
Yes you will always have to change this in your own app, unless you run `npm install` with no awareness of the environment so that `devDependencies` are always installed. Just keep in mind that we are making developers have to change this to `dependencies` when running `npm install` for production. These dependencies are in fact used there because they come from `node_modules`. This is not particular to a node server.

realodix:
I don't think this is necessary, unless you put `public/css/*` and `public/js/*` into the `.gitignore` file

enderandpeter:
Even if you did not ignore the built files, that would not affect the behavior of `npm install --production`. Again, all of those dependencies are used on production so they should really be under `dependencies`. To create the build files on any environment, `laravel-mix` must be present. And if you're using `axios` or `lodash` in your project and `postcss` or `sass` to build CSS, then they must also be installed in production, regardless if the build files are unwisely saved in your repo or if you create them up there as you should.

I think the main point of confusion here comes from lack of awareness of how `npm install --production`, or `npm install` with `NODE_ENV=production` is the proper way to install production dependencies, just like how `composer install --no-dev` installs only `require` dependencies and not `require-dev`. But if you never use the `--no-dev` flag in production, you will always install `require-dev` and be confused about why `laravel/dusk` keeps installing in your production, for instance.

Likewise, once a developer starts putting actual `devDependencies` in thier `package.json` (like `jest` and such) and understand how to run `npm install` for production versus dev, they will see how they will need to set this properly and wonder why this is not already set.

alaminhb:
Laravel online tutorial link,
my name alamin bhuyan
Thanks

PR #5857, opened 2022-03-31
[9.x] Set framework version ^9.6

Aamish99:


driesvints:
The current constraint already installs this.

PR #5860, opened 2022-04-07
Update composer.json

SamuelMwangiW:
Following commit 9986d51, running `composer install` nukes ./bootstrap/cache/packages.php.

Composer seemingly runs package discovery before the `post-install-cmd` script command.
image

This PR proposes to unlink `bootstrap/cache/packages.php` in `pre-install-cmd` instead.

Running `composer update` is not affected.

Thanks for the great work!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         }
     },
     "scripts": {
-        "post-install-cmd": [
+        "pre-install-cmd": [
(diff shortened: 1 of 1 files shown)


driesvints (approved):

taylorotwell:
This could create a problem with the file being regenerated if a request comes in during the installation process. Probably we should move package:discover to both post-install and post-update.

PR #5864, opened 2022-04-14

sinus22:


driesvints:
This isn't correct. This param makes use of the spread operator so each individual argument needs to be a string.

PR #5865, opened 2022-04-15
9.x

mwakalingajohn:
Use "InnoDb" as the default engine for mysql database config. This is to avoid this error on fresh install:

 SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 1000 bytes (SQL: alter table `users` add unique `users_email_unique`(`email`))


driesvints:

PR #5868, opened 2022-04-20
 [9.x] Add missing maintenance to config

ibrunotome:
[#40102](https://github.com/laravel/framework/pull/40102)


ibrunotome:
Maybe a mention in https://laravel.com/docs/9.x/configuration#maintenance-mode should also be added?

PR #5869, opened 2022-04-24
[9.x] Remove @return PHPDoc tag in favor of return types hints

nhedger:
### Description

This PR removes the PHPDoc `@return` tags in favor of return types hints in all service providers under `app/Providers`.

### Why

Now that Laravel requires a minimum PHP version of `8.0`, we are able to use return type hints.


driesvints:

nhedger:
Already do, just thought it would be nice to implement here also, no biggie :)

PR #5870, opened 2022-04-27
Bump laravel/framework version to latest

jnoordsij:
Required for password rule translations to work properly (see https://github.com/laravel/framework/pull/42060)

In the current skeleton, if someone was to update the `password` entries in `validations.php` and resolved an older `laravel/framework` version, the changes would not affect the actual displayed message, which could be quite confusing.


driesvints:
Installing a fresh Laravel app already installs this version but yes, we can do this. Thanks.

PR #5871, opened 2022-04-27
Move laravel tinker to dev deps

Tofandel:
As it's not a production dependency and is only used during development


taylorotwell:

GrahamCampbell:
It can be used in production, and vapor has support for that.

mfn:
Can confirm, multiple times had I the need to "quickly check something" and having tinker was incredible valuable.

PR #5876, opened 2022-05-04
syntax error on line no 49

johnconnor2996:
return Limit::perMinute(60)->by($request->user()?->id ?: $request->ip());
to 
return Limit::perMinute(60)->by($request->user()->id ?: $request->ip());


michaeldyrynda:
This is valid PHP 8 syntax. 

https://www.php.net/releases/8.0/en.php#nullsafe-operator

driesvints:
You're indeed not running php 8 or higher which is a requirement of Laravel 9

PR #5879, opened 2022-05-05
[9.x] Add specific test user in seeder

driesvints:
The very first thing I do when creating a new Laravel app is to setup a specific test user with test credentials that I can use to login without having to register first. I thought to give it a go to not have to do this manually anymore each time. Tried to use a generic as possible name and email address.


jbrooksuk:
Should we also include a password there too?

driesvints:
@jbrooksuk the default for users is `password` which is fine imo.

Helel-arm:
god jab.

PR #5886, opened 2022-05-12
[9.x] add cache option to `view.php`

bernardwiesner:
In this [PR](https://github.com/laravel/framework/pull/41859) `cache` option was added, would be nice to have it in view.php for more visibility and usability.

### DIFF
--- a/config/view.php
+++ b/config/view.php
@@ -19,6 +19,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | View Cache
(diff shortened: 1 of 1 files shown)


Jackysi:
I actually almost lost my mind with view caching when trying to quickly test a code change directly within view.

xvital:
Very useful feature, I appreciate it!

Much easier to debug classic Blade templates now.

PR #5887, opened 2022-05-16
Update composer.json

level39:
Fixes conflict with roave/security-advisories (https://github.com/Roave/SecurityAdvisories/blob/latest/composer.json#L190) during installation or update.

Composer reports:

 Problem 1
    - laravel/laravel is present at version 1.0.0+no-version-set and cannot be modified by Composer
    - roave/security-advisories dev-latest conflicts with laravel/laravel <=5.8.38.
    - roave/security-advisories is locked to version dev-latest and an update of this package was not requested.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,6 @@
 {
     "name": "laravel/laravel",
+    "version": "9.1.8",
     "type": "project",
     "description": "The Laravel Framework.",
     "keywords": ["framework", "laravel"],


taylorotwell:
Not sure why this would be necessary.

mfn:
- ppl like to use roave/security-advisories
- a security advisory at some point was published, invalidating all Laravel <5.8.something installation
- from https://github.com/Roave/SecurityAdvisories/commit/21d0504612c544af1d8ce000e5e0301aa8bbde7c#r72869209
  > The Laravel installer sets the project's "name" in the composer.json to "laravel/laravel". There is also no version specified, so composer assumes "1.0.0.0+no-version-set".
  > 
  > To fix this in your project, just change the "name" in your composer.json to something else.

This can be easily tested with a fresh installation of Laravel:
```
+ $ composer create-project laravel/laravel new-laravel-project
Creating a "laravel/laravel" project at "./new-laravel-project"
…
Application key set successfully.

+ $ cd new-laravel-project

+ $ composer require roave/security-advisories
Using version dev-latest for roave/security-advisories
./composer.json has been updated
Running composer update roave/security-advisories
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - laravel/laravel is present at version 1.0.0+no-version-set and cannot be modified by Composer
    - roave/security-advisories dev-latest conflicts with laravel/laravel <=5.8.38.
    - Root composer.json requires roave/security-advisories dev-latest -> satisfiable by roave/security-advisories[dev-latest].

You can also try re-running composer require with an explicit version constraint, e.g. "composer require roave/security-advisories:*" to figure out if any version is installable, or "composer require roave/security-advisories:^2.1" if you know which you need.

Installation failed, reverting ./composer.json and ./composer.lock to their original content.
```

Workarounds are:
- rename the newly generate project to something else, not laravel/laravel
- provide an explicit version number (like this PR tried).

FWIW, I agree this PR isn't the best solution; not sure what is exactly.

valorin:
Hard-coding the Laravel version in `composer.json` doesn't really make sense here. It'll likely never be updated in projects, so it'll just attract future security vuln reports, that'll be just as confusing with an outdated version number.

It'd be far better to recommend changing the `name` field in `composer.json` on a fresh Laravel install, and recommending Roave add a notice about changing it when using `roave/security-advisories` - it'd be a common problem across frameworks, I'd expect.

PR #5893, opened 2022-05-25
Use shared application instance instead of helper in welcome view

mabdullahsari:
Hello 👋🏼

The reasoning behind this tiny change is to expose a feature unknown to the vast majority of Laravel developers.

Since the welcome view is the first thing opened up by many developers, it would be a nice way of telling "hey, you can also access the app instance like this!". Using helpers is already a very well-known thing, but knowing the alternatives could also be beneficial IMHO.

Thank you for taking your time reviewing this.


taylorotwell:
I'm not sure this is necessary?

PR #5895, opened 2022-05-27
[9.x] Switch to ESM imports

jessarcher:
The Laravel skeleton currently uses Node/CommonJS `require()` statements to load several dependencies. These statements are not supported when using Vite because it exclusively leverages ES Modules (which use the `import` syntax) to achieve lightning-fast speeds.

This PR replaces the `require` statements with equivalent `import` statements to make the transition to Vite smoother. Webpack supports both, so this should have no noticeable impact for Mix users other than improving type information in their IDEs :nerd_face:

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -1,1 +1,1 @@
-require('./bootstrap');
+import './bootstrap';
--- a/resources/js/bootstrap.js
+++ b/resources/js/bootstrap.js
@@ -1,4 +1,5 @@
-window._ = require('lodash');
+import _ from 'lodash';
+window._ = _;
 
 /**
  * We'll load the axios HTTP library which allows us to easily issue requests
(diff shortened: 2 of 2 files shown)


FaridAghili:
What about updating `webpack.mix.js` to use `import`?

```js
const mix = require('laravel-mix');

import mix from 'laravel-mix';
```

jessarcher:
> What about updating `webpack.mix.js` to use `import`?

@FaridAghili You'll get the following error if you `import` in the Mix config:

image

ankurk91:
One can read about ESM
https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

browner12:
For those who may be confused by `import` vs `require` like me, this is called a "side effect import".  It doesn't actually import anything, but rather runs the global code of the referenced file.  So even though `bootstrap.js` isn't *export*ing anything, this is causing any of the global code in there to execute.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#import_a_module_for_its_side_effects_only

PR #5896, opened 2022-05-28
fixed syntax errors in the configureRateLimitimg method by setting th…

Tope19:
…e reguest user and ip to optional


driesvints:
It's already set that way. Most likely you're using an unsupported php version.

PR #5902, opened 2022-06-11
Bump guzzle to fix security vulnerability

fwdmo:
Security advisory: [GHSA-f2wf-25xc-69c9](https://github.com/guzzle/guzzle/security/advisories/GHSA-f2wf-25xc-69c9)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
     "license": "MIT",
     "require": {
         "php": "^8.0.2",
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This is not strictly necessary, in the same way guzzlehttp/guzzle hasn't bumped it's psr7 req to account for a recent vulnerability there.

PR #5903, opened 2022-06-12
? fix

M-Shahbaz:


GrahamCampbell:
This change looks wrong to me. The `?` is here to allow us to fallback to null.

PR #5904, opened 2022-06-13
[9.x] Vite

jessarcher:
This PR makes Vite the default in Laravel.

Please see https://github.com/laravel/framework/pull/42785 for more information.

Please note this PR gitignores the Vite compiled assets directory, so instead of committing your compiled assets, you would build them when you deploy.


.gitignore
```diff
@@ -1,4 +1,5 @@
 /node_modules
+/public/build
```
yob-yob, line 2: why is this added?? I tried to use `npm run build` locally so that my assets are production ready because I need to deploy to production, I was expecting that it should work on prod without the need to install npm packages...

does this mean that i should do an `npm install` then `npm run build` in prod? 

I'm getting error 

```
Vite manifest not found at: /releases/1670917824/public/build/manifest.json
```

Am I missing something here? maybe a link to a docs that helps me with deploying vite assets to prod?

yob-yob: never mind I found this https://github.com/laravel/vite-plugin/blob/main/UPGRADE.md#optional-git-ignore-the-build-directory

This is preference base and Laravel is just more inclined in building assets during deployment.

The `/public/build` can just be commented out.

PR #5909, opened 2022-06-16
Declare strict types

MathiasReker:
Force strict types declaration in all files.


antonkomarev:
Not following PSR-12 code style. And I'm 100% sure it will be rejected

PR #5915, opened 2022-06-18
Added php artisan init for simple project initialization

Afcyy:
It may take some time to run every command separately when you want to initialize existing project so I added php artisan init command which does simple configuration to save some time. 

php artisan init:

> **Copies .env.example to .env**
> **Calls php artisan key:generate**
> **Changes DB_DATABASE name in .env**
> **Creates Database**
> **Calls php artisan migrate**
> **Calls php artisan db:seed**
> **And serves application if wanted**


taylorotwell:

PR #5916, opened 2022-06-18
Add ConsoleOutput to Service Container

bbrist:
I have a use-case where I'm extending the LogManager to redirect logs to console output when running via artisan console. To do this elegantly, its necessary to know what the current verbosity level from the ConsoleOutput.

Registering the ConsoleOutput as a singleton in the service container will allow for that verbosity to be retrieved when desired to analyze the current verbosity level.

This PR adds that registration to the bootstrap/app.php and retrieves that binding in the artisan file.

### DIFF
--- a/artisan
+++ b/artisan
@@ -31,7 +31,7 @@ $app = require_once __DIR__.'/bootstrap/app.php';
 */
 
 $kernel = $app->make(Illuminate\Contracts\Console\Kernel::class);
-$output = $app->make(Symfony\Component\Console\Output\ConsoleOutput::class);
+$output = $app->make(Symfony\Component\Console\Output\ConsoleOutputInterface::class);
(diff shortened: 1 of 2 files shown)


bbrist:
Alternatively, this PR could be altered to register an instance of the ConsoleOutput being used as the argument for the Kernel->handle method in artisan, not unnecessarily registering the object for non-console executions.

i.e.
```
$status = $kernel->handle(
    $input = new Symfony\Component\Console\Input\ArgvInput,
    $app->instance(Symfony\Component\Console\Output\ConsoleOutput::class, new Symfony\Component\Console\Output\ConsoleOutput)
);
```

taylorotwell:
Hey there! You can just make these changes in your own app 👍

PR #5918, opened 2022-06-24
Replace .styleci.yml with first party pint.json

emargareten:


driesvints:
Right now we aren't planning this.

GrahamCampbell:
Pint rules don't precisely match StyleCI, nor do they support fixing the non-PHP files. ;)

PR #5919, opened 2022-06-24
Adding a default timezone option to .env.example and config/app.php

antonioanerao:
Hello. It's my first PR to a big project and I hope I'm doing it the right way. 
I added to .env.example an APP_TIMEZONE field and updated the timezone from config/app.php only use UTC as default if there isn't a APP_TIMEZONE in .env


alteredchimera:
Yes Please. This would be great for the Skeleton moving forwards, and using UTC as a default doesn't break BC. 

I'm assuming however, something would need to change in Carbon to pick this up automatically 🤔

davidbirkin:
Yes Please. This would be great for the Skeleton moving forwards, and using UTC as a default doesn't break BC.

I'm assuming, however, something would need to change in Carbon to pick this up automatically 🤔

antonioanerao:
> Yes Please. This would be great for the Skeleton moving forwards, and using UTC as a default doesn't break BC.
> 
> I'm assuming, however, something would need to change in Carbon to pick this up automatically thinking

What do you think should be changed in Carbon? I'm asking because I always change de config/app.php for my applications to get the APP_TIMEZONE from .env and I never had a problem with this and before I sent the PR I tried running some Carbon and everything went fine.

davidbirkin:
I was under the impression Carbon Defaulted to UTC, regardless of an .env() var. If thats not the case then this is a winner

medeirosrafael:
This probably will not be merged.. Others have tried before.. it's a good practice to keep your database and system in UTC timezone..

davidbirkin:
So it looks like Carbon uses `date_default_timezone_set()` in the CarbonTimezone.php

```
public function __construct($timezone = null)
{
        parent::__construct(static::getDateTimeZoneNameFromMixed($timezone));
}
protected static function getDateTimeZoneNameFromMixed($timezone)
{
        if ($timezone === null) {
            return date_default_timezone_get();
        }

        if (\is_string($timezone)) {
            $timezone = preg_replace('/^\s*([+-]\d+)(\d{2})\s*$/', '$1:$2', $timezone);
        }

        if (is_numeric($timezone)) {
            return static::parseNumericTimezone($timezone);
        }

        return $timezone;
}
```

Would this be a necessary change. Idk if im looking too far into this, but just what I see from looking

```
if (config('app.timezone') !== null) {
    return date_default_timezone_get(config('app.timezone);
}

if ($timezone === null) {
        return date_default_timezone_get();
}
```

Because even here in the Timezone of "America/New_York" a default application always returns a UTC timestamp for `Carbon::now()` or `now()`

davidbirkin:
> This probably will not be merged.. Others have tried before.. it's a good practice to keep your database and system in UTC timezone..

I understand this too. UTC is a transferrable timezone, but also being able to set a default_timezone would be good atleast for display purposes on the UI, which I understand the skeleton is not responsible for

driesvints:
Thanks but right now we're not looking into changing this. Feel free to change this in your own app.

PR #5920, opened 2022-06-26
[9.x] Added needful indexes for `failed_jobs` table

sethsandaru:
## Improvement

It is "nice-to-have" these kinds of indexes for Laravel applications since the beginning when we hit the "laravel new project". 

For debugging, analyze failure and check/re-trigger failed jobs faster, especially in a datetime range.

My current app has like 1M records of failed_jobs but it is a pain to search for something.

Cheers 😉


taylorotwell:
Add them to your project if you need them.

PR #5921, opened 2022-06-28
[9.x] Force https scheme in production environment

userdehghani:
Using https is more common these days and I think it needs special attention.


driesvints:

userdehghani:
Adding this piece of code has become a chore and I am surprised that laravel maintainer still ignores this.
@driesvints

PR #5923, opened 2022-06-30
[9.x] The unused route has been changed.

selcukcukur:
All authorization packages have the dashboard route by default so I think it's a sensible default.


driesvints:

jessarcher:
The link isn't wrapped in an `@if (Route::has('dashboard')` so it will throw a `Route [dashboard] not defined.` exception if the user manually creates a `login` route.

PR #5925, opened 2022-06-30
Vite Hot reload for blade

Wulfheart:
As per [this blog post](https://freek.dev/2277-using-laravel-vite-to-automatically-refresh-your-browser-when-changing-a-blade-file) by @freekmurze I thought to myself that this is a good addition to the default template as it allows the blade views to be hot reloaded on a change. I strongly believe that this belongs here.


Wulfheart:
Nice!

PR #5926, opened 2022-07-01
Add autoinc id column to password_resets table

samlambert:
It is generally good hygiene to have auto-incrementing id columns explicitly defined on all MySQL tables.


taylorotwell:
Feel free to add to your own application if you wish.

PR #5930, opened 2022-07-02
vite: remove comments from output

ankurk91:
Hi, @timacdonald 

Vite uses `esbuild` to minify js; which does not remove legal comments from output

Read more:
https://esbuild.github.io/api/#legal-comments

Removing comments from output can greatly reduce the bundle size and i think it should be default for Laravel.

I tried other options as well, for example

Remove debugger and console from minified build
```js
 drop: ['debugger', 'console'],
```
But it does not work for me.

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -10,5 +10,8 @@ export default defineConfig({
             ],
             refresh: true,
         }),
+        esbuild: {
+           legalComments: 'none',
+        },
     ],
 });


timacdonald:
I feel like we should keep these in place by default.

ankurk91:
What about extracting comments to a separate file?

ikramulmustafa (commented):
I think this should be default

jessarcher:
I agree that we shouldn't be removing these. Additionally, I'm not a fan of including config options that aren't specific to the Laravel integration. Anything that is Laravel-specific can be set by the plugin, but I don't see this as a candidate.

You're of course free to set this in your own project. You may also wish to address this upstream with Vite if there are ways to reduce the bundle size while still meeting licensing requirements.

PR #5931, opened 2022-07-04
Remove repetitions

Doc077:
We can avoid verbose repetitions by destructuring the `env` property.


jessarcher:
The downside with this is Vite will inject all of the `VITE_*` env vars, rather than just those we need.

PR #5934, opened 2022-07-06
laravel have a few bugs in validations in 8.*

farshidah:
hi there,
first in validating nested arrays after level 3,also  if you enabled `Validator::excludeUnvalidatedArrayKeys();` in AppServiceProvider`  , $requested->validated()` return all not validated parameters.
like this:
```
request : [
     "some":[
             "thing1":[
                      "one":"valid",
                      "two":"not_valid"
              ]
    ]
]
```
` $requested->validated()` return all inputs! and cant remove not valid parameters.

//*****************************************************

second bog is:

in validating nested arrays if a parameter is nullable and be in request with null value validation don't return any errors but  ` $requested->validated()` return an empty array!

[
    [
        "title": "this is required",
        "body": "this is nullable",
    ],
    [
        "title": "this is required",
        "body": null,
    ],
]

after validating return any error but ` $requested->validated()` return an empty array [].

if you don't fix this bug in laravel 9.* , I gonna be glad to fix this bugs :)

thank you laravel


farshidah:
@driesvints thank you for your answering 😂😂

timacdonald:
@farshidah if there are bugs in those packages, they should be addressed in the packages. Setting the requirements like this is not how we want to support fixes here.

PR #5935, opened 2022-07-06
Delete hidden class at auth menu

torralbodavid:
I have come across the casuistry of wanting to test the internal sections of my new laravel application but it is not possible to access them from the welcome view because the auth menu is hidden by default, so you can not log in or register into the aplication.

The div has a `sm:block`, but that will only display the information starting at 640px. For example, an iPhone 13 Pro Max in portrait mode is 428px wide, so the menu will never appear. (Please see https://tailwindcss.com/docs/responsive-design#overview).


taylorotwell:
Feel free to adjust as needed in your own application.

PR #5940, opened 2022-07-12
Fix: Typo in a comments

bakai-dev:


driesvints:
This unfortunately breaks the line length of each block.

PR #5941, opened 2022-07-12
Introduce npm run watch

snapey:
Adding a new script entry point for `npm run watch` provides similar behaviour to webpack version.  Assets are built, placed in the public folder and watched for changes to source files.

This is useful for people that don't want hot reloading and want the assets built and able to be committed to their repository without having to remember to stop the `npm run dev` and execute `npm run build`  

It is actually effective to have two windows open, one with `npm run dev` and one with `npm run watch`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
     "private": true,
     "scripts": {
         "dev": "vite",
-        "build": "vite build"
+        "build": "vite build",
+        "watch": "vite build --watch"
     },
     "devDependencies": {
         "axios": "^0.25",


jessarcher:
Hey @snapey, thanks for this. I'm in two minds about it.

Having two commands for development means a new developer must understand each one and make a decision on how to run it. A single command is less friction for the majority. I also personally don't like committing compiled assets, but it's easy enough for people to add this if they want it.

On the other hand, this Vite feature isn't very discoverable on its own.

I think I'd want to see more people making noise about this before we prescribe it.

snapey:
Thanks for the consideration Jess.  FYI, this is where I first raised the idea https://laracasts.com/discuss/channels/vite/equivalent-of-vite-watch  

I am sure there are many like me, that handle multiple small projects for clients that use shared hosting etc and don't want or need the complication of an automated build pipeline. My concern is the risk of deploying a project that does not have the latest css changes.

taylorotwell:
Will hold off for now.

PR #5944, opened 2022-07-14
[9.x] Vite 3 support

timacdonald:
## Dependent PRs

- https://github.com/laravel/vite-plugin/pull/77

> **Warning**: ~~We need to specify the version constraint. I've left it as `TODO` as I was not sure if we would be tagging versions between commit and when we merge this.~~ Jess: Done!

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,9 +6,9 @@
     },
     "devDependencies": {
         "axios": "^0.27",
-        "laravel-vite-plugin": "^0.4.0",
+        "laravel-vite-plugin": "^0.5.0",
         "lodash": "^4.17.19",
         "postcss": "^8.1.14",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -6,9 +6,9 @@
     },
     "devDependencies": {
         "axios": "^0.25",
-        "laravel-vite-plugin": "^0.4.0",
+        "laravel-vite-plugin": "// TODO",
```
timacdonald, line 9: 🚨 This needs to be updated before merge

jessarcher:
I've updated the plugin version constraint to `^0.5.0`. This has not yet been released yet, so I'll leave this PR in draft until it is.

jessarcher (approved):

PR #5949, opened 2022-07-17
replace kernel qualifier with import

rana01645:
This makes kernel.php more clean and readable!


GrahamCampbell:
I don't agree it does. I think this makes it much less clear - it's no longer obvious what middleware are app and which are framework.

driesvints:

PR #5952, opened 2022-07-23
Update font delivery

abenerd:
Seeing the non compliance of Google Fonts to GDPR I thought to update the CDN.
My non-designer eyes could see no difference on the site.


fwdmo:
I don't think that is required no one keep the default welcome view

abenerd:
Fair, but I still think it is a sensible default to have.

PR #5953, opened 2022-07-26
[9.x] Don't need to ignore vite config file

GrahamCampbell:
Whether the list is multi-line is controlled by single-line length. Once it exceeds the threshold, it is moved to multi-line.

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -5,8 +5,5 @@ php:
   finder:
     not-name:
       - index.php
-js:
-  finder:
-    not-name:
-      - vite.config.js
+js: true
 css: true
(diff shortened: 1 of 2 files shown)


driesvints (approved):

jessarcher:
Hi @GrahamCampbell,

I appreciate this change as I didn't like the ignore, but I do wish it would maintain the multi-line if the developer has chosen it.

The [documentation](https://laravel.com/docs/9.x/vite#configuring-vite) references the Vite config using a diff to visualise removing one of the entries. This is much clearer when they are on separate lines. There is also now a mismatch between the docs and what is provided by the skeleton.

Additionally, if you wish to make this same change the Jetstream and Breeze repos, please be aware that the install commands manipulate the Vite config stub depending on the stack, so they expect the stub to be formatted in a specific way. It may be better to have separate config stubs for each stack in that case.

PR #5954, opened 2022-07-26
[9.x] Introduce APP_DOMAIN

ludndev:


driesvints:
See previous attempts

PR #5955, opened 2022-07-26
Add missing favicon to welcome view

manuglopez:
Favicon in welcome view is missing and causing 404 error


driesvints:
A favicon.ico is suppose to be in the public web root. I doesn't needs to be added to this view.

PR #5963, opened 2022-08-03
Remove unused elements

MarvelousMartin:
Following selectors are never used in the `welcome.blade.php`:
- mr-2
- ml-2
- text-gray-200
- text-gray-300


GrahamCampbell:
But then this script is no longer `normalize.css v8.0.1`.

PR #5964, opened 2022-08-03
Add comma before 'so'

datlechin:
This PR is adding a comma before 'so' word


taylorotwell:
Kinda messes up the 3 character cascade.

PR #5968, opened 2022-08-11
Remove extra way

GhostWalkers:
The path is already written in use


timacdonald:
It is Laravel convention to use the FQCN in docblocks.

PR #5970, opened 2022-08-15
add .env.* to .gitIgnore

chongkan:
it should be an explicit conscious decision to remove all variations of .env. from the .gitignore and push it to the repo since the .env files usually contains sensitive information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 /storage/*.key
 /vendor
 .env
+.env.*
 .env.backup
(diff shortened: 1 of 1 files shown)


chongkan:
Note: I requested this a few days ago and if denied, I would like to know the reason to be able to better understand. e.chongkan@gmail.com

driesvints:
Please do not re-send this PR. Feel free to do this in your own app. Amongst other things, this would prevent updates tot he example env file.

PR #5975, opened 2022-08-22
[9.x] feature: `max_digits` and `min_digits` validation translations

danharrin:
laravel/framework#43797

### DIFF
--- a/lang/en/validation.php
+++ b/lang/en/validation.php
@@ -90,6 +90,7 @@ return [
         'numeric' => 'The :attribute must not be greater than :max.',
         'string' => 'The :attribute must not be greater than :max characters.',
     ],
+    'max_digits' => 'The :attribute must not have more than :max digits.',
     'mimes' => 'The :attribute must be a file of type: :values.',
     'mimetypes' => 'The :attribute must be a file of type: :values.',
     'min' => [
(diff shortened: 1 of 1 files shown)


lang/en/validation.php
```diff
@@ -90,6 +90,7 @@
         'numeric' => 'The :attribute must not be greater than :max.',
         'string' => 'The :attribute must not be greater than :max characters.',
     ],
+    'max_digits' => 'The :attribute must not have more than :max digits.',
```
zepfietje, line 93: To match the `max.string` and `digits` rules:
```suggestion
    'max_digits' => 'The :attribute must not be greater than :max digits.',
```

lang/en/validation.php
```diff
@@ -98,6 +99,7 @@
         'numeric' => 'The :attribute must be at least :min.',
         'string' => 'The :attribute must be at least :min characters.',
     ],
+    'min_digits' => 'The :attribute must not have less than :min digits.',
```
zepfietje, line 102:
```suggestion
    'min_digits' => 'The :attribute must be at least :min digits.',
```

lang/en/validation.php
danharrin: "be greater" in a numeric context suggests the value of the number, not its length

zepfietje: Yeah already thought so, probably better as is then.

PR #5978, opened 2022-09-02
Make email unique

martin-ro:
When seeding large amounts of users there is a chance for a duplicate entry SQL error because the [email column is unique](https://github.com/laravel/laravel/blob/64e01753985749fab6626a370b27baa8431e74e6/database/migrations/2014_10_12_000000_create_users_table.php#L19).

### DIFF
--- a/database/factories/UserFactory.php
+++ b/database/factories/UserFactory.php
@@ -19,7 +19,7 @@ class UserFactory extends Factory
     {
         return [
             'name' => fake()->name(),
-            'email' => fake()->safeEmail(),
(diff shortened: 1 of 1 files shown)


driesvints:
I know it was like this before. I'm not sure why we reverted this 🤔

PerryvanderMeer:
For some reason, it's removed over here: https://github.com/laravel/laravel/commit/0c3d1fabe5a9319131aa4e9d74b27f29e74a50fb

driesvints (approved):

PR #5980, opened 2022-09-07
Update UserFactory.php

EGYWEB-Mohamed:
Change Default Password from string to Hash::make("")
for more readability


taylorotwell:
Way slower.

PR #5982, opened 2022-09-10
visible login register link on mobile

AjayYadavSecugile:
Login and register link is not visible on the mobile. updated the code and made is visible on the mobile.


driesvints:
This is intentional.

PR #5983, opened 2022-09-11
Add new `composer command` to restyle the code

michaelnabil230:


davidbirkin:
I like this, however, In the event this is accepted, would it make more sense to make the command either; 

`composer pint` or `composer run-pint` so that there is a little more clarity to what it actually being run behind the scenes?

driesvints:
Feel free to add this to your own project

PR #5992, opened 2022-09-19
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,6 +2,9 @@ name: Tests
 
 on: [push, pull_request]
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   tests:
     runs-on: ubuntu-latest
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,9 +1,14 @@
 name: Tests
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+      - '*.x'
+  pull_request:
 
 permissions:
-  contents: read # to fetch code (actions/checkout)
+  contents: read
 
 jobs:
   tests:
--- a/.github/workflows/update-changelog.yml
+++ b/.github/workflows/update-changelog.yml
@@ -4,6 +4,9 @@ on:
   release:
     types: [released]
 
+permissions: {}
 jobs:
   update:
+    permissions:
+      contents: write # to push (stefanzweifel/git-auto-commit-action)
     uses: laravel/.github/.github/workflows/update-changelog.yml@main
--- a/.github/workflows/update-changelog.yml
+++ b/.github/workflows/update-changelog.yml
@@ -5,8 +5,9 @@ on:
     types: [released]
 
 permissions: {}
+
 jobs:
   update:
     permissions:
-      contents: write # to push (stefanzweifel/git-auto-commit-action)
+      contents: write
     uses: laravel/.github/.github/workflows/update-changelog.yml@main


GrahamCampbell:
Can you show an example how this can be exploited? PR builds surely only push to the fork?

driesvints:
I don't see why this is needed. Right now, no third party extensions are untrusted and as you say forks can't affect the origin repo.

sashashura:
Maybe I'm to late to answer, but still...

It is not a security vulnerability, so you are free to not accept it. It is about running workflows with minimal permissions needed in case any build tool/dependency is compromised. Like not running it with sudo unnecessarily. For example [this workflow run](https://github.com/laravel/laravel/actions/runs/3068508965/jobs/4956060138) was run with the following permissions:
```
GITHUB_TOKEN Permissions
  Actions: write
  Checks: write
  Contents: write
  Deployments: write
  Discussions: write
  Issues: write
  Metadata: read
  Packages: write
  Pages: write
  PullRequests: write
  RepositoryProjects: write
  SecurityEvents: write
  Statuses: write
```

The tests.yml is using `shivammathur/setup-php@v2` action referenced by tag `v2`. In case the repository is compromised and someone pushes malicious changes and re-tags them as v2 your workflow will start running the changes without anything changed in your repository. There is another way to mitigate it as referencing actions by hash, but nonetheless restricting the potential damage is also an option.

I'm not an expert in PHP, but I think `composer install --prefer-dist --no-interaction` installs dependencies from `https://github.com/laravel/laravel/blob/9.x/composer.json`. Any compromised dependency will automatically run on the first `push` trigger with elevated permissions and can modify the repository.

driesvints:
Thanks @sashashura. We have no reason to believe that action will be compromised. Also, if any of the dependencies in our composer.json are compromised then every Laravel app will probably be compromised, not just this one. We have no reason to believe any of that will happen anytime soon.

But I'll have a closer look at this soon and perhaps take some action.

sashashura:
The subtle difference between compromised package sneaking into release is that the release is triggered relatively infrequently, so it may get detected by the time. Pushes happens often and the malicious code may change the repository during elevated CI run.

driesvints:
After talking to the GitHub team about this I'm reconsidering it. We do have the ability to set read/write permissions through the Repo settings but it might be best to do it in the workflow file specifically so it's explicit.

.github/workflows/tests.yml
```diff
@@ -1,6 +1,14 @@
 name: Tests
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+      - '*.x'
+  pull_request:
```
driesvints, line 8: Also added this to only build pushes on *.x branches and master like on framework.

driesvints (approved):

PR #5993, opened 2022-09-26
Prevent silently discarding model attributes by default

tillkruss:
I don't see a reason why this shouldn't be the default.

Related https://github.com/laravel/framework/pull/43893.


driesvints:
Might be a better idea to do this in framework core so people can opt out with a different method call? That way we prevent from adding this to the skeleton.

driesvints:
Setting modelsShouldPreventSilentlyDiscardingAttributes to true by default

taylorotwell:
I do not think this should be enabled in production.

PR #5994, opened 2022-09-27
Allows setting local and public root path

madalinignisca:
This allows on any new started project to be able to provide out of the installation alternative local disks paths, both private and public.

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -32,13 +32,13 @@ return [
 
         'local' => [
             'driver' => 'local',
-            'root' => storage_path('app'),
+            'root' => env('FILESYSTEM_DISK_LOCAL_ROOT', storage_path('app')),
             'throw' => false,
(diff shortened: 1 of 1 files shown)


driesvints:

madalinignisca:
May I ask the rushed answer for not allowing to have some common configuration overridable by environment?

My idea for having almost all configuration using environment variables would allow a better standardization with collaboration between developers and people responsible on deploying the applications.

PS: Isn't this a community oriented project or the rejections come as for this repo is only for the good needs of Laravel LLC?

erayaydin:
@madalinignisca filesystems and its options are not always configure with environments for projects. Your project maybe need that but IMO its not suitable for all the projects. I suggest you ask features in Larachat, IRC, discord or maillist before implementing. `Community-oriented project != accept everything`

PR #5995, opened 2022-09-27
Allow overriding LOG path and LOG daily keep limit

madalinignisca:
This would allow easy to change the LOG file storage path and change the limit of daily config.


driesvints:

PR #5997, opened 2022-09-30
[9.x] Handle precognition requests by default 

driesvints:
> Requires a framework release for https://github.com/laravel/framework/pull/44339

This PR adds the `HandlePrecognitiveRequests` middleware to the default middleware stack. The goal of this is to make it even more frictionless for people to get started with implementing precognition requests.

Precognition requests are invoked with a `Precognition: true` header meaning apps that don't explicitly pass this aren't affected. The only thing that's left to do for users is to implement the `isPrecognitive` checks in their apps. Adding the middleware on a route-per-route basis isn't needed anymore. 

People can still turn off this global middleware and implement on a route-per-route basis but I could not see a problem by adding this by default.

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -21,7 +21,7 @@ class Kernel extends HttpKernel
         \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
         \App\Http\Middleware\TrimStrings::class,
         \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
-        \Illuminate\Foundation\Http\Middleware\HandlePrecognitiveRequests::class,
+        // \Illuminate\Foundation\Http\Middleware\HandlePrecognitiveRequests::class,
     ];
 
     /**
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -21,6 +21,7 @@ class Kernel extends HttpKernel
         \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
         \App\Http\Middleware\TrimStrings::class,
         \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
+        \Illuminate\Foundation\Http\Middleware\HandlePrecognitiveRequests::class,
     ];
 
     /**


habibalkhabbaz:
Hello @driesvints 

By adding `HandlePrecognitiveRequests` to the default middleware stack, the middleware will try to add the `Vary` header on the `web` routes too, which means if we have a response with a binary file such as the following:

```php
return response()->file(Storage::path($payment->attachment));
```

... the middleware will try to to add the `Vary` header here too but it will throw an exception because the response here is an instance of `\Symfony\Component\HttpFoundation\BinaryFileResponse` and not `\Illuminate\Http\Response` that is expected by `appendVaryHeader` function

> Call to undefined method Symfony\Component\HttpFoundation\BinaryFileResponse::header()

### Reference

`appendVaryHeader` function

```php
    /**
     * Append the appropriate "Vary" header to the given response.
     *
     * @param  \Illuminate\Http\Response  $response
     * @return \Illuminate\Http\Response $response
     */
    protected function appendVaryHeader($request, $response)
    {
        return $response->header('Vary', implode(', ', array_filter([
            $response->headers->get('Vary'),
            'Precognition',
        ])));
    }
```

`response()->file` function

```php
    /**
     * Return the raw contents of a binary file.
     *
     * @param  \SplFileInfo|string  $file
     * @param  array  $headers
     * @return \Symfony\Component\HttpFoundation\BinaryFileResponse
     */
    public function file($file, array $headers = []);
```

We can fix this by checking the instance of `response` parameter in the `appendVaryHeader` function and just return the response without adding the header because it's not required here.
If you'd like me to submit a PR, I will be happy to do it.

vini123:
虚幻之物对应着冥冥之路！：《羽》

timacdonald:
In principle I'm okay with this being applied by default, however some things we should consider:

- This does leak implementation on all routes, even when the feature is not used i.e. someone could look at response headers and see the `Vary` header contains `Precognition` and they now know this is a Laravel application powered backend.
- It opens the application to be able to get into weird states that might not be expected.

Take this convoluted scenario:

```php
class MyMiddleware
{
    public function handle($request, $response)
    {
    }
}
```

timacdonald:
@habibalkhabbaz I've already addressed the issue you raised: https://github.com/laravel/framework/pull/44424

habibalkhabbaz:
> @habibalkhabbaz I've already addressed the issue you raised: https://github.com/laravel/framework/pull/44424

Thanks!
By the way, you made something awesome to Laravel 💯

driesvints:
@timacdonald what if we comment out the middleware by default? Like I've done now.

PR #6001, opened 2022-10-04
Bump axios to v1.0

balu-lt:


taylorotwell:
Need to know if all of our libraries work with this?

PR #6003, opened 2022-10-05
[8.x] Comment unused Gate class

muath-ye:
Comment `use Illuminate\Support\Facades\Gate;` class at line number 6.


alteredchimera:
I believe this has been attempted a few times before. Here is an explanation from Taylor with the reasoning; 

https://github.com/laravel/laravel/pull/3832

antonkomarev:
Previous MRs are removing this line, this one in comments out it.

PR #6005, opened 2022-10-06
add 'precognitive' key to `$routeMiddleware`

emargareten:
This will allow to use 'precognitive' in the route instead of the fully qualified class name (this is the [default](https://laravel.com/docs/9.x/middleware#assigning-middleware-to-routes) way of using route middleware)


emargareten:
done

taylorotwell:
Putting this on draft until we get the precognitive frontend helpers out and everything documented.

PR #6009, opened 2022-10-11
Shorten pusher host config

buihanh2304:


taylorotwell:
I still don't understand why it can't just be:

```php
'host' => env('PUSHER_HOST', 'api-'.env('PUSHER_APP_CLUSTER', 'mt1').'.pusher.com'),
```

PR #6015, opened 2022-10-19
Use bcrypt function for generate hash password

devman1917:
because of two reasons:
1 - No need to have a hash inside the code.
2 - The name of the hash algorithm is there for information


michaeldyrynda:
The hardcoded hash is in the factory because it is significantly quicker than generating a random hash for each created user, particularly if you are creating lots of users in a a large test suite.

devman1917:
> The hardcoded hash is in the factory because it is significantly quicker than generating a random hash for each created user, particularly if you are creating lots of users in a a large test suite.

Absolutely correct, thank you
image

PR #6017, opened 2022-10-22
[9.x] Use model key for rate limiting

mnabialek:
Use model key instead of hardcoding it to id


taylorotwell:
Change it in your own application if you wish.

mnabialek:
Yeah, obviously I've changed it in my app but it's pretty easy to miss it and I believe you won't see any error unless you enabled strict type for model.

PR #6018, opened 2022-10-23
[9.x] inline doc language simplify

maherelgamil:


PerryvanderMeer:
This breaks the comment structure.

PR #6019, opened 2022-10-25
[9.x]  Add Cloudflare R2 storage configuration to filesystems.

turbo124:
This PR adds support for Cloudflare R2 storage.

Whilst R2 is billed as a S3 API compatible, the full S3 API is not supported. In particular the x-amx-acl header is not respected, so any attempt to use the public visibility flag will cause file uploads to fail.

This PR adds a base configuration where the visibility is set to private.

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -56,6 +56,18 @@ return [
             'throw' => false,
         ],
 
+        'r2' => [
+            'driver' => 's3',
+            'key' => env('R2_ACCESS_KEY_ID'),
+            'secret' => env('R2_SECRET_ACCESS_KEY'),
+            'region' => env('R2_DEFAULT_REGION'),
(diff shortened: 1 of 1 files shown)


driesvints:
Feel free to add this to your own app 👍

gmgarrison:
CloudFlare is a huge services provider - I think having 'out of the box' support from Laravel would be great (and also decrease the unthinking and expensive reliance on S3 for object storage).

corsair:
I'd also like to see this added to Laravel.

Why be closed-minded and only offer one service when Cloudflare's R2 provides similar features as S3 but far more economically? I'd appreciate objective reasoning for your rejection, @driesvints.

GrahamCampbell:
It is already available out of the box. It's the s3 driver.

PR #6021, opened 2022-10-27
Changing .env to make Pusher work without editing the commented part in the bootstrap.js

cveldman:
This key VITE_PUSHER_HOST existing in the .env file will make it with Vite be available as an empty string in the bootstrap.js, therefore it will always fail on the expression in the bootstrap.js and default will set the wsHost propery for Laravel Echo to an empty string. By removing the key it will fallback to the given cluster with Pusher host again "ws-eu.pusher.com".

I think removing the key from .env.example is better than changing the comparison in the bootstrap.js, because its not a default key given by Pusher when creating a project for Laravel. But maybe checking for a empty string in the bootstrap.js is a fix as well if you guys want to keep this config item in the .env.example file.

it fails on the following expression in the bootstrap.js
`
wsHost: import.meta.env.VITE_PUSHER_HOST ?? ''
`


taylorotwell:
Can you be more clear about what is failing?

cveldman:
Yes, its in the following commented out part in the bootstrap.js of laravel/laravel.

```
window.Echo = new Echo({
    broadcaster: 'pusher',
    key: import.meta.env.VITE_PUSHER_APP_KEY,
    wsHost: import.meta.env.VITE_PUSHER_HOST ?? `ws-${import.meta.env.VITE_PUSHER_APP_CLUSTER}.pusher.com`,
    wsPort: import.meta.env.VITE_PUSHER_PORT ?? 80,
    wssPort: import.meta.env.VITE_PUSHER_PORT ?? 443,
    forceTLS: (import.meta.env.VITE_PUSHER_SCHEME ?? 'https') === 'https',
    enabledTransports: ['ws', 'wss'],
});
```

To set the wsHost property the JavaScript operator ?? will only return the right-hand sided operand when the left-hand side is null or undefined. But the VITE_PUSHER_HOST key just existing in the .env.example with Vite will make it available as import.meta.env.VITE_PUSHER_HOST = "", so not === null.

So the ?? will always pick the left-side and make { wsHost: "" }. and not "ws-eu.pusher.com" as default when no host is provided.

taylorotwell:
Can we change bootstrap.js instead of editing the env file?

cveldman:
Yes, I can change the bootstrap.js file if you like? Also checking for != '' in the bootstrap.js will solve it as wel!

cveldman:
Thanks for fixing the bug! I was still brainstorming about how the solution can become a little more elegant and came to the following solution. The big decider is the different hosts for the server-side and client-side library of Pusher.

If Pusher can provide an PUSHER_HOST and a VITE_PUSHER_HOST when people selected a different cluster than the default mt1, you can remove the PUSHER_APP_CLUSTER and VITE_PUSHER_APP_CLUSTER keys in the .env.example file that are only used by Pusher self.

This way you can default to Pusher and have purpose for the VITE_PUSHER_HOST key that is currently only used by Soketi. This because in pusher-js the default is ws-mt1.pusher.com and in de backend the default will be api-mt1.pusher.com. You can keep the VITE_PUSHER_HOST empty how it is now.

When people choose a different cluster within Pusher they only have to add a PUSHER_HOST and VITE_PUSHER_HOST value.

The bootstrap.js will become this, an little bit elegant in my opinion.
```
window.Echo = new Echo({
    broadcaster: 'pusher',
    key: import.meta.env.VITE_PUSHER_APP_KEY,
    wsHost: import.meta.env.VITE_PUSHER_HOST,
    wsPort: import.meta.env.VITE_PUSHER_PORT ?? 80,
    wssPort: import.meta.env.VITE_PUSHER_PORT ?? 443,
    forceTLS: (import.meta.env.VITE_PUSHER_SCHEME ?? 'https') === 'https',
    enabledTransports: ['ws', 'wss'],
});
```
If you use cluster mt1 of pusher the .env becomes 

```
PUSHER_APP_ID=
PUSHER_APP_KEY=
PUSHER_APP_SECRET=
PUSHER_HOST=
PUSHER_PORT=443
PUSHER_SCHEME=https

VITE_PUSHER_APP_KEY="${PUSHER_APP_KEY}"
VITE_PUSHER_HOST="${PUSHER_HOST}"
VITE_PUSHER_PORT="${PUSHER_PORT}"
VITE_PUSHER_SCHEME="${PUSHER_SCHEME}"
```

And if you use a different cluster of pusher it becomes

```
PUSHER_APP_ID=
PUSHER_APP_KEY=
PUSHER_APP_SECRET=
PUSHER_HOST=api-eu.pusher.com
PUSHER_PORT=443
PUSHER_SCHEME=https

VITE_PUSHER_APP_KEY="${PUSHER_APP_KEY}"
VITE_PUSHER_HOST=ws-eu.pusher.com
VITE_PUSHER_PORT="${PUSHER_PORT}"
VITE_PUSHER_SCHEME="${PUSHER_SCHEME}"
```

So the only change to the .env.example is the removal of the PUSHER_APP_CLUSTER and VITE_PUSHER_APP_CLUSTER keys, and a more elegant bootstrap.js. This is currently an idea, let me know if its interesting!

PR #6026, opened 2022-11-04
[9.x] Add `APP_TIMEZONE` to app.php

zaherg:
It would make update the app timezone easier without editing the config file.


simensen:
Indeed, timezone seems like something that could easily be different for the same application deployed to different environments. And, in general, I like to be able to do as much configuration as possible via environment and this is the only one I regularly need to change when I create a new Laravel app.

driesvints:
Right now, we have no plans for this sorry. Feel free to add this to your own app.

zaherg:
Thanks 🙏

PR #6028, opened 2022-11-08
Adds lowercase validation rule translation

timacdonald:
See: https://github.com/laravel/framework/pull/44883


nshiro:
@timacdonald
As for alphabetical order, just before `lt` looks better.

Thanks always.

PR #6030, opened 2022-11-14
[10.x] throw file-system exceptions by default

ankurk91:
Ref https://github.com/laravel/framework/pull/41308

Laravel 8 used to throw file-system excretions but
Laravel 9 changed this behaviour and made it configurable instead.

This PR brings back the Laravel 8 behaviour and throw exceptions by default.
I think most of Laravel 9 developers are living in dark and unknown about this behaviour.

Exceptions are good :)

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -33,7 +33,7 @@ return [
         'local' => [
             'driver' => 'local',
             'root' => storage_path('app'),
-            'throw' => false,
+            'throw' => true,
         ],
 
         'public' => [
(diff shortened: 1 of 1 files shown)


driesvints:
I'd personally vote against this breaking change as it'd involve a lot of changes to existing apps that use the filesystem.

ankurk91:
@driesvints 
It will not impact on existing apps, since their config file will still have this toggle as `false`.

The exceptions are still configurable, and can be disable to reduce the change need to upgrade to v10

driesvints:
@ankurk91 that's true.

taylorotwell:
We can think about this closer to release but there hasn't been a huge demand for it or anything.

ankurk91:
@taylorotwell 
Would you consider this change now?

PR #6034, opened 2022-11-17
Fix typo in config/session.php

dev:
'a' should be 'an' in this context  🙂


taylorotwell:
This messes up the cascade spacing so I'll have to personally take a look at it.

PR #5654, opened 2021-07-29
[8.x] Unified asset publishing

finagin:
I propose append and promote use additional tag for publishing js and css assets. And automatically publish them after update packages.
#### Related PRs:
https://github.com/laravel/horizon/pull/1045
https://github.com/laravel/telescope/pull/1105

PR #6035, opened 2022-11-18
Add `public/vendor` to gitignore

slavarazum:
According to https://github.com/laravel/laravel/pull/5904, we moved from committing compiled assets, to build them 👍 

In addition Laravel vendor assets publishing fires by default on composer `post-update-cmd` event https://github.com/laravel/laravel/pull/5654

Adding the `public/vendor` directory to .gitignore will be suitable for these changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /node_modules
 /public/build
 /public/hot
+/public/vendor
 /public/storage
 /storage/*.key
 /vendor


driesvints:
These assets are suppose to be committed.

slavarazum:
@driesvints Could you share a reason?

driesvints:
These assets are equivalent to the version that's locked in composer.lock so they need to be committed. Changing this would also require us to change the install docs of nova, telescope, horizon, etc

slavarazum:
From this point of view we can say that the assets in the `build` folder depend on the versions locked in package-lock.json so they need to be commited.
It looks contradictory that we don't commit the `public/build` folder now, but commit the `public/vendor` folder, but not commit `/vendor` folder.

As you have noticed these assets are equivalent to the version that's locked in composer.lock, in other words – just a copy from current vendor versions.
If we now required to launch `npm install && npm run build` to make client side ready to use it would be consistent to supplement this with the publication of the vendor assets.

If the question is on changing the docs, lets make a short conversation and I will do my best 🤝 🙂 

Why wait four or five years (as it was with the main bundle) when we can bring such a simple and consistent change already https://github.com/laravel/laravel/pull/4118 https://github.com/laravel/laravel/pull/4625 https://github.com/laravel/laravel/pull/5646

projct1:
> These assets are equivalent to the version that's locked in composer.lock so they need to be committed. Changing this would also require us to change the install docs of nova, telescope, horizon, etc

That is, after installing the project through `composer create-project` you should always do `git edd.`?
Is this considered the correct way?

PR #6036, opened 2022-11-18
Removed ?  after user() which was causing errors

protendai:
Changed line 49 
from
return Limit::perMinute(60)->by($request->user()?->id ?: $request->ip()); 
to
return Limit::perMinute(60)->by($request->user()->id ?: $request->ip());

The  ?  after user() was causing errors. Noticed it after creating a new project


driesvints:
Not sure how this would be causing errors. Feel free to adjust in your own app.

PR #6037, opened 2022-11-20
Update .gitignore

krishzzi:
devlopment time uses on phpstan.neon only


driesvints:

PR #6042, opened 2022-12-01
Beautify the Badge

SupianIDz:
README should also be beautified

x


driesvints:
Thanks but it's fine.

PR #6049, opened 2022-12-16
Adjusting code explanations through comments.

rafaelstz:
Use a comma before 'so' if it connects two independent clauses (unless they are closely connected and short).


PerryvanderMeer:
This breaks the comment structure.

taylorotwell:
I think it's fine.

PR #6061, opened 2022-12-29
[10.x] Updated git configuration to use LF line endings by default

WendellAdriel:
This change is regarding the discussion **[here](https://github.com/laravel/pint/issues/134)**


adevade:
I'm on Windows and have been doing this in my projects recently with no problems! 👍 Just reporting in my experience for reference.

PR #6064, opened 2023-01-02
[9.x] switch password reset email to a primary key

browner12:
Switching from a normal index here to a primary index behaves the same except for adding a `UNIQUE` constraint.

The `DatabaseTokenRepository` deletes existing records with an email first, before creating a new one, so this additional constraint will be okay. https://github.com/laravel/framework/blob/9.x/src/Illuminate/Auth/Passwords/DatabaseTokenRepository.php#L88

Something similar was tried in #5652, but that PR added an **additional** `id` field as the primary key, which is probably a little overkill since we can just use the existing `email` field.

I have confirmed this solves the problem when the database has `SQL_REQUIRE_PRIMARY_KEY` set to `TRUE`, as it is on DigitalOcean Managed Databases.  I would say this is a good change to the default Laravel skeleton, as it is recommended to have primary keys on all tables for multi-node systems.

### DIFF
--- a/database/migrations/2014_10_12_100000_create_password_resets_table.php
+++ b/database/migrations/2014_10_12_100000_create_password_resets_table.php
@@ -14,7 +14,7 @@ return new class extends Migration
     public function up()
     {
         Schema::create('password_resets', function (Blueprint $table) {
-            $table->string('email')->index();
+            $table->string('email')->primary();
             $table->string('token');
             $table->timestamp('created_at')->nullable();
         });


rcerljenko:
@browner12 hi...
what happens if someone tries to password reset more than once?

PerryvanderMeer:
@rcerljenko previous tokens will become invalid.

browner12:
> The `DatabaseTokenRepository` deletes existing records with an email first, before creating a new one, so this additional constraint will be okay. https://github.com/laravel/framework/blob/9.x/src/Illuminate/Auth/Passwords/DatabaseTokenRepository.php#L88

The way the `DatabaseTokenRepository` currently works is it checks for existing entries on the table with the requested email. If it finds any, it first deletes them before creating the new one.  

Technically, you previously *could* have made a duplicate entry via something like a DB GUI, since there was no `UNIQUE` constraint, but then this code would have deleted ALL matching records upon being invoked.

Now, with this change, you would not be able to manually make this duplicate entry.

rcerljenko:
@browner12 thanks! that was my concern indeed...  :)

browner12:
so I spoke only *slightly* too soon on this fixing the DigitalOcean Managed Databases issue.

The way the schema grammar works (at least for MySQL), if you are adding a primary key that is not defined by one of the "increments" data types, it actually creates two separate queries, the first to create the table with the relevant columns, and then a second to define the primary key. 

This leads to a bit of a race condition. Since we are not creating the primary key immediately in the `CREATE TABLE` query, MySQL thinks we don't have a primary key, and throws its error if `sql_require_primary_key` is true, even though the next query would remedy this.

This PR is still a **good** change, though, since we do want to ensure primary keys on all tables.

I'm going to look into a separate PR to try and define the PRIMARY KEY immediately in the CREATE TABLE query, which would then fully solve this issue.

PR #6069, opened 2023-01-09
[9.x] Updated Example Feature Test to use fluent syntax

WendellAdriel:
I think that the example feature test case could use a more fluent syntax and also use the assert helper methods for the HTTP status code instead of the `assertStatus` with the 200 which feels like a magic number.

### DIFF
--- a/tests/Feature/ExampleTest.php
+++ b/tests/Feature/ExampleTest.php
@@ -14,8 +14,7 @@ class ExampleTest extends TestCase
      */
     public function test_the_application_returns_a_successful_response()
     {
-        $response = $this->get('/');
-
-        $response->assertStatus(200);
+        $this->get('/')
+            ->assertOk();
     }
 }


tests/Feature/ExampleTest.php
```diff
@@ -14,8 +14,7 @@ class ExampleTest extends TestCase
      */
     public function test_the_application_returns_a_successful_response()
     {
-        $response = $this->get('/');
-
-        $response->assertStatus(200);
```
thasleemmji, line 19: ExampleTest is a basic test that any new comer looks into first.
Shouldn't that be kept more descriptive as it was before. Someone starting at the first now will have to go to **assertOk** method to see, ooh it checks 200 status!!

datamweb: You have to differentiate between a beginner who is interested in programming and a beginner who only looks at code.
The first person out of interest will definitely understand very quickly what this code means, besides, s/he will learn something new from Laravel.
The second person just looks at the code, neither interested in learning nor curious to learn.

**LGTM!**

WendellAdriel: I understand your concern @thasleemmji but I do agree with @datamweb here.
The current one is simpler, I understand that, but the change I made is also good for showing:
1. That you can do assertions using a fluent syntax
2. That Laravel provides methods like the `assertOk` that IMO is much better than using `assertStatus` with the status codes as integers.

Now the final decision is not on us, let's wait for someone from the Core Team to review and check if this is going to be accepted or not.

thasleemmji: @datamweb @WendellAdriel Appreciate your explanations.
**Fluent syntax** quite justifiable.

sgphongtx:
I think it is better practice to keep your response under single control.

use Illuminate\Http\Response;

Response::HTTP_OK

antonkomarev:
I think it's good to have separate execute and assert blocks.

taylorotwell:
I think I will keep this as is for now, but thank you for the suggestion.

PR #6074, opened 2023-01-12
fix: correct typo on app bootstrap description

coquer:


driesvints:
This unfortunately breaks the indentation for these comment blocks.

PR #6075, opened 2023-01-12
[9.x] Fix initial database config for SQLite 🔧

michael-rubel:
## About

`.env:`
```php
DB_CONNECTION=sqlite
DB_DATABASE=test
```

After initialization of a new Laravel project with SQLite database, it suggests creating a database before running migrations:

---

But if we try to access this auto-created database, we get:

This PR fixes the configuration in `config/database.php` to ensure the SQLite database creates correctly in the `database` folder.

---

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -38,7 +38,7 @@ return [
         'sqlite' => [
             'driver' => 'sqlite',
             'url' => env('DATABASE_URL'),
-            'database' => env('DB_DATABASE', database_path('database.sqlite')),
+            'database' => database_path(env('DB_DATABASE', 'database').'.sqlite'),
             'prefix' => '',
             'foreign_key_constraints' => env('DB_FOREIGN_KEYS', true),
         ],


timacdonald:
@michael-rubel I'm unable to replicate the issue this is fixing. The following script creates a fresh application, creates a new sqlite database, and is able to connect without any further changes.

```sh
composer create-project laravel/laravel \
  && cd laravel \
  && sed -i "s/DB_CONNECTION=mysql/DB_CONNECTION=sqlite/" .env \
  && php artisan migrate \
  && php artisan tinker --execute="User::create(['name' => 'Tim', 'email' => 'tim@laravel.com', 'password' => 'secret']);"
```

michael-rubel:
@timacdonald sorry, I forgot to add a code snippet for database access.

Add this to your `welcome.blade.php` after running the script:
```blade
@php
App\Models\User::factory()->create();
@endphp
```

timacdonald:
Weird. It works in Tinker but doesn't work in a Blade view.

michael-rubel:
@timacdonald yeah, it looks like this issue exists for some time already, and lots of people are asking about that on [StackOverflow](https://stackoverflow.com/questions/43140711/database-database-database-sqlite-does-not-exist-database-works-from-artisan) but no one reported or PRed that.

timacdonald:
I've found the root problem is the resolution of `realpath` in the Sqlite connector.

Screen Shot 2023-01-13 at 9 49 48 am


If you open tinker and run `realpath('public')` it should resolve a path. If you put this in a view it will return `false`. Looking into a fix for this, which should resolve the issue across the board.

timacdonald:
I've just submitted a PR (https://github.com/laravel/framework/pull/45626) that does not fix the issue, but improves the error message for developers.

I don't believe we can move forward with your proposed fix as it means that the SQLite is expected to live in the database directory, or the path is relative to to database directory.

The path may be an absolute path to anywhere on the filesystem. It doesn't need to live in the project itself.

config/database.php
```diff
@@ -38,7 +38,7 @@
         'sqlite' => [
             'driver' => 'sqlite',
             'url' => env('DATABASE_URL'),
-            'database' => env('DB_DATABASE', database_path('database.sqlite')),
+            'database' => database_path(env('DB_DATABASE', 'database').'.sqlite'),
```
stasadev, line 41: @michael-rubel, you should improve current workflow without breaking the previous logic.

Something like this will cover everything:

```suggestion
            'database' => strpos(env('DB_DATABASE') ?? '', DIRECTORY_SEPARATOR) !== false
                ? env('DB_DATABASE')
                : (! empty(env('DB_DATABASE'))
                    ? database_path(env('DB_DATABASE', 'database').'.sqlite')
                    : env('DB_DATABASE')),
```

```env
# the current logic in 9.x branch
DB_DATABASE=/path/to/database.sqlite # /path/to/database.sqlite
DB_DATABASE=test # test
DB_DATABASE=null # null
DB_DATABASE= # ''

# after my changes
DB_DATABASE=/path/to/database.sqlite # /path/to/database.sqlite
DB_DATABASE=test # /path/to/project/database/test.sqlite
DB_DATABASE=null # null
DB_DATABASE= # ''
```

But looking at complexity, I don't believe that it can be merged. Maybe someone can come with a better solution.

michael-rubel:
@timacdonald Shouldn't the config be manually edited if the user wants the database outside of the project?
At least this fixes the onboarding process without touching anything in the existing codebase.

config/database.php
michael-rubel: @stasadev this is only a starter template? Existing projects shouldn't be affected by this change.

stasadev: @michael-rubel I agree that it can be useful for beginners, but it's a breaking change.
Imagine when it is accepted, it will completely ruin every online guide that you can find on SQLite in Laravel.

taylorotwell:
Tim's PR has been merged to provide a better error message. For the record when I use SQLite I just delete DB_DATABASE from my env file entirely.

PR #6080, opened 2023-01-24
function required for migration adjustments after every installation

canerksk:


driesvints:

PR #6085, opened 2023-01-29
Made default queue for database and beanstalkd connections configurable via env

Zakini:
This PR allows for the default queue for the database and beanstalkd queue connections to be configured using environment variables, similar to how the Redis and SQS connections are set up.


taylorotwell:
Add this to your own application if you wish.

PR #6087, opened 2023-02-02
[9.x] adding json condition for RedirectIfAuthenticated

milwad-dev:
When add ```guest``` middleware into routes and add guard for example ```sanctum```, if send request as json the middleware redirect to ```/home``` url.
But for api is better to return json response.

Before:

After:


taylorotwell:
Feel free to add to your own app if you wish!

renatokira:
I made my own

image

renatokira:
My own
image

PR #6089, opened 2023-02-06
[10.x] sets ASSET_URL to use `/` as the fallback value

timacdonald:
## Problem

If there is no `ASSET_URL` for a Laravel project, Laravel will fallback to appending the current request's host. It ignores the `APP_URL`.

```php
APP_URL=https://app-url.com
ASSET_URL=
```

If the site is served at `http://127.0.0.1:8008`...

```php
asset('style.css');

// http://127.0.0.1:8000/style.css 
```

This works great for the backend, but Vite will never know what the host of the request is, as all its knowledge must be known at build time. So Vite and Laravel can not truly share the asset URL across the build time / runtime divide _when `ASSET_URL` is empty_.

**Why is this a problem?**

Laravel outputs preload tags for all the required assets for the initial page load (hashes removed for readability).

```html
<link rel="modulepreload" href="http://127.0.0.1:8000/build/assets/Register.js">
<link rel="modulepreload" href="http://127.0.0.1:8000/build/assets/GuestLayout.js">
<link rel="modulepreload" href="http://127.0.0.1:8000/build/assets/TextInput.js">
<link rel="modulepreload" href="http://127.0.0.1:8000/build/assets/PrimaryButton.js">
<link rel="modulepreload" href="http://127.0.0.1:8000/build/assets/ApplicationLogo.js">
```

Internally we use the `asset` helper to generate these URLs. You can see they have appended the incoming request host.

When Vite boots and on subsequent page navigation Vite will also preload all required assets that _have not yet been seen_.

Vite does not know about `http://127.0.0.1:8000/build/assets/app.css`. Vite only knows about `/build/assets/app.css`. This results in duplicate preload tags populating the document on initial page load.

```html
<link rel="modulepreload" href="http://127.0.0.1:8000/build/assets/Register.js">
<link rel="modulepreload" href="http://127.0.0.1:8000/build/assets/GuestLayout.js">
<link rel="modulepreload" href="http://127.0.0.1:8000/build/assets/TextInput.js">
<link rel="modulepreload" href="http://127.0.0.1:8000/build/assets/PrimaryButton.js">
<link rel="modulepreload" href="http://127.0.0.1:8000/build/assets/ApplicationLogo.js">

<!-- ... -->

<link rel="modulepreload" as="script" href="/build/assets/Register.js">
<link rel="modulepreload" as="script" href="/build/assets/GuestLayout.js">
<link rel="modulepreload" as="script" href="/build/assets/TextInput.js">
<link rel="modulepreload" as="script" href="/build/assets/PrimaryButton.js">
<link rel="modulepreload" as="script" href="/build/assets/ApplicationLogo.js">
```

This isn't hugely problematic, as the browser knows that both of these URLs point to the same resource, i.e. the browser doesn't try to preload 2 sets of resources. The Vite preloads are a noop in the case that Laravel has already output its own preload tags.

## Solution

If we have the `ASSET_URL` fallback to a `/` (which was not possible in `9.x` but is in `10.x`) we get uniform output from both Laravel and Vite, and thus Vite does not output duplicate tags.

```html
<link rel="modulepreload" href="/build/assets/Register.js">
<link rel="modulepreload" href="/build/assets/GuestLayout.js">
<link rel="modulepreload" href="/build/assets/TextInput.js">
<link rel="modulepreload" href="/build/assets/PrimaryButton.js">
<link rel="modulepreload" href="/build/assets/ApplicationLogo.js">
```

We could look at changing this in the framework, but I figure doing it in the skeleton allows for ultimate control for developers if they want to restore the previous functionality of using the incoming request's host.

## Not just a Vite problem

Although this is motivated by a Vite related problem, this will benefit other and future tools. 

I believe that a known static value for `ASSET_URL` makes total sense as a default. The asset helper is meant to link to assets, and very often those assets are built by a 3rd party. This allows a more expected shared value across the runtime / build time divide.

I also don't believe this should be problematic in anyway, because the request host and `/` should be equivalent unless you are doing something really funky.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -56,7 +56,7 @@ return [
 
     'url' => env('APP_URL', 'http://localhost'),
 
-    'asset_url' => env('ASSET_URL'),
+    'asset_url' => env('ASSET_URL', '/'),
 
     /*
     |--------------------------------------------------------------------------


jimgwhit:
Not everyone uses vite, please remember this is a php framework.  I use custom css.

ankurk91:
@timacdonald 

After making the change to my Laravel 9 project, I see double slash in asset url

```php
// configs
'asset_url' => env('ASSET_URL','/'),
```

```blade
// blade template
<img src="{{asset('assets/logos/light-rectangle.svg')}}" width="200" decoding="async">
```

Result in web browser
```html
<img src="//assets/logos/light-rectangle.svg" width="200" decoding="async">
```

The image failed to load.

Can we strip the extra slash from the framework core?

timacdonald:
@ankurk91 setting a `/` is not supported in Laravel 9.x. It is only supported in Laravel 10+

ankurk91:
Okay, thanks for info

dwightwatson:
I've run into a slight issue with this which required me to revert this change for my app and thought it was worth noting here for anyone else affected.

I use the `asset` helper in my emails to link to a logo image. With this change applied the relative URL doesn't work and the image asset is broken.

(I'm not using a CDN, rather letting Cloudflare handle caching static assets for me on my root URL).

So I suppose the easiest alternatives here are to set ASSET_URL in your app to an empty string or your APP_URL.

ankurk91:
I am facing the issue in emails too.
Images in email are not being loaded due to broken URL. :(

timacdonald:
Hey folks, I've reverted this one. Would much rather have emails work "out of the box" than dedupe some noop HTML tags.

timacdonald:
Appreciate the feedback.

mathieutu:
This one also broke the inertia version handling, as it uses `config('asset_url')` if present. 
(I installed the project just when this was merged, and investigated why the cache busting was not working)

PR #6090, opened 2023-02-06
[10.x] sets package.json type to module

timacdonald:
Full description: https://github.com/laravel/vite-plugin/pull/189

See:

- https://github.com/laravel/vite-plugin/pull/189
- https://github.com/laravel/breeze/pull/246
- https://github.com/laravel/jetstream/pull/1242

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,12 +1,13 @@
 {
     "private": true,
+    "type": "module",
     "scripts": {
         "dev": "vite",
         "build": "vite build"
     },
     "devDependencies": {
         "axios": "^1.1.2",
-        "laravel-vite-plugin": "^0.7.2",
+        "laravel-vite-plugin": "^0.7.5",
         "vite": "^4.0.0"
     }
 }


driesvints (approved):

fabio-ivona:
@timacdonald after this PR, my [vite livewire plugin](https://github.com/defstudio/vite-livewire-plugin) fails to load with this error:

> failed to load config from /var/www/vite.config.js
> error when starting dev server:
> TypeError: livewire is not a function
>     at file:///var/www/vite.config.js.timestamp-1685515480958-63f946afe0d4e.mjs:10:5
>     at ModuleJob.run (node:internal/modules/esm/module_job:194:25)

any hint on how can I solve this?

timacdonald:
Looks like you are building a [CJS](https://github.com/defstudio/vite-livewire-plugin/blob/master/tsconfig.json#LL5C11-L5C11) package.

You will want to build both a CJS and MJS version to support both systems.

fabio-ivona:
I guess this is the way, was just hoping for a way to support both

thanks @timacdonald for the hint!

timacdonald:
@fabio-ivona you should be able to support both if you build both of those types 👍 that is what we are doing in Laravel's Vite plugin

fabio-ivona:
yep, I'm looking at your code, I think I can do the same, thanks @timacdonald!

PR #6091, opened 2023-02-06
[10.x] Refine skeleton

taylorotwell:
This PR refines the skeleton a bit by renaming the `bootstrap` directory to `.laravel`, which indicates that the directory is for Laravel internal files and we may use it for other internal purposes in the future.

This also removes the `lang` directory by default in light of the new `lang:publish` command in Laravel 10.x and new language line fallback behavior that makes the `lang` directory optional. I think that shipping "internationalization" in the skeleton for every single application feels a bit heavy-handed and this is part of an overall effort to slim down the application the skeleton to be more light-weight in general.


ankurk91:
What about moving the `bootstrap/cache/*.php` files to `storage/framework/` ?

After this change, there will be not need to track file permission for `bootstrap/cache` folder.

quanghuybest2k2:
### I don't like this, it's inconvenient, please stop

datlechin:
> ### I don't like this, it's inconvenient. Please stop!!!

Just don't use Laravel

dinhquochan:
What about `$app->bootstrapPath()`? Is it rename to `$app->laravelPath()` or `$app->internalPath()`?

xuandung38:
I'd like to provide a feedback about the renaming of the `bootstrap` directory to `.laravel`. I disagree with this decision as it adds a personal touch and could cause confusion for users in the future. Additionally, using a directory name starting with a dot also creates inconvenience for developers on **MacOS as the directory will be hidden.**

I hope that a better structured decision with a clearer purpose can be made and minimize any difficulties for developers. Thank you for considering my opinion.

dinhquochan:
> Additionally, using a directory name starting with a dot also creates inconvenience for developers on MacOS as the directory will be hidden.

Also Linux/Unix users, because dot files/folders in the File Manager or `ls` command is hidden by default.

driesvints:
> What about moving the `bootstrap/cache/*.php` files to `storage/framework/` ?
> 
> After this change, there will be not need to track file permission for `bootstrap/cache` folder.

So the reason you don't want to do this is because the storage directory is meant to be persistent across deploys while the `bootstrap/cache` directory is meant to be cleared out when deploying a new build. They can't live together.

mbabker:
> > Additionally, using a directory name starting with a dot also creates inconvenience for developers on MacOS as the directory will be hidden.
> 
> Also Linux/Unix users, because dot files/folders in the File Manager or `ls` command is hidden by default.

As much as I personally agree with this and don't think there's a true benefit in renaming `bootstrap` to `.laravel`, that ship has already sailed thanks to the large number of dot files and folders that already exist in an application (the skeleton has `.github`, `.editorconfig`, and `.styleci.yml` (among other files)), and depending on your workflows, you may also add files such as `.php-cs-fixer.php`, `.npmrc`, or `.eslint*`.  And of course, there's the `.env` file.

You can argue that aside from `.env`, none of that stuff is part of your application so it doesn't have the same semantics as the app loading resources from a `.laravel` directory, but for a large number of PHP and Laravel applications, it's already a pretty common practice to have several dot files and folders in their project.  So the only thing this rename really changes is the default visibility of the bootstrap directory in a Linux environment.

driesvints:
@dododedodonl existing `bootstrap` directories are still supported. You don't "have" to change to the dots directory.

dododedodonl:
I get it has nothing to do with existing applications, and that changing it is possible.

What I tried to highlight is when one is maintaining several Laravel applications and creating new ones, it is in my view most easy when setups are as similar as possible. I personally try to migrate existing applications to new defaults for this reason.

Wouldn't this change be a nice opportunity to also move to a situation where `storage/*` is the only directory in a default Laravel application that has to be writable?

andreapollastri:
For me this solution is OK! Maybe you could also split /bootstrap into /.laravel and /cache directories or into /laravel.php (app.php) file and /cache directory.

matsava:
I don't really mind the change but what does this solve exactly?

taylorotwell:
To further explain my thinking about `bootstrap` / `.laravel` ... the `bootstrap` directory in general is 100% irrelevant to the developer. There is no reason to go in there or modify that code for any reason. So, I don't like that it is front-and-center part of your application. It's Laravel plumbing code. Contrast this to `storage` where you actually do have reasons to go in there to see what files are stored, or read your logs, or check your cache.

Therefore, I want to rename `bootstrap` to `.laravel` to give better indication this is a folder that belongs to the framework and you generally shouldn't mess with it.

alaminfirdows:
> To further explain my thinking about `bootstrap` / `.laravel` ... the `bootstrap` directory in general is 100% irrelevant to the developer. There is no reason to go in there or modify that code for any reason. So, I don't like that it is front-and-center part of your application. It's Laravel plumbing code. Contrast this to `storage` where you actually do have reasons to go in there to see what files are stored, or read your logs, or check your cache.
> 
> Therefore, I want to rename `bootstrap` to `.laravel` to give better indication this is a folder that belongs to the framework and you generally shouldn't mess with it.

I completely agree with you. Renaming the folder to .laravel sounds like a great idea to me and I think it will be helpful in indicating the purpose of the folder.

kocoten1992:
Hi, in the past Jeffrey Way recommend a custom `helpers.php` (for custom functions) in bootstrap directory and load via composer.json
```php
"files": [
    "bootstrap/helpers.php"
]
```

If not messing with with `.laravel` dir now, should that file just be put in `app` ?

vlakoff:
I'm afraid many problems would happen because it is an hidden directory. And with both names, there's equally one folder to "manage". Thus, I think the troubles because the directory is hidden outweight the (minor) annoyance of seeing a rarely-interacted directory.

taylorotwell:
I don't understand this talk about it being a hidden directory. Are we professional software developers or are we not? Are we able to understand these basic computing concepts or not? Multiple files in this project are already "hidden".

inmanturbo:
> I'm afraid many problems would happen because it is an hidden directory. And with both names, there's equally one folder to "manage". Thus, I think the troubles because the directory is hidden outweight the (minor) annoyance of seeing a rarely-interacted directory.
> 
> Also, all hidden files/directories are metadata basically. Some of them are of course important, but if they were all deleted the application would still be working. The proposed renaming would change this pattern, as a crucial part of the application (though not user-interacted) would now be contained in such a directory.

Good point! many natives apps though that create and manage hidden files avoid errors by recreating them at runtime if are needed and they were accidentally deleted. This is easily done for the cache paths, but the presence of  `app.php` (or `laravel.php` as it may become)  makes it a literal `bootstrap` and a first class inhabitant. Picking yourself up by the bootstraps  is kind hard without the bootstraps.

It's a really cool idea though to hide away internals into a hidden, branded folder. What if the framework held onto a stub for  `app.php`,  which  `artisan.php` and `index.php` could copy into place in the event that it was missing? And some warnings and notes in the docs for those hacking on internals of course.

vlakoff:
Are there use cases to edit `bootstrap/app.php`?

If not, a possibility might be to move the code of `bootstrap/app.php` somewhere into the framework, so the `bootstrap` dir would contain only the `cache` dir, which contains regenerable data, then we could rename this `bootstrap/cache` dir to `.laravel`.

inmanturbo:
> Are there use cases to edit `bootstrap/app.php`?
> 
> If not, a possibility might be to move the code of `bootstrap/app.php` somewhere into the framework, so the `bootstrap` dir would contain only the `cache` dir, which contains regenerable data, then we could rename this `bootstrap/cache` dir to `.laravel`.

There are use cases yes, though perhaps not technically supported. Most common is probably renaming, moving or adding kernels. Sometimes in large projects to restructure the skeleton to fit some DDD pattern.

moamahfouz:
I remember my first days using the framework I never show this directory's contents and I wondered why is it here if I am not going to use it. I think this is a good refactor. 👍

PR #6095, opened 2023-02-12
[10.x] Remove lodash

emargareten:
I know this has been attempted many times in the past, but I think the time has come (especially now that we have first party starter kits)

Previous PRs:
https://github.com/laravel/laravel/pull/5466
https://github.com/laravel/laravel/pull/5545
https://github.com/laravel/laravel/pull/3925

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,6 @@
     "devDependencies": {
         "axios": "^1.1.2",
         "laravel-vite-plugin": "^0.7.2",
-        "lodash": "^4.17.19",
         "postcss": "^8.1.14",
         "vite": "^4.0.0"
     }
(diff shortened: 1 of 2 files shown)


taylorotwell:
Do any of our starter kits depend on this?

ankurk91:
I was going to send a similar PR. 
I agree that it should be removed or at-least should not be imported this way.

There are better ways to import such modules for better tree shaking, for example

```diff
- import * as _ from "lodash";
+ import get from 'lodash/get';
```

emargareten:
> Do any of our starter kits depend on this?

@taylorotwell I just double checked and didn't find any usage of lodash

PR #6098, opened 2023-02-13
Use Hash facade to hash password in default User factory

zobay:
## Description
- Use Hash facade to hash password in default User factory


driesvints:
Please see many previous attempts.

PR #6105, opened 2023-02-14
[10.x] Add PHPUnit result cache to gitignore

itxshakil:


driesvints (approved):
I guess people are still using PHPUnit v9 on Laravel v10 so we should maybe add this yeah.

PR #6106, opened 2023-02-15
Allow php-http/discovery as a composer plugin

nicolas-grekas:
[php-http/discovery](https://packagist.org/packages/php-http/discovery) is widely used by SDK authors to provide auto-discovery of PSR-17/18 implementations. It's widely used in all PHP communities (Laravel, Symfony, etc.)

Since v1.15, it became a composer plugin, see https://github.com/php-http/discovery/pull/208. Adding this line improves DX by removing one needless prompt by Composer when the package is (transitively) required when installing a dependency. This takes a similar approach to #5959.

You might want to check the README of https://github.com/FriendsOfPHP/well-known-implementations for back ground on why this became a plugin.

For cross-ref, Symfony's skeleton also enabled this by default, see https://github.com/symfony/skeleton/pull/214

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,8 @@
         "preferred-install": "dist",
         "sort-packages": true,
         "allow-plugins": {
-            "pestphp/pest-plugin": true
+            "pestphp/pest-plugin": true,
+            "php-http/discovery": true
         }
     },
     "minimum-stability": "stable",


driesvints (approved):

patricksebastien:
When installing sail I get this:
```
In PluginManager.php line 744:                                                           
  php-http/discovery contains a Composer plugin which is blocked by your allo  
  w-plugins config. You may add it to the list if you consider it safe.        
  You can run "composer config --no-plugins allow-plugins.php-http/discovery   
  [true|false]" to enable it (true) or disable it explicitly and suppress thi  
  s exception (false)                                                          
  See https://getcomposer.org/allow-plugins
```

I am installing sail via docker:
```
docker run --rm     -v $(pwd):/opt     -w /opt laravelsail/php81-composer:latest     bash -c "composer require --dev laravel/sail && composer install && php artisan sail:install --with=mysql,mailhog,redis"
```

nicolas-grekas:
The patch in this PR solves this issue for new projects.

For existing projects, you need to apply it manually or run:
`composer config --no-plugins allow-plugins.php-http/discovery true`

driesvints:
I just released this so should be fixed now for new projects.

PR #6114, opened 2023-02-17
Missing cluster option in broadcasting.php

erstevn:
The documentation also need to be updated because cluster option in pusher/pusher-js#655 is mandatory.

So even if you use another websocket client like ably, you still need to define the "cluster" option as a string (you can have empty string) in the bootstrap.js


taylorotwell:
Did you actually have a problem in a real app? This value you are updating here is not sent to pusher-js. We already send the cluster to pusher-js.

PR #6115, opened 2023-02-17
Revert "[10.x] add ses-v2 mailer in config"

taylorotwell:
Reverts laravel/laravel#6112


ankurk91:
Any reason behind this?

Or atleast we can document this, so that users can switch to new version.

PR #6116, opened 2023-02-17
fix(docblock): fix destructured method argument type

shehi:
`...$guards` isn't `string`, but rather an array of strings.


shehi:
I see CS tool is complaining, but honestly I don't understand about what. Doesn't seem to be an alignment issue. Why would it require 2x spaces?!

driesvints:
The current type hint is correct as this method unpacks the arguments passed down to it.

shehi:
As you wish, but FYI: this isn't how many IDEs interpret these arguments, and many open source libs accept that as a norm.

PR #6118, opened 2023-02-18
[10.x] improving the performance and maintainability of the codebase

Nejcc:
reduces the likelihood of errors related to duplicate declarations and avoids unnecessary file inclusions


driesvints:
This isn't needed, thanks.

GrahamCampbell:
This change is incorrect, FYI. It is very important for correctness that those could be required more than once.

PR #6125, opened 2023-02-22
[10.x] Add api prefix in name route

gustavosobrinho01:
When downloading a Laravel project, our API routes do not come with the api name.

This means that we have to change the RouteServiceProvider file to meet this requirement.

I see it as necessary and prudent for api routes to have the api prefix. in their names.

Ex.: execute `php artisan route:list` command and see:

Before: 
Captura de Tela 2023-02-22 às 19 23 52

After:
Captura de Tela 2023-02-22 às 19 24 06

### DIFF
--- a/app/Providers/RouteServiceProvider.php
+++ b/app/Providers/RouteServiceProvider.php
@@ -29,6 +29,7 @@ class RouteServiceProvider extends ServiceProvider
         $this->routes(function () {
             Route::middleware('api')
                 ->prefix('api')
+                ->name('api.')
                 ->group(base_path('routes/api.php'));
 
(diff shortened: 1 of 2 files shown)


rojtjo:
Personally I prefer seeing the full name in the routes file, it just reduces the cognitive load for me.

I've used the route name prefixes in the past on a few bigger projects and when working on those projects again after a certain time it takes a lot of effort to remember all those details without reading through all my service providers that register routes.

gustavosobrinho01:
@rojtjo I understand your point of view, but the idea is to segregate only the api routes with the web routes.

You will continue to see the full name of the route.

PR #6130, opened 2023-02-26
[10.x] Specify facility in the syslog driver config

nicolus:
This is not mentioned in the documentation so there was no way to know how to specify a facility like `local0` unless you go digging in `\Illuminate\Log\Logmanager`'s source code, which IMHO should not be necessary to figure out what options are available.

Another solution would be to add it to the documentation, but I believe it's simpler to just have it in the default config file.

### DIFF
--- a/config/logging.php
+++ b/config/logging.php
@@ -102,6 +102,7 @@ return [
         'syslog' => [
             'driver' => 'syslog',
             'level' => env('LOG_LEVEL', 'debug'),
+            'facility' => LOG_USER,
         ],
 
         'errorlog' => [


ankurk91:
`LOG_USER` is a php native constant

nicolus:
Yes, thanks for clarifying. It's a native PHP constant that represents the "user" facility. It's actually what it defaults to in the [createSyslogDriver](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Log/LogManager.php#L345) method : 
```php
protected function createSyslogDriver(array $config)
{
    return new Monolog($this->parseChannel($config), [
        $this->prepareHandler(new SyslogHandler(
            Str::snake($this->app['config']['app.name'], '-'),
            $config['facility'] ?? LOG_USER, $this->level($config)
        ), $config),
    ]);
}
```

So this commit doesn't change anything behavior wise, it just make the configuration more visible, which would have saved me 20 minutes of searching for how to change it.

PR #6140, opened 2023-03-09
feat: add phpcs workflow

Abdel-Monaam-Aouini:


GrahamCampbell:
Laravel already uses StyleCI. phpcs doesn't actually have a proper Laravel preset.

PR #6143, opened 2023-03-14
Allow updating session cookie Same-Site attributes

aran112000:
Currently, the majority of `Session` options can be configured via environment variables, but this doesn't extend to the `Same-Site` attribute which can be an important security tool to tighten.

Options such as `http_only` aren't changed as allowing easier modification of these values would allow a less-secure posture more easily which we don't want to encourage.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -196,6 +196,6 @@ return [
     |
     */
 
-    'same_site' => 'lax',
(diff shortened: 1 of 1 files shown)


ankurk91:
Feel free to modify it in your project.

valorin:
I don't see the need for this change. `SameSite` isn't something most apps will need to change on a per-environment basis, and `Lax` is the best default for most apps.

PR #6146, opened 2023-03-20
[10.x] Remove -i in mail.php

W-TT-W:
Removed -i as the behaviours is unspecified with -bs, and only relevant wuth -t used in [8.x]

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -65,7 +65,7 @@ return [
 
         'sendmail' => [
             'transport' => 'sendmail',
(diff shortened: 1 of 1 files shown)


W-TT-W:
Based on information gathered on mails options from these links. 

https://refspecs.linuxbase.org/LSB_3.0.0/LSB-PDA/LSB-PDA/baselib-sendmail-1.html 
https://linux.die.net/man/1/mail

taylorotwell:
If you're not having an actual problem in your real app I don't want to change this right now.

PR #6152, opened 2023-04-11
Add directories for those using NetBeans

guymarshall:
Add the following NetBeans-specific directories to .gitignore:

build/
nbbuild/
dist/
nbdist/
.nb-gradle/


GrahamCampbell:
I don't think many people use this. I would not include this by default.

driesvints:

PR #6154, opened 2023-04-12
Remove commented namespace

vc-dhavaljoshi:


Cornatul:
So why was this closed ?

PR #6160, opened 2023-04-18
Fix laravel/framework constraints for Default Service Providers

Jubeki:
It seems like the constraints where not updated.

But the class `DefaultProviders` only exists since 10.8


claudiodekker:
For reference, here's a `prefer-lowest` composer install run that's now failing: https://github.com/claudiodekker/laravel-auth/actions/runs/4734577989/jobs/8403696328?pr=37

driesvints:
Released. Thanks!

PR #6162, opened 2023-04-23
Use Composer ^2.2

lloricode:
Base on laravel/framework#46096 it require composer 2.2 for a new project, 

regards


GrahamCampbell:
👎 This is not necessary. The framework already requires it.

lloricode:
> 👎 This is not necessary. The framework already requires it.

i see, thanks will close this PR now

PR #6164, opened 2023-04-25
Fix  419 | PAGE EXPIRED with ob_start(); on index.php

muath-ye:
I added ob_start(); to index resolve the issue of 419 | PAGE EXPIRED

The issue of expired page happened on multiple Laravel versions.
It is a common issue, see:

https://stackoverflow.com/a/69458738

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -1,5 +1,7 @@
 <?php
 
+ob_start();
+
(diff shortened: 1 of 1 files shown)


driesvints:
I very much doubt this is necessary sorry.

PR #6169, opened 2023-05-01
Use static middleware methods in api routes file

jnoordsij:
Uses syntax introduced in https://github.com/laravel/framework/pull/46362 within this skeleton, to promote the use of more modern syntax.


taylorotwell:
We will make this transition ourselves once we document things.

PR #6171, opened 2023-05-10
Add hashed cast to user password

emargareten:
This PR adds the new `hashed` cast to the user password field.

I left the user factory with the already hashed password since this cast first checks if the field needs rehash (this check doesn't affect the performance).


clementmas:
For reference: [ [10.x] Add 'hashed' cast #46947 ](https://github.com/laravel/framework/pull/46947)

PR #6172, opened 2023-05-11
Slimmer Application Skeleton

taylorotwell:


app/Http/Controllers/Controller.php
```diff
@@ -2,11 +2,9 @@
 
 namespace App\Http\Controllers;
 
-use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
-use Illuminate\Foundation\Validation\ValidatesRequests;
 use Illuminate\Routing\Controller as BaseController;
 
 class Controller extends BaseController
```
antonkomarev, line 7: @taylorotwell Do we need to have an extendable empty controller class in user space at all? The first  thing I'm always doing in fresh install - deleting it.
If you decide to keep it, I propose to make it abstract then.

config/sanctum.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+use Illuminate\Cookie\Middleware\EncryptCookies;
+use Illuminate\Foundation\Http\Middleware\ValidateCsrfToke;
```
dinhquochan, line 4:
```suggestion
use Illuminate\Foundation\Http\Middleware\ValidateCsrfToken;
```

app/Providers/EventServiceProvider.php
```diff
@@ -2,24 +2,10 @@
...
-use Illuminate\Support\Facades\Event;
 
 class EventServiceProvider extends ServiceProvider
 {
-    /**
-     * The event to listener mappings for the application.
-     *
-     * @var array<class-string, array<int, class-string>>
-     */
-    protected $listen = [
```
ankurk91, line 17: i would like to keep empty `$listen` array here

ankurk91:
Whole community is habitual to use `app/Exceptions/Handler.php` file from years now.
It is big move to remove this file.

I will have to create my own middlewares by extending from core in all of my projects now. It creates extra work.

lazer-hybiz:
Regarding loading the broadcast routes file automatically, I think it's better to have the filepath as a config rather than hardcoding it in the core.

app/Http/Controllers/Controller.php
driesvints: I agree. We can delete this. Keep the directory with a `.gitkeep` file?

app/Http/Kernel.php
```diff
@@ -60,7 +62,7 @@ class Kernel extends HttpKernel
         'can' => \Illuminate\Auth\Middleware\Authorize::class,
         'guest' => \App\Http\Middleware\RedirectIfAuthenticated::class,
```
driesvints, line 63: @taylorotwell you added this middleware in https://github.com/laravel/framework/pull/47040/files#diff-6a190f6e68c89dafd0a08692e63e48982a2bf570884ef58f525c2637ea516ae1R10 but forgot to update it here.

```suggestion
        'guest' => \Illuminate\Auth\Middleware\RedirectIfAuthenticated::class,
```

driesvints: Probably also delete the `App\Http\Middleware\RedirectIfAuthenticated` middleware class after this.

driesvints (approved):
This is the way.

innocenzi:
I love most of this, but honestly not a big fan of removing the exception handler. For Inertia apps, it's super common to override the `render` function entirely.

dillingham:
Sentry and other exception tooling document app/Exceptions/Handler.. may be worth looping them in

https://docs.sentry.io/platforms/php/guides/laravel/

crynobone:
@dillingham you can register it within `bootstrap/app.php`

app/Http/Kernel.php
```diff
@@ -14,27 +14,29 @@ class Kernel extends HttpKernel
...
      */
     protected $middlewareGroups = [
         'web' => [
-            \App\Http\Middleware\EncryptCookies::class,
+            \Illuminate\Cookie\Middleware\EncryptCookies::class,
             \Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse::class,
             \Illuminate\Session\Middleware\StartSession::class,
             \Illuminate\View\Middleware\ShareErrorsFromSession::class,
-            \App\Http\Middleware\VerifyCsrfToken::class,
+            \Illuminate\Foundation\Http\Middleware\ValidateCsrfToken::class,
```
milwad-dev, line 39: Sometimes some developers need to add some routes to VerifyCsrfToken middleware to not expect `csrf_token`!

milwad-dev (commented):
Great

app/Http/Kernel.php
rojtjo: A new static method `except` was added for this: https://github.com/laravel/framework/pull/47040/files#diff-f144212c54668b3942147079c0cd571035548d29ce57c4d18ab1e5b3fd754241

davidbirkin:
If some of these changes, such as the `app/Exceptions/Handler` are a breaking change to **__ANY__** app in the 10.x existing frameworks, it should be done on the 11.x release with Breaking Changes/Upgrade Guide. 

Just my 2 cents.

michaeldyrynda:
The app skeleton is not covered by semantic versioning and applies only to newly-created apps that use it. 

If you choose to retrospectively make the changes yourself, that's not a consideration for the framework.

milwad-dev:
@davidbirkin 
I don't think break any changes, because if you use Handler (Exception) when you upgrade to 11.x it's still exists!

app/Http/Controllers/Controller.php
lazer-hybiz: why do we need this directory? it should be created the first time a controller is created just like requests, jobs, mail and most of the other things.

driesvints:
@davidbirkin did you read the very first line of Taylor's post? 😅

davidbirkin:
> 

I guess I meant; and this theoretical only as I don't know internal timeline of events. 

Lets say these changes happened with 10.12. People upgrading from 10.10, I was wondering if these were affected or classed as `Existing applications`.

taylorotwell:
@innocenzi agree - the exception handling is the part I'm still thinking about the most.

app/Http/Controllers/Controller.php
driesvints: @lazer-hybiz for people just starting with PHP and Laravel it's nice to see the directory structure of where Laravel places things.

bootstrap/app.php
```diff
@@ -1,55 +1,20 @@
...
-| the calling script so we can separate the building of the instances
-| from the actual running of the application and sending responses.
+| the service container that can resolve all classes and components.
 |
 */
 
-return $app;
+return Application::create()
+    ->withBroadcasting()
+    ->withExceptionHandling(function ($handler) {
```
ralphschindler, line 18: If this is going to be an invitation to modify `$handler` perhaps type-hint what $handler is so that someone can start interacting with $handler by typing `$handler->...` and get code completion?

If it is not super common to modify the $handler, perhaps we don't need `->withExceptionHandling()` here?

jrd-lewis:
I wonder if this would be better suited for 11.x instead of 10.x, seeing the number of changes this would make in both the framework and skeleton?

config/sanctum.php
tylernathanreed, line 1: Why does this file ship with a vanilla Laravel install? I would expect the Sanctum service provider to publish a default implementation.

bootstrap/app.php
tylernathanreed: This approach feels awkward, as now all of the custom reporting and rendering logic is now biased towards being placed in a bootstrap file, rather than a dedicated location for error handling (which was the previous approach).

Could we add a `withExceptionHandler` method that accepts a class name and binds it to the handler contract? I feel like that would be an easier alternative for those looking to substitute the handler, rather than manipulating the default one.

lancepioch: @tylernathanreed You can just pass in a callable.

```php
return Application::create()
    ->withBroadcasting()
    ->withExceptionHandling(YourCustomExceptionHandler::handle(...));
```

driesvints:
@jrd-lewis there's no reason to. There's no breaking changes at all.

config/sanctum.php
driesvints: To better facility the install process of Jetstream.

krzysztofrewak:
I would love to see moving default config files to vendor and provide a simple way to extend only ones I need (maybe with handy php artisan config:publish database?). Something similar was working in Lumen, as far as I recall. That would simplify skeleton a lot!

bootstrap/app.php
lazer-hybiz: https://github.com/laravel/framework/pull/47040#discussion_r1193762425

config/sanctum.php
tylernathanreed: But isn't the point of this PR to trim the fat? Not everyone uses Jetstream. Why not just make Jetstream publish this file?

jrd-lewis:
> @jrd-lewis there's no reason to. There's no breaking changes at all.

Even without breaking changes, the scope of the two PRs feels bigger than a minor patch version should handle. It's also similar to when Vite replaced Mix in the middle of 9.x, which was poorly communicated and did end up breaking lots of projects at the time.

tontonsb:
> @jrd-lewis there's no reason to. There's no breaking changes at all.

It would break some tutorials and other instructions like the aforementioned Sentry docs.

database/migrations/0001_01_01_000000_create_users_table.php
zepfietje, line 1: Feels kinda weird to keep the `create_users_table` name.
Not too sure, but maybe something along the lines of `create_user_tables` could work.
Same for jobs.

database/seeders/DatabaseSeeder.php
```diff
@@ -12,11 +13,11 @@ class DatabaseSeeder extends Seeder
      */
     public function run(): void
     {
-        // \App\Models\User::factory(10)->create();
+        // User::factory(10)->create();
```
zepfietje, line 16: We might as well remove this line?

database/migrations/0001_01_01_000000_create_users_table.php
mabdullahsari: I'd even propose `create_authentication_tables`.

Riley19280:
With all these things removed, I would love to see the process to override these framework defaults documented, so that Laravel feels flexible and easy as it always has

chu121su12:
Any consideration in updating both `public/index.php` and `artisan`? Why will/will not update those?

app/Http/Kernel.php
asanikovich: why don't setup this by config files?

tylernathanreed:
I'm seeing that the new PR doesn't currently allow the community to weigh in on it.

I hope that the concerns that we all brought up are still addressed.

4m1n0s:
Hey,

what do you think about eliminate bootstrap directory

move app.php to app or config folder and add default app provider to withProviders configurations

and cache folder to storage framework cache directory

```php
<?php

use Illuminate\Foundation\Application;
use Illuminate\Foundation\Configuration\Exceptions;
use Illuminate\Foundation\Configuration\Middleware;

/*
| Create The Application
|
| The first thing we will do is create a new Laravel application instance
| which serves as the brain for all of the Laravel components. We will
| also use the application to configure core, foundational behavior.
|
*/

return Application::configure()
    ->withProviders([
        App\Providers\AppServiceProvider::class,
    ])
    ->withRouting(
        web: __DIR__.'/../routes/web.php',
        // api: __DIR__.'/../routes/api.php',
        commands: __DIR__.'/../routes/console.php',
        // channels: __DIR__.'/../routes/channels.php',
    )
    ->withMiddleware(function (Middleware $middleware) {
        //
    })
    ->withExceptions(function (Exceptions $exceptions) {
        //
    })->create();
```

joelharkes:
this is the followup PR: https://github.com/laravel/framework/pull/47309

would this mean the laravel skeleton would no longer be necessary or greatly reduced?

PR #6174, opened 2023-05-12
[10.x] Bring back pusher cluster config option

jesseleite:
We use Laravel Echo and Pusher in our first party Statamic Collaboration Addon https://github.com/statamic/collaboration, and notice that users are having problems connecting via Laravel Echo websocket in newer Laravel versions where `cluster` no longer exists in the default Pusher config.

We notice it was [added to the example in bootstrap.js](https://github.com/laravel/laravel/pull/6059) because pusher-js 8.0 now requires `cluster`. With this requirement, and with our Collaboration Addon reading from the users' Laravel config, we're thinking that it should be accessible through the default Laravel config as well, and so that the config can be cached, etc.

Your thoughts?

### DIFF
--- a/config/broadcasting.php
+++ b/config/broadcasting.php
@@ -36,6 +36,7 @@ return [
             'secret' => env('PUSHER_APP_SECRET'),
             'app_id' => env('PUSHER_APP_ID'),
             'options' => [
+                'cluster' => env('PUSHER_APP_CLUSTER'),
                 'host' => env('PUSHER_HOST') ?: 'api-'.env('PUSHER_APP_CLUSTER', 'mt1').'.pusher.com',
                 'port' => env('PUSHER_PORT', 443),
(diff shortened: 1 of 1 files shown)


jasonvarga:
To clarify:

- Cluster isn't required on the PHP side, if the host is provided. [Here's the Pusher code proving this](https://github.com/pusher/pusher-http-php/blob/4ace4873873b06c25cecb2dd6d9fdcbf2f20b640/src/Pusher.php#L98-L107).
   This is probably why it was removed in #5897
- But on the JS side it is required, which is why it was added back in #6059.

- We are passing the PHP broadcast config down to JS/Echo in our package, where having the `cluster` option is missing and causing some headaches.
- Having both `host` and a `cluster` in the options shouldn't hurt anything, since the `host` will take precedence on the PHP side.

jesseleite:
Thank you @taylorotwell! 🎉

PR #6180, opened 2023-05-20
Avoid mention ses-v2 in config/mail.php docblock

ziming:
I'm not sure why Taylor removed ses-v2 from mail.php (see link below) but since it is removed, I also remove via the docblock as I got a gotcha trying to specify ses-v2 for my MAIL_MAILER today

See:
https://github.com/laravel/laravel/pull/6115

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -28,7 +28,7 @@ return [
     | sending an e-mail. You will specify which one you are using for your
     | mailers below. You are free to add additional mailers as required.
     |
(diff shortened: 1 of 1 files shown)


nunomaduro (approved):

taylorotwell:
Hmm, maybe we should add ses-v2 back into config. I feel like there was some gotcha that caused me to not advertise ses-v2 much.

ziming:
this might be the 1st time I'm happy seeing my PR closed 😂

PR #6182, opened 2023-05-26
fix: change must-verify-email import in user model

frans-slabbekoorn:
When uncommenting the `MustVerifyEmail` and adding it as trait to the `User` model, results in the following error: `App\Models\User cannot use Illuminate\Contracts\Auth\MustVerifyEmail - it is not a trait`.

This PR changes the import to the `MustVerifyEmail` trait.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -2,7 +2,7 @@
 
 namespace App\Models;
 
-// use Illuminate\Contracts\Auth\MustVerifyEmail;
+// use Illuminate\Auth\MustVerifyEmail;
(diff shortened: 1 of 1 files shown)


rojtjo:
The trait is already included in the base user: https://github.com/laravel/framework/blob/10.x/src/Illuminate/Foundation/Auth/User.php

To enable the email verification all you need to do is implement the interface on your own `User` model, no need to add the trait again (unless you are no longer extending the base user).

frans-slabbekoorn:
Okay sorry for my misunderstanding, I will close this issue.

PR #6183, opened 2023-05-27
Punctuation fix in config/app.php on line 180

dyanakiev:
CleanShot 2023-05-27 at 11 26 28


driesvints:
This unfortunately breaks the indentation pattern of these lines.

dyanakiev:
@driesvints 
> This unfortunately breaks the indentation pattern of these lines.

Can you elaborate, so i can fix it?

PR #6185, opened 2023-05-29
Added comment in .gitignore

MoroAlberto:
In general, commenting the .gitignore file is a great practice to enhance the readability and understanding of the project. 
It helps to make it clearer why certain files or directories are excluded from version control with Git.


driesvints:

PR #6187, opened 2023-06-01
Fix HTML encoding in welcome template

szepeviktor:
Just that.

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -8,7 +8,7 @@
 
         <!-- Fonts -->
         <link rel="preconnect" href="https://fonts.bunny.net">
-        <link href="https://fonts.bunny.net/css?family=figtree:400,600&display=swap" rel="stylesheet" />
+        <link href="https://fonts.bunny.net/css?family=figtree:400,600&amp;display=swap" rel="stylesheet" />
 
         <!-- Styles -->
         <style>


driesvints:
This isn't needed.

szepeviktor:
Thank you for your comment @driesvints!

> Attribute values are a mixture of [text](https://html.spec.whatwg.org/multipage/syntax.html#syntax-text) and [character references](https://html.spec.whatwg.org/multipage/syntax.html#syntax-charref), except with the additional restriction that the text cannot contain an [ambiguous ampersand](https://html.spec.whatwg.org/multipage/syntax.html#syntax-ambiguous-ampersand).

[The standard](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2) has another stance.

husseinalhammad:
Based on the definition of an ambiguous ampersand:

> An ambiguous ampersand is a U+0026 AMPERSAND character (&) that is followed by one or more [ASCII alphanumerics](https://infra.spec.whatwg.org/#ascii-alphanumeric), followed by a U+003B SEMICOLON character (;), where these characters do not match any of the names given in the [named character references](https://html.spec.whatwg.org/multipage/named-characters.html#named-character-references) section.

This is an ambiguous ampersand:

```html
<link href="a&b;">
```

This is not an ambiguous ampersand:

```html
<link href="a&b">
```

szepeviktor:
Okay-okay. I got caught ⚡ 

I am zero risk taker, so I go the explicit way.

PR #6196, opened 2023-06-13
Update .gitignore to add .DS_Store

TheBlckbird:
It is really annoying when macOS creates `.DS_Store` files and you always have to edit the `.gitignore` to include them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ yarn-error.log
 /.fleet
 /.idea
 /.vscode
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Because it's an OS specific ignore. What stops me making a Graham OS that makes .graham files. Should I go PR that to every .gitignore file on github? No - my point is that OS specific files have nothing to do with individual projects. The ignore file for projects is... for projects.

PR #6200, opened 2023-06-15
Update Controller.php

JustSteveKing:
Making the default controller `abstract` as it is only ever extended not used directly

### DIFF
--- a/app/Http/Controllers/Controller.php
+++ b/app/Http/Controllers/Controller.php
@@ -6,7 +6,7 @@ use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
 use Illuminate\Foundation\Validation\ValidatesRequests;
 use Illuminate\Routing\Controller as BaseController;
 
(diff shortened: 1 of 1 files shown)


JustSteveKing:
@driesvints this is just a small paper-cut that seems to trigger me 😆

localheinz (commented):
👍

a-bashtannik:
I don't think this should be merged. Nothing blocks from using `\App\Http\Controllers\Controller` as a single controller with a couple of methods. There is an abstract parent behind this class, which defines some default mechanics, but this class can be used as fully working thing with the single responsibility.

danharrin:
@a-bashtannik 99% of people would never do that. If you want to, that's fine, there's nothing stopping you from removing the `abstract`. It's userland code.

taylorotwell:
I'm working on *a lot* of skeleton changes for 11.x - will lump this into that. 👍

PR #6202, opened 2023-06-17
Add seeder call example

paulredmond:
I had someone new to Laravel ask how to call other seeders after using `make:model --seed`. Though it's documented I thought it might help demonstrate how to do it.

No worries if you'd rather just leave this only in the docs and not here ❤️


taylorotwell:
Will just leave in docs for now.

PR #6203, opened 2023-06-19
Remove unused import

Jubeki:
A change/bugfix in PHP-CS-Fixer `3.18` will now remove this line. See https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/6909

For Laravel Pint I created a PR which upgrades to the new PHP-CS-Fixer version: See https://github.com/laravel/pint/pull/188

This line is responsible for a failing formatting Test in my package `jubeki/laravel-code-style`
See https://github.com/Jubeki/laravel-code-style/actions/runs/5306386155

Other options would be:
- Using a case sensitive `Event` somewhere in the file.
- To comment the line, instead of removing it.

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,7 +1,5 @@
 php:
   preset: laravel
-  disabled:
-    - no_unused_imports
   finder:
     not-name:
       - index.php
--- a/app/Providers/EventServiceProvider.php
+++ b/app/Providers/EventServiceProvider.php
@@ -5,7 +5,6 @@ namespace App\Providers;
 use Illuminate\Auth\Events\Registered;
 use Illuminate\Auth\Listeners\SendEmailVerificationNotification;
 use Illuminate\Foundation\Support\Providers\EventServiceProvider as ServiceProvider;
-use Illuminate\Support\Facades\Event;
 
 class EventServiceProvider extends ServiceProvider
 {


GrahamCampbell (changes requested):
This repo intentionally has unused imports. https://github.com/laravel/laravel/blob/10.x/.styleci.yml#L4

Jubeki:
Oh, I didn't see that. Thanks for pointing it out!

driesvints:
@Jubeki are you saying that with the next Pint update, Pint will fail on the default skeleton? If so then we still need to fix this.

Jubeki:
@driesvints Yeah this line will be removed with the laravel/pint update. I will reopen the PR and leave it to your decision.

driesvints (approved):

Jubeki:
@driesvints You can see here what will happen with PHP-CS-Fixer `3.18`: https://github.com/Jubeki/laravel-code-style/actions/runs/5306386155

(Rules are the same as pint, just with installing all dependencies instead of a pint binary)

GrahamCampbell:
In which case this PR should also update the styleci config.

Jubeki:
@GrahamCampbell Do you think these lines can also be removed?

```yaml
  finder:
    not-name:
      - index.php
      - server.php
``` 
  
Because `server.php` was removed from the skeleton and I don't see a reason for not formatting `index.php`
(I will create another PR for the configuration changes)

PR #6207, opened 2023-06-27
[10.x] Remove unused import from EventServiceProvider

zepfietje:
Latest version of Pint with the Laravel preset removes the `Event` facade import from the `EventServiceProvider` since it's unused.

This PR removes the import from the skeleton, since there's not much use in having the import when it gets removed on first run of Pint anyway.

### DIFF
--- a/app/Providers/EventServiceProvider.php
+++ b/app/Providers/EventServiceProvider.php
@@ -5,7 +5,6 @@ namespace App\Providers;
 use Illuminate\Auth\Events\Registered;
 use Illuminate\Auth\Listeners\SendEmailVerificationNotification;
 use Illuminate\Foundation\Support\Providers\EventServiceProvider as ServiceProvider;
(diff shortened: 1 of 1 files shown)


Jubeki:
This is there for convenience. See https://github.com/laravel/laravel/pull/6203

zepfietje:
Thanks for referencing your PR, @Jubeki.

The discussion is inconsistent though.
In other places with unused imports (e.g. the user model) the import is commented out: https://github.com/laravel/laravel/blob/3ac233abb21e29c94727c56c027067f3da9cbe32/app/Models/User.php#L5.

zepfietje:
> I think this will just be left in Laravel v10 skeleton for now. For v11 there are the skeleton slims planned and I think, there this EventServiceProvider will no longer exist. (Can still change)

Good point. 👍

PR #6211, opened 2023-06-30
Shorten relative paths in phpunit.xml

spawnia:
Shorten all relative paths in `phpunit.xml`, as already done with `bootstrap="vendor/autoload.php"`.


taylorotwell:
Will look at it for the 11.x skeleton I'm working on.

PR #6217, opened 2023-07-18
Remove `/public/build` from .gitignore

mtawil:
Hi,

This PR is for removing `/public/build` from the `.gitignore`. This change will prevent errors during deployment when compiled files are missing.

FYR:
- https://github.com/laravel/laravel/pull/3805
- https://github.com/laravel/laravel/commit/86b4b1b65659a7b0f4b3dad1e753aa1efa5a7642#commitcomment-83530235

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 /.phpunit.cache
 /node_modules
-/public/build
 /public/hot
 /public/storage
 /storage/*.key


GrahamCampbell (changes requested):
You are supposed to do the build during the deployment process, not commit the files. Of course, if you want to do that in your own app, you can, but I don't think we're gonna recommend that bad pattern.

mtawil:
> You are supposed to do the build during the deployment process, not commit the files. Of course, if you want to do that in your own app, you can, but I don't think we're gonna recommend that bad pattern because over time, it makes the repo massive.

Since the `node_modules` is already ignored, I have to install the packages every time I deploy, which doesn't make sense and slows down the deployment process.

GrahamCampbell:
That's a decision you are welcome to make on your own app, but for everyone else, we definitely want to make people run `npm ci` then build at deploy time. Other advantages are security related, making sure no bad files can be included in the compiled files, or possibly the developer has old dependencies in their node_modules.

PR #6218, opened 2023-07-18
Update composer.json

arttiger:
ext-pdo require because class PDO used in config/database.php

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,8 @@
         "guzzlehttp/guzzle": "^7.2",
         "laravel/framework": "^10.10",
         "laravel/sanctum": "^3.2",
-        "laravel/tinker": "^2.8"
+        "laravel/tinker": "^2.8",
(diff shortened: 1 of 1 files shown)


buismaarten:
It's optional

siarheipashkevich:
@buismaarten why it is optional, if it is used out of the box?

image

buismaarten:
@siarheipashkevich because the function `extension_loaded` is used in this situation, but the static analyzer does not recognize this.

The requirement `ext-pdo` is used in the composer.json of the Laravel database package.   That said, the skeleton does not need the `ext-pdo` requirement.

siarheipashkevich:
@buismaarten should I require this extension in my app, if I use MySQL?

buismaarten:
@siarheipashkevich Laravel supports MySQL out of the box.  As long as the php_mysql extension is installed and you have no errors, everything is fine.

PR #6220, opened 2023-07-23
update readme add icons to title

ibrataha8:


driesvints:
Thank you but we're not making this change right now.

ibrataha8:
Why ?

PR #6221, opened 2023-07-28
fixed type in app.php

Asociateone:
the ',' was missing between `loaded` and `so`.


driesvints:
This unfortunately breaks the comment spacing.

PR #6227, opened 2023-08-16
can set guzzlehttp/guzzle Can I use the current version? 🔨

64Dev:


crynobone:
`"^7.2"` should already allows `7.7.0`

PR #6228, opened 2023-08-21
Postmark mailer configuraiton update.

ninjaparade:
as mentioned in the following
ref https://github.com/craigpaul/laravel-postmark/issues/141

Postmark allows the mailer configuration to accept the `stream_id`, this is helpful if you sent transaction and broadcast notifications from Laravel with Postmark.


ninjaparade:
image

PR #6229, opened 2023-08-22
Update RedirectIfAuthenticated middleware for API support

alvarofvr:
RedirectIfAuthenticated middleware not work correctly in the context of a Laravel Breeze application scaffolded with api option (php artisan breeze:install api) or in general in the context of a client-driven authentication where Laravel act as a pure Backend.

See this issue for more explanations: https://github.com/platformsh-templates/laravel/issues/50

### DIFF
--- a/app/Http/Middleware/RedirectIfAuthenticated.php
+++ b/app/Http/Middleware/RedirectIfAuthenticated.php
@@ -21,6 +21,9 @@ class RedirectIfAuthenticated
 
         foreach ($guards as $guard) {
             if (Auth::guard($guard)->check()) {
+                if ($request->expectsJson()) {
+                    return response()->json(['message' => 'Already authenticated.']);
+                }
                 return redirect(RouteServiceProvider::HOME);
             }
         }
--- a/app/Http/Middleware/RedirectIfAuthenticated.php
+++ b/app/Http/Middleware/RedirectIfAuthenticated.php
@@ -24,6 +24,7 @@ class RedirectIfAuthenticated
                 if ($request->expectsJson()) {
                     return response()->json(['message' => 'Already authenticated.']);
                 }
+                
                 return redirect(RouteServiceProvider::HOME);
             }
         }
--- a/app/Http/Middleware/RedirectIfAuthenticated.php
+++ b/app/Http/Middleware/RedirectIfAuthenticated.php
@@ -24,7 +24,7 @@ class RedirectIfAuthenticated
                 if ($request->expectsJson()) {
                     return response()->json(['message' => 'Already authenticated.']);
                 }
-                
+
                 return redirect(RouteServiceProvider::HOME);
             }
         }


alvarofvr:
The middleware by default is aliased as "guest" and I think it's right for both cases:
`'guest' => \App\Http\Middleware\RedirectIfAuthenticated::class,`
clearly the scope is manage the routes that not require an authentication. 
In my opinion adding a new middleware with a different name that conceptually does the same checks but with a different return it's just duplication and I think there is no confusion in a clean check on the response type.  

**In Laravel we have other middleware that performs the same type of check and therefore covers both cases.**

Also having a new middleware will lead to more collateral things...add a new alias, chose a new name for the alias, the middleware is not always needed in all cases  etc...and side effects on other projects like scaffolding tools (see original issue https://github.com/platformsh-templates/laravel/issues/50). 

Lastly, in my opinion, there is no need to add user information in the response because the middleware only check authentication. The response can also be left blank but a simple message like "You are already authenticated." (nothing more) can help.

taylorotwell:

begueradj:
@514sid  We expect a JSON response when we chose:
```cmd
php artisan breeze:install --api
```
and not to be redirected anywhere

michaeldyrynda:
Perhaps we consider whether or not the `--api` switch in the Breeze installer overwrites the middleware as part of its scaffolding. 

In doing so, the middleware could also be renamed to something like `EnsureGuest`. 

Might be something for Laravel 11.x, though.

begueradj:
@michaeldyrynda exactly, you understood the issue

alvarofvr:
> Perhaps we consider whether or not the `--api` switch in the Breeze installer overwrites the middleware as part of its scaffolding.
> 
> In doing so, the middleware could also be renamed to something like `EnsureGuest`.
> 
> Might be something for Laravel 11.x, though.

@michaeldyrynda the scaffolding project (https://github.com/platformsh-templates/laravel/issues/50#issuecomment-1679371574) as said by a contributor pull the middleware file directly by this repo.

uz-bueno:
This is interesting. My understanding is that the middleware has behavior for redirecting when working with a traditional web application, and that is fine. 

Now, when having the backend and frontend separated, the backend becomes a stateless API, so I'm not sure if this middleware makes sense for a Laravel application that is just an API. Note that I'm assuming that Breeze (with --api) behaves like this, maybe I'm wrong.

Whenever you call the auth endpoint it won't know if the user is already logged in, so the expected behavior is to have a new auth token if credentials are valid.

I think that the behavior of knowing if the user is already logged in and redirection belongs to the frontend client.

Shujee:
With Laravel 11, since these middleware have been moved to the framework instead of being in user-land code, it is even more important that `RedirectIfAuthenticated` behaves correctly in case of API calls. I'm using Laravel as API in my SPA and getting affected by this even without using Breeze.

begueradj:
That is why we do not understand why this has been rejected. @Shujee

Tamas-hi:
This is still an issue for us as well. If Laravel acts as an API, this is not the behavior expected from this middleware. Any idea how can we communicate this issue better towards the developers?

begueradj:
@Tamas-hi This pull request includes a solution to this issue. It was unfairly rejected though.

Tamas-hi:
I would like to give this a second go and tag @taylorotwell @driesvints to reconsider this for one of the next (major) version.

Here is my personal take:

The middleware naming 'EnsureGuest', proposed by @michaeldyrynda, would perfectly capture its dual purpose. Under this name, the middleware could handle user authentication more intuitively by either:

- Redirecting an authenticated user (as it currently does for traditional Laravel web applications).

- Returning an 'Already authenticated' message in scenarios where the middleware is applied in a Laravel backend API.

This name makes the middleware's function unambiguous, offering clear benefits to both web apps and APIs.

For context, please take it into account that this two-way handling is already effectively managed in the reverse scenario: for example when accessing an authenticated route without being logged in:

```
protected function unauthenticated($request, array $guards)
{
    throw new AuthenticationException(
        'Unauthenticated.',
        $guards,
        $request->expectsJson() ? null : $this->redirectTo($request), // two-way handling
    );
}
```

I think even staying with the original name can be an option, but handling the _$request->expectsJson(_) as OP did in this PR makes a lot of sense for us in this case as well. 

Thanks if you reconsider. :)

dsebastien:
This is indeed still an issue. I added an API route with the 'guest' middleware, and stumbled upon this. It's far from logical or intuitive for newcomers. Could you reconsider?

smares:
@taylorotwell this is super frustrating for API-only backends.

PR #6230, opened 2023-08-22
[slim-skeleton-11.x] Schedules `inspire` command directly on the definition

nunomaduro:
More info: https://github.com/laravel/framework/pull/48137.


driesvints (approved):

taylorotwell:
Going to think about this one a bit more. You still need a place to schedule class based commands so giving an example of using Schedule facade is somewhat helpful. Merged framework though.

May come back to this later.

PR #6236, opened 2023-09-10
Create BaseModel.php for common functionality

ahmedalmory:


GrahamCampbell (changes requested):
I don't agree this makes sense to include in the default app.

muhammadmp97 (commented):
If I want to have some common functionalities on models, I simply create such a class and extend it, or use traits to make it more obvious to developers what it is about. The first case is rare and I have never needed to do that because the trait concept/Laravel global scopes/etc are enough.

driesvints:
No plans for this sorry.

PR #6240, opened 2023-09-15
Revert "Fix incorrect collation for MySQL 8"

Jubeki:
Reverts laravel/laravel#6239

It seems this collation is not supported for MariaDB:  
https://mariadb.com/kb/en/supported-character-sets-and-collations/

I also tested it with Laravel Sail + MariaDB and it doesn't work:

Screenshot 2023-09-15 at 16 22 29

In the original PR, @binaryfire asked if it possible to add a new config section for mariadb.


taylorotwell:
Hmm, I think we still need a solution for MySQL 8 here. It may be time to define a separate connection for Maria?

PR #6241, opened 2023-09-18
[10.x] Let database handle default collation

Jubeki:
I tested with Laravel Sail both MySQL and MariaDB both seem to accept the null value and then each use their own default collation.

Here  what I did:

<details><summary>MySQL</summary>
<p>

```
laravel new mysql-collation-test
cd mysql-collation-test

# Edit config/database.php and update collation in mysql to null

php artisan sail:insall
# Select only MySQL

sail up

# In another terminal
sail artisan migrate
sail artisan db

# SQL Command for showing collation defaults
SHOW VARIABLES LIKE 'collation%';
```

Result:

```
+----------------------+--------------------+
| Variable_name        | Value              |
+----------------------+--------------------+
| collation_connection | utf8mb4_0900_ai_ci |
| collation_database   | utf8mb4_0900_ai_ci |
| collation_server     | utf8mb4_0900_ai_ci |
+----------------------+--------------------+
3 rows in set (0.00 sec)
```

</p>
</details> 

<details><summary>MariaDB</summary>
<p>

```
laravel new mariadb-collation-test
cd mariadb-collation-test

# Edit config/database.php and update collation in mysql to null

php artisan sail:insall
# Select only MariaDB

sail up

# In another terminal
sail artisan migrate
sail artisan db

# SQL Command for showing collation defaults
SHOW VARIABLES LIKE 'collation%';
```

Result:
```
+----------------------+--------------------+
| Variable_name        | Value              |
+----------------------+--------------------+
| collation_connection | utf8mb4_general_ci |
| collation_database   | utf8mb4_general_ci |
| collation_server     | utf8mb4_general_ci |
+----------------------+--------------------+
3 rows in set (0.00 sec)
```

</p>
</details> 

You can so configure your defaults outside of Laravel if wanted, while still having the option to explicitly set a default for your application.

Risks:
- You have upgraded your Database and may have not updated your default collation. This could result in unexpected behaviour.
- Different collations between development / production environments

Alternative: Duplicate the mysql entry to mariadb and then change the default collation.
- Remaining TODOS: Update Docs and Sail installation for MariaDB
- while this is not a breaking change for existing applications, some developers may be confused if mysql is suddenly changed to mariadb

@binaryfire does this solution also work in your case?

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -53,7 +53,7 @@ return [
             'password' => env('DB_PASSWORD', ''),
             'unix_socket' => env('DB_SOCKET', ''),
             'charset' => 'utf8mb4',
-            'collation' => 'utf8mb4_unicode_ci',
+            'collation' => null,
             'prefix' => '',
             'prefix_indexes' => true,
             'strict' => true,


Jubeki:
Another thing to be aware of:
Laravel 10 supports MySQL 5.7 were the default collation is `utf8mb4_general_ci` see https://dev.mysql.com/doc/refman/5.7/en/charset-charsets.html

MySQL 5.7 also doesn't support the collation `utf8mb4_0900_ai_ci` and will reach end of support in October 2023. It is also not supported in Laravel 11 anymore. See https://laravel.com/docs/master/database#introduction

So setting the default for Laravel 10 `utf8mb4_0900_ai_ci` would only be possible in Laravel 11 for MySQL.

binaryfire:
Thanks for testing @Jubeki 

I definitely think Laravel 11 should drop support for MySQL 5.x. There have been several other changes in v8, for example the [password hashing algorithm](https://dev.mysql.com/doc/refman/8.0/en/sha256-pluggable-authentication.html). 8.0 was released in 2018 so it’s been out for 5 years now. And supporting backwards compatibility with 5.x would mean reduced compatibility with 8.0.

Personally I still think separate mysql and mariadb connections make sense. Oracle are focused on making MySQL a hybrid NoSQL/SQL solution with features like the [document store](https://www.mysql.com/products/enterprise/document_store.html) and a bunch of extra JSON functions. MariaDB is mainly focused on being a pure SQL db. I’m pretty sure v9 will see an even bigger divergence between the two.

binaryfire:
Letting the db handle the default collation could be a good stopgap for v10 though. I just meant separate connections would be better long term.

driesvints:
@Jubeki we reverted this one

PR #6245, opened 2023-09-21
[10.x] Increase bcrypt rounds to 12

valorin:
[PHP is increasing the default bcrypt cost](https://wiki.php.net/rfc/bcrypt_cost_2023) to either 11 or 12 to keep up with increases in computing, so we should do the same within Laravel. The current default of 10 was set in PHP 11 years ago, which is no longer a suitable default.

12 appears to be the sweet spot between performance and security, [as confirmed by a member of the Hashcat team](https://phpc.social/@tychotithonus@infosec.exchange/111104389164750098). [Symfony uses a cost of 13](https://github.com/symfony/password-hasher/blob/6.3/Hasher/NativePasswordHasher.php#L36C1-L37C1), however that may be too high for some servers.

Due to the way hashing works, there are no backwards compatibility issues - older passwords with lower rounds will still be handled properly, and code that automatically rehashes passwords will upgrade them over time. It's also worth pointing out that since rounds are defined in `config/hashing.php`, existing projects won't automatically get the new rounds cost and thus won't have any performance impacts. [The RFC contains hash calculation timings](https://wiki.php.net/rfc/bcrypt_cost_2023) if you'd like more information on the impacts.

Increasing rounds to 12 in `config/hashing.php` should be a recommended upgrade step for Laravel 11 (and possibly added to the guide for 10?).

Framework PR: https://github.com/laravel/framework/pull/48494

### DIFF
--- a/config/hashing.php
+++ b/config/hashing.php
@@ -29,7 +29,7 @@ return [
     */
 
     'bcrypt' => [
-        'rounds' => env('BCRYPT_ROUNDS', 10),
+        'rounds' => env('BCRYPT_ROUNDS', 12),
     ],
 
     /*


GrahamCampbell:
Probably we should change the default from bycrypt to argon in laravel 11.

valorin:
PHP recently discussed changing the default to Argon2, but were advised that Argon2 is not appropriate for use within web applications. See: https://externals.io/message/120993 & https://twitter.com/TerahashCorp/status/1155119064248913920

roycewilliams:
Concur with @valorin here. To summarize, even some of the PHC judges acknowledge that, in order to achieve auth speeds in alignment with UX tolerance studies (500ms to 1 second of auth delay per user), both the Argon2 family and scrypt have to be tuned "downward" so much that they end up being _less resistant to offline attack than bcrypt at equivalent auth speeds_(!)

Paradoxically, for the multi-user interactive login use case, and the "resist offline attack of multiple hashes", properly tuned bcrypt remains superior.

roycewilliams:
One more note: between CPU speed increases and the rise of GPU password cracking in the decade since PHP's default bcrypt work factor was set to 10, it would have been better to have moved to bcrypt cost 11 in 2017 or so. bcrypt cost 12 is the more defensible choice in 2023.

PR #6246, opened 2023-09-22
[10.x] Add `:8000` to .env.example

egyjs:
"Why not enhance Laravel experience by including `:8000` to the `APP_URL=http://localhost` configuration? 
After all, these settings are conveniently provided as default values in the .env.example file."

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,7 +2,7 @@ APP_NAME=Laravel
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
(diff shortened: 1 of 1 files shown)


Stijnbak:
People don't necessarily use port 8000.

egyjs:
> People don't necessarily use port 8000.

Yes, but as a default its `8000`.

Also people don't necessarily use localhost as the domain but as default yes

PR #6247, opened 2023-09-22
[10.x] Use 12 bcrypt rounds for password in UserFactory

Jubeki:
So that `needsRehashing` doesn't default to `true` on a fresh installation.

created using `php artisan tinker` and then verified:

```php
Hash::make('password');
Hash::check('password', '$2y$12$Z/vhVO3e.UXKaG11EWgxc.EL7uej3Pi1M0Pq0orF5cbFGtyVh0V3C');
```

### DIFF
--- a/database/factories/UserFactory.php
+++ b/database/factories/UserFactory.php
@@ -21,7 +21,7 @@ class UserFactory extends Factory
             'name' => fake()->name(),
             'email' => fake()->unique()->safeEmail(),
             'email_verified_at' => now(),
-            'password' => '$2y$10$92IXUNpkjO0rOQ5byMi.Ye4oKoEa3Ro9llC/.og/at2.uheWG/igi', // password
+            'password' => '$2y$12$Z/vhVO3e.UXKaG11EWgxc.EL7uej3Pi1M0Pq0orF5cbFGtyVh0V3C', // password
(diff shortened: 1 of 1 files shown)


valorin:
Given that PHPUnit overrides the rounds to 4 for performance reasons, and `UserFactory` is a testing helper, wouldn't it make more sense to use a hash generated with 4 rounds instead of 12?

I haven't tested it, but presumably it means any tests that hit the login flow will be forced to rehash the password, incurring a massive speed hit as it hashes with `rounds=12` and then rehashes for `rounds-4`. But maybe I'm missing something?

Jubeki:
@valorin you are right that the `phpunit.xml` sets the rounds to 4.

But the same hashed password string is also used for seeding the local database, were the 12 rounds will be used. The Unit Tests `needsrehashing` is probably only run during successful login tests which are also not so many in my opinion.

Probably a decision which @taylorotwell should make, how many rounds should be used in the UserFactory.

valorin:
Ah good point about local dev seeding.

The benefit I see in picking 4 is tests will be slightly faster, but as you said, only the ones doing a login flow.

I'm happy either way. 🤷

Jubeki:
@valorin See #6259 for the updated cost during tests.

PR #6248, opened 2023-09-27
[10.x] Fix typo in the comment for token prefix (sanctum config)

yuters:
This pull request fixes a simple typo in a recently added comment, as well as change some other words in order to not break the secret rule.


taylorotwell:
❤️

PR #6255, opened 2023-10-15
Remove `personal_access_tokens` migration because it is registered by Sanctum package

bram-pkg:
The `create_personal_access_tokens_table` migration is not needed in the skeleton, because it is already registered by the Sanctum package (https://github.com/laravel/sanctum/blob/3.x/src/SanctumServiceProvider.php#L43).

Keeping the migration updated with the package is also easier if it's not copied into the skeleton. Removing the migration also makes it easier for people to remove sanctum from their project, if needed.


taylorotwell:
Evaluating these changes for Laravel 11.

bram-pkg:
Cool, thanks!

PR #6258, opened 2023-10-23
[10.x] Hash improvements

timacdonald:
Proposing we make the algorithm check on by default for new applications. This means an exception will be thrown when checking a hash against a driver that did not create the hash.

Screenshot 2023-10-23 at 11 12 27 am

If an application decides to change hashing algorithm, then this can be turned off, however I feel having it on is the sensible and expected default.


datamweb (commented):
LGTM!

driesvints (approved):

PR #6262, opened 2023-10-31
Session blocking

zobay:
## Description 
- Add the `block_lock_seconds` key in config.
- Add the `block_wait_seconds` key in config.

The above feature was added in [this PR](https://github.com/laravel/framework/pull/48795).


taylorotwell:
Would need to follow comment conventions.

PR #6267, opened 2023-11-03
[10.x] Add suggestion to enable AddLinkHeadersForPreloadedAssets middleware

jnoordsij:
In https://github.com/laravel/framework/pull/44096 the `AddLinkHeadersForPreloadedAssets` middleware was added. It might provide a small performance bonus over using just the html headers when enabled. Therefore I propose to add it as a hint to the skeleton so people are more aware of it (I only encountered it while explicitly looking for this feature, which I doubt too many will do).

An alternative might be to even enable it by default within the skeleton, as I see basically no downsides to enabling it.

Note: I think this should be in the global stack, but possibly the `web` stack might be more appropriate.


taylorotwell:
Will think about this for the Laravel 11.x skeleton since we're working on that now.

PR #6273, opened 2023-11-15
[10.x] Ability to define a PostgreSQL DB Schema via environment variables.

ArdaKaraderi:
As discussed on: https://www.enterprisedb.com/blog/new-public-schema-permissions-postgresql-15

Postgres 15< requires the DB user to be the DB owner in order to use the public schema. However, it is not always a good idea to make the application's DB user the DB owner. Therefore, now with the new versions when the out-of-the-box setup is used with PostgreSQL, the migrations fail due to permission issues.

This update enables a user to directly change the search_path (formerly known as schema) to another value than the static one via an environment variable called "PG_SEARCH_PATH". This has been a method adopted by the community, however, it has not been standardised by Laravel yet. Therefore, this PR aims to provide a standardized way of doing so.

More discussions: https://stackoverflow.com/questions/69674451/postgresql-how-to-give-the-database-owner-permissions-to-public-schema


taylorotwell:
You can add this to your own application.

PR #6276, opened 2023-11-20
[slim-skeleton-11.x] Adds `abstract` keyword on Controller class

nunomaduro:
This pull request adds the `abstract` keyword on Controller class for Laravel 11.x.


Jubeki:
Would this not be something viable for Laravel 10.x?

PR #6280, opened 2023-11-23
Make timezone configurable from .env

nurmuhammet-ali:
I believe setting the timezone from .env will be more useful and convenient.


driesvints:
Thank you but please see many previous attempts. We believe this is something that shouldn't be changed.

PR #6283, opened 2023-11-28
add support for DB_SCHEMA environment variable for PostgreSQL databases

ssn:
PostgreSQL supports the use of schemas within the same database.
To use different schemas, 'search_path' needs to be set.
Laravel default configuration sets 'search_path' to 'public', and any change to the schema requires a change in this file.

With this update, the schema can be defined using a DB_SCHEMA environment variable.


driesvints:
Thank you but please see previous attempts. Feel free to add this to your own app.

PR #6284, opened 2023-11-29
[10.x] Update BCRYPT_ROUNDS in phpunit.xml match with hash config

hosseinmoghaddam:
I don't know if this is to lower the cost of testing or not but I have to change BCRYPT_ROUNDS every new project because it doesn't match the hashing file

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -19,7 +19,7 @@
     </source>
     <php>
         <env name="APP_ENV" value="testing"/>
-        <env name="BCRYPT_ROUNDS" value="4"/>
+        <env name="BCRYPT_ROUNDS" value="10"/>
         <env name="CACHE_DRIVER" value="array"/>
         <!-- <env name="DB_CONNECTION" value="sqlite"/> -->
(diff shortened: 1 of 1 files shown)


Jubeki:
Because Hashing with many rounds is an expensive operation, you want fewer rounds in your tests. Therefor there is a different value in `phpunit.xml` which is only used during tests.

hosseinmoghaddam:
> Because Hashing with many rounds is an expensive operation, you want fewer rounds in your tests. Therefor there is a different value in `phpunit.xml` which is only used during tests.

This is true, but if someone is new to Laravel, after running `User::factory()->create()` they will get the following error, and I think it can be confusing.
image

Jubeki:
Could be related to: https://github.com/laravel/laravel/pull/6258

But it works for me in a new laravel application. Do you have changed some configuration regarding hashing?

Jubeki:
Or are you setting a different hash round value in your `.env` ?

crynobone:
@hosseinmoghaddam please update your `UserFactory` class.

hosseinmoghaddam:
It's true, it was my mistake, thank you

PR #6287, opened 2023-12-11
Use `assertOk()` instead of `assertStatus(200)` in tests

TENIOS:
I think it's a better way for Laravel 11.x concepts

### DIFF
--- a/tests/Feature/ExampleTest.php
+++ b/tests/Feature/ExampleTest.php
@@ -14,6 +14,6 @@ class ExampleTest extends TestCase
     {
         $response = $this->get('/');
 
-        $response->assertStatus(200);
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks!

TENIOS:
I think also need apply this change to the `.stub` file: https://github.com/laravel/framework/blob/10.x/src/Illuminate/Foundation/Console/stubs/test.stub#L18

And add this line as a comment https://github.com/laravel/framework/blob/10.x/src/Illuminate/Foundation/Console/stubs/test.stub#L6
(like this https://github.com/laravel/laravel/blob/10.x/tests/Feature/ExampleTest.php#L5 or delete this line)

PR #6076, opened 2023-01-17
Utilise the Route::view method for the default welcome view

ryanxedi:
Very clean and supersedes the current implementation.

PR #5387, opened 2020-09-01
[7.x] Use `Route::view` for the initial route

mitogh:
Define the usage of a view route on the initial view instead of a callback.

PR #5347, opened 2020-07-20
Use Route::view()

zaratedev:

PR #6290, opened 2023-12-17
[10.x] Refactor Default route to use Route::view()

mr-punyapal:
In this stage, the view() method makes more sense than get().

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -13,6 +13,4 @@ use Illuminate\Support\Facades\Route;
 |
 */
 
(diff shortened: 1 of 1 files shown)


rojtjo:
This has been attempted many times. 

Here are some of the previous attempts:
#4701
#4844
#4883
#5340
#5347
#5371
#5387
#6076

mohammed-17:
Ok great 😃👍

mr-punyapal:
Make sense 🫡

PR #6295, opened 2023-12-18
[11.x] Rename migration tables

Jubeki:
Was it intended that, that the users migration was 0, cache 2 and jobs 3, leaving out the 1?

In my opinion, the migration file names should reflect their intention and not only a single table in this case.


fylzero:
Sorry, pointed to wrong branch. Updated PR [here](https://github.com/laravel/laravel/pull/6334).

PR #6296, opened 2023-12-18
[11.x] Refresh Database in Tests

Jubeki:
Due to the changes in `.env.example` were you use session and cache tables, you also need to migrate the database.

What is your opinion on including the `LazilyRefreshDatabase` Trait into the General TestCase by default?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,6 +16,16 @@ jobs:
   tests:
     runs-on: ubuntu-latest
 
+    services:
+      mysql:
+        image: mysql:8
+        env:
+          MYSQL_ALLOW_EMPTY_PASSWORD: yes
+          MYSQL_DATABASE: laravel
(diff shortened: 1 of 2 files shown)


driesvints:
Not really a fan of requiring mysql for the default test suite run... wondering if we can tackle this differently.

Jubeki:
Taylor changed the defaults in https://github.com/laravel/laravel/commit/feded74d9eea2cde9715b75412ae8e5a5342f52d

So now you are requiring some kind of database for the default skeleton.
SQLite would probably then be the only alternative.

Jubeki:
Or even better use the `:memory:` option from sqlite, then you don't really have a database.

driesvints:
I think @nunomaduro's PR #6297 might be a better alternative here. Thanks!

Jubeki:
That is also an option! Thanks for the fast response.

PR #6301, opened 2023-12-23
[10.x] Add roundrobin transport driver config

me-shaon:
In this PR, I'm adding the 'roundrobin' Mail transport driver config which was added in this PR: https://github.com/laravel/framework/pull/49435

### DIFF
--- a/config/mail.php
+++ b/config/mail.php
@@ -50,16 +50,16 @@ return [
             'transport' => 'ses',
         ],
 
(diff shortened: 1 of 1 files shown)


u01jmg3:
@me-shaon: looks like this change never made it to the [`11.x`](https://github.com/laravel/laravel/blob/11.x/config/mail.php) branch. Should it be added?

PR #6307, opened 2023-12-29
Fix RuntimeException: Could not verify the hashed value's configuration

gnovaro:
Update phpunit.xml


taylorotwell:
This shouldn't be necessary.

valorin:
The rounds are kept short in testing to make tests run faster.

A better fix is to update your factories to use rounds of 4, or do something like https://github.com/laravel/laravel/blob/10.x/database%2Ffactories%2FUserFactory.php

PR #6308, opened 2024-01-01
[10.x] Update guzzlehttp/guzzle to latest version

anasmorahhib:
Update guzzlehttp/guzzle to latest version


driesvints:
Thanks but the current constraints already pull this in.

PR #6309, opened 2024-01-02
[11.x] Simplifies console routes

nunomaduro:
This pull request uses https://github.com/laravel/framework/pull/48137, to simplify the console routes. 💅🏻

```php
})->purpose('Display an inspiring quote')->hourly();
```


Jubeki:
I would be in favour of leaving the `Schedule Command` Section with the following commented out for example:

```php
// Schedule::command('model:prune')->daily();
```

I am also not sure, if anything should be scheduled by default. (Does `inspire` fill up the log files?)

taylorotwell:
Don't think we should do this atm.

PR #6311, opened 2024-01-03
[11.x] Bumps versions on `composer.json`

nunomaduro:
This pull request concerns L11, and updates all versions in `composer.json` for aesthetic reasons. Primarily to display the latest versions of those packages in the `composer.json` file.

In addition, it *always* use the pattern `Major.Minor` for all requirements. I've kept 1.6.3 for `axios`, because: https://github.com/laravel/laravel/pull/6306.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,8 +6,8 @@
         "build": "vite build"
     },
     "devDependencies": {
-        "axios": "^1.6.1",
-        "laravel-vite-plugin": "^1.0.0",
-        "vite": "^5.0.0"
+        "axios": "^1.6",
+        "laravel-vite-plugin": "^1.0",
+        "vite": "^5.0"
     }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -6,8 +6,8 @@
         "build": "vite build"
     },
     "devDependencies": {
-        "axios": "^1.6.1",
-        "laravel-vite-plugin": "^1.0.0",
-        "vite": "^5.0.0"
+        "axios": "^1.6",
```
GrahamCampbell, line 9: This undoes https://github.com/laravel/laravel/commit/6f6cfb020414bb084eaffeb910ac013095949e22. We should be aligned on if we want such PRs at all.

GrahamCampbell (changes requested):

GrahamCampbell (approved):

PR #6316, opened 2024-01-11
[11.x] Simplify PHP comments

nunomaduro:
Hi everyone! This pull request proposes a small change the PHP comments in our skeleton. They've been around for a while, and we're all pretty used to them. But, I'm curious about your thoughts on reshaping our docs to align more with the style used in "laravel/framework:" (and in any other repository really), while simultaneously do the three-character rule.

Here is an example:

Screenshot 2024-01-11 at 17 27 08

Not that, I am not fully convinced; however, if this pull request gets merged, I will go around to the other repositories and adjust the PHP comments accordingly via PR.

### DIFF
--- a/artisan
+++ b/artisan
@@ -5,29 +5,19 @@ use Symfony\Component\Console\Input\ArgvInput;
 
 define('LARAVEL_START', microtime(true));
 
-/*
-|--------------------------------------------------------------------------
-| Register The Auto Loader
-|--------------------------------------------------------------------------
-|
-| Composer provides a convenient, automatically generated class loader for
-| this application. We just need to utilize it! We'll simply require it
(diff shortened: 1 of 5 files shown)


taylorotwell:
Took it a bit further and deleted all comments from routes files and integrated your previous scheduler simplification suggestion.

renatofrota:
> Took it a bit further and deleted all comments from routes files and integrated your previous scheduler simplification suggestion.

@taylorotwell ,

Respectfully, I didn't appreciate the comments being removed. Laravel stops being friendly to newbies, like me, when a file is left without any explanation as to why it exists. The fact that it is absolutely simple makes it difficult to understand its function. It is precisely in this situation that comments are most useful.

PR #6319, opened 2024-01-16
[10.x] Don't install unnecessary symfony polyfills

philbates35:
At the time of writing, when creating a fresh Laravel app the following symfony polyfill packages end up being installed:

* `symfony/polyfill-ctype`
* `symfony/polyfill-intl-grapheme`
* `symfony/polyfill-intl-idn`
* `symfony/polyfill-intl-normalizer`
* `symfony/polyfill-mbstring`
* `symfony/polyfill-php72`
* `symfony/polyfill-php80`
* `symfony/polyfill-php83`
* `symfony/polyfill-uuid`

The project's `composer.json` requires a minimum of PHP 8.1, and according to https://laravel.com/docs/10.x/deployment#server-requirements both the `ctype` and the `mbstring` extension are required, meaning that there is no need to polyfill those extensions or PHP <= 8.1. The others should continue to be polyfilled.

For the PHP version polyfills, we could just just add the `php72` and `php80` polyfills to the `"replace"` section, as they are the only ones that end up being installed on a fresh project, but that felt incomplete and also doesn't future proof for the future when installing a new package that happens to pull in e.g. `polyfill-php56` which would be installed if we didn't explicitly add it here.

From https://github.com/symfony/polyfill#design:

> If your project requires a minimum PHP version it is advisable to add polyfills for lower PHP versions to the replace section of your `composer.json`. This removes any overhead from these polyfills as they are no longer part of your project. The same can be done for polyfills for extensions that you require.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,20 @@
         "phpunit/phpunit": "^10.1",
         "spatie/laravel-ignition": "^2.0"
     },
+    "replace": {
+        "symfony/polyfill-ctype": "*",
+        "symfony/polyfill-mbstring": "*",
+        "symfony/polyfill-php54": "*",
+        "symfony/polyfill-php55": "*",
+        "symfony/polyfill-php56": "*",
+        "symfony/polyfill-php70": "*",
+        "symfony/polyfill-php71": "*",
+        "symfony/polyfill-php72": "*",
+        "symfony/polyfill-php73": "*",
+        "symfony/polyfill-php74": "*",
+        "symfony/polyfill-php80": "*",
+        "symfony/polyfill-php81": "*"
+    },
     "autoload": {
         "psr-4": {
             "App\\": "app/",


driesvints:
I don't feel we should add this to the skeleton by default. You can add this to your own project if you like.

GrahamCampbell:
Yeh, I don't think we should do this, and actually some of those polyfills you may need if you don't have the extensions.

philbates35:
> I don't feel we should add this to the skeleton by default. You can add this to your own project if you like.

@driesvints Yep of course you're right it can be added individually to everyone's project based on personal preference, but before the PR is closed I just want to highlight that its currently impossible to create a fresh Laravel app using PHP < 8.1 or when `ctype` / `mbstring` are missing due to the requirements in this project's `composer.json` and laravel/framework's `composer.json`, meaning really its just wasteful having these packages downloaded on every `composer install` for every new Laravel project in development, prod and in CI. As well as any (admittedly micro) performance hit it take to have these packages included in the autoloader, as mentioned in the Symfony docs.

> Yeh, I don't think we should do this, and actually some of those polyfills you may need if you don't have the extensions.

@GrahamCampbell If its just the extension polyfills that are your main concerns, would you be open to me dropping `polyfill-mbstring` and `polyfill-ctype` from the PR, but keeping the `polyfill-phpxx`'s?

driesvints:
> its currently impossible to create a fresh Laravel app using PHP < 8.1

PHP < 8.1 isn't supported anymore. PHP 8.1 is the minimum for Laravel v10: https://github.com/laravel/framework/blob/10.x/composer.json#L18

philbates35:
@driesvints Yep exactly, which is why I submitted the PR in the first place - to stop polyfills for PHP <8.1 from ever being installed because its impossible to run the app on PHP < 8.1 as you say.

To put it another way, when creating a new Laravel project there's no point having a polyfill for PHP 7.2, PHP 8.0 etc installed because we can guarantee they must be using PHP >= 8.1.

taylorotwell:
Yeah, not sure I want this in the skeleton by default.

PR #6323, opened 2024-01-22
[11.x] Uses `sqlite` by default on testing

nunomaduro:
This pull request aligns the `phpunit.xml` file with this pull request: https://github.com/laravel/laravel/pull/6322.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -22,8 +22,8 @@
         <env name="APP_MAINTENANCE_DRIVER" value="file"/>
         <env name="BCRYPT_ROUNDS" value="4"/>
         <env name="CACHE_STORE" value="array"/>
-        <!-- <env name="DB_CONNECTION" value="sqlite"/> -->
-        <!-- <env name="DB_DATABASE" value=":memory:"/> -->
+        <env name="DB_CONNECTION" value="sqlite"/>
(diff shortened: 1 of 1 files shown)


driesvints (approved):

taylorotwell:
My main reservation with this is it potentially puts you on a different database platform for running tests than your actual app in local dev.

First, even without this PR, you would **already** be running tests on SQLite, just on in-memory, since it would use your default database driver which would be SQLite.

But, imagine you use the Laravel installer to create a new Laravel project, and you select MySQL as your database when the installer prompts you. Your default database in your .env file would be updated to MySQL, but with this PR your tests would still be running on SQLite. I don't think that's a great idea. I think your local dev database should match your test database.

driesvints:
@taylorotwell as long as the actual database in MySQL is different than your local dev environment I think that's reasonable.

nunomaduro:
> you would already be running tests on SQLite, just in-memory

@taylorotwell Maybe we can - update the DB_CONNECTION - in runtime, if people are using SQLite as database driver?

taylorotwell:
Maybe...

I think for now I will leave this as-is, where your testing database is always pegged to your application's database by default, which I think makes sense. It still means your testing on SQLite.

I think if we are going to go down the road offered in this PR the installers need to be a bit smarter about updating this file as well.

PR #6326, opened 2024-01-26
Update .gitignore

nurul-komor:
We don't require these two lines with fresh Laravel. even not from anyone. We need the versions that match with our system. so, that we don't face difference problems before setup laravel sail and so on

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ yarn-error.log
 /.fleet
 /.idea
 /.vscode
+/composer.lock
+/package-lock.json


decadence:
[Commit your composer.lock file to version control](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control)

remotemerge:
I noticed you made changes to ignore `composer.lock` and `package-lock.json` from Git. While your intention is understandable, it's generally not a recommended practice. Including lock files is crucial for consistent dependency management, reproducibility, and security.

driesvints:
These are definitely required when building applications.

iamafasha:
I would prefer to have these files ignored especially for the `composer.lock `. when you run `composer install` on any machine, it generates or updates the `composer.lock`  accordingly.  

The problem with keeping causes a lot of conflicts when you are working in teams.

michaeldyrynda:
`composer install` uses the `composer.lock` file to ensure a known set of packages are installed at expected versions. 

Without the lockfile, `composer install` does nothing. Using `composer update` will update your installed dependencies, and in turn, the `composer.lock` file. 

Running `composer update` as part of deployment is risky because it will lead to installing newer versions of dependencies than your application is tested against, and likely (eventually) errors in production due to untested dependencies. 

Conflicts in the lockfile are expected on occasion, but need to be managed either by controlling who does updates or coordinating them such they can be minimised.

If two people run `composer update` or install new dependencies, then yes, conflicts between the versions of the file will occur. 

Similar is true for the `package-lock.json` file. Both are for consistent and stable installation of vendor code. 

For app code, they should always be committed as part of the repository.

perisicnikola37:
@nurul-komor not a good practice to remove these from .gitignore

nurul-komor:
My mistake. 
 Laravel doesn't provide these two lines. But, I wanna suggest adding these to .gitignore file

remotemerge:
> My mistake. Laravel doesn't provide these two lines. But, I wanna suggest adding these to .gitignore file

Programming is just like hiking a wonderful mountain through a slippery path with some beautiful people. ⛰️🚶 So walk carefully, slowly, and intelligently.💡

PR #6327, opened 2024-01-26
Slim example env

mattdfloyd:
This removes the Pusher specific env variables from the example env file. These variables would be used in the broadcasting config _if_ the application is using the pusher broadcast driver. By default, the driver is null so these would not be used in a fresh installation.

I delete these from my env files on each project. I wonder if Pusher is event the de facto platform with the rise of Ably? If a developer does choose to use Pusher, the broadcast config file still gives them clear direction on which env variables ought to be set.


MrPunyapal:
You forgot to remove `VITE_APP_NAME="${APP_NAME}"`

taylorotwell:
I'll consider this closer to release.

PR #6328, opened 2024-01-27
Expose an env variable to configure Postgres search_path 

YasienDwieb:


YasienDwieb:
@antonkomarev Maybe removal of the variable from .env.example could be a good idea, while keeping the config code, no?

antonkomarev:
Maybe the team will accept it, I don't know.

PR #6333, opened 2024-02-04
[11.x] Completely remove dates in migrations

fylzero:
I realize this is extremely nit picky, but do we even need the 1s here? Can't we just make these numeric 1, 2, 3?


fylzero:
Sorry, pointed to wrong branch. Updated PR [here](https://github.com/laravel/laravel/pull/6334).

EzraDechwong0:
Open> I realize this is extremely nit picky, but do we even need the 1s here? Can't we just make these numeric 1, 2, 3?

PR #6334, opened 2024-02-04
[11.x] Completely remove date concept in migrations

fylzero:
I realize this is extremely nit picky, but do we even need the 1s here? Can't we just make these numeric 1, 2, 3?


davidbirkin:
If this is merged, it would be nice to have a config that allows us to revert back to dates. I use the dates to know when changes were made. Yes, I understand I can look at commit history, but dates are handy at a glance when developing in the IDE

fylzero:
> If this is merged, it would be nice to have a config that allows us to revert back to dates. I use the dates to know when changes were made. Yes, I understand I can look at commit history, but dates are handy at a glance when developing in the IDE

This PR is only suggesting to change this for the stock/initial Laravel migrations - not migrations as a whole

davidbirkin:
> > If this is merged, it would be nice to have a config that allows us to revert back to dates. I use the dates to know when changes were made. Yes, I understand I can look at commit history, but dates are handy at a glance when developing in the IDE
> 
> This PR is only suggesting to change this for the stock/initial Laravel migrations - not migrations as a whole

Understood, plese disregard my comment then! Thanks for clarification

PR #6335, opened 2024-02-07
Implement L11 welcome page design

jasonlbeggs:
Once this PR is approved, we can upload the two screenshots to a CDN and I'll replace the links.

Then I'll submit PRs for the Vue and React versions in Breeze and Jetstream.


mateusjatenee:
nice

jasonlbeggs:
I am curious to hear Taylor and the team's thoughts on color variables. I used pretty much all JIT colors here since the styles are inlined and we don't have a Tailwind config. With Breeze and Jetstream, there is a Tailwind config and I suppose we might want to update the auth/dashboard/profile views to match the new colors? Or do we just leave those as is and let the welcome page use slightly different colors? I'm open to thoughts and suggestions.

antonkomarev:
@jasonlbeggs could you provide screenshots before and after?

timacdonald:
> we can upload the two screenshots to a CDN and I'll replace the links

Can we make sure to have a fallback here in case of CDN failure, i.e., an `onerror` hook that replaces the image tag with something else so that it looks good or re-styles the block in some way? This will ensure that the page looks good offline as well.

I also note that the size of the background SVG means that, at least on my machine (M1, 2020 Mac mini) in both Firefox and Safari, the browser is _noticeably_ sluggish when resizing the browser window. It is usually only developers dragging browser widow corners, so usually I don't care about that kinda thing too much, but as this page is targeted at developers it feels important to leave a good first impression. We don't want a sluggish "basic" welcome page IMO. Maybe we should just capture the SVG as a PNG?

I also note that the SVG background image renders differently across Safari, Chrome, and Firefox. 

<details>

<summary>Desktop screenshots</summary>

## Firefox

Screenshot 2024-02-12 at 1 21 38 pm

## Chrome

Screenshot 2024-02-12 at 1 22 00 pm

## Safari

Screenshot 2024-02-12 at 1 21 42 pm

</details>

Lastly, due to the white Laravel logo and the way Safari renders the background it means the logo is rather hard to make out on mobile.

<details>

<summary>Mobile screenshot</summary>

IMG_DDF0FAE4129D-1

</details>

jasonlbeggs:
Thanks for the feedback @timacdonald! Ugh I thought I fixed the background image rendering differently in each browser. Yeah a PNG would probably be the easiest option for that. We can upload it to the CDN too.

I'll experiment with adding an `onerror` listener and see what looks best.

Do you have any opinions on the color naming that I mentioned in my previous comment?

timacdonald:
I don't love the use of all inline colors, but I don't think it is a problem - just a feeling.

I don't think we should have different colors between the welcome page and Breeze an Jetstream. We had that previously and it felt rather janky because the "Welcome" page here becomes part of Breeze and Jetstream. A fresh Breeze install would see one color scheme for the homepage and a different color scheme for the login / password reset pages.

I previously standardised the colors for this reason: https://github.com/laravel/laravel/pull/6002

jasonlbeggs:
I can try refactoring to default Tailwind colors (except maybe the brand red) if @taylorotwell is cool with the design not matching perfectly. I'll also have to adjust the colors on the docs site screenshots to match.

resources/views/welcome.blade.php
```diff
@@ -2,7 +2,7 @@
 <html lang="{{ str_replace('_', '-', app()->getLocale()) }}">
     <head>
         <meta charset="utf-8">
-        <meta name="viewport" content="width=device-width, initial-scale=1">
+        <meta name="viewport" content="width=device-width, initial-scale=1" />
```
Seb33300, line 5: FYI, W3C validator is raising an `info` message when using trailing slashes and the recommendation is to omit it.
https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements#trailing-slashes-directly-preceded-by-unquoted-attribute-values

jasonlbeggs:
Ok @timacdonald  I've updated the design to use the default Tailwind colors and added an onerror hook to one of the screenshots so if it doesn't load, the page will adjust layout.

Normal circumstances:
image
image


Images don't load:
image
image

I also moved the background image to a real `` tag to get around the sluggishness and browser differences.

Check it out and let me know if we're ready to move on to adding it to Breeze and Jetstream!

driesvints:
@avinmaster no need for that

nmedelmehdi:
Hello
I think in line **19**, className has been forgotten from a React code. <div **className**="bg-gray-50 text-black/50 dark:bg-black dark:text-white/50"> to be replaced by <div class="bg-gray-50 text-black/50 dark:bg-black dark:text-white/50">

PR #6338, opened 2024-02-11
Vite config HMR Hostname Update

JamShady:
This updates the default vite.config file to use the defined APP_URL (from the appropriate .env file, depending on what 'mode' the app is set up to use) as the HMR hostname.

I use Laravel in a dockerised setup (specifically, Laradock), and HMR requests always fail without this.

Hosts also need to be exposed via --hosts, i.e. `npx vite --hosts`

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -1,11 +1,23 @@
-import { defineConfig } from 'vite';
+import { defineConfig, loadEnv } from 'vite';
 import laravel from 'laravel-vite-plugin';
 
-export default defineConfig({
-    plugins: [
-        laravel({
-            input: ['resources/css/app.css', 'resources/js/app.js'],
-            refresh: true,
-        }),
-    ],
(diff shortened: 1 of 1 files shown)


JamShady:
Defaulted to the 10.x branch instead of master accidentally

timacdonald:
Recommend setting this in your own application if you need it.

driesvints:

JamShady:
No problem to do it in your own app - I tend to agree.

My reason for making this PR though was to give some kind of instruction or documentation for this common problem to save other developers wasting time researching the solution.

Is the objective here to be completely minimalistic, or to provide solutions to common problems. Perhaps this could be added in a commented out form?

ahinkle:
> My reason for making this PR though was to give some kind of instruction or documentation for this common problem to save other developers wasting time researching the solution.

Sounds a PR for the docs rather than the code itself.

PR #6339, opened 2024-02-12
[11.x] Use list instead of array<int, string>

phh:
Instead of writing `array<int, Type>` one can take advantage of `list<Type>` if index is starting at 0 (like if the index is generated automatically).

This is both supported by [PHPStan](https://phpstan.org/writing-php-code/phpdoc-types#lists) and [Psalm](https://psalm.dev/docs/annotating_code/type_syntax/array_types/#lists).

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -14,7 +14,7 @@ class User extends Authenticatable
     /**
      * The attributes that are mass assignable.
      *
-     * @var array<int, string>
+     * @var list<string>
      */
     protected $fillable = [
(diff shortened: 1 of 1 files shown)


taylorotwell:
No plans to change this right now since we use this syntax a lot of places.

phh:
@ahinkle Feel free to reopen with new considerations. This PR was made due to just taken advantage of the `list` over array, however, seems like with the Larastan 3 release it makes a bit more sense.

PR #6342, opened 2024-02-15
Update .gitignore with 'local' folder

vasymus:
add some 'local' folder (which should be ignored) in the root of the project,
so developers could add their own temporary files for testing / notes / comparing etc


driesvints:

vasymus:
It could be considered as "good default".
some of the other folders from gitignore
are not in every project :-)
but as you wish

PR #6343, opened 2024-02-17
propose database seeder change

ArielMejiaDev:
In 7+ Years working with Laravel, there are no projects that I see without the `call` method in the database seeder, and in many cases, this is the only method invoked.

This PR adds a comment to make this as easy to use as the `User` factory comment that is added as some kind of boilerplate.


datamweb (commented):
LGTM!

taylorotwell:

PR #6345, opened 2024-02-19
[11.x] Move Providers Folder

Jubeki:
This is only a proposal / idea. Not sure if its for the best.

The AppServiceProvider is kind of for bootstrapping the application, so it would make sense to move it / all ServiceProviders into the `bootstrap` folder.


driesvints:
I don't think we should do this. It's been in the app directory for so long. No need to move this. So many tutorials across the web will need to be updated just for this change.

PR #6346, opened 2024-02-22
Added APP_LOCALE and APP_TIMEZONE to .env

devrfc83:
Most of the settings on the **config/app.php** file are customizable from **.env**, but the 'locale' and 'timezone' are not.

The modification in the config/app.php file is harmless: it reads an environment variable if it's set, and if not it falls back to the default values that were included anyways.

I also modified the **.env.example** file, adding both variables just in case.


driesvints:
This is coming in Laravel v11 👍

PR #6347, opened 2024-02-22
AllowLanguages middleware

vishal-prajapati-yudiz:
The AllowLanguages middleware dynamically sets the application's language based on the request's preferred language.
It retrieves the allowed languages from the controller and checks the request's x-language header to determine the language to use.
If the requested language is allowed, it sets the application's locale to that language; otherwise, it defaults to the default language.


vishal-prajapati-yudiz:
The AllowLanguages middleware dynamically sets the application's language based on the request's preferred language.
It retrieves the allowed languages from the controller and checks the request's x-language header to determine the language to use.
If the requested language is allowed, it sets the application's locale to that language; otherwise, it defaults to the default language.

driesvints:

PR #6354, opened 2024-02-28
Recompile styles

jasonlbeggs:


driesvints:
Thanks! Do we also need to do this for Breeze or Jetstream?

jasonlbeggs:
I don't think so. They use actual compiled styles with Vite. Thanks!

PR #6355, opened 2024-03-05
Add DB_CHARSET + DB_COLLATION

jasonmccreary:
This adds `DB_CHARSET` and `DB_COLLATION` environment variables.

Since the default values are changing between Laravel 10 and Laravel 11, having these would allow applications to further slim their configuration files when this is the only customized option, or prevent publishing their entire database driver config just to change these single values.


jasonmccreary:
If merged, I'm glad to put these in the .env.example and issue a similar PR to the framework base configs.

PR #6356, opened 2024-03-06
Add Laravel Herd

mpociot:
This PR adds a link to Herd to the first-party tools section of the welcome view.

image


driesvints:
We'll need to update the welcome views on Breeze and Jetstream as well.

PR #6367, opened 2024-03-13
[11.x] Changing DB_COLLATION

imariusalin:
Tested on windows & nixos. On default installation this error persists.

`SQLSTATE[HY000]: General error: 1273 Unknown collation: 'utf8mb4_0900_ai_ci' (Connection: mysql, SQL: select table_name as `name`, (data_length + index_length) as `size`, table_comment as `comment`, engine as `engine`, table_collation as `collation` from information_schema.tables where table_schema = 'crm' and table_type in ('BASE TABLE', 'SYSTEM VERSIONED') order by table_name)`

Changing the DB_COLLATION to utf8mb4_unicode_ci fixes the issue!


crynobone:
I believe you are using MySQL 5.7?

driesvints:
@imariusalin what DB version are you using?

driesvints:
@imariusalin seems highly unlikely that this is failing for the mentioned engine versions since our test suites test these versions. You're 100% sure you're not running MySQL 5.7?

crynobone:
Wouldn't it be better to use `DB_CONNECTION=mariadb`?

driesvints:
@imariusalin I don't understand. You're using the MYSQL connection for MariaDB while there is a MariaDB connection? Does the issue also happen for you if the MariaDB connection? It seems not from your posts above but want to double check.

Jubeki:
In the first image it says `unknown database "teste"` it seems like the database is not correctly created?


As for your original report shows, that you tried to use a MySQL 8 Collation on either a MySQL 5.7 or MariaDB database. (Like @driesvints and @crynobone said)

MySQL 8: `utf8mb4_0900_ai_ci` (not supported by MariaDB and MySQL 5.7)
MariaDB: `utf8mb4_uca1400_ai_ci` (not supported by MySQL 8 and MySQL 5.7)

Older and deprecated: `utf8mb4_unicode_ci`, that was also the reason why it was changed in Laravel 11.

Jubeki:
Maybe instead of changing the default collation, it would be better to show a reasonable error, like:

It seems you are using MariaDB and also try to use a MySQL 8.0 only collation, we recommend changing the driver to mariadb or change the collation at your own risk

Or show what Database Version you are using, so that it is easier debugable.

sairiz:
Hi. My server setup in Laravel Forge use MariaDB 10.11

By default creating a table will use
Type : Innodb
Encoding : utf8mb4
Collation : utf8mb4_unicode_ci

The screenshot below is taken to show all the option available. Since this is from Forge, it should be a good reference of a live setup from Laravel ecosystem. I did have to set DB_COLLATION=utf8mb4_unicode_ci

Screenshot 2024-03-13 at 10 41 46 PM

davidoskay:
@Jubeki 

There was **wrong assumption**  for MariaDB version:
https://github.com/laravel/framework/pull/48455#issuecomment-1729278985
> uca1400_ai_ci is based on Unicode 14 and available [in MariaDB since 10.10.1](https://jira.mariadb.org/browse/MDEV-27009) (Laravel 11.x will support
> MariaDB 10.11+ so we're covered https://laravel.com/docs/master/database).

So wrong version assumtion (10.11+) or we have wrong info in docs:
Minimum MariaDB **10.3+** https://laravel.com/docs/11.x/database#introduction
but uca1400_ai_ci is from **10.10.1**
newest **xampp** version MariaDB **10.4.32** (for windows users xampp is very common It's why I'm relating to this)

Jubeki:
It seems the supported mariadb versions were reverted in the docs with https://github.com/laravel/docs/pull/9113

I also did the PR with 10.11+ in mind.

taylorotwell:
What is the actual solution here @Jubeki

Jubeki:
@taylorotwell 

XAMPP is often used with Windows and the only alternative would be Laragon (which uses MySQL 5.7 so a similar Problem).

MariaDB 10.4, 10.5 and 10.6 are still supported for up to two years (see https://endoflife.date/mariadb).

There are two options in my opinion without reverting the default collation to the older one:

1. Contribute to XAMPP / Laragon: Raise the versions to MySQL 8.0 or MariaDB 10.11, not sure if these changes will be accepted.

2. While installing a fresh application or running migrations detect the used Database Version and make approbiate collation changes (or provide needed changes), while also informing the developer about theses changes. The biggest Problem with this is, that the development environment and production environment can be different. If a Developer doesn't know which database version is used locally or simply uses a different version than in production. Changing the collation can simply be done in the `.env`


I would have loved to recommend Laravel Herd when it is released for Windows, but the database service is only available in the PRO version.

taylorotwell:
@Jubeki can we just revert the MariaDB collation back to what it was for Laravel 10?

Jubeki:
`null` should probably be a valid option, I will test if it works with MariaDB 10.4 (using Docker)

taylorotwell:
Thanks - let me know

Jubeki:
@taylorotwell Switching to null is not an option, because it will result in `latin1_swedish_ci` which is worse than `utf8mb4_unicode_ci`

Laravel 10 used `utf8mb4_unicode_ci`, so reverting to that is an option.

It would be interesting to know, what people use in production and/or local development.

> [!Note]
> This PR currently reverts the collation for MySQL and for MariaDB.

taylorotwell:
Updated MariaDB default.

taylorotwell:
Let's wait to see if we get issues with the 5.7 thing.

stasadev:
I see `Unknown collation: 'utf8mb4_0900_ai_ci'` for MySQL 5.7 with `DB_CONNECTION=mysql`.

If the docs states that Laravel 11 supports MySQL 5.7+ OOTB, it is better to revert it too.

grooverdan:
Yep, XAMPP (https://github.com/ApacheFriends/xampp-build/issues/17) appears to be moving slowly. and Laragon development appears to have stalled.

> It seems in the future MariaDB wants to use the short name `uca1400_ai_ci` for multiple collations.

Technically multiple character sets, but yes, short names ensure accidental character set miss-match doesn't occur.

> The full collation name with the charset prefix is still available:

I did some prototype wordpress work on pulling a modern collation from what was available here: https://github.com/grooverdan/wordpress-develop/commit/86500d31df7a7b10d417b630b90768f11a85c408#diff-101f4101613ae810fc8e54675d9437f9aed83f89dfe2bd3068b2ed2964eece03R896-R938 (that I should get back to, feel free to use).

Apart from that, not much I can add that you haven't covered already, uca1400 isn't accessible on < 10.10.1 MariaDB versions.

[found another user in distress](https://stackoverflow.com/questions/78154387/laravel11-mysql-migration-problem-with-collation-general-error-1273-unknown-c).

PR #6372, opened 2024-03-14
[11.x] Revert collation change

driesvints:


Smef:
I'm noticing this change now. What was the reason for changing the default back to the old collation and not using the MySQL 8 default?

The default collation for MySQL 8 is `utf8mb4_0900_ai_ci`.
https://dev.mysql.com/blog-archive/mysql-8-0-collations-migrating-from-older-collations/

There was an old PR where this was changed, and this one reverts that change. In this thread there's discussion about having a separate MariaDB connection, which does exist. https://github.com/laravel/laravel/pull/6239

PR #6378, opened 2024-03-16
[11.x] Add DB_ENGINE config

mhfeizi:
This adds DB_ENGINE environment variable.


MrPunyapal:
this will be helpful ☑️

balu-lt:
It should be just `env('DB_ENGINE')` because `null` is already there by default.

saturnphp:
yeah good idea

mhfeizi:
https://github.com/laravel/framework/pull/50620

PR #6379, opened 2024-03-17
fix/annoying-scrollbar

neokofg:
This pull request addresses the issue of an unwanted scrollbar on the right side of the screen. To resolve this, the bottom padding from the Laravel and PHP versions was removed, and a max-height was added to the images.


jrd-lewis:
Just my two cents: I think the `pb-16` class should be `pt-16` since it's more common for `pb` utility classes to apply bottom padding and `pt` classes to apply top padding.

PR #6387, opened 2024-03-28
[11.x] Make meta color scheme follow user theme preference

ericmp33:
Minor contribution here! I think we should use color scheme to suit the user's theme preferences. In this case, for the welcome view, if user prefers dark, scrollbar will be dark.

| Before  | After |
| ------------- | ------------- |
|   |   |


perisicnikola37:
Seems nice to me

Bosphoramus:
Seems a little bit weird to me that this PR is being closed when i previosuly had this PR merged to the Laravel docs which is similar
https://github.com/laravel/laravel.com/pull/246

PR #6388, opened 2024-03-30
update users table migration

larschinkel:
make the sessions table name dynamic depending on the SESSION_TABLE environment variable if stated like it already is inside the sessions config


Nidacoke:
29d1d41af57beb13ed8c0f8e058368e3e58dec4f

antonkomarev:
This would not work, because as soon as config files are cached — env method always will return null

msamgan:
yes, i was about to comment that, this has to be updated from config("file.key") for it to work. @larschinkel

PR #6389, opened 2024-04-03
Fix PHPUnit constraint

szepeviktor:
there was a BC break in PHPUnit https://github.com/sebastianbergmann/phpunit/issues/5690

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "laravel/sail": "^1.26",
         "mockery/mockery": "^1.6",
         "nunomaduro/collision": "^8.0",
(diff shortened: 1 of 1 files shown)


szepeviktor:
...resulting in

> Non-readonly class Illuminate\Testing\Constraints\ArraySubset cannot extend readonly class PHPUnit\Framework\Constraint\Constraint

driesvints:
Thanks! Normally this constraint is already pulled in but can't hurt to have as the default.

szepeviktor:
Thank you for merging it.
I encourage developers to run tests with lowest/locked/highest dependencies.

PR #6390, opened 2024-04-07
[11.x] Using once for password in UserFactory.php

devman1917:
improve syntax with same result
image


taylorotwell:
Not quite the same behavior.

PR #6397, opened 2024-05-03
LOG_DEPRECATIONS_CHANNEL is set to stack in .env.example

nastoychev:
Setting `LOG_DEPRECATIONS_CHANNEL=stack` instead of `LOG_DEPRECATIONS_CHANNEL=null` is beneficial because it ensures that deprecation warnings and errors from PHP or packages are logged, helping to catch potential issues early on and avoid overlooking them during the initial stages of development with Laravel.


driesvints:

nastoychev:
> Feel free to do this in your own app 👍 

I've successfully implemented it, and now I can track warnings in my laravel.log file right from the inception of development.

PR #6404, opened 2024-05-22
Use str helper function for app locale formatting

joshbonnick:
This PR replaces the use of `str_replace()` for formatting the app locale in `welcome.blade.php` with the `str` helper / strings API version.


taylorotwell:
I think it's fine!

PR #6406, opened 2024-05-27
[11.x] Removes `--dev` dependencies

nunomaduro:
This pull request adjusts `--dev` dependencies.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
     "license": "MIT",
     "require": {
         "php": "^8.2",
-        "laravel/framework": "^11.0",
+        "laravel/framework": "^11.9",
         "laravel/tinker": "^2.9"
     },
     "require-dev": {
@@ -15,8 +15,7 @@
         "laravel/sail": "^1.26",
         "mockery/mockery": "^1.6",
         "nunomaduro/collision": "^8.0",
-        "phpunit/phpunit": "^11.0.1",
-        "spatie/laravel-ignition": "^2.4"
+        "phpunit/phpunit": "^11.0.1"
     },
     "autoload": {
         "psr-4": {


Jubeki:
I am not sure if it is a good idea, to move all dev dependencies into the framework.

How about creating a new package named e.g. `laravel/dev` which can be added to the skeleton, where you manage dependencies such as the error page or collision, etc.

So that the framework focus on the production/staging environment and dev focuses on local development, debugging and testing.

datlechin:
> How about creating a new package named e.g. `laravel/dev` which can be added to the skeleton, where you manage dependencies such as the error page or collision, etc.
> 
> So that the framework focus on the production/staging environment and dev focuses on local development, debugging and testing.

Love this idea

driesvints:
@Jubeki think you misunderstood this because that's not what this PR does

Jubeki:
@driesvints yeah, I know that this PR only removes the dependency to Spatie Laravel Ignition. But in combination with the already merged PR https://github.com/laravel/framework/pull/51261 the "Dev Dependency" was moved into the framework. (Which means it is also installed in production environments)

Back to the point I wanted to make, you could merge collision and ignition into a new laravel/dev package, instead of maintaining the code in framework. (Splitting production / staging dependencies and dev dependencies).

sheadawson:
Can I ask why ignition was removed? This symphony exception page that now shows has zero helpful information on it and it took quite a while for me to work out that it was because ignition has been removed. Is the symphony exception what I'm supposed to be seeing? I notice it only shows if the error occurs when stemming from a blade template context - I do see the nice new laravel error page with dark/light mode options when blade templates are not involved...

Screenshot 2024-08-05 at 9 02 32 PM

Jubeki:
@sheadawson this is not the error page you should see if you run the latest laravel/framework version. Because the Error Page was moved into the framework. What versions are you running on?

Screenshot 2024-08-05 at 11 13 43

Jubeki:
@sheadawson or can you share a repository, where the symphony error page is shown (saw to late that you edited your comment).

PR #6407, opened 2024-05-28
Updating BCRYPTS ROUNDS from 4 to 12 in phpunit.xml file

yatharth-vataliya:
I am updated one of my project from 10.x to 11.x and after updating everything works fine except pest tests.
Pest tests are failing with error mentioned in screen shot
 
image

and I found a solution on internet that there is mismatch in BCRYPTS_ROUNDS in .env.example and phpunit.xml and that  difference is still in 11.x branch and because of that difference pest test cases are failing and that's why I  am submitting  this pull request.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -20,7 +20,7 @@
     <php>
         <env name="APP_ENV" value="testing"/>
         <env name="APP_MAINTENANCE_DRIVER" value="file"/>
-        <env name="BCRYPT_ROUNDS" value="4"/>
+        <env name="BCRYPT_ROUNDS" value="12"/>
         <env name="CACHE_STORE" value="array"/>
         <!-- <env name="DB_CONNECTION" value="sqlite"/> -->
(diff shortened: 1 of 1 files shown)


GrahamCampbell (changes requested):
I don't believe this change to be correct. The tests intentionally use a different value. The .env.example file is not for running the tests.

Jubeki:
Take a look at your factories:
https://github.com/laravel/laravel/blob/b6d55576d1c9fcf15adebc76bcb6d8cb476a4418/database/factories/UserFactory.php#L14-L33

That is uses Hash::make() so that the correct configuration is applied. Or do a check that you use a value with 4 rounds in testing while still using 12 rounds in local or production.

yatharth-vataliya:
@Jubeki I found the problem and that is with `'password' => static::$password ??= Hash::make('password'),` where I am using hashed password instead of doing Hash::make('password') which causes the error. And if we want to use that approach then we can solve error by just bumping BCRYPT_ROUNDS to 12 in phpunit.xml file.

PR #6426, opened 2024-06-20
Add or update the Azure App Service build and deployment workflow config

simran1358:


TENIOS:
Please, @simran1358, stop doing that.

PR #6428, opened 2024-07-01
fix: password hashing in UserFactory

aifodu:
The User model already casts password as `hashed`

Hashing it in the factory again results in a repetition which makes the default password unusable.

This is resolved by not hashing password in the factory.


driesvints:
Please see previous attempts.

LisandroMoura (commented):
its ok

PR #6430, opened 2024-07-10
Unquote `null`

u01jmg3:
- As it stands, [`session.php`](https://github.com/laravel/laravel/blob/4ef5e2f89e987f84b33b62f79e96485dcaa8f209/config/session.php#L197-L199) suggests `null` as an unquoted option.
- I realise I've updated suggested **driver** options but should they too be unquoted? Thought it was worth bringing to your attention for you to take a look. I may be wrong.

https://github.com/laravel/laravel/blob/4ef5e2f89e987f84b33b62f79e96485dcaa8f209/config/session.php#L197-L199

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -27,7 +27,7 @@ return [
     | same cache driver to group types of items stored in your caches.
     |
     | Supported drivers: "array", "database", "file", "memcached",
-    |                    "redis", "dynamodb", "octane", "null"
+    |                    "redis", "dynamodb", "octane", null
     |
     */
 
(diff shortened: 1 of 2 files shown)


taylorotwell:

u01jmg3:
This could have been a learning moment, but I've been given the dreaded autofill response instead. 😞

GrahamCampbell:
This is not the same thing. The name for these drivers is actually the string `'null'`. The session config is not naming a driver, it's saying to use the browser's default.

GrahamCampbell:
https://github.com/laravel/framework/blob/213a370b703592587bafcd52d38a0ad772ff7442/src/Illuminate/Queue/QueueServiceProvider.php#L116-L121

u01jmg3:
Thanks :v:

PR #6437, opened 2024-07-30
[11.x] Add SESSION_SECURE_COOKIE to .env.example

fabricecw:
With the introduction of the new configuration skeleton in Laravel 11, the `SESSION_SECURE_COOKIE` setting can easily be overlooked. However, enforcing secure cookies is crucial for maintaining security in production environments. This PR adds the `SESSION_SECURE_COOKIE` key to the `.env.example` file to increase its visibility and encourage developers to configure it appropriately when going from local to productive environments.

The default value shall remain `false` for local development convenience and to avoid breaking changes.


taylorotwell:
Can add to your own app if you wish.

valorin:
@taylorotwell Why can't this live in `.env`? It's an important security setting that **should be enabled** any time you deploy anywhere other than localhost. Compared to some of the other options in `.env` which many apps will never need to touch, this is one that every app should be setting when they deploy.

I would even argue it should be enabled by default. The dev can manually disable it if they have non-HTTPS local dev, and `php artisan serve` could probably automatically disable it. Tools like Herd make local HTTPS dev trivial, and browsers often require HTTPS for some of their features to work.

fabricecw:
> Can add to your own app if you wish.

While we already add this key via our installation scripts, I think it's important to consider the broader impact of having `SESSION_SECURE_COOKIE` absent from the default configuration.

`SESSION_SECURE_COOKIE` is essential for every application moving from a local to a production environment to ensure cookies are transmitted securely over HTTPS. Its absence in the default configuration may lead to vulnerabilities. I'm confused a key like `SESSION_DOMAIN` lives in the skeleton but not this one.

By adding this key, we can proactively increase security awareness and help our community to adhere to best practices, thereby reducing the risk of insecure deployments.

valorin:
> Thanks @valorin for sharing the same thoughts. Regarding the default value `true`, I think the trade-off between security and dev convenience needs some more considerations.

Fair point.

I think a suitable compromise would be what Symfony does with `auto`, which sets `Secure` on HTTPS requests and leaves it off for HTTP. That shifts the responsibility to HTTPS onto the web server / environment and serves up the best cookie regardless.

fabricecw:
Makes sense. If @taylorotwell comes back to this PR I will start working on the implementation in the framework.

valorin:
@fabricecw I don't think Taylor checks closed PRs, so I'd suggest either resubmitting a new PR with this change and a more detailed description, or making the PR with the `auto` feature and submitting that for his consideration.

I'm happy to help out with either if you'd like my help. 🙂 
[Reach out directly](pinkary.com/@valorin) or make a draft PR and ping me.

PR #6439, opened 2024-08-13
Remove axios due to high severity security issue

kejedi:
Axios is a dangerous package to include by default.

Here is the latest NPM audit report:

```
# npm audit report

axios  >=1.3.2
Severity: high
Server-Side Request Forgery in axios - https://github.com/advisories/GHSA-8hc4-vh64-cxmj
fix available via `npm audit fix --force`
Will install axios@1.3.1, which is a breaking change
node_modules/axios

1 high severity vulnerability
```

We don't need axios, we can use native `fetch`.

This also cleans the `resources` folder, as the extra `bootstrap.js` file was always confusing. Now the `app.js` file is empty just like `app.css`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
         "build": "vite build"
     },
     "devDependencies": {
-        "axios": "^1.6.4",
         "laravel-vite-plugin": "^1.0",
         "vite": "^5.0"
     }
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -1,1 +1,1 @@
-import './bootstrap';
+
--- a/resources/js/bootstrap.js
+++ b/resources/js/bootstrap.js
@@ -1,4 +1,0 @@
-import axios from 'axios';
-window.axios = axios;
-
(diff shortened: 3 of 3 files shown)


GrahamCampbell (changes requested):
This change is incorrect. The latest version of axios has no known vulnerabilities.

kejedi:
> This change is incorrect. The latest version of axios has no known vulnerabilities.

Ok I guess npm audit is wrong and the open PR currently in the axios repo to fix the issue is there for no reason.

GrahamCampbell:
I believe the CVE was initially incorrectly filed without an upper-bound, and then your npm has cached the bad entry.

kejedi:
> I disagree this is high risk

My cats breath smells like cat food

taylorotwell:
Will keep an eye on this but don't see an emergency need to remove axios as the library is still maintained and receiving security fixes.

laserhybiz:
v1.7.4 released

kejedi:
> v1.7.4 released

yup, good on you for the PR.

glad they resolved the issue quickly and Taylor was right to hold out in favour of a security patch.

PR #6442, opened 2024-08-20
Update welcome.blade.php

JordanchoEftimov:
In modern web development, the HTML5 convention of omitting the self-closing slash (/>) is generally considered better practice.

Also, changed it for better consistency of the link tags, since the first tag is without `/` and the second tag was with `/`.


xtrasmal:
Can you explain why it is considered better practice? What are these considerations?

driesvints:
I don't think this is really needed.

Motifsky:
I prefer to always add the slash. Because, it provides a visual clue of non-closing tags.

PR #6443, opened 2024-08-22
[11.x] fix: wrong redirect from POST to GET

SocolaDaiCa:
when I POST domain.com/api/v1/users it ok,
but when I POST domain.com/api/v1/users/ it alway redirect to GET domain.com/api/v1/users (without last slash), It crash my request.

I think feature `redirect Trailing Slashes If Not A Folder` should only work for GET method, not POST PUT PATCH DELETE

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -10,6 +10,7 @@
     RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
 
     # Redirect Trailing Slashes If Not A Folder...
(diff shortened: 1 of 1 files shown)


taylorotwell:

SocolaDaiCa:
@taylorotwell but it's a .htaccess config issue, not code related so I can't package it

decadence:
@SocolaDaiCa it's default answer from Taylor. Just add this line in your own application.

PR #6445, opened 2024-08-26
Add AWS_ENDPOINT for SQS queue

stasadev:
- https://github.com/aws/aws-sdk-php/issues/2846
- https://github.com/aws/aws-sdk-php/issues/2947

After SQS migrated from query to JSON protocol, Laravel lost the ability to change the host with `SQS_PREFIX`.

To reproduce:

1. `composer require aws/aws-sdk-php`
2. set this in `.env`:
	```dotenv
    QUEUE_CONNECTION=sqs
	SQS_PREFIX=http://foobar/000000000000
	```
3. try to run a job, see this error (the endpoint is not overridden):
    ```
	Error executing "SendMessage" on "https://sqs.us-east-1.amazonaws.com";`
	```
	It worked fine with SDK <3.286.2:
	```
	composer require aws/aws-sdk-php:3.286.1
	```
	With 3.286.1 you see what you expect:
	```
	Error executing "SendMessage" on "http://foobar/000000000000/default"
	```

This PR adds a missing endpoint (see related issues for further explanation).

### DIFF
--- a/config/queue.php
+++ b/config/queue.php
@@ -60,6 +60,7 @@ return [
             'queue' => env('SQS_QUEUE', 'default'),
             'suffix' => env('SQS_SUFFIX'),
             'region' => env('AWS_DEFAULT_REGION', 'us-east-1'),
+            'endpoint' => env('AWS_ENDPOINT'),
             'after_commit' => false,
         ],
 


driesvints:

stasadev:
Quoting AWS PHP SDK maintainer:

> SQS recently migrated from query to json protocol, which also required changes from 3rd party SQS implementations

https://github.com/aws/aws-sdk-php/issues/2846#issuecomment-1832330363

---

> Feel free to add this to your own app.

Just because no one has raised this issue in 8 months doesn't mean that no one needs it but me.

I think it would be useful to add because SQS has worked with `SQS_PREFIX` for years, but now it doesn't work as expected.

Please consider changing something here, even if you don't want to add the `endpoint`, as the following `SQS_PREFIX` gives false hope that it can be used to override the host (but this doesn't work with AWS PHP SDK 3.286.2+):

https://github.com/laravel/laravel/blob/c12fd185e64b6fd652243e06f290f438164ddde5/config/queue.php#L59

PR #6448, opened 2024-09-03
Adding some comments for service provider

coder-bhavesh-here:
- Adding return type in docblock
- Added comments which can guide how to bind a service to the container and how to register routes or event listeners


driesvints:
Thanks but no plans for this sorry.

PR #6451, opened 2024-09-12
[11.x] Add PHPStan / Larastan

bert-w:
I was surprised that the boilerplate application did not include PHPStan (Larastan) by default. We should point developers towards the right direction to write correct and working code, and this is one way that helps with that.

I also wanted to add the `routes/` directory to the `phpstan.neon` paths (since Laravel 11 people will now start dumping application code into these files as well, so it wouldn't be bad to have it included in the phpstan scan also), but it introduced an error with the current placeholder code in `routes/console.php`:

```php
})->purpose('Display an inspiring quote')->hourly();
```

The `$this` context is undefined here and I can't properly typehint it using phpstan.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -9,7 +9,11 @@ use Illuminate\Notifications\Notifiable;
 
 class User extends Authenticatable
 {
-    use HasFactory, Notifiable;
+    /**
+     * @use \Illuminate\Database\Eloquent\Factories\HasFactory<\Database\Factories\UserFactory>
+     */
+    use HasFactory;
+    use Notifiable;
 
     /**
      * The attributes that are mass assignable.
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,7 @@
     },
     "require-dev": {
         "fakerphp/faker": "^1.23",
+        "larastan/larastan": "^2.9",
         "laravel/pint": "^1.13",
         "laravel/sail": "^1.26",
         "mockery/mockery": "^1.6",
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,0 +1,18 @@
+includes:
+    - vendor/larastan/larastan/extension.neon
+
+parameters:
+
+    paths:
+        - app/
+
+    # Level 9 is the highest level
+    level: 7
+
+#    ignoreErrors:
+#        - '#PHPDoc tag @var#'
+#
+#    excludePaths:
+#        - ./*/*/FileToBeExcluded.php
+#
+#    checkMissingIterableValueType: false


phpstan.neon
```diff
@@ -0,0 +1,18 @@
+includes:
+    - vendor/larastan/larastan/extension.neon
+
+parameters:
+
+    paths:
+        - app/
+
+    # Level 9 is the highest level
+    level: 7
```
bert-w, line 10: Note that `level: 9` also passes, but it is a real pain to maintain this level for the average application. Anywhere between 5-7 seems acceptable to me.

taylorotwell:
Curious to get @nunomaduro's thoughts on this.

phpstan.neon
onairmarc: It might be worth adding `"phpstan/extension-installer": "^1.3"` to the `require-dev` block so that way all PHPStan/Larastan Extensions are automatically discovered during `post-autoload-dump` . Adding this would remove the need for lines 1 and 2 here.

jorqensen:
Came here from the [Twitter poll](https://x.com/taylorotwell/status/1835762519522082931), but wanted to elaborate on my vote for no.

I personally think this is way too intrusive to add as a default to the app skeleton. This will most likely result in some bad experiences for new-comers, or even people who are not using static analysis today.

I get that @bert-w has good intentions with this PR, but I think static analysis should be **opt in** not **opt out**.

If anything, could we not consider a `php artisan install:analysis` command, that sets up a "sensible default"? Sanctum also used to be default, but is now moved to the `install:api` command.

mrpritchett:
> Came here from the [Twitter poll](https://x.com/taylorotwell/status/1835762519522082931), but wanted to elaborate on my vote for no.
> 
> I personally think this is way too intrusive to add as a default to the app skeleton. This will most likely result in some bad experiences for new-comers, or even people who are not using static analysis today.
> 
> I get that @bert-w has good intentions with this PR, but I think static analysis should be **opt in** not **opt out**.
> 
> If anything, could we not consider a `php artisan install:analysis` command, that sets up a "sensible default"? Sanctum also used to be default, but is now moved to the `install:api` command.

1000x this.

simivar:
@jorqensen out of curiosity, how can this be of burden to new comers if in order to run the analysis you still need to run the  phpstan analysis command anyway?

kauffinger:
> @jorqensen out of curiosity, how can this be of burden to new comers if in order to run the analysis you still need to run the phpstan analysis command anyway?

For example, it might be well integrated into their IDE if installed and will show warnings by default. Even level 5 is, in my opinion, already too much for somebody new to Laravel/PHP to be burdened by.

canvural:
@jorqensen It is opt in already in my opinion. Developers have the choice to run static analysis or not. But if they choose to run, this will help them get started faster.

Motifsky:
Opt in if desired, not built in. Unless the Laravel Team feels this is THE way to go.

jorqensen:
> @jorqensen out of curiosity, how can this be of burden to new comers if in order to run the analysis you still need to run a separate command (phpstan analysis) anyway?

> @jorqensen It is opt in already in my opinion. Developers have the choice to run static analysis or not. But if they choose to run, this will help them get started faster.

Valid point. Though as @kauffinger mentioned, it might be integrated and checking errors from your IDE by default. I am not sure if this is the case in PHPStorm, as I am a VSC user myself.

Either way, personally I think an installation command makes more sense for a complete opt in. It would go nicely with the currently existing install commands like `install:api` and `install:broadcasting`. Sounds like a win for both sides. 🤷‍♂️

If this becomes a default, it'll probably be *okay*. Just taking the time to raise concerns, rather than having it be an afterthought.

RobertBoes:
> @jorqensen out of curiosity, how can this be of burden to new comers if in order to run the analysis you still need to run a separate command (phpstan analysis) anyway?

You say including this would help people writing correct and working code, then that statement falls in the same realm as it being a burden.

And arguably, this wouldn't result in people writing better code, as PHP is just loosely typed, so eventually lots of it comes down to comments especially higher levels of phpstan. The first step would be writing those comments correctly, if that's not done properly it would introduce a false sense of security.

bert-w:
Don't listen to the hobbyists @taylorotwell ! 🐤

djaiss:
You've moved away from Spatie's exception handler so you are not dependant of third party tools. Why should we go back on this?

Also, why would a framework take the responsibility to include a static analysis tool? I love Laravel to be opinionated, but not **that** opinionated.

CoolGoose:
For what it's worth. I think this should be added in since it's not forcing anyone to use it.
If your IDE is smart enough to figure out phpstan, so be it, you can always silence those.

The number of errors I saw because people just forget to do basic things like returning a response in a controller is too high for comfort 🙈 . 

My only concern is if this shouldn't cover phpstan and psalm and you can pick either or or both

bert-w:
In all seriousness, this is the template application from which people build their own projects. It would just be another sensible default besides `laravel/sail` and `laravel/pint`. It is easily removable if you don't plan on using it, so imho it would only provide a benefit.

Of course the same reasoning can be used to _not_ add it (but then new programmers wouldnt know about it so easily).

jasonmccreary:
@bert-w, _Sail_ and _Pint_ are first-party packages. That's a big difference - Laravel is committed to their upkeep.

cleptric:
A good middle ground could be to add `phpstan/phpstan` to [`suggest`](https://getcomposer.org/doc/04-schema.md#suggest) in the `composer.json`.

bert-w:
> @bert-w, _Sail_ and _Pint_ are first-party packages. That's a big difference - Laravel is committed to their upkeep.

Hm so would you consider `larastan` a "second-party" package? I'm all for the best tool for the job really. It doesn't have to 100% depend on Laravel core developers.

You have a good point for any major Laravel upgrade, but then again there is lots of php typing yet to be added to the framework itself (which will probably take years) so the transition won't be easy anyway.

Personally I would vouch _against_ an install command, since in that case you might as well run `composer require larastan/larastan` and copy over the default `phpstan.neon` file (its just very little work).

composer.json
```diff
@@ -11,6 +11,7 @@
     },
     "require-dev": {
         "fakerphp/faker": "^1.23",
+        "larastan/larastan": "^2.9",
```
ondrejmirtes, line 14: phpstan/phpstan should be added here too. Larastan is “just” a PHPStan extension, the main thing you’re writing the configuration against and what you’re running is PHPStan so the user should be aware they’re using it and in what version.

benjibee:
There's an argument to be made for starting a new project with PHPStan / Larastan from the get-go. Adding it later on is both a pain and a boon, but writing code with this feedback from the start can save a lot of headaches and lead to a much more stable codebase.

phpstan.neon
ondrejmirtes: Personally I’d choose level 5 for Laravel apps.

phpstan.neon
```diff
@@ -0,0 +1,18 @@
...
+
+#    ignoreErrors:
+#        - '#PHPDoc tag @var#'
+#
+#    excludePaths:
+#        - ./*/*/FileToBeExcluded.php
+#
+#    checkMissingIterableValueType: false
```
ondrejmirtes, line 18: Remove this line, this option is already deprecated.

ondrejmirtes (commented):
Just a few of my cents as PHPStan author :)

michael-rubel:
Just do it. Enforcing static analysis on the framework level is a good thing. It's a dev dependency like pint or faker, and if someone doesn't want it for some reason, he/she will remove it in the initial stage of the project bootstrapping.

colincameron:
I would see analysis in the same vein as tests - a lot of developers produce good, working code without it, but it's an important tool to produce reliable, safe code.

Laravel should adopt best practices by default - PHPUnit/Pest is there ready to use but you're free to ignore it, so why shouldn't PHPStan/Larastan? Inclusion in the default installation doesn't harm anyone or force anyone to it, but it makes these tools more accessible to developers who might not have considered using them.

phpstan.neon
michael-rubel, line 10: Consider enabling [Octane Compatibility](https://github.com/larastan/larastan/blob/e1b64a0f3120cf40fca9703a718e29b0cef89f9f/src/Rules/OctaneCompatibilityRule.php) check from Larastan by default (or at least leave it as a comment, so people know it exists).

nikuscs:
I agree level 5 would be more than welcome, as this would prevent some really basic mistakes for new users or even users that rushed the code at some point. Plus i believe Laravel 11 already brings some nice generics, and i would love to see more of those and more types in the next versions! ✅

phpstan.neon
ondrejmirtes: Levels 6 and up are more for "type enthusiasts" I'd say. Level 5 strucks a great balance for being useful even for the "uniniated".

Level 5 is about "I can't believe I made this stupid mistake".

Level 6 is about "Why does it want me to add a typehint here!?"

Of course we're just talking about the default here. Anyone is free to change it in their app.

michael-rubel: Don't you think a decent app must have at least basic typing? I'd enforce level 6 to make future codebases cleaner. It's not that much of an effort to write a type hint. Those are not generics after all. 😄

filljoyner:
I like the idea but don't like the idea of it baked in. Just thinking, ultimately, where is the line for things that get baked in to an installation and things that can be optionally installed.

IMHO it could be a useful addition to the Laravel installer and a nice addition as a console command to install. Having a console install command sets a good pathway for additional community recommendations. If you want to use it, install/configure by a handy console command.

dillingham:
I think with pestphp.com being the default testing framework.. it opens the consideration that if it had it's [type coverage package](https://pestphp.com/docs/type-coverage) installed by default so then would laravel and any new features to it would also come to the framework.. emphasizing pest as the testing enviroment for all things testing

pushpak1300:
While the concept is interesting, I believe it should be an optional installation rather than an automatic one. When I'm just hacking around on a project, I don't want to deal with fixing Larastan issues in my application. I know I can switch it off anytime, but that adds some friction to the overall tinkering process in Laravel, which I'd rather avoid.

Fludem:
## Why I (an idiot) wouldn't add PHPStan by default

### The silent majority's code is absolute slop (I'm the worst dev btw, probably the lowest IQ in this thread)

Most of the commercial products built with Laravel and any other language/framework look like they've been obfuscated, and it's that bad. You can buy some scripts online for a self-hosted Imgur alternative, for example, from some well-known sites, open the code base, and realise that it has more vulnerabilities than lines of code.

This shows that it's not just the beginners writing slop, as we all have when we were beginners, but actual full-time developers shipping to thousands of servers.


### Increases perceived difficulty of learning Laravel

PHPStan might intimidate newbies who are trying to learn and make the mountain they mountain they want to climb look steeper.

Making PHPStan more prevalent when someone is a beginner highlights their knowledge gaps and makes them overly aware of how much they don't know too soon. The more you do this, the more people you scare away.

As an anecdotal example, at a company I'm working with, we reduced the churn and support load massively by hiding the advanced settings unless the user required them.

We all know nearly anyone can learn how to develop software, but one of the core things they need is self-belief. Adding more unnecessary complexity too soon will chip away at that self-belief.

## My Crap Idea

Adding an argument to the CLI for Installing the framework that shows the more advanced options should be added instead, hiding the complexity from those who probably don't need to see it.

It also means we can reduce the amount of commands involved when setting up a project, as this could be used for more than PHPStan.

I believe there's a CLI installer using Laravel Prompts anyway, I've just enslaved herd for too long.

Fludem:
We could also just put PHPStan at level 9, and add rector with everything turned on. Then, make the serve command run both before serving to confuse newcomers.

This would lower the amount of Laravel developers over time, thus artificially creating scarcity so we can all get paid more

michaeldyrynda:
I think making it easily-installable as the "blessed approach" to static analysis is the way to go.

Ensuring that the skeleton adheres to and is compliant with whatever the appropriate level - probably level 5 as a balance between "this is helpful" and "this is needlessly pedantic" - should be the way to go.

Laravel by and large wants to help people ship faster, and reduced what is in the skeleton app to a bare minimum over the years. Anything that you'd need to add in aid of optional static analysis could start to see some of that leaking back into it. Docblocks and comments added so satisfy the analysis tool, for example, would be a negative for the cleanliness of the default install.

It's only then that you start considering obtuse error messages, having to add docblocks or `@var` annotations to guide the static analysis tooling. It can lead to much overwhelm when onboarding newcomers.

I don't think "safe" or "correct" code at the expense of working code is a good approach, simply because we have well over a decade of documentation on how to appropriately test code to ensure it does what it should, where static analysis is likely to come in and start introducing "errors" for otherwise working code.

princejohnsantillan:
Maybe as an option during Laravel install. The same way pest is installed only when we choose it.

shirkerohit:
My two bits,

I like the core idea of having quality checks in place but I think it shouldn't be enforced. Like I will be too overwhelming for someone who starts freshly and all of sudden haunted by 100's of warnings and errors. Plus, mostly if someone is totally unaware of the tool like phpstan this will make them more overwhelming to start with the framework itself.

A better approach like others mentioned is let people choose it at the time of installation or have commands to later add it when needed like the starter kits.

FoksVHox:
If Laravel were to add Larastan & PHPStan, then it should be an opt-in via a command, and not shoved down your throat per default. Laravel did a bold move when they released Laravel 11, with the skeleton slimming. However, they found a perfect middle ground by adding install commands.

colincameron:
@Fludem 

> Most commercial products built with Laravel and **any other language/framework** look like they've been obfuscated, and it's that bad. You can buy some scripts online for a self-hosted Imgur alternative, for example, from well-known sites, open the code base, and realise that it has more vulnerabilities than lines of code.
> 
> This shows that it's not just the beginners' writing slop like we all did when we first started, but actual full-time developers shipping to thousands of servers.

I'd say that this is a great argument for including PHPStan. Bad code should be harder to write. We're all dealing with users' data in some form - maybe we should be forced to take better care of it.

troccoli:
For what is worth I am in favour of adding it as default. Laravel is highly opinionated, so why not force you to write better code? A command to opt-in would be pointless IMO, as it's very easy to install it anyway.

phh:
I recently got onboarded on a project without PHPStan from the get-go. Starting with the baseline and level 5 helped us discover new and potential errors.

At level 6 we were "forced" into defining types, which I personally prefer (and what the best for out project).
Level 5 gave us the most insights to errors in our actual code, while level 6 gave us insights in missing types and errors based on those types.

To me level 5 would be a perfect start for any new project. If you're into types (or know your way around PHPStan), you know how to level 6+ right away.

corazzi:
> @Fludem
> 
> I'd say that this is a great argument for including PHPStan. Bad code should be harder to write. We're all dealing with users' data in some form - maybe we should be forced to take better care of it.

The framework is opinionated on how to do certain things, but all this is going to do is increase friction. I'm all for bad code being harder to write – but for me. That should be my decision to enforce that.

People are going to write bad code, anyway. There's several outcomes from having this in the core that I can see:

**New or inexperienced devs will be put off or confused**, especially if they're just trying the framework or even the language for the first time. Modern PHP is a joy to write, especially compared to 10, 15, 20 years ago. Let people write it, and don't let perfect be the enemy of good.

**People will get annoyed** and just disable it anyway, or complete ignore it altogether, defeating the point and not solving the problem.

**People who already use these tools will be fine** – but it takes 2 minutes to install and configure into a fresh or even existing project.

For what it's worth, I use Larastan and would support it being in the installer as an optional add-on or something, as long as it doesn't get in the way.

colincameron:
I'm not sure making it optional at install-time would achieve anything, as anyone who doesn't want to use it will just never run the analyse command. If it's installed by default - **and importantly, explained in the docs** - then it's an encouragement for writing better code. Exactly the same as PHPUnit or Pest.

michaeldyrynda:
I don’t think it sufficient to document how to run it, simply because running it is the easy part. 

Trying to document all the rules, restrictions, quirks, and whatnot, not to mention it would almost definitely conflict with the existing documentation for what works vs  what is “correct” is a much bigger undertaking than just adding phpstan to `composer.json` and its config. 

Adding it has implications on the Laravel project’s support burden in doing so, beyond the affordance of “good code”.

ralphjsmit:
I would also argue that the entire direction of the Laravel framework the past 1-2 years is to make it easier for beginners to get started. Consider Herd, for Mac and Windows, consider the reduced skeleton, consider Laravel Cloud. Adding in a static analysis tool to a starting framework that sometimes will provide hard-to-understand errors for beginners, would seem to conflict with the supposed goal of making it easier to get started and enjoyable for devs to adopt the Laravel framework.

RobertBoes:
I genuinely feel like these kinds of changes shouldn't be added on a whim, as it has a significant impact.

Ignoring errors when things aren't implemented:
I tried this setup with Laravel Breeze: 3 errors, even on level 5 it generates 1 error. Haven't tried all Breeze flavors or Jetstream, but it would likely yield similar results.
I created a model (with `php artisan make:model`): 1 error.
Created a model with all options (request, controller, policy etc): 8 errors (due to the controller)

Don't get me wrong, I'm not against PHPStan (or other static analyzers), but I don't think static analyzers are some sort of magical thing that would make code better. Arguably on the contrary, as with many things in the framework, such as Eloquent, there's a lot of magic going on and PHPStan simply ignores things like `$user->random_non_existing_column`

app/Models/User.php
```diff
@@ -9,7 +9,11 @@
 
 class User extends Authenticatable
 {
-    use HasFactory, Notifiable;
+    /**
+     * @use \Illuminate\Database\Eloquent\Factories\HasFactory<\Database\Factories\UserFactory>
```
badasukerubin, line 13: This is better defined in the factory itself, for example:
```php
use Illuminate\Database\Eloquent\Factories\Factory;
use Returfs\Marketplace\Shared\Internal\Models\User;

/**
 * @extends Factory<User>
 *
 * @phpstan-import-type SettingDataArray from User
 */
class UserFactory extends Factory
{
```

badasukerubin (commented):
We use PHPStan at level 9 in our projects, and it is fantastic, our code is well-typed, structured, and documented, and we also catch many errors faster in development.
Enforcing PHPStan in the framework is kind of like enforcing TS in a core JS framework, it takes away the freedom of choice for developers. I'll outline some more issues below:

- PHPStan has its learning curve, the robust documentation testifies to this and it may be unnecessary to learn i.e I don't need to learn TS to write good JS.
- We've got a couple of other static analyzers in PHP, e.g. Phan which is also fantastic. How do we handle developers who prefer the others?
- There are also 101 ways to do things in PHPStan (See my review), how do we reconcile this?
Imo, if we are to take this route, we need to ensure that there's a maintainable, generic but standard/defined way to achieve static analysis in Laravel. This PR doesn't do that.

iamlasse:
Why not make it an option where you can select from similar to the option of using Pest or Phpunit? There are a few options out there besides Larastan, it could be good if one could have a choice. Or deselect as well.

taylorotwell:
Going to hold off on this one. 👍

inmanturbo:
> Going to hold off on this one. 👍

Excellent choice to hold off. Personally I would find it convenient. I would also find a .github/ folder with default workflows for pint, phpunit/pest, and larastan pretty convenient. But I don't feel the burden falls on the framework. Some kind of plugin system to add extensibility to the Laravel install command would be pretty neat, then we could have community maintained installers. I don't any see that happening though to be fair, as the Laravel team will likely favor the commercial GUI installer.

andreapollastri:
There are a lot of packages that could be installed by default in Laravel and I don't agree with Larastan by default! A framework has to be agnostic and this choose may scare dummy level developers! Sometimes also the default for user model and auth logic is an issue for me in Laravel so please do not add larastan! In simple projects it could be a pain in the ass!

michaeldyrynda:
> I would also find a .github/ folder with default workflows for pint, phpunit/pest, and larastan pretty convenient.

There is [this repo](https://github.com/laravel/.github) for this kind of stuff if you're interested @inmanturbo

abkrim:
_"La letra con sange entra" (You learn through pain?)_ is a Spanish saying, which refers to the fact that the path of punishment is the one that makes them learn. (A saying from the 16th century)

In my day to day life I struggle with the lack of compliance with good practices of team developers, but I think it is because I live in a country where business practices do not lead to the proper training of programmers (nor to the payment of their adequate salaries) which leads to a generation of programmers who end up becoming code writers, full of speed and demands from above, which leads to bad coding practices.

But at the same time, forcing an entire community to use something, instead of making it "installable" if desired, is not the best way for another generation to advance and love Laravel, and with it learn to program following good practices.

I did it when I was 53, when I started with Laravel 5, and with the little time I have left, I continue learning day by day, and there are still times when I don't have time to keep up to date.

My vote is for a `php artisan install:xxxxxx`

"La letra con sange entra" is a Spanish saying, which refers to the fact that the path of punishment is the one that makes them learn. (A saying from the 16th century)

In my day to day life I struggle with the lack of compliance with good practices of team developers, but I think it is because I live in a country where business practices do not lead to the proper training of programmers (nor to the payment of their adequate salaries) which leads to a generation of programmers who end up becoming code writers, full of speed and demands from above, which leads to bad coding practices.

But at the same time, forcing an entire community to use something, instead of making it "installable" if desired, is not the best way for another generation to advance and love Laravel, and with it learn to program following good practices.

I did it when I was 53, when I started with Laravel 5, and with the little time I have left, I continue learning day by day, and there are still times when I don't have time to keep up to date.

My vote is for a `php artisan install:xxxxxx`

Sorry if I'm not a great developer yet, but I am a big fan of the Laravel ecosystem.

inmanturbo:
> There is [this repo](https://github.com/laravel/.github) for this kind of stuff if you're interested @inmanturbo

@michaeldyrynda Yes thanks. That is not quite what I would be after but thank you. A minimalistic pint.yml, stan.yml and tests.yml to automatically be included as a starting point when running `laravel new` with github options would be nice, but I wouldn't expect it to be handled by the official installer, but rather some extension thereof. I understand also how org dotfiles work in github, but in my case there are a lot of different orgs, and requirements are per project not per org.

nasirkhan:
** It should not be added as a default.** 

At the same time it would be great to have a page on Laravel doc where useful and popular packages will be listed.

composer.json
fylzero: Larastan already requires phpstan as a sub-dependency, so I don't believe there is an additional need to require it here as well.

PR #6452, opened 2024-09-13
Run optimize in tests

JurianArie:
Our testsuite was able to catch the bug in https://github.com/laravel/framework/issues/52779 thanks to running optimize like we do in prod. It might also be beneficial for Laravel to run this in the skeleton testsuite.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,5 +43,8 @@ jobs:
       - name: Generate app key
         run: php artisan key:generate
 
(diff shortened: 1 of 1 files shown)


Jubeki:
This Github Action is only for this repository. It is not included in a fresh Laravel Application, as the `.github` folder is listed in the export ignore list:

https://github.com/laravel/laravel/blob/e0421a4ec94be0044217e83b9c839676337139ae/.gitattributes#L9

PR #6457, opened 2024-10-02
[11.x] add Eloquent strictness

browner12:
Make the Eloquent strictness default in non-production environments.

I think this is a safe, sensible default for new applications, for both experienced and new programmers.  Will help teach new programmers best practices for performance, and avoid bugs caused by silent failures. I'm guessing most experienced programmers are already including this in their app anyway, and it can be easily removed/adjusted as needed.

I'm not sure if this has been discussed before, but I could not find any old PRs regarding this, so figured I give it a go.

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -21,8 +21,8 @@ class AppServiceProvider extends ServiceProvider
     public function boot(): void
     {
         //
-        Model::preventLazyLoading(!$this->app->isProduction());
-        Model::preventSilentlyDiscardingAttributes(!$this->app->isProduction());
-        Model::preventAccessingMissingAttributes(!$this->app->isProduction());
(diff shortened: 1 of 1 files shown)


taylorotwell:
I really hate most of these methods. 😬 They cause such a headache. I think the prevention of silently discarding attributes is the only helpful one.

browner12:
I'd be curious to hear why you hate these methods, especially the lazy loading one. Is it just because people get confused on there use?  I find them insanely useful.

decadence:
I add this in every project except `preventSilentlyDiscardingAttributes`. `preventAccessingMissingAttributes` I use even in production because it helps immediately detect the error, rather than looking for the cause of the strange behavior.

> I'd be curious to hear why you hate these methods

I think this is due to the difficulty of maintaining code with these methods, since sometimes they used to work when they shouldn't.

PR #6460, opened 2024-10-06
Add .htaccess for Improved Hosting Compatibility

brunoribeiro-lab:
This commit adds a new .htaccess file in the root of the Laravel project. It ensures that the application functions correctly in hosting environments without direct access to the vhost configuration. For Apache, it should be DocumentRoot /var/www/laravel/public.

With this .htaccess file, installation on hosting services that do not provide full control over the directory structure is simplified, redirecting all requests to the public directory.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,5 @@
+<IfModule mod_rewrite.c>
+    RewriteEngine On
+    RewriteCond %{REQUEST_URI} !^/public/
+    RewriteRule ^(.*)$ /public/$1 [L]
+</IfModule>


CallMeLeon167:
I tested the whole thing on nginx and got a 403 straight away :)

brunoribeiro-lab:
The .htaccess file is exclusive to Apache and is not compatible with Nginx, which handles URL rewriting directly in the server configuration. The purpose of creating the .htaccess file is to ensure greater compatibility with hosting services that use Apache, making redirection and the proper functioning of the application easier in those environments

GrahamCampbell (changes requested):
This is the wrong place for the .htaccess file. One should **never** point apache at the project root like this. It's critical to only use the public directory, which indeed already has a .htaccess file.

PR #6462, opened 2024-10-08
Update welcome.blade.php to add missing alt tag

mezotv:
Every image should have an alt tag for accessibility reasons, but this one did not, so this commit aims to add the missing one.


mezotv:
While this is a tiny addition I think the example view should folllow best practices and I would consider this to be one!

PR #6463, opened 2024-10-10
Add Tailwind, "composer run dev"

taylorotwell:
This PR does two things...

First, it adds a basic Tailwind configuration out of the box. This lets you start using Tailwind immediately without installing any starter kit. Useful if you just want to mess around or build things from scratch.

Second, it adds a `composer run dev` script, which starts `php artisan serve`, `php artisan queue:listen --tries=1`, `php artisan pail` (now a dev dependency by default), and `npm run dev` all in one command, with color coded output in the terminal using `concurrently`.

CleanShot 2024-10-10 at 09 07 08@2x

This gives you one command to run the application and have all of the basic stuff you need.


masterfiras101:
What about pail in windows

juanborras:
Would it be possible to add --host and --port?

msamgan:
@juanborras, you can add the --port, etc, in the dev command manually. in the php artisan serve section.

AhmedAlaa4611:
I hat pail to be a dev dependency by default!!!

luisprmat:
> What about pail in windows

**laravel pail** and **laravel horizon** require **PHP-PCNTL** extension *Not supported for Windows*. In my PC definitely i'm using **WSL** over Ubuntu 24.04 y work fine!

jrd-lewis:
There needs to be a flag added to the installer to do new installs without having Tailwind installed.

PR #6470, opened 2024-10-15
[11.x] Fixes `pail` command killing other dev commands

nunomaduro:


Shaz3e:
Thanks @taylorotwell and @nunomaduro

PR #6471, opened 2024-10-16
[11.x] npm check in composer run dev command

amdad121:
I installed a fresh Laravel app and tried to run the composer run dev command, but it showed an error saying 'vite not found,' indicating that npm is not installed. I have added a check to see if npm is installed, and if not, it will automatically install npm.

laravel

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
         ],
         "dev": [
             "Composer\\Config::disableProcessTimeout",
-            "[ -f package-lock.json ] || npm install",
+            "[ -f package-lock.json ] || npm ci",
             "npx concurrently -c \"#93c5fd,#c4b5fd,#fb7185,#fdba74\" \"php artisan serve\" \"php artisan queue:listen --tries=1\" \"php artisan pail\" \"npm run dev\" --names=server,queue,logs,vite"
         ]
     },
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -48,6 +48,7 @@
         ],
         "dev": [
             "Composer\\Config::disableProcessTimeout",
+            "[ -f package-lock.json ] || npm install",
```
crynobone, line 51: Should it better to use `npm ci` here?

amdad121: Yes, I think it would be better.

taylorotwell:
Does this file check work on Windows?

PerryvanderMeer:
@taylorotwell Nope.

afbeelding

amdad121:
> Does this file check work on Windows?

Unfortunately, the old code wasn't working on the Windows machine, so I changed the logic. I've now tested it on both my macOS and Windows machines, so it should work on Linux as well.

Screenshot 2024-10-16 212332
Screenshot 2024-10-16 212203
Screenshot 2024-10-16 at 9 37 44 PM
Screenshot 2024-10-16 at 9 38 56 PM

amdad121:
@taylorotwell It's working fine in my machine as i said. Please check out the attached images.

PR #6478, opened 2024-10-22
Inconsistency in Github action names

amirbabaeii:
Test Action starts with a capital letter, while the others don't, so I changed them for better consistency.


.github/workflows/pull-requests.yml
```diff
@@ -1,4 +1,4 @@
-name: pull requests
+name: Pull Requests
```
loveyou001, line 1: Rum

loveyou001 (commented):
Test

PR #6479, opened 2024-10-23
shorten the npm run dev a bit by concurrency shortcut.

msamgan:
The concurrency package provides some shortcuts for npm. npm:dev -> npm run dev. 
It will make the composer run command a little bit shorter, not much, but it might be towards betterment.


taylorotwell:
I think it's ok as is 👍

msamgan:
thanks for the consideration.

PR #6480, opened 2024-10-23
adding Dockerfile and main

elaria-2017:


Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+COPY . .
+
+# Install application dependencies
+RUN echo "which composer"  # Check if Composer is in PATH
+RUN composer install --no-interaction --optimize-autoloader
+
+
+# Set application key
+RUN php -r "file_exists('.env') || copy('.env.example', '.env');"
+RUN php artisan key:generate --force
```
joshmanders, line 36: This should not happen in the docker file.

PR #6483, opened 2024-10-28
Optional PORT & HOST in composer run dev

david-valdivia:
Allows PORT and HOST on laravel serve through composer run dev.

i know that it's too early, but i was thinking in add a port and host optionals on composer run dev, it is an amazing tool and i think it could by more versatil.

Propuse to use:

╰─➤  HOST=localhost  composer run dev // runs on default port
╰─➤  PORT=8082  composer run dev // runs default host port 8082
╰─➤  HOST=localhost PORT=8084 composer run dev // runs on port 8084 and host localhost
╰─➤  composer run dev   // runs without --port and --host params                                                                   

Any ideas to make it better? @msamgan

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
         ],
         "dev": [
             "Composer\\Config::disableProcessTimeout",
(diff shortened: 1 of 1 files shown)


taylorotwell:
You can update this in your own application 👍

PR #6484, opened 2024-11-01
Add schema property to enhance autocompletion for composer.json

octoper:
This PR introduces a quality of life improvement by adding `"$schema": "https://getcomposer.org/schema.json"` to `composer.json` in order to have autocompletion in editors like VS Code, PhpStorm etc.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,4 +1,5 @@
 {
+    "$schema": "https://getcomposer.org/schema.json",
     "name": "laravel/laravel",
     "type": "project",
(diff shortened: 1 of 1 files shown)


octoper:
@mfn  Haven't noticed that as I mostly use VS Code, thanks for pointing this out!

martinbean:
@octoper Thanks! I always add this in new Laravel projects I create, and didn’t even think to PR it 😅

PR #6487, opened 2024-11-08
[11.x] make password resets use the "cache driver" by default

browner12:
my intent with https://github.com/laravel/framework/pull/53428 was to make the Cache driver the default for password resets, as I think it's the better option. if on board, this should not be merged until that PR is released.

wondering if we should add a dedicated cache store by default?

- drop the `password_reset_tokens` migration.
- update the auth config to use the 'cache' driver by default

### DIFF
--- a/config/auth.php
+++ b/config/auth.php
@@ -92,8 +92,8 @@ return [
 
     'passwords' => [
         'users' => [
+            'driver' => 'cache',
             'provider' => 'users',
-            'table' => env('AUTH_PASSWORD_RESET_TOKEN_TABLE', 'password_reset_tokens'),
             'expire' => 60,
             'throttle' => 60,
         ],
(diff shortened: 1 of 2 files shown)


Jubeki:
I am against this change.

There are many applications who clear the cache in between deployments using `cache:clear`

As far as I understand it, this will also purge the password reset tokens.

(Or am I misunderstanding how the cache driver works in this case?)

browner12:
you are correct with the default case. which is why I noted in the original PR that I highly recommend creating a dedicated cache "store" so you aren't clearing these.

I'm definitely open to creating an explicit "store" by default, but I was not sure what "driver" we would pick for that.

Jubeki:
I think creating a different store (not exactly sure what that would imply) would be unneeded complexity for most of the Laravel Apps.

You currently only need one sqlite file for creating a new App (cache is also stored in the database). So no need for further complexity if a simple database table is enough for most use cases in my opinion.

joshmanders:
I think the `store` property addresses this concern and the only real argument against using it (which is far more superior) is "we have to educate people on the new changes" and that's flaky at best.

taylorotwell:
I don't think we should do this.

PR #6493, opened 2024-11-16
Enhance database configuration with additional connections for better…

Rubakumardirushan:
This pull request updates the `database.php` configuration file to include:
- Additional MySQL connection options (`mysql-old`) for compatibility with older MySQL setups.

These changes aim to improve compatibility and offer developers more flexibility in configuring their databases without modifying the default setup.

Please review and consider merging this update into the project. Thank you!

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -62,6 +62,26 @@ return [
             ]) : [],
         ],
 
+     'mysql-old' => [
+            'driver' => 'mysql',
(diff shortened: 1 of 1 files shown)


Jubeki:
The only difference is the charset `utf8mb4` and `utf8` I think a developer should be able to an adjust the configuration to his needs. In my opinion is the current one a sensible default. You can always add your changes to your own application.

PR #6494, opened 2024-11-17
Update 0001_01_01_000000_create_users_table.php

lesichkovm:
This change fixes an issue, where authentication succeeds, but subsequently the user is left unauthenticated.

It happens only when the ID of the user not an integer (i.e. string) and the session driver is set to database (in my case MySQL).

The reason it occurs is the current setup generates a column of type bigint of length 20. As a result the database session driver is unable to update the session entry, and silently fails.

My current workaround is another migration to fix this:

```php
Schema::table('sessions', function (Blueprint $table) {
    $table->dropIndex('sessions_user_id_index');
});
Schema::table('sessions', function (Blueprint $table) {
    $table->string(column: 'user_id', length: 40)->nullable()->index()->nullable()->change();
});
```

But of course its best to be fixed at the origin, other than having to resolve to black magic :)

Also it would be good to have a fix to the silent failure to update the session table, probably add an entry to the logs.

### DIFF
--- a/database/migrations/0001_01_01_000000_create_users_table.php
+++ b/database/migrations/0001_01_01_000000_create_users_table.php
@@ -29,7 +29,7 @@ return new class extends Migration
 
         Schema::create('sessions', function (Blueprint $table) {
             $table->string('id')->primary();
-            $table->foreignId('user_id')->nullable()->index();
+            $table->string(column: 'user_id', length: 40)->nullable()->index()->nullable();
             $table->string('ip_address', 45)->nullable();
             $table->text('user_agent')->nullable();
             $table->longText('payload');


joparara:
While I see your point, it will break the foreign key for existing users that don't have an issue with the current setup. 

I suggest to file a ticket with details on how to reproduce it, then propose the solution forward

crynobone:
I believe if you are changing the schema for User model from `$table->primary();` to `$table->uuid();` then you should also change the schema for session (on the same file) as well.

lesichkovm:
> While I see your point, it will break the foreign key for majority of existing users that don't have an issue with the current setup where they use file based sessions or other session drivers that do not have a problem with this..

The foreign key constraint is removed yes, which brings it on par with the file based session solution.

The way I see it any session driver should work exactly in the same way as the other, so that they can be interchangeable.

The proposed solution, does make the database session driver and the file based session driver work interchangeably.

lesichkovm:
> I believe if you are changing the schema for User model from `$table->primary();` to `$table->uuid();` then you should also change the schema for session (on the same file) as well.

This is exactly what we are trying to avoid here. The default Laravel installation should allow the file and database session drivers to be swapped without extra modification out of the box by default.

taylorotwell:
No plans to change this atm.

lesichkovm:
@taylorotwell has this been considered for the next version of Laravel?

PR #6495, opened 2024-11-18
[11.x] match `HidesAttributes` docblocks

browner12:
the docblock in `HidesAttributes` was updated in #42512, so this child class should be using the same.

otherwise PHPStan throws a "PHPDoc type array<int, string> of property App\Models\User::$hidden is not covariant with PHPDoc type list<string> of  property Illuminate\Database\Eloquent\Model::$hidden" error

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -26,7 +26,7 @@ class User extends Authenticatable
     /**
      * The attributes that should be hidden for serialization.
      *
-     * @var array<int, string>
+     * @var array<string>
      */
     protected $hidden = [
         'password',


browner12:
nevermind, this doesn't fix the PHPStan error, but it does still sync them up.

app/Models/User.php
```diff
@@ -26,7 +26,7 @@ class User extends Authenticatable
     /**
      * The attributes that should be hidden for serialization.
      *
-     * @var array<int, string>
+     * @var array<string>
```
ahinkle, line 29:
```suggestion
     * @var list<string>
```

`list` solves the issue with defaults of Larastan 3.0 to prevent the covariant issue you are describing.

 However, this pattern currently isn't consistent across the framework.

browner12: thanks for pointing this out @ahinkle. yah, looks like very limited use in the FW so far, but maybe something we want to look into.

PR #6496, opened 2024-11-20
Update welcome.blade.php

syofyanzuhad:
Add meta description to improve SEO.
source: https://developer.chrome.com/docs/lighthouse/seo/meta-description/


Jubeki:
Nice change. But I hope this page will not be visible in the end product being used in production.

PR #6498, opened 2024-11-27
Upgrade to Vite 6

timacdonald:


package.json
```diff
@@ -12,6 +12,6 @@
         "laravel-vite-plugin": "^1.0",
```
Jubeki, line 12:
```suggestion
        "laravel-vite-plugin": "^1.1",
```

This should probably be changed too, as I assume that v1.1 will support Vite 6 right?
Or will this be a major version bump?

timacdonald: We shouldn't need to bump this.

Running NPM install with `"laravel-vite-plugin": "^1.0"` and `"vite": "^6.0"` should install the compatible version, and latest version, of the Laravel Vite plugin.

PR #6499, opened 2024-12-05
improvment welcome blade

amirhosseintajafari:
Sorting html codes in welcome.blade.php


antonkomarev:
@amirhosseintajafari how does this improvement help? I suppose people just delete this view right after app installation.

amirhosseintajafari:
@antonkomarev In my opinion, every code should be in order, even if you are right

dcgitx:
I'm glad this is closed - I'm not sure it's even real, dependent on the devs code formatter this would all change anyway, that's all it is.. confused..

PR #6500, opened 2024-12-05
[11.x] Update vite.config.js to ignore storage path

nexxai:
## Problem

I'm working on an app that downloads HTML and CSS files from various sites and temporarily stores them in `storage/...`.  As each file is downloaded in a development environment, Vite's hot reloading sees these new files and tries to rebuild the `app.js` and `app.css` files for each matching file, causing a ton of unnecessary refreshes and, in some cases, actually breaking the page until I can restart the `npm run dev` process.

## Solution

I don't know if this is the *best* solution (I am certainly all ears if there is a better way to accomplish this) but by adding the storage folder to Vite's server watch ignore list, it means that Vite will only watch developer-generated code for changes, rather than anything user-submitted/ephemeral that lives in the `storage_path()` and its children.  I prevented the unnecessary scanning by setting Vite's `server.watch.ignore` option to include the storage folder.

If this PR is accepted, I will submit a documentation update to explain the change.

## Example

This is what the Vite log looks like prior to this change
Screenshot 2024-12-05 at 12 31 13 PM

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -10,7 +10,7 @@ export default defineConfig({
     ],
     server: {
         watch: {
-            ignored: ['**/storage/**'],
+            ignored: ['storage/**'],
         },
     },
 });
--- a/vite.config.js
+++ b/vite.config.js
@@ -10,7 +10,7 @@ export default defineConfig({
     ],
     server: {
         watch: {
-            ignored: ["**/storage/**"],
+            ignored: ['**/storage/**'],
         },
     },
 });
--- a/vite.config.js
+++ b/vite.config.js
@@ -8,4 +8,9 @@ export default defineConfig({
             refresh: true,
         }),
     ],
+    server: {
+        watch: {
+            ignored: ["**/storage/**"],
+        },
+    },
 });


timacdonald:
This is because [Vite performs a full page refresh when a html file that is not part of the module graph is updated](https://github.com/vitejs/vite/blob/7a0758cbbc825717851962d4e6f82e369b8e800b/packages/vite/src/node/server/hmr.ts#L586-L601).

I'm not against ignoring some directories by default, including `storage`. I've [suggested this in the past](https://github.com/laravel/vite-plugin/issues/152#issuecomment-1937979427) because we currently watch the entire project, which can [hit upper limits on applications with a large number of files in storage and vendor](https://github.com/laravel/vite-plugin/issues/152#issue-1393080738). 

I feel it could be good to ignore the following directories by default in the Vite plugin:

```
/bootstrap/**
/database/**
/storage/**
/tests/**
/vendor/**
```

However, we cannot blanket ignore the `storage` or `vendor` directory as the Vite Plugin is watching some files for the page reload feature:

```
/vendor/laravel/framework/src/Illuminate/Pagination/resources/views/*.blade.php
/storage/framework/views/*.php
```

I would like to see this built into the `Vite-plugin` rather than in the `vite.config.js` and provide an escape hatch for those who do want Vite to watch these directories. I imagine we would respect the user's `server.watch.ignore` config and fallback to our sensible default for Laravel applications.

This would likely require a `v2` of the plugin because it could break some peoples workflows. That being said, it only breaks things in dev mode and not in production, so I'm personally not against throwing it in `v1`. I can't imagine it would impact many people.

joshmanders:
@timacdonald I like the sounds of all that!

nexxai:
Sorry team, I am barely passable at vanilla JavaScript, let alone TypeScript or how I'd even go about modifying the Vite plugin.  If one of you wants to give me a hand, I would try to take a stab at it, but I'd definitely need some guidance (more like hand-holding, to be honest).  Failing that, I'll just stick to adding the entry in my own config as needed.

timacdonald:
I haven't got space to do this right now. Would appreciate someone having a go at a PR.

I'll circle back to it when I can.

timacdonald:
Hey folks,

There is actually a PR open for this now, thanks to @natsuki-engr.

If you had space, it would be great if you could test it and see if you find a difference.

I've dropped on comment on how to test it in the PR: https://github.com/laravel/laravel/pull/6500

joshmanders:
Correction: PR is laravel/vite-plugin#349

timacdonald:
Shesh, I'm an idiot. Thanks, Josh.

PR #6501, opened 2024-12-07
Add comment suggestion

NavidSaeidi:
Added a comment to define the start time for performance monitoring, which will help with understanding the code better for new contributors.


NavidSaeidi:
Added a comment to define the start time for performance monitoring, which will help with understanding the code better for new contributors.

public/index.php
```diff
@@ -14,4 +15,4 @@
 
 // Bootstrap Laravel and handle the request...
 (require_once __DIR__.'/../bootstrap/app.php')
-    ->handleRequest(Request::capture());
+    ->handleRequest(Request::capture());
```
antonkomarev, line 18: Why did you remove the empty line at the end of the file?

NavidSaeidi: I removed the empty line at the end of the file to maintain consistency with the rest of the code. In many coding standards, extra empty lines are avoided to keep the files clean and concise.

datamweb: @NavidSaeidi Using **LF** instead of **CRLF** in PHP files is recommended because:

1. **Cross-Platform Compatibility**: LF is the standard on Linux and macOS, ensuring consistent behavior across environments.
2. **Smaller File Size**: LF uses one character, while CRLF uses two, reducing file size.
3. **Git-Friendly**: Git prefers LF, avoidin
4. g unnecessary changes in diffs or merges.
5. **Better Standards**: Modern tools and languages work seamlessly with LF.

Switching to LF avoids compatibility issues and ensures smoother collaboration and development.

PR #6505, opened 2024-12-12
[11.x] Storage:link source and target directory statement updated

ChetanTechnource:
Updated key values source and target directory statement for symlinks array


ChetanTechnource:
@taylorotwell Please review see if this is helpful

PR #6509, opened 2024-12-16
set default scheme for smtp mailer

Khuthaily:
This PR addresses an issue where the smtp mailer fails with the message `The "" scheme is not supported; supported schemes for mailer "smtp" are: "smtp", "smtps".`


crynobone:
Please upgrade to latest version of Laravel Framework, or report the bug to Framework with reproducing code

PR #6510, opened 2024-12-23
Add Functional Index Support to Schema Builder

ahsan-ghalib:
This PR adds support for functional indexes in Laravel's schema builder. Functional indexes allow developers to use MySQL expressions like LOWER, UPPER, etc., directly in their indexes.


taylorotwell:

parvin528:
Great

PR #6511, opened 2024-12-26
fix: fixed the lenght of the columns

pedrovian4:
## Description

When Running really large  caches got overflow on the cache  table  I think might be handy to have on default it as longtext

### DIFF
--- a/database/migrations/0001_01_01_000001_create_cache_table.php
+++ b/database/migrations/0001_01_01_000001_create_cache_table.php
@@ -12,14 +12,14 @@ return new class extends Migration
     public function up(): void
     {
         Schema::create('cache', function (Blueprint $table) {
-            $table->string('key')->primary();
-            $table->mediumText('value');
+            $table->longText('key')->primary();
+            $table->longText('value');
             $table->integer('expiration');
         });
 
         Schema::create('cache_locks', function (Blueprint $table) {
-            $table->string('key')->primary();
-            $table->string('owner');
+            $table->longText('key')->primary();
+            $table->longText('owner');
             $table->integer('expiration');
         });
     }


database/migrations/0001_01_01_000001_create_cache_table.php
```diff
@@ -12,14 +12,14 @@
     public function up(): void
     {
         Schema::create('cache', function (Blueprint $table) {
-            $table->string('key')->primary();
-            $table->mediumText('value');
+            $table->longText('key')->primary();
```
GrahamCampbell, line 15: This looks incorrect for the key. We for sure don't want a text column for the pk.

GrahamCampbell (changes requested):

database/migrations/0001_01_01_000001_create_cache_table.php
pedrovian4: @GrahamCampbell  Oh I see, when the key is really large do you see any alternative? Oh my company we got some problems on our cache tables where the keys are really large? Maybe some kinda of compression?

joshmanders: @pedrovian4 can you give an example? That's an insanely large key if so. Are you mistakingly encoding the value into the key?

pedrovian4: @joshmanders  Yes, really large key, actually i am not,  because the app was really cache dependent, we found this when by mistake switched the env keys from REDIS to Database on production, to avoid down time and rebuild the app on the Cluster we changed the type of the key.

pedrovian4: > @pedrovian4 it's not uncommon to use a hash function for such things

Do we have some Native support on laravel for such thing? I think I should start with a issue instead a PR, I am sorry.

PR #6512, opened 2024-12-31
Remove base abstract controller class

abbas-roholamin:


joshmanders:
No explanation why? Taylor will surely merge this in. /s

PR #6515, opened 2025-01-06
Update .gitignore to not ignore auth.json in the lang directory.

Tjoosten:
Hey There, 

I made this pull request for the following. For now if i perform the  `php artisan lang:publish` command i will get the the default english translation files from the framework. 

But if i want to a new language and want to use JSON file. the auth.json file will not be added to my commit due to the `auth.json` entry in the `.gitignore` file. This commit will allow to commit the ´´lang/nl/auth.json´ file but still isgnore the  `auth.json` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,9 +13,9 @@
 .phpunit.result.cache
 Homestead.json
 Homestead.yaml
-auth.json
 npm-debug.log
 yarn-error.log
+/auth.json
 /.fleet
 /.idea
 /.nova


perisicnikola37:
Honestly, this is kind of okay, but I would leave it as user preference. I would not include this in the framework itself

Tjoosten:
> IMO this is kind of okay, but I would leave it as user preference. I would not include this in the framework itself

I understand, but it is a small low-effort PR. So i let the core team decide on this on. 
I also understand that they reject this change. But in my opinion i see no reason to ignore the lang/nl/auth.json file by default.

joshmanders:
> IMO this is kind of okay, but I would leave it as user preference. I would not include this in the framework itself

This is already apart of the framework, this PR is just fixing it to work correctly. Ignoring composer auth.json should not ignore framework specific code.

PR #6529, opened 2025-01-29
moving `laravel/pail` from 'require-dev' to 'require'

aagjalpankaj:
`laravel/pail` can be useful for digging into production issues like tinker. Normally we don't install dev dependency on production.


taylorotwell:
You can do this on your application if you want. 👍

PR #6535, opened 2025-02-11
Move npm production dependencies out of devDependencies

Jamesking56:
This was born out of an issue I had recently with [Renovate](https://www.mend.io/renovate) dependency updater which tried to update a bunch of my sites from Tailwind v3 to v4 automatically (and auto-merging PRs to do this).

Because Tailwind was in `devDependencies`, the updater thought it was perfectly fine to do a major upgrade.

Considering both Axios and Tailwind are usually production dependencies where a major update would break most websites, I think these should be marked as ordinary NPM dependencies so that updaters like these now know that a major update is breaking and should be manually reviewed.

This updates `laravel/laravel` so that these dependencies are marked as non-dev by default, thus eliminating this issue and making it clear that these dependencies are directly used on the site.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,1589 @@
+{
+    "name": "laravel",
+    "lockfileVersion": 3,
+    "requires": true,
+    "packages": {
+        "": {
+            "dependencies": {
+                "axios": "^1.7.9",
+                "tailwindcss": "^4.0.6"
+            },
+            "devDependencies": {
+                "autoprefixer": "^10.4.20",
+                "concurrently": "^9.0.1",
+                "laravel-vite-plugin": "^1.2.0",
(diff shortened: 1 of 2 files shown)


taylorotwell:
I don't think this actually matters?

joshmanders:
While I agree that axios should be in dependencies and not devdependencies because it's an actual dependency in the "frontend" aspect of the boilerplate, and not a dependency used to generate the bundle, such as vite and autoprefixer, etc.

I don't think Tailwind should be in it. And the reasoning being that an automated program assumed that dev dependencies can have major version updates done automatically is kind of a problem with the automation and not so much Laravel's package.json.

taylorotwell:
I'll revisit this for Laravel 12.x.

PR #6536, opened 2025-02-12
[11.x] remove `APP_TIMEZONE` environment variable

browner12:
force the timezone to be defined in the config file, rather than deferring to an environment variable.

IMO having the timezone dependent on an environment variable adds an unnecessary amount of risk for your average user who may be unaware of the effects from changing this value.

some scenarios where this could cause issues:

- devs set this value to their local timezone on their local dev machines, and they are writing `Carbon:create()` commands or similar that now are working only under the assumption of that timezone value
- you have multiple production servers setup across the country and each has their own timezone value according to their location, but they all talk to a central database. now the database is loaded with mixed timezone values

having an explicit value defined once for the application removes these risks.

reverts part of #6188


a-h-abid:
Funny, it was added after long request by various users, now its revert back to old way.

Expect more PR requests for adding it back again. 😮‍💨

GrahamCampbell:
People can just edit their own config file to their preference, and pull things from env variables if they want. The framework can keep with best practice and good defaults.

ItsRD:
Maybe this is something that should be added to the upgrade guide — I guess the one for upgrading from 11 to 12?

In my case, I used Laravel Shift to upgrade my project from 10 to 11 a while back, which deleted my config/app.php file. After upgrading to Laravel 12 yesterday, I noticed that the dates were incorrect. The APP_TIMEZONE was previously set to Europe/Amsterdam, but the dates were in UTC. That’s when I came across this issue and added the config/app.php file back into my project.

CoolGoose:
I think this is a problem for shift @ItsRD , on a framework update, laravel doesn't update your config files

ItsRD:
@CoolGoose When Laravel 11 was released, did newly installed Laravel application have the `config/app.php` file installed? I [thought they didn't](https://github.com/laravel/laravel/commit/428a190050866c53043900dee91c579b0fa40768#diff-950397cc5ef29707b54d0c774bd94af85836bab1a27f2f86533d832d25caf224). If so, people are going to have this issue when upgrading their Laravel application to the newest 11 or even 12 version, right? Manually or by using Laravel Shift.

wjvankesteren:
Still, this change looks strange to me.

Other settings remains in the .env file, e.g. the database connection, which could have a specific timezone set which will be reflected in the .env variable APP_TIMEZONE. It will be strange at that point to have to change code for that.

Secondly, it is simply not consistent: some settings can be overriden by .env settings ans some not; very strange

browner12:
some things **should not** be overwritten at the environment level.

Anticom:
> I think this is a problem for shift @ItsRD , on a framework update, laravel doesn't update your config files

@CoolGoose I disagree with this statement. I expect shift to be aligned with laravel standards. If you guys choose to drop it, it's not a shift issue IMHO.

---

In terms of general discussion on this topic I also disagree with the justifications given for this change by @browner12 :

> devs set this value to their local timezone on their local dev machines, and they are writing `Carbon:create()` commands or similar that now are working only under the assumption of that timezone value

Does not make any sense at all. If your local dev environment is using another timezone throughout all data stored in local DB (which _should_ be decoupled from any production data when it's a local dev environment) I see hardly any scenario where you would end up with something that would work in your local environment but break in production.

The only two cases where this might impose issues I can think of are the following two cases:
1. You _have to_ use a production dump due to relying on real-world data for development or debugging any issues.
2. Requests to upstream APIs your app is consuming.

That being said if any of the cases mentioned above cause issues I have strong feelings about this being a **developer responsibility** in this case.

> you have multiple production servers setup across the country and each has their own timezone value according to their location, but they all talk to a central database. now the database is loaded with mixed timezone values

I don't know of a single OS etc. that set `APP_TIMEZONE` environment on the system automatically to specify machine-global timezone settings. If this is true for any of laravel's hosting services, that's. Also here devs / ops should be responsible for actually managing application environment correctly. Using this an argument can be said about every single environment variable required for a production deployment to function correctly.

To me both arguments sound like you're expecting developers / ops to be stupid and then blame the framework. If this is the way we want to move forward with laravel, we may as well revert back `CACHE_STORE` to `CACHE_DRIVER` since it may break production deployments if we don't adjust our environment config after upgrading.

---

Lastly simply looking at the reactions on the comments in this discussion we can simply observe that this has been an unpopular change. This should be taken into consideration for future changes.

PR #6539, opened 2025-02-13
Re-add nunomaduro/collision

jasonmccreary:
In preparing for the Laravel 12.x Shift, I noticed the dependency on `nunomaduro/collision` was removed. But I don't see it within `laravel/framework` or another first-party dependency. So I'm assuming maybe this was accidently removed when resolving [this merge conflict](https://github.com/laravel/laravel/commit/295bdfeee961c760425a392942273febe1eb6cc9), and should be re-added. 😅

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "laravel/pint": "^1.13",
         "laravel/sail": "^1.41",
         "mockery/mockery": "^1.6",
-        "nunomaduro/collision": "^8.1",
+        "nunomaduro/collision": "^8.6",
         "phpunit/phpunit": "^11.5.3"
     },
     "autoload": {
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
         "laravel/pint": "^1.13",
         "laravel/sail": "^1.41",
         "mockery/mockery": "^1.6",
+        "nunomaduro/collision": "^8.1",
         "phpunit/phpunit": "^11.5.3"
     },
     "autoload": {


joshmanders:
If I'm not mistaken it was removed because there's now internal handling of errors and if you want the old collision package version instead of internal you can just reinstall it and it'll get used over the internal one.

composer.json
```diff
@@ -16,6 +16,7 @@
         "laravel/pint": "^1.13",
         "laravel/sail": "^1.41",
         "mockery/mockery": "^1.6",
+        "nunomaduro/collision": "^8.1",
```
crynobone, line 19:
```suggestion
        "nunomaduro/collision": "^8.6",
```

Only Collision 8.6 and above can be used on Laravel 12.

crynobone (changes requested):

composer.json
crynobone: It was remove since it's not possible to be used with Laravel 12. Check `conflict` section in Collision's composer.json

jasonmccreary: Gotcha. Thanks. Guess I'll close then.

crynobone: Actually your PR is valid, just 8.1-8.5 will be ignored to conflict. You can use 8.6 and above

jasonmccreary: Oh, I understand now. I'll reopen.

jonnott:
I've confused - why do we need nunomaduro/collision back in the skeleton as a dependency if "it was removed because there's now internal handling of errors and if you want the old collision package version instead of internal you can just reinstall it and it'll get used over the internal one" (@joshmanders) ?

crynobone:
Because of this line: https://github.com/nunomaduro/collision/commit/c1ad0f1185e0b9a40243c34b3888347038d9891e

Same with Laravel 13, it would be able to install with `nunomaduro/collision` until conflict get removed when we are close to release.

PR #6544, opened 2025-02-20
Migration key was too long issue.

Archana-1993:
Whenever I install a fresh Laravel project and run migrations with any MySQL version, I face the issue of the unique key being too long. To avoid manually adding the schema every time, I have added it in the App Service Provider. I am facing this issue on Windows OS.

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -3,6 +3,7 @@
 namespace App\Providers;
 
 use Illuminate\Support\ServiceProvider;
+use Illuminate\Support\Facades\Schema;
 
 class AppServiceProvider extends ServiceProvider
 {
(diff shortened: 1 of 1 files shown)


lloricode (commented):
I believe you are using an outdated version of MySQL. You can add this to your project configuration to address the issue when running older MySQL versions.

Archana-1993:
Hello @lloricode 

I have read your above added comments.

But, I think this change is great for newer Laravel versions, It would be beneficial if it also supported older versions. Many projects still rely on older Laravel versions due to long-term maintenance and integration with legacy systems. Just as PHP maintains compatibility across versions, it would be helpful if this change could work for older Laravel versions as well. If there’s any way to adjust this for broader compatibility, that would be great. Alternatively, if you have another solution to achieve this, please consider to adding it.

Thank you so much.

PR #6545, opened 2025-02-21
use base_path to reference route files

dextermb:
It makes sense to me to use the `base_path` function to point towards the route files. The syntax seems cleaner than using relative paths and string concatenation.


taylorotwell:
I think it's OK. 👍

PR #6550, opened 2025-02-26
Display Laravel and PHP versions on the welcome page

shishirsteadfast:


antonkomarev:
Why composer.lock and package.lock files are updated?

shishirsteadfast:
I ran composer update and npm install, which is why the composer.lock and
package-lock.json files were updated.  If it's not necessary, I can roll
back the updates to the composer.lock and package-lock.json files since
they're not mandatory

Mohammedzain244:
Amazing

taylorotwell:
I don't think this is necessary.

PR #6551, opened 2025-02-26
[11.x] Fix broken assets in welcome page after Laravel 12 release

alaminfirdows:
After the release of Laravel 12, the default welcome.blade.php view in Laravel 11 references three assets that no longer exist, causing 404 Not Found errors. This PR replaces the missing image URLs with inline SVG images.

Note: I initially tried using `<svg>` images directly, but the `object-*` class doesn’t apply to them. That’s why I used base64-encoded images instead.

Fixes: https://github.com/laravel/framework/issues/54802

--

After solving this issue:
image

image


alaminfirdows:
The `object-top` or `object-position: top` won’t work on an inline `<svg>` element because object-position is only applicable to replaced elements like `, <video>, or <iframe>`. Inline `<svg>` is not a replaced element, so CSS properties like `object-fit` and `object-position` won’t affect it.

https://developer.mozilla.org/en-US/docs/Web/CSS/object-position

taylorotwell:
I'll just make those assets available again.

PR #6552, opened 2025-02-26
feat: Add command to clear all caches in the application

JCarlosRAlmeida:
This command clears all caches from the application, including configuration, route, and view caches. By executing `php artisan app:clear`, you ensure that the application is using the most up-to-date settings and data, improving performance and system integrity.


antonkomarev:
I found command name is pretty confusing.

PR #6554, opened 2025-02-26
feat: add type hints for $this in routes/console.php

AJenbo:
## Summary
This PR adds a type hint for `$this` in `routes/console.php`, ensuring better IDE support and improved static analysis when working with console commands.

## Why This Change?
- **Improves Developer Experience**: IDEs can now provide autocompletion and proper type inference for `$this` within console command closures.
- **Enhances Static Analysis**: Helps tools like PHPStan understand the type of `$this`, reducing potential false positives in type checks.
- **Better Code Readability**: Explicitly declaring `$this` as an instance of `Illuminate\Console\Command` makes the code easier to understand for new developers.

## Testing & Compatibility
- No runtime behavior changes; this is a documentation-level improvement.
- Fully backward compatible with existing Laravel applications.

### DIFF
--- a/routes/console.php
+++ b/routes/console.php
@@ -3,6 +3,7 @@
 use Illuminate\Foundation\Inspiring;
 use Illuminate\Support\Facades\Artisan;
 
+/** @var Illuminate\Console\Command $this */
 Artisan::command('inspire', function () {
     $this->comment(Inspiring::quote());
 })->purpose('Display an inspiring quote');


routes/console.php
```diff
@@ -1,8 +1,10 @@
 <?php
 
+use Illuminate\Console\Command;
 use Illuminate\Foundation\Inspiring;
 use Illuminate\Support\Facades\Artisan;
 
+/** @var Command $this */
```
JackNoordhuis, line 7: Using the FQN `\Illuminate\Console\Command` here and removing the import would be better IMO, the class isn't actually used in any code.

AJenbo: done

AJenbo:
@bert-w this solves the issue you where having with route files missing context for `$this`.

timacdonald:
This sets `$this` within the entire `routes/console.php` file to be an instance of `Illuminate\Console\Command`.

However, `$this` is an instance of `Illuminate\Foundation\Console\Kernel` in the `routes/console.php` file.

```php
<?php

use Illuminate\Foundation\Inspiring;
use Illuminate\Support\Facades\Artisan;

})->purpose('Display an inspiring quote');

dd($this::class);
```

Screenshot 2025-03-02 at 16 12 48


If this was wanted, the docblock would need to be contained within each command's anonymous Closure, otherwise it is invalid.

```diff
- /** @var Illuminate\Console\Command $this */
Artisan::command('inspire', function () {
+   /** @var Illuminate\Console\Command $this */
    $this->comment(Inspiring::quote());
})->purpose('Display an inspiring quote');
```

AJenbo:
Hmm I se what's happening is that the context is rebound between the function being defined and later being executed:

https://github.com/laravel/framework/blob/7b8e12d8f4218615370d329a258f0b7c4afc6237/src/Illuminate/Foundation/Console/ClosureCommand.php#L64

This change is going to be obscure to a lot of tools which expects the context to be bound at the time the closure is being defined and so expect that it's using the context of the file. I don't think your solution of adding the docblock inside of the anonymous function is going to work in general for the same reason.

I'll close this PR for now.

PR #6555, opened 2025-02-27
Make the github test action run out of the box independent of the choice of testing framework

ndeblauw:
# Summary
This PR changes the test runner for the github action to `php artisan test`

# Why This Change?
The included github testing action executes `./vendor/bin/phpunit` but when pest is chosen, this results in an error. 
I know it's easy to fix (and the error message tells you exactly what to do, that is, run `./vendor/bin/pest`), but wouldn't it be nicer if it would just work out of the box?

# Testing & Compatibility
No runtime behavior changes
Compatible with existing Laravel applications.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,4 +44,4 @@ jobs:
         run: php artisan key:generate
 
       - name: Execute tests
-        run: vendor/bin/phpunit
+        run: php artisan test


flavio-schoute:
It doesn't matter much, because `.github` doesn't come along anyway when you create a new Laravel project, it's convenient in itself, suppose you copy `tests.yml`, yeah then you don't have to modify it....

michaeldyrynda:
The starter kits don’t have the `.github` directory set to `export-ignore`, but the main Laravel template repo does.

Possibly an oversight - would need to remove it from `.gitattributes` here as well.

maru0914:
I think this PR solved this issue, and Laravel Herd will be updated with latest laravel/installer. 
https://github.com/laravel/installer/pull/401

PR #6556, opened 2025-02-27
Add schedular work to composer dev script

Mubeen142:
This PR adds the `php artisan schedule:work` command to the dev script so that in local development you don't need to setup cronjobs or run the schedule work command manually.


taylorotwell:
Add this to your own application if you want. 👍

PR #6562, opened 2025-03-04
Hint outer scope in console routes

AJenbo:
This is a follow up to https://github.com/laravel/laravel/pull/6559

It restores the `@var Kernel $this` annotation, which was initially included in my original PR but later removed during review. Omitting this annotation causes issues with some developer tools, as they fail to recognize that this file is within the scope of a class. Without it, the inner `$this` overwrites (`@var ClosureCommand $this`) lead to incorrect assumptions by static analyzers and IDEs.

By explicitly declaring `@var Kernel $this`, we ensure proper type inference and improve the developer experience.

### DIFF
--- a/routes/console.php
+++ b/routes/console.php
@@ -1,9 +1,11 @@
 <?php
 
 use Illuminate\Foundation\Console\ClosureCommand;
+use Illuminate\Foundation\Console\Kernel;
 use Illuminate\Foundation\Inspiring;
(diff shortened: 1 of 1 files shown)


jhm-ciberman:
> causes issues with some developer tools

Can you show a real screenshot showing the problem?

taylorotwell:
This just looks so bad though I'm not sure I can stomach it.

PR #6563, opened 2025-03-04
Cast env to string in config

AJenbo:
This PR explicitly casts values returned by `env()` to string where `strings` are expected. Since `env()` can also return `bool` which then gets implicitly cast to a string, this makes it clear that we expect a string conversion in these cases.

#### Why?

This helps static analysis by removing ambiguity, making it easier to catch real issues.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -101,7 +101,7 @@ return [
 
     'previous_keys' => [
         ...array_filter(
-            explode(',', env('APP_PREVIOUS_KEYS', ''))
+            explode(',', (string) env('APP_PREVIOUS_KEYS', ''))
         ),
     ],
 
--- a/config/cache.php
+++ b/config/cache.php
@@ -103,6 +103,6 @@ return [
     |
     */
 
-    'prefix' => env('CACHE_PREFIX', Str::slug(env('APP_NAME', 'laravel'), '_').'_cache_'),
+    'prefix' => env('CACHE_PREFIX', Str::slug((string) env('APP_NAME', 'laravel'), '_').'_cache_'),
 
 ];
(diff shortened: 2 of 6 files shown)


GrahamCampbell:
I don't agree this change makes sense. This makes the code slower and obscures real bugs.

AJenbo:
>  This makes the code slower 

Over a million iterations, the difference was 0.005 seconds. Even with six instances, this would amount to just 0.000000003 seconds. Even if you consider this significant, it shouldn’t matter much since the configs are cached in production and only evaluated once during deployment.

> obscures real bugs

How does it obscure bugs? The explicit cast is a clear, intentional choice that avoids any ambiguity, making it obvious that this isn’t an oversight. This way, we can focus on real issues without false positives cluttering up the list.

sawoodscraperrs:
> This PR explicitly casts values returned by `env()` to string where `strings` are expected. Since `env()` can also return `bool` which then gets implicitly cast to a string, this makes it clear that we expect a string conversion in these cases.
> 
> #### Why?
> This helps static analysis by removing ambiguity, making it easier to catch real issues.

Laravel already handles implicit type conversions correctly, so explicit casting isn’t necessary in cases like `env()`. Since PHP naturally converts values when passed to functions like `explode()`, this change doesn’t provide a functional benefit. Additionally, Laravel encourages handling type casting at the configuration level rather than in the code itself.

AJenbo:
Yes, the implicit cast works, the point is to make it explicit so that it's clear to analyzers that it's intentional. It's not about functionality but tooling.

taylorotwell:

AJenbo:
Thank you for the boilerplate response

PR #6564, opened 2025-03-04
Improve Default Unit Test

AJenbo:
This PR updates the default unit test to validate an actual part of the Laravel skeleton. Instead of asserting `true === true`, it now tests that the `User` model correctly handles attribute assignment.

#### Why?

- The original test fails PHPStan because it performs a redundant check.
- Ensures the test can actually fail, making it a better example.

Together with my two other PRs, this resolves all reported issues from PHPStan on a default Laravel installation.

### DIFF
--- a/tests/Unit/ExampleTest.php
+++ b/tests/Unit/ExampleTest.php
@@ -2,6 +2,7 @@
 
 namespace Tests\Unit;
 
+use App\Models\User;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think it's OK for now.

AJenbo:
It's not the worst, maybe it will encourage people to write a proper test for there system, or just delete it I guess either one are fine.

PR #6581, opened 2025-03-15
[12.x] Clear and concise implementation

AhmedAlaa4611:
Removing unnecessary type annotation that do not provide significant benefits.

Changes:
- Removed the inline type annotation.
- Make the implementation more minimal.

Reasons:
The additional lines did not provide any functional improvements. Simplifying the file keeps the codebase clean, concise, and easier to maintain.

These changes help maintain clarity and avoid unnecessary complexity in the file.

### DIFF
--- a/artisan
+++ b/artisan
@@ -1,7 +1,6 @@
 #!/usr/bin/env php
 <?php
 
-use Illuminate\Foundation\Application;
 use Symfony\Component\Console\Input\ArgvInput;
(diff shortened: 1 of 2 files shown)


rojtjo:
I wouldn't call specifying a type as unnecessary complexity. It helps with static analysis and it also allows us to easily navigate to method definitions in our editor.

PR #6582, opened 2025-03-15
Add user(): helper function

imliam:
This adds a `user()` helper function to the base application, which is type-hinted to return an instance of the `App\Models\User` class by default.

**Why?**

`auth()->user()` is handy, and widely used in Laravel projects, but its return typehint is for a `Illuminate\Contracts\Auth\Authenticatable` instance.

While this may be valid from a framework perspective, and there are many uses for Authenticatables that are not the User model, most Laravel applications use the defaults. Therefore, users expect an instance of the User model returned.

This means that, without the help of third-party tools like [laravel-ide-helper](https://github.com/barryvdh/laravel-ide-helper), as developers we get a poor experience when getting the authenticated user and expecting the model:
- No IDE support for chaining methods/properties
- Static analysis fails if we chain anything not on the Authenticatable interface
- We have to typehint `/** @var \App\Models\User $user */` everywhere we use the helper

By adding this to the basic application scaffold, the new `user()` helper can get proper typehinting for the out-of-the-box User model that ships with a fresh application, supporting proper typehinting. As it is in the userland instead of the framework, it can be changed to support other Authenticatables if the developer wishes.

**Why should this be in the base application?**

Yes, users could always add this helper manually to their app, but having a standard way the framework does it will eventually bring consistency to new applications—and I believe this utility is used commonly enough that it warrants being in the base.

There's clearly [some interest](https://x.com/LiamHammett/status/1900568350599885168) in it, and people have a dozen of their own home-grown solutions for the same problem, something this could provide.

**Why not in the framework?**

`App\Models\User` is defined in the user's application, so the framework has no knowledge of it or if it changes names/namespaces/etc.

### DIFF
--- a/bootstrap/helpers.php
+++ b/bootstrap/helpers.php
@@ -1,0 +1,22 @@
+<?php
+
+use App\Models\User;
+use Illuminate\Auth\AuthenticationException;
+
+/**
+ * Get the currently authenticated user model.
+ */
+function user(): User
+{
+    $currentUser = auth()->user();
+
+    if ($currentUser === null) {
+        throw new Exception('The current user is not authenticated.');
+    }
+
+    if (!$currentUser instanceof User) {
+        throw new Exception('The currently authenticated user is not an instance of ' . User::class);
+    }
+
+    return $currentUser;
+}
--- a/bootstrap/helpers.php
+++ b/bootstrap/helpers.php
@@ -1,7 +1,6 @@
 <?php
 
 use App\Models\User;
-use Illuminate\Auth\AuthenticationException;
 
 /**
  * Get the currently authenticated user model.
--- a/bootstrap/helpers.php
+++ b/bootstrap/helpers.php
@@ -14,8 +14,8 @@ function user(): User
         throw new Exception('The current user is not authenticated.');
     }
 
-    if (!$currentUser instanceof User) {
-        throw new Exception('The currently authenticated user is not an instance of ' . User::class);
+    if (! $currentUser instanceof User) {
+        throw new Exception('The currently authenticated user is not an instance of '.User::class);
     }
 
     return $currentUser;
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
             "App\\": "app/",
             "Database\\Factories\\": "database/factories/",
             "Database\\Seeders\\": "database/seeders/"
-        }
+        },
+        "files": ["bootstrap/helpers.php"]
     },
     "autoload-dev": {
         "psr-4": {


James4645:
I don't get the point of this. Seems like yet another worthless file I'll have to remove from every app I make like the console routes file.

joshmanders:
Yeah this should be either apart of the framework's existing helpers or added yourself to your own project.

With the new `--using` flag in the installer you can make your own starter with this

AhmedAlaa4611:
We already have Laravel's default middleware `auth` and `guest`, this is useless.

bootstrap/helpers.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+use App\Models\User;
+use Illuminate\Auth\AuthenticationException;
```
rodrigopedra, line 4: This exception is not thrown anywhere.

Maybe the Exceptions being thrown below should be an instance of this class?

bootstrap/helpers.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+function user(): User
+{
+    $currentUser = auth()->user();
+
+    if ($currentUser === null) {
+        throw new Exception('The current user is not authenticated.');
+    }
+
+    if (! $currentUser instanceof User) {
```
rodrigopedra, line 17: As the function has a return type of `App\Models\User` this check is unnecessary. 

If an object which is not an instance of `App\Models\User` is returned, PHP will error out automatically.

rodrigopedra:
I usually add a `helpers.php` file to my app's `./app` folder and namespace it on `App`, something like this:

```php
<?php

namespace App;

function foo() {}
```

It is still short enough to use in views (`{{ \App\foo() }}`), and in regular code I can import it. And has the benefit of not polluting the global namespace.

I would vote to have the helpers file under `./app`, as it is application's code.

I can see the benefit of having this file included on the likes to show users how easy, and useful, it is to have a helpers file. 

Much like what we have in `./route/console.php`, which ships with the `inspiring` command, and anyone who doesn't want that command can easily remove it on their app.

bootstrap/helpers.php
imliam: It would throw a TypeError you're right, but I made it explicit here so that static analysis tools can tell what's going on

bootstrap/helpers.php
imliam: My bad - I've removed this unused namespace now.

There could be a better exception for this but I'm not entirely sure which - the existing uses of AuthenticationException certainly don't seem to line up for this so they get handled differently by different parts of the framework. For now, I've kept it as a basic exception.

bootstrap/helpers.php
balu-lt: We already have the user model defined in the config: `auth.providers.users.model`.
It should use that.

I’m experiencing the same frustration with PHPStan. I’ll probably implement something like this in my projects, but I feel like a better version should be implemented at the framework level.

msamgan:
It also has the additional benefit of an added helper file that will come with the project, most of the devs had to add that file to their project manually. Maybe ``bootstrap/`` is not the right place, but the idea in itself is not bad.

bootstrap/helpers.php
```diff
@@ -0,0 +1,21 @@
...
```
Usmanzahidcode, line 16: Helpful idea, but I think using `App\Models\User` directly could be limiting. In many cases, the authenticated user isn't always a `User` instance — it can be any `Authenticatable` implementation, defined via `auth.providers.users.model` and possibly located in a different namespace.

Also, when calling `auth()->user()` (or this helper if added), developers typically expect an `Authenticatable`, not specifically a `User` model. This might lead to confusion or unexpected type errors in custom setups.

bootstrap/helpers.php
```diff
@@ -0,0 +1,22 @@
...
+        throw new Exception('The currently authenticated user is not an instance of '.User::class);
+    }
```
mavenleo, line 18: This breaks in apps with multiple auth guards. It assumes only User gets authenticated.

MR-Sharifi:
In my opinion, it's not a necessary functionality. Even if it was, it does not consider all possible situations. For example, what if someone (like me) uses a custom-curated class for the user entity that does not extend the App\Models\User?

bootstrap/helpers.php
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Get the currently authenticated user model.
+ */
+function user(): User
+{
+    $currentUser = auth()->user();
+
+    if ($currentUser === null) {
+        throw new Exception('The current user is not authenticated.');
```
shaedrich, line 13: Maybe, this could be a dedicated `UnauthenticatedException` or the like 🤔

bootstrap/helpers.php
```diff
@@ -0,0 +1,21 @@
...
+    }
```
shaedrich, line 14: If you want, you could shorten this
```suggestion
    throw_if($currentUser === null, Exception::class, 'The current user is not authenticated.');
```

inmanturbo:
I think this is an issue for the vscode extension honestly. There has got to be a good way somehow to configure static analysis to bind preferred, current or default implementations for contracts in general.

I honestly feel the OP's pain. I do, but this is part of a larger problem. `Authenticatable` is one of many cross cutting, integral contracts provided and fullfilled by the framework. If only we could globally configure the one's we're most concerned with somehow using some kind of config map? Prefferably not in yaml. :sweat_smile:

RyanPaiva56:
> > PHPStorm wont do anything and defers everything to the IDEHelper (BLESS BARRYV) but it adds a ton of bloat.
> 
> The Laravel IDEA plugin is worth every cent, and works incredible well with (most) type analysis needed on a Laravel project.
> 
> PHP Storm offers the Laravel IDEA plugin at a discounted rate when licensing both the IDE and the plugin together.
> 
> _I am not affiliated to PHP Storm nor to the Laravel IDEA plugin, just a satisfied client of both tools_

I've paid for both and went to vscode anyway. Caleb's make vscode awesome already puts it ahead of phpstorm, and with intelephense (the paid version) and the Laravel vscode extension, it's miles ahead. If we can invest more in a 1st party extension, it solves all of these issues. Case in point, we have talented developers sending pull requests for functions that have been around from day 1 because IDEs are still painful even for veterans.

benfaerber:
I always create a method called `User::auth()` on my user to get proper hints.

```php
class User {
    public static function auth(): self {
        $user = auth()->user();
        throw_if(! $user, Exception::class, "Called from invalid auth context!");
        throw_if(! $user instanceof self, Exception::class, "This is not a user model!");
        return $user;
    }
}

```

shaedrich:
> I always create a method called `User::auth()` on my user to get proper hints.

This might be interesting to add to [`\Illuminate\Foundation\Auth\Access\Authorizable`](https://github.com/laravel/framework/blob/12.x/src/Illuminate/Foundation/Auth/Access/Authorizable.php), however, that won't solve @imliam's original problem 🤔

nazmulpcc:
Ignoring other issues, this will need manual tweaking as soon as someone uses another Authenticatable class.
What I do is add something like this in a `.ide.php` file

```php
namespace Illuminate\Contracts\Auth {

    interface Guard
    {
        public function user(): \App\Models\User;
    }
}
```
I don't usually reach for `barryvdh/laravel-ide-helper` in simple cases like this.

benfaerber:
> > > I always create a method called `User::auth()` on my user to get proper hints.
> > 
> > 
> > This might be interesting to add to [`\Illuminate\Auth\Authenticatable`](https://github.com/laravel/framework/blob/12.x/src/Illuminate/Auth/Authenticatable.php), however, that won't solve @imliam's original problem 🤔
> 
> If you add `public static function current(): ?static` to `Authenticatable` and call `User::current()` you get `null|\App\Models\User`

I guess there's nothing wrong with that. `auth()->user()` can return `null` so it will be consistent. The user will just have to manage the null check.

SanderMuller:
With the Laravel Idea Plugin being free now in PhpStorm, the vscode extension should fix this if it doesn't already, then this helper shouldn't be needed. I also think having a `user()`, `auth()->user()` and `request()->user()` is too confusing for new developers.

imliam:
Plenty of comments here about using the PHPStorm/VSCode plugin, but that has 2 problems:
1. You shouldn't have to rely on a very specific IDE to get a good experience for what can be supported with a fundamental PHP feature
2. That doesn't solve the problem for third-party tools outside those IDEs, like static analysis (which currently have to do workarounds for this common use case)

> @MadBox-99 - `auth()->user() can be Auth::user()` so u dont need helper, mostly try to use all time facade for geting authed user.

The issue this aims to resolve is to get the right _type_ hinted for the user returned, which those functions don't do for the scenario described.

danielh-official:
https://github.com/laravel/framework/blob/9f6700b7d3a108d2cd446c5a73fefa5675f1cb53/src/Illuminate/Contracts/Auth/Guard.php#L26

In the phpdoc for this line, can we achieve the same thing just by changing it to the following:

```php
    /**
     * Get the currently authenticated user.
     *
     * @return \Illuminate\Contracts\Auth\Authenticatable|\App\Models\User|null
     */
    public function user();
```

The problem with this approach is, who's to say they won't use a different user model? Can this be overwritten in a Laravel project. And if so, why even make the change in the first place? Better to just update the documentation to mention, "Oh, if you don't like auth()->user() showing an error in your IDE because it doesn't know it can be \App\Models\User, override the phpdoc here."

shaedrich:
@igorleszczynski Well, for this, Laravel is the wrong place 😂 The Framework strives for convenience and constantly breaks with this and other principles

rupeshstha:
I think this PR correctly identifies a real DX problem `auth()->user()` returning `Authenticatable|null` leads to poor IDE and static analysis support in the vast majority of Laravel apps that use the default user model.

However, I don’t think a global `user()` helper hardcoded to `App\Models\User` is the right layer to solve it.

Once this ships in the skeleton, it becomes implicitly canonical, even if “**editable**”. That introduces assumptions which break down quickly in:

- apps that swap `auth.providers.users.model`
- modular or domain-driven architectures
- applications where the authenticated entity is not always a User (e.g. Admin, Customer, or other Authenticatable models per guard)

In those cases, a globally typed `user()` helper either becomes misleading or forces runtime checks and exceptions, which weakens the type guarantees it’s trying to provide.

To me, this feels less like an authentication issue and more like a type-inference gap between Laravel and static analysis tools. I wonder if a slightly different approach could preserve the benefits you’re aiming for while remaining more flexible. for example:

- A model-centric approach for eg:`User::current(): ?static`
- Expanding and documenting typed injection like `#[CurrentUser] User $user`
- A more generic / tooling-focused solution that maps `auth()->user()` to the configured model

Overall, I think you’ve identified an important DX problem worth addressing at the framework level. I’m just not convinced a hard typed global helper is the safest long-term shape for it but I’d love to see this evolve into a solution that gives us strong typing without baking in assumptions about application structure.

aliali05340-tech:
Thank you, you did a great job, may God reward you.

elmarzouguidev:
What about Multi-Guard? 100% the helper will fail. 
In my opinion, this is a personal philosophy. Not a framework philosophy, so you can add a custom helper function in your root project.

indigoram89:
My User model is located in App/Services/Users/Models/User

dillowhenrick:
LGTM

composer.json
```diff
@@ -24,7 +24,8 @@
             "App\\": "app/",
             "Database\\Factories\\": "database/factories/",
             "Database\\Seeders\\": "database/seeders/"
-        }
+        },
+        "files": ["bootstrap/helpers.php"]
```
matmper, line 28: I don't think that would be the best approach; it destroys the framework's pattern. We have the option of working via other methods, as is already the case with url(), for example.

I have this helper in my application, but I made it something closed, knowing that it won't cause problems in multi-guard, closed scope. This helper is flawed!

ahmed-aliraqi:
@imliam From my perspective, adding a global `user()` helper to the framework might not be necessary. Laravel already provides `$request->user()` and `auth()->user()`, which are clear and consistent with the framework’s design.

For IDE autocomplete, tools like Laravel IDE Helper already solve this problem without introducing a new global helper.

If you still find it useful, I’d recommend keeping it as a project-specific helper instead of adding it to the framework. Also, it would be better to make it flexible by allowing a guard argument, for example:

```php
function user(?string $guard = null): \App\Models\User|\Illuminate\Contracts\Auth\Authenticatable|null
{
    return auth($guard)->user();
}
```

This way it stays consistent and gives you the flexibility to switch guards when needed.

jonnott:
Agree with those saying this should be added on a project-specific basis as and when something like it is needed. I manage dozens of completely different Laravel projects with all sorts of different user/auth/guard setups, and many which are pretty vanilla Laravel in this regard - and I can't think of a single one which would need a helper like this vs something like $request->user() or auth()->user() (or auth()->check() for that matter).

Also, I think having a helpers file in the skeleton (as opposed to the framework itself, which obviously has one), is a bit of a code smell. A project helpers file should be project-specific (if needed at all), and not out of the box. It would encourage inexperienced Laravel devs to just create global non-OOP methods to solve any problem they come across..

XavielCesarJoao:
> [!TIP]
>

bootstrap/helpers.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+use App\Models\User;
+
+/**
+ * Get the currently authenticated user model.
+ */
+function user(): User
+{
+    $currentUser = auth()->user();
```
gusdecool, line 10: IIRC, `auth()->user()` could return `nullable`. If that so, we should follow the convention to make this function to also return nullable `?User` to not confuse the user.

composer.json
gusdecool, line 28: We didn't need this at all if the function have namespace or in static class as composer auto loader will automatically load it.

e.g.

```php
class UserHelper
{
  public static function getAuthUser(): ?User {....}
}

// then call it with
UserHelper::getAuthUser();
```

bootstrap/helpers.php
bloomlive: Not only this, but it is assuming the authenticated model would be User, which might not be the case even with a single guard.

Again, we are using generic exception.

composer.json
amoos-cresco: I agreed it should be opened bootstrap/helpers.php could be open so anybody can add their own helpers

PR #6583, opened 2025-03-16
create action class with command

amirtavakolian:
in this pr i have added a new command class called ActionMakeCommand which you can create action classes with this command.
in ArtisanServiceProvider.php i have just add  'ActionMake' => ActionMakeCommand::class, in line 185 
and registerActionMakeCommand method in line 285

### DIFF
--- a/vendor/laravel/framework/src/Illuminate/Foundation/Console/ActionMakeCommand.php
+++ b/vendor/laravel/framework/src/Illuminate/Foundation/Console/ActionMakeCommand.php
@@ -1,0 +1,66 @@
+<?php
+
+namespace Illuminate\Foundation\Console;
+
+use Illuminate\Console\GeneratorCommand;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Input\InputOption;
+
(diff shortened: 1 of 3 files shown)


nadlambino:
Can't we already create a class using make:class command?

amirtavakolian:
> Can't we already create a class using make:class command?

Yes, we can use make:class, 
but if we do, we’ll have to manually move the class to the Actions folder. 
I think it’s better to have the Actions directory ready and generate action classes directly in the right place, 
avoiding the extra step.

nadlambino:
We don't really have to if we add the namespace in make:class command. `make:class Actions\DoSometing`. But if the action stub contains some boilerplates, like using some traits to make it queueable, that would be great

amirtavakolian:
> We don't really have to if we add the namespace in make:class command. `make:class Actions\DoSometing`. But if the action stub contains some boilerplates, like using some traits to make it queueable, that would be great

That’s true, but i think using make:action ensures consistency and avoids the need to specify the namespace every time. 
It also makes it easier for teams to follow a convention without worrying about structure manually.

PR #6523, opened 2025-01-23
Upgrade to Tailwind CSS v4.0

datlechin:

PR #6588, opened 2025-03-28
[12.x] Remove not necessary tailwindcss source declarations

rozsazoltan:
In PR #6523, the settings ensuring TailwindCSS v4 compatibility were added to version 12.x. However, three of the declared sources are unnecessary and only increase the codebase size.

TailwindCSS v4 automatically detects all sources that are not in `.gitignore` and are not special files. See more details here:
* [TailwindCSS v4 - which files are scanned](https://tailwindcss.com/docs/detecting-classes-in-source-files#which-files-are-scanned)

### Resources

As a result, there is no need to declare `resources/**/*.blade.php` and `resources/**/*.js` as sources.

### Storage Views

Additionally, scanning cached files in `storage/framework/views/*.php` is unnecessary, as TailwindCSS can discover the required classes by mapping the original source files.

Mapping cached files can lead to incorrect results if the developer dynamically constructs class names, e.g., `bg-red-$custom`, where `$custom` might default to `500`. In this case, due to the storage source, `bg-red-500` will be added to the prod, but the other classes won't, and the developer won't understand the explanation.

To avoid this, the TailwindCSS v4 documentation specifically advises developers to avoid such dynamic declarations.
* [TailwindCSS v4 - Dynamic class names](https://tailwindcss.com/docs/detecting-classes-in-source-files#dynamic-class-names)

So, if the classes are properly declared in the source (whether in app, resources, or elsewhere), they will be correctly included in the compiled CSS without the need to consider the storage folder.

### DIFF
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -1,9 +1,6 @@
 @import 'tailwindcss';
 
 @source '../../vendor/laravel/framework/src/Illuminate/Pagination/resources/views/*.blade.php';
-@source '../../storage/framework/views/*.php';
-@source '../**/*.blade.php';
-@source '../**/*.js';
 
 @theme {
     --font-sans: 'Instrument Sans', ui-sans-serif, system-ui, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji',


rozsazoltan:
To be honest, I would also remove the source pointing to pagination; it's understandably included since `.gitignore` excludes the contents of the `vendor` folder, and thus TailwindCSS does not automatically map it.

The missing of the source in this case creates an extra issue for developers using pagination when adding it.

Having the source included, on the other hand, adds a few classes that developers who never use pagination might never utilize.

By the way, the pagination documentation covers the correct implementation, so it might even make sense to remove it by default:
* [Laravel 12.x - Pagination with Tailwind](https://laravel.com/docs/12.x/pagination#tailwind)

rozsazoltan:
* #6823

PR #6596, opened 2025-04-08
Get app timezone from env

fpicosm:


taylorotwell:
Should just be UTC by default.

fpicosm:
@taylorotwell Why? I mean, all variables in `config/app` have their related env variable. 

By default the timezone should be UTC, ok, this is covered with `'timezone' => env('APP_TIMEZONE', 'UTC'),` but what if I want to use "Europe/Madrid" for example? I need to edit the config file instead of the env file, like we do with the rest of variables

GrahamCampbell:
It's generally not good practice to run the backend on anything other than UTC, regardless of the location of your business or customers. If you'd really like to change this, against recommendations, it's still possible to do by making this change directly in your own config file.

fpicosm:
> It's generally not good practice to run the backend on anything other than UTC, regardless of the location of your business or customers. If you'd really like to change this, against recommendations, it's still possible to do by making this change directly in your own config file.

I understand. But in that case, I don't get the point to have a "timestampsTz" option creating migrations, for example 😂

PR #6603, opened 2025-04-18
[12.x] PHPUnit 12

Jubeki:


composer.json
```diff
@@ -17,7 +17,7 @@
         "laravel/sail": "^1.41",
         "mockery/mockery": "^1.6",
         "nunomaduro/collision": "^8.6",
-        "phpunit/phpunit": "^11.5.3"
+        "phpunit/phpunit": "^12.1.2"
```
crynobone, line 20: PHPUnit 12 requires PHP 8.3 as minimum

crynobone (changes requested):

composer.json
Jubeki:
```suggestion
        "phpunit/phpunit": "^11.5.3 || ^12.1.2"
```

@crynobone Yeah, I just saw that too.

Should it be kept at PHPUnit 11 in the skeleton, or allow both?

GrahamCampbell (changes requested):
I think it's best to just have phpunit 11 as the default. I don't see a big rush to upgrade for most people.

Jubeki:
Yeah, agree that this is not a blocker, maybe something for Laravel 13.

Though it seems Symphony 8.0 will still require PHP 8.2, so I am not sure if it is necessary to bump Laravel to require PHP 8.3 for PHPUnit 12.

PR #6606, opened 2025-04-24
[12.x] Add cache lock_table default

jackbayliss:
It would be nice to have a default here, so then it's possible to do:

 `config('cache.stores.database.lock_table')` 
 
As we can already do: `config('cache.stores.database.table') // returns cache `

The default is like for like as it is [here](https://github.com/laravel/framework/blob/44e6a294e4441e9e3338008af0288979b3f677e8/src/Illuminate/Cache/CacheManager.php#L251)

If this is merged I'm happy to also add it into the framework repo, but thought I'd target the skeleton first 🤷🏻


taylorotwell:
Breaking change.

PR #6614, opened 2025-05-15
[12.x] Reverts #6613

AhmedAlaa4611:
Description
---
In practice, editors provide autocomplete and validation for `package.json` without explicitly specifying the `$schema`.

Validation
---
Screenshot (820)

Autocomplete
---
Screenshot (822)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,4 @@
 {
-    "$schema": "https://json.schemastore.org/package.json",
     "private": true,
     "type": "module",
     "scripts": {


taylorotwell:
So, is this causing problems?

AhmedAlaa4611:
@taylorotwell  It’s not causing any problems, but since most editors already provide validation and autocomplete for `package.json` by default, the `$schema` line doesn’t offer additional value in practice.

PR #6622, opened 2025-05-27
config/app.php timezone should be configurable based on .env var APP_TIMEZONE

developerByDay:
Unless we override the timezone in config/app.php in our project directory, UTC in not configurable and becomes the APP_TIMEZONE. Which in our case, caused many problems. Since we use expirable tokens across multiple applications, it caused serious issues in our native APIs.


GrahamCampbell (changes requested):
This change has been proposed before, and is not wanted to be included by default. Changing the time-zone on the server side is almost always the wrong answer. There are rare exceptions, but UTC time-zone is very strongly recommended.

PR #6638, opened 2025-07-03
[12.x] Update config/database.php to match the latest core configuration

AhmedAlaa4611:
Description
---
The `config/database.php` file in the project was outdated compared to the core's `database.php`. This update ensures it remains up to date with the latest changes, including any new options or improvements added to the framework.

Note
---
I'm not sure whether we should add this or not.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -36,6 +36,7 @@ return [
             'url' => env('DB_URL'),
             'database' => env('DB_DATABASE', database_path('database.sqlite')),
             'prefix' => '',
+            'prefix_indexes' => null,
             'foreign_key_constraints' => env('DB_FOREIGN_KEYS', true),
             'busy_timeout' => null,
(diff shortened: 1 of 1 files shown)


antonkomarev:
@taylorotwell maybe we should not have config files copies in this project, and copy them from the framework on first install?

AhmedAlaa4611:
Hello @antonkomarev, I think this could cause issues for users who use the `composer install` command when setting up a new project.

antonkomarev:
I think there are all solvable issues.

AhmedAlaa4611:
Also, this will break the slim skeleton we have made in Laravel 11, because we will have all the config files in every fresh installation project.

I think a good solution is to keep in `laravel/framework` only those files that don't present in the `laravel/laravel` and eliminate duplicates from the `laravel/framework`.

PR #6640, opened 2025-07-13
Updating .env.example

woganmay:
Avoids unnecessary docker errors when using Sail + installing all available dependencies at once.


woganmay:
Sorry, submitted this one in error - should have tested more first.

PR #6659, opened 2025-08-20
Add UniqueWithTrashed validation rule

yaminlam:
This PR adds a new custom validation rule UniqueWithTrashed that allows developers to enforce uniqueness of a column while ignoring soft-deleted records (deleted_at is not null).

Why it’s useful:

Laravel’s built-in unique rule doesn’t ignore soft-deleted rows by default.

This rule allows for cleaner validation logic in applications that use soft deletes.

Usage Example:

```
use App\Rules\UniqueWithTrashed;

$request->validate([
    'name' => ['required', new UniqueWithTrashed('zones', 'name')],
]);

```

This checks that the name column in the zones table is unique, excluding any soft-deleted rows.

### DIFF
--- a/app/Rules/UniqueWithTrashed.php
+++ b/app/Rules/UniqueWithTrashed.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace App\Rules;
+
+use Closure;
+use Illuminate\Contracts\Validation\ValidationRule;
+
+class UniqueWithTrashed implements ValidationRule
+{
+    protected string $modelClass;
+    protected string $column;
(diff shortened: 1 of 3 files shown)


yaminlam:
> You can use Rule::unique for this.

The built-in unique rule does not automatically ignore soft-deleted records. This means if a record exists with a non-null deleted_at, the default unique validation would still fail.

My UniqueWithTrashed rule automatically ignores soft-deleted items, making it easier to enforce uniqueness only among active records without repeating custom query logic each time. It’s especially helpful in applications that heavily use soft deletes.

PR #6661, opened 2025-08-21
Feature/unique with trashed

yaminlam:


yaminlam:
The built-in unique rule does not automatically ignore soft-deleted records. This means if a record exists with a non-null deleted_at, the default unique validation would still fail.

My UniqueWithTrashed rule automatically ignores soft-deleted items, making it easier to enforce uniqueness only among active records without repeating custom query logic each time. It’s especially helpful in applications that heavily use soft deletes.

yaminlam:
The UniqueWithTrashed rule is helpful because it ignores soft-deleted items (deleted_at not null) when checking uniqueness. This ensures that only active records are considered, which the default unique rule doesn’t do automatically.

taylorotwell:
Rule::unique()->withoutTrashed()

PR #6663, opened 2025-08-24
chore(scripts): add lint and fix commands for Pint

AmadulHaque:
### Summary
This PR adds two new Composer scripts to improve the developer workflow:

- `composer lint` → runs Pint to check code style
- `composer fix` → runs Pint in dirty mode to automatically fix issues

### Why
Having dedicated lint and fix scripts:
- Provides a faster way to run Pint
- Encourages consistent coding standards
- Improves DX (developer experience)

### Changes
- Updated `composer.json` scripts section:
  - Added `lint`: "vendor/bin/pint"
  - Added `fix`: "vendor/bin/pint --dirty"

### Testing
- Run `composer lint` → code style issues will be reported
- Run `composer fix` → issues will be auto-fixed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,7 +54,9 @@
         "test": [
             "@php artisan config:clear --ansi",
             "@php artisan test"
-        ]
+        ],
+        "lint": "vendor/bin/pint",
+        "fix": "vendor/bin/pint --dirty"
     },
     "extra": {
         "laravel": {


AhmedAlaa4611:
I’m a bit hesitant about adding these scripts though. Since **Pint** already has simple commands (pint / pint --dirty), keeping things as is avoids an **extra layer of indirection** and ensures contributors learn the actual tool commands.

This could add some redundancy and potentially confuse new contributors who may not realize they’re just calling Pint directly.

AmadulHaque:
> I’m a bit hesitant about adding these scripts though. Since **Pint** already has simple commands (pint / pint --dirty), keeping things as is avoids an **extra layer of indirection** and ensures contributors learn the actual tool commands.
> 
> This could add some redundancy and potentially confuse new contributors who may not realize they’re just calling Pint directly.

> 
Thanks for the feedback! 🙌 I see your point about redundancy and agree that Pint is already simple to run directly.

My intention with adding these scripts was mainly to:

Provide a consistent, easy-to-remember entry point (composer lint / composer fix)

Align with how many projects already expose common tooling (like composer test)

Lower the barrier for contributors who may not be familiar with Pint at first

That said, I totally understand the concern about introducing indirection. If the team prefers sticking to the raw Pint commands to keep things straightforward, I’m happy to adjust or close this PR. ✅

taylorotwell:

PR #6666, opened 2025-08-26
[12.x] Add default Redis retry configuration

mateusjatenee:
Allows Redis connections to be retried automatically. 

The additional configuration is only applied if the `phpredis` extension is loaded. It depends on constants that are not available without it — for users running other cache drivers, or running Predis, nothing changes.

`decorrelated_jitter` is used by default to avoid "retry storms" — [Here's an AWS article](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/)

With the default configuration, retry delays would be ~300ms, ~600ms, ~1000ms, for a ~2 second total retry window.

This is a bit of a beefy config, mostly because of that `match` statement — maybe we could pass strings and parse them on `framework` — that'd also solve having to be careful with calling the `Redis` namespace

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -158,12 +158,10 @@ return [
             'password' => env('REDIS_PASSWORD'),
             'port' => env('REDIS_PORT', '6379'),
             'database' => env('REDIS_DB', '0'),
-            ...extension_loaded('redis') ? [
-                'max_retries' => env('REDIS_MAX_RETRIES', 3),
-                'backoff_algorithm' => env('REDIS_BACKOFF_ALGORITHM', 'decorrelated_jitter'),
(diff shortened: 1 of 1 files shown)


mateusjatenee:
@taylorotwell docs PR here: https://github.com/laravel/docs/pull/10787

I kept it very brief, but I can expand if you think it deserves more space. 

As for this PR, I removed the `extension_loaded('redis')` conditional in favor of making the config smaller — it doesn't break if the user is running `predis`, it just doesn't do anything. Let me know if we should be more explicit there.

xtrasmal:
Careful with this one @taylorotwell 
PR has the 👹👹👹 mark of the beast

PR #6672, opened 2025-09-08
Fix CVEs affecting vite

faissaloux:
These are the two CVEs affeting vite
https://github.com/vitejs/vite/security/advisories/GHSA-g4jq-h2w9-997c
https://github.com/vitejs/vite/security/advisories/GHSA-jqfw-vq24-v9c3

#### Affected versions
\>=7.0.0,<=7.0.6


GrahamCampbell:
Sure, but no existing users with their existing lock files will get getting this change.

PR #6690, opened 2025-10-18
Decouple TailwindCSS from Default Laravel Installation

MasihTak:
## Summary

This pull request updates the default Laravel installation process to remove TailwindCSS as a mandatory dependency. The goal is to preserve Laravel’s core philosophy of developer freedom and provide a cleaner, more neutral starting point for all project types.

## Motivation

Not all Laravel applications require TailwindCSS or any frontend styling framework. Many projects are:

- API-only, or backend-driven
- Using Vue, React, or custom SCSS setups
- Minimal services that don’t need compiled CSS at all

Bundling TailwindCSS by default adds unnecessary dependencies, configuration files, and build steps for these cases.

Fix: https://github.com/laravel/framework/issues/57415

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,11 +7,9 @@
         "dev": "vite"
     },
     "devDependencies": {
-        "@tailwindcss/vite": "^4.0.0",
         "axios": "^1.11.0",
         "concurrently": "^9.0.1",
         "laravel-vite-plugin": "^2.0.0",
-        "tailwindcss": "^4.0.0",
         "vite": "^7.0.7"
     }
 }
--- a/vite.config.js
+++ b/vite.config.js
@@ -1,6 +1,5 @@
 import { defineConfig } from 'vite';
 import laravel from 'laravel-vite-plugin';
-import tailwindcss from '@tailwindcss/vite';
 
 export default defineConfig({
     plugins: [
@@ -8,6 +7,5 @@ export default defineConfig({
             input: ['resources/css/app.css', 'resources/js/app.js'],
             refresh: true,
         }),
-        tailwindcss(),
     ],
 });


MasihTak:
@taylorotwell  Respectfully, the statement that “there is no other viable way to maintainably and scalably style web applications” is not accurate at all, and it undermines the diversity of modern frontend architecture.

There are many proven, scalable approaches — from SCSS and CSS Modules to design systems powered by frameworks like Bootstrap, Vuetify, or custom atomic setups. Dismissing all alternatives as “non-viable” doesn’t align with Laravel’s philosophy of developer freedom and flexibility.

I also can’t ignore how this decision appears to align closely with Tailwind’s sponsorship and ecosystem ties. If that’s influencing the framework’s defaults, it raises valid concerns about neutrality. Laravel should not enforce a styling tool based on partnerships or popularity, but on architectural independence and developer choice.

For anyone interested in a deeper analysis, I’ve written an article that illustrates some of the real-world drawbacks of TailwindCSS in maintainability and scalability contexts:
[The Dark Side of TailwindCSS: A Real-World Example](https://medium.com/@masihtak/the-dark-side-of-tailwindcss-real-world-example-249b1064b89c)

punkfairie:
> There is no other viable way to maintainably and scalably style web applications.

I am incredibly disappointed to see a statement like this from you. If you, personally, feel this way and only use Tailwind in your own projects, that is fine. But to make a blanket statement like that and refuse to consider this pull request is a slap in the face to your users who care about CSS proper. There is a line between opinionated and forced conformity, and the fact that Tailwind is still included when no starter kit is selected crosses that line.

Laravel is a PHP framework. Trying to dictate how developers write CSS is outside the scope of a PHP framework. If you want to include Tailwind in the starter kits that is fine, especially since we now have the option of community starters. But forcing Tailwind even with no kit is too much. *Especially* when Laravel supports use as an API backend, why would those projects need CSS styling?

If you, personally, do not care to learn how to use actual CSS, that's your prerogative. Insisting that CSS is not scalable or maintainable just shows that you don't know how to do so. That is fine, not everyone needs to learn everything. You are certainly better at PHP than I am. But it's simply an untrue statement and a lot of very talented people have built careers on learning to maintain CSS at scale.

I have a lot of love for Laravel, but things like this have made me seriously consider finding another framework. You cannot run an open source project and refuse to be open to discussions like this.

izdrail:
@punkfairie it's all about the money . 
That lambo needs gasoline.

benodeveloper:
Actually, it should be up to the developer to decide what to use and what not to use, whether it is between Tailwind CSS, another framework, or using native CSS styling. The same applies to Axios,forcing me to use Axios when there are other alternative solutions out there, such as the native JavaScript fetch API, requires a bit more consideration

vini123:
虚幻之物对应着冥冥之路！：《羽》

ersin-demirtas:
I don't understand the need for this PR. How hard was it for you to remove tailwind? I use tailwindcss for 90% of my project. I am against this PR. Clearly money is not the problem here. You already have your public repo with the changes use that to create your projects.

punkfairie:
@ersin-demirtas It's not about having to do the work of removing it. Yes it is simple to remove, but why is it there in the first place? If someone is choosing no starter kit, they are either creating an API backend (which has zero to minimal need for CSS), or they want to use a different stack then is offered in the kits. 

There is no reason for Tailwind to be included and just adds bloat that must be removed. It's like if Ubuntu server edition insisted on installing Gnome and there was no way to prevent it. It's utterly unnecessary.

MasihTak:
I cleaned up the messy styles using pure CSS. If you want, I can make it even cleaner with SCSS!
https://github.com/MasihTak/laravel

PR #6759, opened 2026-02-25
[13.x] Ensure setup script creates SQLite database file before migration

prateekbhujel:
## Summary

Ensure `composer setup` creates `database/database.sqlite` before running migrations.

## Problem

On fresh installs using the default SQLite setup, `composer setup` may run `php artisan migrate --force` before `database/database.sqlite` exists, which can cause setup to fail.

## Solution

Add a pre-migration step in the `setup` script:

- Create `database/database.sqlite` if it does not exist.

This mirrors the safeguard already used in `post-create-project-cmd`.

## Validation

- `composer validate --no-check-publish --strict`
- Confirmed the setup script now guarantees SQLite file presence before migration.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
             "composer install",
             "@php -r \"file_exists('.env') || copy('.env.example', '.env');\"",
             "@php artisan key:generate",
+            "@php -r \"file_exists('database/database.sqlite') || touch('database/database.sqlite');\"",
             "@php artisan migrate --force",
             "npm install",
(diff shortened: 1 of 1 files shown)


taylorotwell:
The migrate command prompts you to create the database?

prateekbhujel:
Thanks for the clarification. You’re right — since `composer setup` runs `migrate --force`, the missing SQLite database file is already created by the command, so this change is redundant. Appreciate the quick review.

PR #6763, opened 2026-03-11
Add explicit PDO import for consistency with Pdo\Mysql import

cspatric:
PR #6760 introduced `use Pdo\Mysql` to use the new OOP PDO API on PHP 8.5+, but the PHP < 8.5 fallback still references `PDO::MYSQL_ATTR_SSL_CA` without an explicit `use PDO` import. While PHP resolves `PDO` from the global namespace correctly, adding the explicit import makes the file consistent and the intent clear.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -1,6 +1,7 @@
 <?php
 
 use Illuminate\Support\Str;
+use PDO;
 use Pdo\Mysql;
 
 return [


cspatric:
config/database.php currently imports `Pdo\Mysql` (added in #6760) but does
not import `PDO`, which is referenced in the same file via `PDO::MYSQL_ATTR_SSL_CA`.

**Why?**

PHP resolves `PDO` from the global namespace automatically since no namespace
is declared in the file — so this is not a bug. However, having `use Pdo\Mysql`
without `use PDO` creates an inconsistency: one class is explicitly imported,
the other is silently resolved. A developer reading the imports at the top of
the file gets an incomplete picture of the file's dependencies.

Adding `use PDO` makes the imports self-documenting and consistent with the
pattern established by #6760.

While a small change, keeping imports explicit and consistent is the kind of detail that makes a codebase easier to reason about at a glance.

PR #6764, opened 2026-03-13
[12.x] Bump Vite to v8

datlechin:
Vite 8 was released with Rolldown replacing esbuild and Rollup under the hood. The default Laravel config requires no changes, follow up https://github.com/laravel/vite-plugin/pull/341

https://vite.dev/guide/migration


taylorotwell:
We'll bump this when we have sorted out the breaking changes.

PR #6767, opened 2026-03-19
[13.x] Update `postmark` Config Key

devajmeireles:
According to the latest version of the Postmark Laravel package, the correct value is `services.postmark.token`. We can confirm it here: https://github.com/craigpaul/laravel-postmark/blob/3.x/src/PostmarkServiceProvider.php#L28

### DIFF
--- a/config/services.php
+++ b/config/services.php
@@ -15,7 +15,7 @@ return [
     */
 
     'postmark' => [
-        'key' => env('POSTMARK_API_KEY'),
+        'token' => env('POSTMARK_TOKEN', env('POSTMARK_API_KEY')),
     ],
 
     'resend' => [


mjamilasfihani:
I think it's 3rd party issue? because Postmark driver I believe it was handled by framework itself. here is the catch :

https://github.com/laravel/framework/blob/13.x/src/Illuminate/Mail/MailManager.php#L356-L381

mjamilasfihani:
and yes, `token` key was used in laravel 10.x

https://github.com/laravel/framework/blob/10.x/src/Illuminate/Mail/MailManager.php#L344

taylorotwell:
We support both.

PR #6770, opened 2026-03-25
Feat: add LOG_CHANNEL=null to phpunit.xml

Anthony14FR:
Tests were writing to storage/logs/laravel.log instead of discarding logs. Other test drivers (CACHE_STORE, MAIL_MAILER, QUEUE_CONNECTION, SESSION_DRIVER) are already configured for testing. LOG_CHANNEL should also be set to 'null' to prevent disk I/O and keep test logs clean.


GrahamCampbell (changes requested):
I don't think makes a good default. Many people want the tests to write to the logs to aid with debugging.

PR #6771, opened 2026-03-25
Feat: Add APP_TIMEZONE configuration

Anthony14FR:
Made the application timezone configurable via APP_TIMEZONE env variable, consistent with APP_LOCALE, APP_FALLBACK_LOCALE, and APP_FAKER_LOCALE which are already configurable. Defaults to UTC to maintain backward compatibility.


GrahamCampbell (changes requested):
See previous attempts at this. This is not encouraging out of the box because almost always setting this to not UTC is a terrible idea. You can of course put this in your own app if you know what you are doing.

PR #6773, opened 2026-03-26
[13.x] Add session driver to cache config

jackbayliss:
upgrading some open source stuff.

I tend to go to the Framework as the source of truth rather than the skeleton and noticed a bit of drift. 

Perhaps this was done for a reason, but I dont imagine it was?

This copies the same as laravel/framework.

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -39,6 +39,11 @@ return [
             'serialize' => false,
         ],
 
+        'session' => [
+            'driver' => 'session',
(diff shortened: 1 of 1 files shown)


taylorotwell:
It's basically intentional. There are some configs that exist at framework level that I think rarely need to be modified and it keeps skeleton configs a bit slimmer.

jackbayliss:
@taylorotwell Aha, got ya! thanks

PR #6774, opened 2026-03-27
Update .gitignore

Cegem-360:
Laravel ide helper is the most used IDE support package for laravel more than 120M download and without in VScode and PHPstorm isn't getting a Powerfull Facades helper.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 /storage/*.key
 /storage/pail
 /vendor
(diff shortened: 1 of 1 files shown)


martinbean:
This will need updating to a prefix match pattern (`_ide_helper*`) as ide-helper generates other files. For example, when using with the `-N` flag it will create a file named `_ide_helper_models.php`.

PR #6776, opened 2026-03-31
[security] pin axios version

NickSdot:
Active supply chain attack https://github.com/axios/axios/issues/10604

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     },
     "devDependencies": {
         "@tailwindcss/vite": "^4.0.0",
-        "axios": "^1.11.0",
+        "axios": ">=1.11.0 <=1.14.0"
         "concurrently": "^9.0.1",
         "laravel-vite-plugin": "^3.0.0",
         "tailwindcss": "^4.0.0",


NickSdot:
@timacdonald @jessarcher tagging you for a faster merge since you must be awake.

crynobone:
NPM has already removed the affected releases; Therefore, this doesn't affect new installations.

CleanShot 2026-03-31 at 14 12 19@2x

NickSdot:
> NPM has already removed the affected releases; Therefore, this doesn't affect new installations.

Doesn't mean that the compromised accounts cannot publish other infected versions. Better safe than sorry until the situation is fully resolved, IMO. ✌️

paulmax-os:
Given the constant increase of supply chain attacks, I'd even go a step further and pin all dependencies to provide a safe(r) default. Who knows what's next and it doesn't always have to be a main dependency, that gets compromised.

I know this will create some sort of maintenance hell, but simply running `laravel new axios-breach` could have been enough to get pwned.

SanderSander:
What about using the `min-release-age` setting in a .npmrc as default configuration? This way huge supply chain attacks can be prevented.

https://github.com/npm/cli/pull/8965

paulmax-os:
> What about using the `min-release-age` setting in a .npmrc as default configuration? This way huge supply chain attacks can be prevented.
> 
> [npm/cli#8965](https://github.com/npm/cli/pull/8965)

I guess this would be a nice addition, but it only helps in case you're running a version of npm that actually supports this relatively new flag.

francoism90:
Is axios actually still needed? I don't see it in the vue-starter for example: https://github.com/laravel/vue-starter-kit/blob/main/package.json

Nothing against axios, it can still be useful. But maybe replace it with fetch instead?

Does pinning actually work? When someone has access, I thought they could re-pin (tag) versions?

paulmax-os:
> Is axios actually still needed? I don't see it in the vue-starter for example: https://github.com/laravel/vue-starter-kit/blob/main/package.json
> 
> Nothing against axios, it can still be useful. But maybe replace it with fetch instead?

I've asked myself the same question for the basic Laravel (no starterkits) and I can't actually find a good reason for why it is installed. The starterkits around Inertia would pull it in as a (dev) dependency though (https://github.com/inertiajs/inertia/blob/3.x/packages/vue3/package.json).

Anyway this would not protect against future supply chain attacks on other packages.

Regarding the question whether pinning does work: As far as I know, the number you're seeing in a package.json (or lock) is not a tag, but a version number (assuming install through registry and not from git). This version number is immutable. The concept of tags also exists on npm, but there a tag is something like `latest`. However, if dependencies are installed from Github instead of the npm, the situation changes: branches are mutable, tags can be moved, and only a commit SHA is truly fixed.

In other words, it depends on the ecosystem, you're using:
- Github Actions -> pin with commit SHA
- Docker Images -> pin with digest (sha256)
- NPM Package -> pin with version number + lockfile and use `npm ci` in order to stick to lockfile

So for npm I would combine pinning with a  recent version of npm (latest of 11) and .npmrc file that holds:
```
ignore-scripts=true
save-exact=true
min-release-age=7
allow-git=none
```

`min-release-age` and `allow-git` are relatively new as already mentioned above. So you want to make sure your npm version actually supports those.

PR #6778, opened 2026-03-31
Remove axios and enable ignore-scripts

WendellAdriel:
Removing axios as a dependency and adding `.npmrc` with `ignore-scripts=true` to improve security by preventing third-party postinstall scripts from running during `npm install`.


resources/js/bootstrap.js
pushpak1300, line 1: we also need to remove this https://github.com/laravel/laravel/blob/e7fc00c74719d80a15aca462373c9024b0f86517/resources/js/app.js#L1 ?

pushpak1300:
```
const originalFetch = window.fetch;

window.fetch = (url, options = {}) => {
  options.headers = {
    'X-Requested-With': 'XMLHttpRequest',
    ...(options.headers || {})
  };
  return originalFetch(url, options);
};
``` 

I'm pretty sure we don't need this header anymore correct ?

WendellAdriel: I don't think it's needed, but if needed, I can add it

pataar:
It might be wise to also enable minimumReleaseAge. 

```
min-release-age=2 # days
```

Most affected packages are less that that age old. This also applies for transitive packages.

Available since 11.10 on npm: https://github.com/npm/cli/pull/8965

pushpak1300:
> It might be wise to also enable minimumReleaseAge.
> 
> ```
> min-release-age=2 # days
> ```
> 
> Most affected packages are less that that age old. This also applies for transitive packages.
> 
> Available since 11.10 on npm: [npm/cli#8965](https://github.com/npm/cli/pull/8965)

this is double edge swords. this means if any critical fixes or security fixes will be fixed after 2 days.

pataar:
> this is double edge swords. this means if any critical fixes or security fixes will be fixed after 2 days.

That's the tradeoff. It is an effective measure against supply chain attacks though. 1 day would be sufficient as well. 


More information: https://pnpm.io/supply-chain-security
(This is an article about pnpm, but the same principles and ideas apply to npm)

pataar:
Also, note that ignore-scripts is not a golden bullet right now:
https://github.blog/changelog/2026-02-18-npm-bulk-trusted-publishing-config-and-script-security-now-generally-available/

You might need `allow-git=none` too.

SanderSander:
> > It might be wise to also enable minimumReleaseAge.
> > ```
> > min-release-age=2 # days
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Most affected packages are less that that age old. This also applies for transitive packages.
> > Available since 11.10 on npm: [npm/cli#8965](https://github.com/npm/cli/pull/8965)
> 
> this is double edge swords. this means if any critical fixes or security fixes will be fixed after 2 days.

Critical security fixes can be detected with a simple `npm audit` or any other tool, and when a package is compromised it  can be updated immediately. I think it would be better to apply good practices in this repository, it can always be changed afterwards if needed by a developer in their project. (I also think this is how people get more aware of features like this). 

Only the `ignore-scripts=true` won't protected for a full 100% when a packages is compromised, the min-release-age would be a valuable extra protection here, and is created specifically  to add additional protection against supply chain attacks.

paulmax-os:
As already said in a different PR, the min-release-age and allow-git flags require users to be on a recent version of npm. Not arguing against it, just mentioning it.

But that's why I would combine that with pinned deps for those who are not using a recent version of npm.

taylorotwell:
This one has conflicts.

vegax87:
Don't forget to check the older branches, as axios hasn't been removed from their package.json yet

browner12:
for anyone using Bun, it will respect specific values in the `.npmrc` file, including the `ignore-scripts` value.

https://bun.com/docs/pm/npmrc#ignore-scripts-skip-lifecycle-scripts

you can also set it explicitly in the `bunfig.toml`

https://bun.sh/docs/runtime/bunfig#install-ignorescripts

PR #6785, opened 2026-04-08
Update comment in config/session.php

jonnott:
To say that JSON is the default is misleading and incorrect.

The framework code reveals that the actual default is 'php': https://github.com/search?q=repo%3Alaravel%2Fframework%20session.serialization&type=code

The opportunity was there with the launch of Laravel 13 to change the actual default in the framework, rather than having this situation where a supposed new 'default' is only actually applied if a config value is added to a project's config file.

This would of course have needed mentioning in the Upgrade Guide, but it probably should be covered in the Upgrade Guide anyway, as the effect of this config file 'default' change is significant for projects that already stored PHP objects in the session prior to Laravel 13, which is probably a 'medium' likelihood situation.

Essentially, if during a project's Laravel 13 upgrade, the session.serialization value is added to the config file as 'json', then it's a breaking change. However, if the value is not added during upgrade, the project keeps working. But only because 'php' is the actual framework default, contrary to what this comment suggests.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -219,10 +219,10 @@ return [
     | Session Serialization
     |--------------------------------------------------------------------------
     |
-    | This value controls the serialization strategy for session data, which
-    | is JSON by default. Setting this to "php" allows the storage of PHP
-    | objects in the session but can make an application vulnerable to
-    | "gadget chain" serialization attacks if the APP_KEY is leaked.
+    | This value controls the serialization strategy for session data. Setting
+    | this to "php" allows the storage of PHP objects in the session but can
+    | make an application vulnerable to "gadget chain" serialization attacks
+    | if the APP_KEY is leaked.
     |
     | Supported: "json", "php"
     |


jonnott:
One example of where this issue cropped up was where it broke laracasts/flash, which does store PHP objects (in fact, a Laravel collection of objects) in the session.

@JeffreyWay then made this commit to address this (the change to 'json' in project config) https://github.com/laracasts/flash/commit/4fc18fed3152d8910ed64589b0f5d049fd6a0806 .. however, this only actually fixes a project *if* that project had session.serialization added during Laravel 13 upgrade as 'json'. Otherwise, the framework default of 'php' still applies, and laracasts/flash is now in fact *broken* for such a project.

taylorotwell:
This comment is also misleading. The default is literally set to `json` just below this comment.

jonnott:
@taylorotwell But you're wilfully ignoring the point here - that the actual default value is that within the framework code, and this difference between framework-level default and config file setting is very much causing problems throughout the ecosystem .. there's just been a PR merged into Passport too to address this exact problem.

PR #6794, opened 2026-04-17
Allow timezone to be configured via APP_TIMEZONE env variable

bangbangda:
This PR changes the `timezone` value in `config/app.php` from a hardcoded `'UTC'` string to `env('APP_TIMEZONE', 'UTC')`, bringing it in line withevery other top-level config value in the same file.


Youssef-Amjad:
Why is it not accepted?

izdrail:
because you should configure that at the production lvl , plus utc makes sense

PR #6798, opened 2026-04-22
Replaced database options filter and added option for certificate verification

juanparati:
I noticed that MariaDB/MySQL PDO options are filtered using "array_filter".

This can be problematic when negatives values such "0" or "false" are used because those values are also filtered. 

## Changes

- Replaced "array_filter" by "Arr::whereNotNull" so options like MYSQL_ATTR_SSL_VERIFY_SERVER_CERT are not filtered.
- Added the option "ATTR_SSL_VERIFY_SERVER_CERT" / "MYSQL_ATTR_SSL_VERIFY_SERVER_CERT" with the value "false" as default, so if self-signed certificate is used. 

## Impact

- False values like "0" and "false" are accepted.

## Notes 

- PDO will work as usually when "MYSQL_ATTR_SSL_VERIFY_SERVER_CERT" is used with any false value and "MYSQL_ATTR_SSL_CA" / "MYSQL_ATTR_SSL_CA"  is not available.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -1,5 +1,6 @@
 <?php
 
+use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Pdo\Mysql;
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Is this a real issue you encountered in the real world?

PR #6802, opened 2026-04-27
[13.x] Adds pao by default

nunomaduro:
This pull request adds `nunomaduro/pao` by default on the skeleton; it's already in all starter kits since last week, with 0 problems reported so far: https://github.com/laravel/maestro/pull/81.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
         "laravel/pint": "^1.27",
         "mockery/mockery": "^1.6",
         "nunomaduro/collision": "^8.6",
+        "nunomaduro/pao": "^1.0.5",
         "phpunit/phpunit": "^12.5.12"
     },
     "autoload": {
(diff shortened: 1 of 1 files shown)


izdrail:
I’m not entirely sure how to frame this, so I’ll share my thoughts directly.

I recognize that you, @nunomaduro, are a core team member and active contributor to Laravel. However, having this package—even as a development dependency—under your personal namespace strikes me as unprofessional from a governance and trust perspective.

Perhaps we should consider making this a standalone Laravel package, officially maintained under the Laravel umbrella.

To highlight a potential risk: what would happen if either of the two packages currently under your name were compromised in a supply chain attack, and you were no longer working on Laravel at that time?

nunomaduro:
@izdrail thanks the feedback, going to move this project to the laravel organization.

PR #6811, opened 2026-05-06
feat: add env to timezone config

amirmms:
add a env load for timezone config


rojtjo:
See #6536 and [the many attempts to bring it back](https://github.com/laravel/laravel/pulls?q=APP_TIMEZONE+created%3A%3E2025-02-12)

amirmms:
> See #6536 and [the many attempts to bring it back](https://github.com/laravel/laravel/pulls?q=APP_TIMEZONE+created%3A%3E2025-02-12)

Interesting! Requiring the app’s timezone to always be set to UTC isn’t very appealing, and it takes away the flexibility to run multiple instances of the app in different timezones.

I wish this capability would come back.

crynobone:
The recommended approach is to use UTC and only convert to local timezone on response.

amirmms:
> The recommended approach is to use UTC and only convert to local timezone on response.

You’re right, but when logs are written, the timestamps end up incorrect in the log files. And for every Laravel instance we install, we have to tweak this config again.

PR #6821, opened 2026-05-15
refactor: reverse drop order in migrations

braxey:
Drop migration tables in reverse order.


izdrail:
what is this supposed to do ?

braxey:
> what is this supposed to do ?

It just makes more sense to drop tables in reverse order to unwind the migration.

PR #6824, opened 2026-05-21
[13.x] Render JSON exceptions for API routes by default

LucasCavalheri:
## Summary

This PR configures new Laravel applications to render exceptions as JSON for `api/*` routes by default.

It uses the existing `shouldRenderJsonWhen(...)` API and does not introduce any new framework surface area.

API routes now return JSON exception responses even when clients such as Postman, Insomnia, or curl do not explicitly send an `Accept: application/json` header.

Non-API routes preserve the existing `expectsJson()` behavior.

---

## Context

This PR follows the discussion from the previously proposed framework-level change:

[PR](https://github.com/laravel/framework/pull/60208)

In that discussion, Taylor suggested that this behavior may fit better in the `laravel/laravel` application skeleton rather than introducing a new helper API in `laravel/framework`.

This implementation keeps the change minimal and scoped to newly generated applications.

---

## Implementation

Adds the following exception rendering configuration to `bootstrap/app.php`:

```php
$exceptions->shouldRenderJsonWhen(
    fn (Request $request) => $request->is('api/*') || $request->expectsJson(),
);
```

This ensures:

- `api/*` routes render JSON exceptions by default
- existing `expectsJson()` behavior remains unchanged for all other routes

---

## Why

A common source of confusion for new Laravel API applications is receiving HTML exception responses from API routes when testing with tools that do not automatically send:

```http
Accept: application/json
```

This improves the out-of-the-box API experience while preserving existing behavior for web applications.


taylorotwell:
Can you revert changes to `composer.lock`?

LucasCavalheri:
> Can you revert changes to `composer.lock`?

Reverted the unintended composer.lock changes from the PR. Thanks for catching that.

PR #6825, opened 2026-05-21
[13.x] remove compiled views from Tailwind source scanning

browner12:
all of these compiled view files should be covered already by Tailwind automatically scanning the `/resources` directory.

I kept this separate from #6823 because it is kind of change of behavior, as this path will no longer be scanned. but I think it's okay for it to no longer be scanned.

### DIFF
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -1,7 +1,6 @@
 @import 'tailwindcss';
 
 @source '../../vendor/laravel/framework/src/Illuminate/Pagination/resources/views/*.blade.php';
-@source '../../storage/framework/views/*.php';
 
 @theme {
     --font-sans: 'Instrument Sans', ui-sans-serif, system-ui, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji',


taylorotwell:
Just going to leave it.

browner12:
this means we are scanning uncompiled and compiled Blade files, which will result in Tailwind working harder for no change.

francoism90:
@browner12 Isn't this useful for packages that don't use the resources/ folder? When you optimize the views, it will probably cache does views, meaning they are included in the final build.

rozsazoltan:
@francoism90 Analyzing the cache is not the right approach. It encourages developers to generate even dynamically constructed class names [[1]](https://stackoverflow.com/q/69687530/15167500) [[2]](https://stackoverflow.com/q/79769249/15167500) (e.g. `"bg-" . $color`) "because it works" (via cache), while in reality this should be avoided.

Tailwind CSS is capable of discovering its class names in any file, not necessarily only under resources, but anywhere, whether it is a PHP file, C, Rust, or something else.

If your source files come from `vendor` or `node_modules`, since these are excluded by `.gitignore`, you need to register the relevant package path with `@source` [[3]](https://stackoverflow.com/a/79433685/15167500) in that specific project. So by default, there is no need to ship for this.

francoism90:
@rozsazoltan I agree with your approach.

The only thing for keeping it, are these edge cases where someone forgets to register it or where it may be missing somehow. 

I think that's the reason why it's defined as source? 

```bash
php artisan optimize:view
npm run build
```

This would do a final scan I think?

PR #6836, opened 2026-06-11
[13.x] Use PHPUnit Test attribute in example tests

congkv:
This PR updates the skeleton's example tests to use the PHPUnit `#[Test]` attribute instead of the `test_` method-name prefix.
  
PHPUnit 12.5 (required by this skeleton) fully supports attributes as the modern, recommended way to mark test methods, and this keeps the generated test files aligned with current PHPUnit conventions for new applications.
  
  - Add `use PHPUnit\Framework\Attributes\Test;`
  - Mark each example method with `#[Test]` and drop the redundant `test_` prefix
  
  No behavior changes.

### DIFF
--- a/tests/Feature/ExampleTest.php
+++ b/tests/Feature/ExampleTest.php
@@ -3,6 +3,7 @@
 namespace Tests\Feature;
 
 // use Illuminate\Foundation\Testing\RefreshDatabase;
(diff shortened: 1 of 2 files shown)


taylorotwell:
Maybe on the next major release.

PR #6839, opened 2026-06-18
DOC: remove Mailgun from config/services.php

asanikovich:
Removed an outdated comment/mention of Mailgun from config/services.php.


taylorotwell:
I'll fix! Need to maintain the spacing.

PR #6840, opened 2026-06-20
Add 'array' as a supported maintenance mode driver doc block

jackbayliss:
Follows what we did here: https://github.com/laravel/framework/pull/60490 

Cus its the skeleton, feels nice / consistent to me

Also nicer for AI I suppose?


MatchZone2026 (commented):
B

PR #6843, opened 2026-06-26
Use `artisan dev` command in `composer dev` script

ace-of-aces:
Same as with https://github.com/laravel/maestro/pull/171

---

Replaced `npx concurrently ...` with [the new `artisan dev` command](https://github.com/laravel/framework/pull/60412) in `composer dev`, as it provides the same defaults and removes the need to keep two configurations in sync.

Raised the minimum Laravel dependency to ^13.17 to ensure `artisan dev` and `artisan dev:list` are available.

Kept the `concurrently` Node dependency, as `artisan dev` uses it under the hood.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "license": "MIT",
     "require": {
         "php": "^8.3",
-        "laravel/framework": "^13.8",
+        "laravel/framework": "^13.17",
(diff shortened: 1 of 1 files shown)


taylorotwell:
We will swap this when we're ready.

PR #6846, opened 2026-07-14
Update APP_URL to use port 8000

dipesh79:


taylorotwell:
This shouldn't usually be necessary?

dipesh79:
> This shouldn't usually be necessary?

Yeah but when i am cloning ```.env.example``` locally then it might be useful.

nicotravassos:
@taylorotwell can we have the next release please last one was 27th May I want to create a new project without patching it with new stuff. Thank you.

PR #6851, opened 2026-08-06
Clarify fixed timezone rationale

QistiAmal1212:
Add explanatory comment to config/app.php stating the timezone is intentionally not read from the environment because it must remain identical across environments/servers that share a database to avoid misinterpreting stored dates.


QistiAmal1212:
Regarding the repeated PRs and discussions on this issue, I think it's fair to address it in the comments here.

PR #6852, opened 2026-08-06
Remove unnecessary concurrently dependency

michaelnabil230:
This PR removes the `concurrently` dependency from `package.json`.

Since it is no longer used following the changes in #6849, it can be safely removed, reducing the project's npm dependencies.


taylorotwell:
php artisan dev still uses concurrently.

PR #6870, opened 2026-09-21
Make composer run setup idempotent by guarding key:generate

ametad:
**Description**

`composer run setup` bootstraps an environment, but it is only mostly safe to run more than once. Every step converges to the same end state on a repeat run, except `key:generate`:

- `composer install` installs exactly what `composer.lock` specifies.
- The `.env` copy is guarded with `file_exists('.env') ||`.
- `migrate --force` only applies pending migrations.
- `npm install` and `npm run build` are deterministic for a given lockfile and source tree.
- `artisan key:generate` generates a new `APP_KEY` on every invocation.

Outside production, `key:generate` does not check for an existing key. It only calls `confirmToProceed()`, which auto-confirms, so it silently overwrites the key (see `KeyGenerateCommand::setKeyInEnvironmentFile()`). In production it prompts, which fails under `--no-interaction`.

**Why this matters**

Re-running setup rotates the encryption key. Active sessions are invalidated, signed URLs stop validating, and anything encrypted can no longer be decrypted. Because the other steps are safe to repeat, setup looks like a routine "setup my environment" command while hiding one destructive side effect.

**Reproduce**

```sh
composer run setup
grep ^APP_KEY= .env   # APP_KEY=base64:AAAA...

composer run setup
grep ^APP_KEY= .env   # APP_KEY=base64:BBBB...  <- key changed
```

**Change**

`key:generate` only runs when `.env` has no `APP_KEY` or an empty one:

```diff
-    "@php artisan key:generate",
+    "@php -r \"if (!preg_match('/^APP_KEY=.+/m', file_get_contents('.env'))) { passthru(PHP_BINARY.' artisan key:generate --ansi', \\$c); exit(\\$c); }\"",
```

- A plain `@php artisan key:generate` cannot be made conditional: composer scripts have no conditionals, and `@php` is only expanded at the start of a command. The guard therefore uses the same `@php -r` style as the existing `.env` copy step.
- `PHP_BINARY` runs artisan with the same PHP that composer uses.
- `passthru()` streams artisan's output (including ANSI colors) straight to the terminal.
- The exit code is passed on with `exit($c)`, so a failing `key:generate` still fails the script. `$c` is escaped (`\\$c`) because composer runs the line through a shell that would otherwise expand it.

**Result**

`composer run setup` can be run any number of times and converges to the same state, keeping the existing key. Destructive resets stay explicit and separate.

**Testing**

```sh
composer run setup
grep ^APP_KEY= .env   # APP_KEY=base64:AAAA...

composer run setup
grep ^APP_KEY= .env   # APP_KEY=base64:AAAA...  <- key not changed
```

**Scope note**

The guard only inspects `.env`. If `APP_KEY` comes purely from a real environment variable, `.env` may still get a generated key. That is harmless, since real environment variables take precedence over `.env` in Laravel.


ametad:
Hi @taylorotwell , 
If my PR is really not bringing anything constructive to the table, than I understand the decision this is declined. But at present, I am guessing why this PR is cancelled. Would you be so kind te tell me why it is turned down? Perhaps I can change something to make it better. Or explain some things that are unclear. 

Is it perhaps because I used ai to produce this PR? Because I have used it, clearly. It is not something I tried to disguise. But I've not used it out of lazyness, I've checked out the answers because I know it can make mistakes. 

I've read the contribution guide again and I found this text:

> Pull request descriptions must be written entirely by the contributor. Pull requests with AI-generated descriptions will be closed.

When I wrote and coded the PR, my focus was on understanding and testing. And I've reviewed the text and code several times, made some changes. 

May I create another PR with the same commits, but with a rewritten description, without being blocked from the repository?

I hope to learn something so I can contribute, in a way that is really contributing.

Have a nice day.

iruoy:
Why not add a `--missing` option to the actual command in the framework and then use that option here? Then you won't need to use the `passthru(PHP_BINARY.' artisan')` hack.

ametad:
> Why not add a `--missing` option to the actual command in the framework and then use that option here? Then you won't need to use the `passthru(PHP_BINARY.' artisan')` hack.

That could be a solution as well. And it is already possible with the `\Illuminate\Console\Prohibitable` trait to configure the Command to not overwrite the key. There are multiple solutions. 

But the point I wanted to make first is: can we make the Composer script `setup` **idempotent**?

